PR #14, opened 2023-05-30
Add a simple build and test CI

j2kun:


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,17 @@
+name: Build and Test
+on: [push]
```
asraa, line 2: nit: recommend adding default read-only perms at the top-level of the workflow

```
permissions: read-all
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "This job was triggered by a ${{ github.event_name }} event."
+      - run: echo "Running on OS ${{ runner.os }}"
+      - run: echo "Checking out branch ${{ github.ref }}."
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: "Run `bazel build`"
+        run: |
+          bazel build //...
```
asraa, line 14: Is bazel already installed? I don't *think* it is, and you might need a bazel installation action

https://github.com/bazelbuild/setup-bazelisk maybe this?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,17 @@
...
+on: [push]
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "This job was triggered by a ${{ github.event_name }} event."
+      - run: echo "Running on OS ${{ runner.os }}"
+      - run: echo "Checking out branch ${{ github.ref }}."
+      - name: Check out repository code
+        uses: actions/checkout@v3
```
asraa, line 11: nit: in a follow-up I can pin actions by hashes & use depandabot to update them

.github/workflows/build_and_test.yml
j2kun: It is installed! https://github.com/actions/runner-images/pull/490

j2kun:
The action succeeded https://github.com/google/heir/actions/runs/5126658016/jobs/9221436296

Though it took 1.5 hours. @asraa what do you think about merging this as-is and, by possibly ignoring it for internal CLs, keep going and have a followup ticket to use a hosted runner and/or cache. I want to do a bit more research into how the caching works in bazel and what other MLIR Google projects do.

asraa:
> keep going and have a followup ticket to use a hosted runner and/or cache.

Yes! Check out the GitHub Action cache - if that's not good enough, then probably RBE is the way.

.github/workflows/build_and_test.yml
asraa: thanks!

.github/workflows/build_and_test.yml
j2kun: just did it, using pin-github-action https://github.com/mheap/pin-github-action, manually confirming it's the right hash https://github.com/actions/checkout/commit/8e5e7e5ab8b370d6c329ec480221332ada57f0ab

.github/workflows/build_and_test.yml
j2kun: Done

PR #24, opened 2023-06-15
Add basic HEIR website

j2kun:
It's a hugo site with the docsy theme, some basic placeholder content, and a github workflow that publishes to the gh-pages branch when main is touched.


j2kun:
Tested at https://j2kun.github.io/heir/

PR #74, opened 2023-07-28
Add a polynomial attribute

j2kun:
This PR adds a polynomial attribute definition to the poly dialect, which can be used to annotate types or ops with information about the polynomial ring in which the operation occurs.

~~An example attribute is added to `poly.mul` for demonstration, with a TODO left to remove it, and the poly_ops.mlir test updated to check for it.~~ Update: I had time to add `#poly.ring` so I went ahead and did it.

Notable aspects of this PR:

- Polynomial is an internal representation of a polynomial as a list of Monomials, using the llvm::TrailingObjects storage paradigm. Polynomial is not meant to be used for polynomial arithmetic _per se_ (i.e., we shouldn't be doing large computations with it). I had originally added `Polynomial::operator+` expecting to do some simple operations with it, and then removed it when I realized it was unnecessary and we can just use `fromMonomials` and construct the monomials directly.
- The underlying coefficient/exponent types are all APInt with a 64-bit width, the idea being that if we need to increase it later we can do so easily. I also saw [MPInt](https://github.com/llvm/llvm-project/blob/b299ec16661f653df66cdaf161cdc5441bc9803c/mlir/include/mlir/Analysis/Presburger/MPInt.h#L87) in the Presburger library, and I think that might be a better type to use for this because it doesn't assert that the bit width of the two operands to be the same. But since it's hidden inside Presburger I wasn't sure if it is sensible/stable to depend on. I asked @groverkss in the mlir chat about this.

### DIFF
--- a/include/Dialect/Poly/IR/PolyAttributes.td
+++ b/include/Dialect/Poly/IR/PolyAttributes.td
@@ -1,0 +1,66 @@
+#ifndef HEIR_INCLUDE_DIALECT_POLY_IR_POLYATTRIBUTES_TD_
+#define HEIR_INCLUDE_DIALECT_POLY_IR_POLYATTRIBUTES_TD_
+
+include "PolyDialect.td"
+
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/AttrTypeBase.td"
+
+class Poly_Attr<string name, string attrMnemonic, list<Trait> traits = []>
+    : AttrDef<Poly_Dialect, name, traits> {
+  let mnemonic = attrMnemonic;
+}
+
+def Polynomial_Attr : Poly_Attr<"Polynomial", "polynomial"> {
+  let summary = "An attribute containing a single-variable polynomial.";
+  let description = [{
+     #poly = #poly.poly<x**1024 + 1>
+  }];
+
+  let parameters = (ins "Polynomial":$value);
+
+  let builders = [
+    AttrBuilderWithInferredContext<(ins "Polynomial":$value), [{
+      return $_get(value.getContext(), value);
+    }]>
+  ];
+  let extraClassDeclaration = [{
+    using ValueType = Polynomial;
+    Polynomial getPolynomial() const { return getValue(); }
+  }];
+
+  let skipDefaultBuilders = 1;
+  let hasCustomAssemblyFormat = 1;
+}
+
+def Ring_Attr : Poly_Attr<"Ring", "ring"> {
+  let summary = "An attribute specifying a ring.";
+  let description = [{
+    An attribute specifying a polynomial quotient ring with integer
+    coefficients, $\mathbb{Z}/n\mathbb{Z}[x] / (p(x))$.
+
+    `cmod` is the coefficient modulus $n$, and `ideal` is the ring ideal
+    $(p(x))$. Because all ideals in a single-variable polynomial ring are
+    principal, the ideal is defined by a single polynomial.
+
+      #ring = #poly.ring<cmod=1234, ideal=#poly.polynomial<x**1024 + 1>>
+  }];
+
+  let parameters = (ins "APInt": $cmod, "Polynomial":$ideal);
+
+  let builders = [
+    AttrBuilderWithInferredContext<(ins "APInt": $cmod, "Polynomial":$ideal), [{
+      return $_get(ideal.getContext(), cmod, ideal);
+    }]>
+  ];
+  let extraClassDeclaration = [{
+    Polynomial ideal() const { return getIdeal(); }
+    APInt coefficientModulus() const { return getCmod(); }
+  }];
+
+  let skipDefaultBuilders = 1;
+  let hasCustomAssemblyFormat = 1;
+}
+
+
+#endif  // HEIR_INCLUDE_DIALECT_POLY_IR_POLYATTRIBUTES_TD_
--- a/include/Dialect/Poly/IR/PolyOps.td
+++ b/include/Dialect/Poly/IR/PolyOps.td
@@ -1,6 +1,7 @@
 #ifndef HEIR_INCLUDE_DIALECT_POLY_IR_POLYOPS_TD_
 #define HEIR_INCLUDE_DIALECT_POLY_IR_POLYOPS_TD_
 
+include "PolyAttributes.td"
 include "PolyDialect.td"
 include "PolyTypes.td"
 include "mlir/Interfaces/InferTypeOpInterface.td"
@@ -56,7 +57,9 @@ def Poly_MulOp : Poly_Op<"mul", [SameOperandsAndResultType]> {
   let summary = "Multiplication operation between polynomials.";
 
   let arguments = (ins
-    Variadic<Poly>:$x
+    Poly:$lhs,
+    Poly:$rhs,
+    Ring_Attr:$ring
   );
 
   let results = (outs
--- a/include/Dialect/Poly/IR/PolyTypes.td
+++ b/include/Dialect/Poly/IR/PolyTypes.td
@@ -2,13 +2,10 @@
 #define HEIR_INCLUDE_DIALECT_POLY_IR_POLYTYPES_TD_
 
 include "PolyDialect.td"
+include "PolyAttributes.td"
 
 include "mlir/IR/DialectBase.td"
 include "mlir/IR/AttrTypeBase.td"
-
-//===----------------------------------------------------------------------===//
-// Poly type definitions
-//===----------------------------------------------------------------------===//
 
 // A base class for all types in this dialect
 class Poly_Type<string name, string typeMnemonic>
@@ -21,10 +18,8 @@ def Poly : Poly_Type<"Polynomial", "poly"> {
 
   let description = [{
     A type for polynomials in a polynomial quotient ring.
-
-    The type is parametrized by the degree of the polynomial
-
   }];
+
 }
 
 #endif  // HEIR_INCLUDE_DIALECT_POLY_IR_POLYTYPES_TD_
--- a/lib/Dialect/Poly/IR/PolyAttributes.cpp
+++ b/lib/Dialect/Poly/IR/PolyAttributes.cpp
@@ -1,0 +1,182 @@
+#include "include/Dialect/Poly/IR/PolyAttributes.h"
+
+#include "include/Dialect/Poly/IR/Polynomial.h"
+#include "llvm/include/llvm/ADT/SmallSet.h"           // from @llvm-project
+#include "llvm/include/llvm/ADT/StringExtras.h"       // from @llvm-project
+#include "llvm/include/llvm/ADT/StringRef.h"          // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"           // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace poly {
+
+void PolynomialAttr::print(AsmPrinter &p) const {
+  p << '<';
+  p << getPolynomial();
+  p << '>';
+}
+
+/// Try to parse a monomial. If successful, populate the fields of the outparam
+/// `monomial` with the results, and the `variable` outparam with the parsed
+/// variable name.
+ParseResult parseMonomial(AsmParser &parser, Monomial &monomial,
+                          llvm::StringRef *variable, bool *isConstantTerm) {
+  APInt parsedCoeff(APINT_BIT_WIDTH, 1);
+  auto result = parser.parseOptionalInteger(parsedCoeff);
+  if (result.has_value()) {
+    if (failed(*result)) {
+      parser.emitError(parser.getCurrentLocation(),
+                       "Invalid integer coefficient.");
+      return failure();
+    }
+  }
+
+  // Variable name
+  result = parser.parseOptionalKeyword(variable);
+  if (!result.has_value() || failed(*result)) {
+    // we allow "failed" because it triggers when the next token is a +,
+    // which is allowed when the input is the constant term.
+    monomial.coefficient = parsedCoeff;
+    monomial.exponent = APInt(APINT_BIT_WIDTH, 0);
+    *isConstantTerm = true;
+    return success();
+  }
+
+  // Parse exponentiation symbol as **
+  // We can't use caret because it's reserved for basic block identifiers
+  // If no star is present, it's treated as a polynomial with exponent 1
+  if (failed(parser.parseOptionalStar())) {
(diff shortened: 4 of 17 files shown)


AlexanderViand:
Looking forward to digging into this properly next week! I think the decision to go with 64 bits is reasonable for now, with a proper answer making a good agenda item for Tuesday's meeting :) Kunwar (@Groverkss) also kindly agreed to join for the meeting so we can quiz him directly.

I already chatted with a few people about what the "correct" answer to [this question](https://github.com/google/heir/discussions/55#discussioncomment-6556643) is. So far, no one had a good use case for the non-RNS (i.e., multiprecision) variants of the schemes, but maybe they are relevant for some more esoteric/advanced applications of FHE. Even for RNS variants, I guess there could be (HW) implementations that use RNS limbs with more than 64 bits? However, again, I'm unaware of any of the top of my head. If we can do MPInt without too much overhead in the "common" 64 bit case, it's probably the safer/more future proof choice.

PS: @j2kun Could you please request the review/give any necessary repo permission to my work account (@AlexanderViand) instead? Thanks!

j2kun:
> PS: @j2kun Could you please request the review/give any necessary repo permission to my work account (@AlexanderViand) instead? Thanks!

Sent the invite to @AlexanderViand-intel and will add that account as a reviewer after you accept.

WORKSPACE
```diff
@@ -109,3 +109,22 @@ http_archive(
...
+#
+#   bazel run @hedron_compile_commands//:refresh_all
+#
+# to ingest new data into these tools.
+#
+# See the project repo for more details and configuration options
+# https://github.com/hedronvision/bazel-compile-commands-extractor
+http_archive(
+    name = "hedron_compile_commands",
+    url = "https://github.com/hedronvision/bazel-compile-commands-extractor/archive/3dddf205a1f5cde20faf2444c1757abe0564ff4c.tar.gz",
```
asraa, line 125: small nit: can you add a comment date here? appears to be 2023-05-13

include/Dialect/Poly/IR/PolyAttributes.td
```diff
@@ -0,0 +1,71 @@
...
+
+class Poly_Attr<string name, string attrMnemonic, list<Trait> traits = []>
+    : AttrDef<Poly_Dialect, name, traits> {
+  let mnemonic = attrMnemonic;
+}
+
+class Poly_Attr_With_Custom_Format<string name, list<Trait> traits = []>
+    : AttrDef<Poly_Dialect, name, traits> {
+  let mnemonic = ?;
+}
```
asraa, line 17: can you remove this one please?

include/Dialect/Poly/IR/PolyAttributes.td
```diff
@@ -0,0 +1,71 @@
...
+  }];
+
+  let parameters = (ins "APInt": $cmod, "Polynomial":$ideal);
+
+  let builders = [
+    AttrBuilderWithInferredContext<(ins "APInt": $cmod, "Polynomial":$ideal), [{
+      return $_get(ideal.getContext(), cmod, ideal);
+    }]>
+  ];
+  let extraClassDeclaration = [{
```
asraa, line 61: might be nice to have an extra getter for degree of the polynomial? WDYT?

speaking of: because degree is usually determined as `n` maybe the coefficient modulus should be `q` (and we can specificy that it doesn't necessarily require being prime)

include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -56,7 +57,9 @@ def Poly_MulOp : Poly_Op<"mul", [SameOperandsAndResultType]> {
   let summary = "Multiplication operation between polynomials.";
 
   let arguments = (ins
-    Variadic<Poly>:$x
+    Poly:$lhs,
```
asraa, line 60: +1 to the change from variadic to two individual poly's - let's keep it simple.

include/Dialect/Poly/IR/PolyAttributes.td
```diff
@@ -0,0 +1,71 @@
...
+  let mnemonic = attrMnemonic;
+}
+
+class Poly_Attr_With_Custom_Format<string name, list<Trait> traits = []>
+    : AttrDef<Poly_Dialect, name, traits> {
+  let mnemonic = ?;
+}
+
+def Polynomial_Attr : Poly_Attr<"Polynomial", "polynomial"> {
+  let summary = "An attribute containing a polynomial.";
```
asraa, line 20:
```suggestion
  let summary = "An attribute containing a single-variable polynomial.";
```

asraa: maybe it also needs to be specified that exponents are positive valued only?

lib/Dialect/Poly/IR/PolyAttributes.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+/// Try to parse a monomial. If successful, populate the fields of the outparam
+/// `monomial` with the results, and the `variable` outparam with the parsed
+/// variable name.
+ParseResult parseMonomial(AsmParser &parser, Monomial &monomial,
+                          llvm::StringRef *variable, bool *isConstantTerm) {
+  APInt parsedCoeff(APINT_BIT_WIDTH, 0);
+  auto result = parser.parseOptionalInteger(parsedCoeff);
+  if (!result.has_value()) {
+    parsedCoeff = APInt(APINT_BIT_WIDTH, 1);
```
asraa, line 24: nit/ my own understanding: since it's an optional integer to parse (so it doesn't set `parsedCoeff` to something unexpected when it doesn't encounter an integer), why not make 1 the default value and then only check if it has a failure value to emit error?

lib/Dialect/Poly/IR/Polynomial.cpp
```diff
@@ -0,0 +1,73 @@
...
+  auto assignCtx = [context](detail::PolynomialStorage *storage) {
+    storage->context = context;
+  };
+
+  // A polynomial's terms are canonically stored in order of increasing degree.
+  llvm::OwningArrayRef<Monomial> monomials_copy =
+      llvm::OwningArrayRef<Monomial>(monomials);
+  std::sort(monomials_copy.begin(), monomials_copy.end());
+
+  StorageUniquer &uniquer = context->getAttributeUniquer();
```
asraa, line 28: not sure if there is a good answer here but what is the difference / purpose of the attribute uniquer versus any other type?

tests/poly_ops.mlir
```diff
@@ -1,25 +1,28 @@
-// RUN: heir-opt %s | FileCheck %s
+// RUN: heir-opt %s > %t
+// RUN: FileCheck %s < %t
 
 // This simply tests for syntax.
 
+#my_poly = #poly.polynomial<1 + x**1024>
+#my_poly_2 = #poly.polynomial<t**3 + t + 2>
+#my_poly_3 = #poly.polynomial<2>
+#my_poly_4 = #poly.polynomial<x>
```
asraa, line 9: were these needed?

asraa (commented):
Thank you! This is great

WORKSPACE
j2kun: I have been copy/pasting this into all my in-progress PRs now that I'm doing this dev in open source, so let me just pull it out into a separate one and we can get it in.

include/Dialect/Poly/IR/PolyAttributes.td
j2kun: will change, just adding a comment: IMO the generally accepted definition of a polynomial requires all terms to have positive exponents, and when the exponents can be negative they're more specifically called Laurent series. (and the ring notation is different R[[x]] instead of R[x].

include/Dialect/Poly/IR/PolyAttributes.td
j2kun: Do you mean the polynomial's actual degree (which is data dependent; `x^3 + 1` has degree 3 in `Z[x]/(x^8 + 1)`) or just a single-method helper that calls `getPolynomial().getDegree()`? If the latter, I don't see two chained method calls as a bad thing, as it keeps the API surface smaller. Maybe we might want to tighten down the Polynomial API though, and not expose the Monomial list directly....

lib/Dialect/Poly/IR/PolyAttributes.cpp
j2kun: Ah, yes, this was an artifact of me learning as I went. I initially declared `APInt parsedCoeff` with no initial value, but not specfiying the bit width (using the default bit width) causes assertion failures when it tries to compare two values with different bit widths. So then I added the initializer to fix that and didn't notice that the control flow could be simplified as a result. Will fix.

lib/Dialect/Poly/IR/Polynomial.cpp
j2kun: So I don't know why the MLIRContext has multiple uniquers (they're all the same type), but I figure they might be for conflict avoidance of some sort? Like if you want separate instances of a type de-duped for attributes vs types vs affine maps? Not really sure.

tests/poly_ops.mlir
j2kun: I intended to have them to handle all the parser edge cases, but now I'm realizing I didn't add any integer coefficients to the terms with nonzero degree, so I am updating this one slightly to have that.

include/Dialect/Poly/IR/PolyAttributes.td
j2kun: Done.

WORKSPACE
j2kun: Migrated to https://github.com/google/heir/pull/82

include/Dialect/Poly/IR/PolyAttributes.td
asraa: Ah yeah, that's a fair point - it came up more in my mind as an element of a polynomial quotient ring where the negative exponents could just be a representation of the same with positive. But this should be a generic poly. Thanks!

include/Dialect/Poly/IR/PolyAttributes.td
asraa: Yeah, here it would be `ideal().getPolynomial().getDegree()` - that's fine! agree let's keep the API and tablegen as minimal as possible.

j2kun:
@AlexanderViand-Intel any opposition to submitting this PR? If you're short on time, we can always chat about it post submission. I think Asra wants use of the polynomial attribute for her bgv dialect.

asraa (approved):

j2kun:
I'm going to make a judgement call and merge this one in. Happy to follow up afterward with any additional comments and improvements.

AlexanderViand-Intel (commented):
I think the polynomial attribute itself looks great, I'm not so sure about the `!poly.poly` type, though.  I don't think I agree with having this be completely parameter-less with a ring attribute on the operation makes sense here. This MLIR/LLVM idiom of "operations specify type semantics" seems to be rooted in a pretty low-level view of computation, where a lot of data is basically just 32/64/word size bits and individual CPU instructions will in fact happily take these bits and re-interpret them.

In the FHE world, however, there's a massive difference between a polynomial from a degree 64k, 600-bit cmod ring and one from an n=2k, 32-bit ring, and "instructions" (both software and hardware implementations of, e.g., NTT) can't just ignore these differences.

AlexanderViand-Intel:
Sorry for the late response, I was at USENIX all week and forgot to submit a review beforehand, I submitted it now (see above).

include/Dialect/Poly/IR/PolyTypes.td
```diff
@@ -2,14 +2,11 @@
 #define HEIR_INCLUDE_DIALECT_POLY_IR_POLYTYPES_TD_
 
 include "PolyDialect.td"
+include "PolyAttributes.td"
 
 include "mlir/IR/DialectBase.td"
 include "mlir/IR/AttrTypeBase.td"
 
-//===----------------------------------------------------------------------===//
-// Poly type definitions
```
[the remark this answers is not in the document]
j2kun: It was mostly to keep the PR small enough. I think now that I'm deep in the dialect conversion framework, I see that, at the very least, we need the size of the underlying storage type and the coefficient bit width available in the type in order to use the TypeConverter.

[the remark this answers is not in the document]
AlexanderViand-Intel: Mh, I'm not sure we want details of the TypeStorage to leak into the IR, rather the coefficient bit width should determine (in some implementation defined way) how the TypeStorage stores this.  My motivation for this is to have a clear separation between "MLIR the concept" and "MLIR the C++ code base", making sure the things we standardized depend only on the former (of course, with a reference implementation using the latter).

In order to make the TypeConverter work, you could add a `storageType()` method to the (C++ implementation of) the type. Alternatively, the type could just figure this out internally in it's `create`/`build`/constructor, especially if we end up using something like `MPint` anyway?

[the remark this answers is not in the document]
AlexanderViand-Intel: I see! I guess in an ideal world we'd want some kind of `Z_q` type for the second part (i.e., instead of `i32`)? Sort of a `baseQ` (taking inspiration from the `base2` dialect proposed by @KFAFSP)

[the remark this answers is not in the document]
j2kun: I think the polynomial attribute in this PR would be sufficient to put on the type. And then we just mandate that each polynomial type is associated with a specific ring, and if we want to switch rings we have to add some special ops for that.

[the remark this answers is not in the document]
KFAFSP: For example, `base2.fixed<signed 32,16>` (alias `base2.si32_16`) is only usable with a compile-time fixed precision. Is this desired here, or do you need a type to model $\mathbb{Q}$ instead of $\mathbb{Z}_N \cdot 2^{-E}$?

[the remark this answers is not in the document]
AlexanderViand-Intel: > I think the polynomial attribute in this PR would be sufficient to put on the type. And then we just mandate that each polynomial type is associated with a specific ring, and if we want to switch rings we have to add some special ops for that.

I'm not sure I'm following. In my understanding, an attribute is a compile-time value (usually closely associated with a type, with a `dialect.constant` operation that "turns" an attribute into a value of that type). Having a `#poly.polynomial` attribute as a type parameter for the `!poly.polynomial` type would be parameterizing the type by a specific polynomial? Sorry if I'm misunderstanding things here!  I guess we could parameterize the `!poly.polynomial` type by `#poly.ring` if that's what you meant?

PR #78, opened 2023-07-31
Add Secret dialect

j2kun:
This PR implements a `secret` dialect to represent generic, high level computation on encrypted data, along with a demonstration lowering pass that drops the secret types.

It defines a single new type, `secret.secret<AnyType>` which wraps a non-secret type, and centers around the `secret.generic` op, which "lifts" a computation on the plaintext types to a computation on the corresponding secret types, with some metadata attached to the high level operation encapsulated by a `secret.generic` so that it can be lowered to custom ops in specific dialects.

I also defined a pass, `--secret-forget-secrets`, which simply drops the secrets from all types and returns an equivalent program that operates on plaintext data.

Some notes:

- In my examples I have `secret.generic`'s body containing a single op, but I think more realistically we will have front ends that convert large regions of code into a large `secret.generic`, and then passes in this dialect that transform it by doing things like extracting loops with static bounds outside of the body of a `secret.generic` and only having the inner-most loop body contain `secret.generic` ops (which could then be further analyzed/optimized with partial loop unrolling and fusing), or something like transforming the body of a `secret.generic` into a combinational circuit, then applying a circuit optimizer to it.
- Forgetting secrets is only intended to be useful for testing, e.g, to ensure a program is functionally equivalent before and after any FHE passes are applied. But it also helps me understand how to write passes that operate on the mildly complex IR I've defined for `secret.generic`.
- I'd like to eventually generalize the forget-secrets pass to operate on an interface(s), and then apply that interface to lower level dialects, so that at any point during a program's compilation lifecycle, we can exit the "encrypted" part of the pipeline and test functional equivalence.

### DIFF
--- a/include/Dialect/Secret/IR/BUILD
+++ b/include/Dialect/Secret/IR/BUILD
@@ -1,0 +1,101 @@
+# Secret, a dialect defining secretnomials in a secretnomial ring and their operations.
+
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "td_library")
+
+package(
+    default_applicable_licenses = ["@heir//:license"],
+    default_visibility = ["//visibility:public"],
+)
+
+exports_files(
+    [
+        "SecretDialect.h",
+        "SecretOps.h",
+        "SecretTypes.h",
+    ],
+)
+
+td_library(
+    name = "td_files",
+    srcs = [
+        "SecretDialect.td",
+        "SecretOps.td",
+        "SecretTypes.td",
+    ],
+    includes = ["@heir//include"],
+    deps = [
+        "@llvm-project//mlir:BuiltinDialectTdFiles",
+        "@llvm-project//mlir:ControlFlowInterfacesTdFiles",
+        "@llvm-project//mlir:InferTypeOpInterfaceTdFiles",
+        "@llvm-project//mlir:OpBaseTdFiles",
+        "@llvm-project//mlir:SideEffectInterfacesTdFiles",
+    ],
+)
+
+gentbl_cc_library(
+    name = "dialect_inc_gen",
+    tbl_outs = [
+        (
+            ["-gen-dialect-decls"],
+            "SecretDialect.h.inc",
+        ),
+        (
+            ["-gen-dialect-defs"],
+            "SecretDialect.cpp.inc",
+        ),
+    ],
+    tblgen = "@llvm-project//mlir:mlir-tblgen",
+    td_file = "SecretDialect.td",
+    deps = [
+        ":td_files",
+    ],
+)
+
+gentbl_cc_library(
+    name = "types_inc_gen",
+    tbl_outs = [
+        (
+            ["-gen-typedef-decls"],
+            "SecretTypes.h.inc",
+        ),
+        (
+            ["-gen-typedef-defs"],
+            "SecretTypes.cpp.inc",
+        ),
+        (
+            ["-gen-typedef-doc"],
+            "SecretTypes.md",
+        ),
+    ],
+    tblgen = "@llvm-project//mlir:mlir-tblgen",
+    td_file = "SecretTypes.td",
+    deps = [
+        ":dialect_inc_gen",
+        ":td_files",
+    ],
+)
+
+gentbl_cc_library(
+    name = "ops_inc_gen",
+    tbl_outs = [
+        (
+            ["-gen-op-decls"],
+            "SecretOps.h.inc",
+        ),
+        (
+            ["-gen-op-defs"],
+            "SecretOps.cpp.inc",
+        ),
+        (
+            ["-gen-op-doc"],
+            "SecretOps.md",
+        ),
+    ],
+    tblgen = "@llvm-project//mlir:mlir-tblgen",
+    td_file = "SecretOps.td",
+    deps = [
+        ":dialect_inc_gen",
+        ":td_files",
+        ":types_inc_gen",
+    ],
+)
--- a/include/Dialect/Secret/IR/SecretOps.td
+++ b/include/Dialect/Secret/IR/SecretOps.td
@@ -1,0 +1,171 @@
+#ifndef HEIR_INCLUDE_DIALECT_SECRET_IR_SECRETOPS_TD_
+#define HEIR_INCLUDE_DIALECT_SECRET_IR_SECRETOPS_TD_
+
+include "SecretDialect.td"
+include "SecretTypes.td"
+include "mlir/IR/BuiltinAttributeInterfaces.td"
+include "mlir/Interfaces/ControlFlowInterfaces.td"
+include "mlir/Interfaces/InferTypeOpInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+class Secret_Op<string mnemonic, list<Trait> traits = []> :
+        Op<Secret_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::secret";
+}
+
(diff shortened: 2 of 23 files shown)


j2kun:
Note to self: check out llvm-project/mlir/lib/Dialect/Linalg/Transforms/Detensorize.cpp as a guiding pass, and use the DialectConversion framework to make the secret type illegal.

j2kun:
There are a handful of TODOs still littered throughout the PR, for improved documentation and my confusion about the dialect conversion framework. But I think now is a good time for an initial review by @asraa and @AlexanderViand-Intel

include/Dialect/Secret/IR/BUILD
```diff
@@ -0,0 +1,97 @@
...
+
+gentbl_cc_library(
+    name = "ops_inc_gen",
+    tbl_outs = [
+        (
+            ["-gen-op-decls"],
+            "SecretOps.h.inc",
+        ),
+        (
+            ["-gen-op-defs"],
```
asraa, line 86: We might want to add `gen-op-doc` - maybe the other dialects don't have it..?

include/Dialect/Secret/IR/SecretOps.td
```diff
@@ -0,0 +1,161 @@
...
+}
+
+def Secret_ConcealOp : Secret_Op<"conceal", [Pure]> {
+  let summary = "Convert a non-secret value into a secret";
+  let description = [{
+    Convert a value to a secret containing the same value.
+
+    This op represents a scheme-agnostic encryption operation, as well as a
+    "trivial encryption" operation which is needed for some FHE schemes. This
+    op is also useful for type materialization in the dialect conversion
```
asraa, line 23: Should we file a follow-up issue for this for discussion to resolve when it feels like we know? Because in terms of lowering a noiseless vs true encryption will end up different - but maybe the case is that a lowering would be able to understand which is which depending on context? (if a secret key is not available it must be a trivial encryption)

include/Dialect/Secret/IR/SecretOps.td
```diff
@@ -0,0 +1,161 @@
...
+  let assemblyFormat = "$cleartext attr-dict `:` type($cleartext) `->` type($output)";
+
+  let builders = [
+    // Builder to infer output type from the input type
+    OpBuilder<(ins "Value":$cleartext)>
+  ];
+
+  // TODO: add a folder that collapses secret<secret<T>> into secret<T>
+  // and removes intermediate conceal ops.
+  // let hasFolder = 1;
```
asraa, line 44: Could be a fairly easy follow-up

include/Dialect/Secret/IR/SecretOps.td
```diff
@@ -0,0 +1,161 @@
...
+  Arguments<(ins Variadic<AnyType>:$values)> {
+  let summary = "Secret yield operation";
+  let description = [{
+    `secret.yield` is a special terminator operation for blocks inside regions
+    in `secret` generic ops. It returns the cleartext value of the
+    corresponding private computation to the immediately enclosing `secret`
+    generic op.
+  }];
+  let builders = [OpBuilder<(ins), [{ /* nothing to do */ }]>];
+  let hasCustomAssemblyFormat = 1;
```
asraa, line 83: this might be a candidate for a verifier (i remember when we were pair programming)

include/Dialect/Secret/IR/BUILD
j2kun: Done.

include/Dialect/Secret/IR/SecretOps.td
j2kun: I think this dialect will probably change significantly before we can do that, e.g., by adding information about which secret key something is encrypted wrt, or equivalently a client/server distinction as to what is private to whom. Let's move that to https://github.com/google/heir/discussions/96 maybe.

include/Dialect/Secret/IR/SecretOps.td
j2kun: Added https://github.com/google/heir/issues/108 and TODO refs

include/Dialect/Secret/IR/SecretOps.td
j2kun: Added to the TODOs and a note about SingleBlockImplicitTerminator

j2kun:
I think there is some benefit to having this checked in even if we think doing it in full generality is too ambitious. It will be useful for some experimentation and @asraa and I are interested in trying it out as an entry point for some ML-specific dialects.

include/Dialect/Secret/IR/SecretOps.td
```diff
@@ -0,0 +1,171 @@
...
+        %z = arith.addi %x, %y: i32
+        secret.yield %z : i32
+      } -> (!secret.secret<i32>)
+    ```
+
+    A `secret.generic` op also has a corresponding `library_name` attribute
+    that provides semantic information about what op is computed inside the
+    region. This can be used by lowering passes to target specific
+    instructions in lower level dialects that support them. For example, an FHE
+    dialect may have special support for an encrypted matmul op.
```
AlexanderViand-Intel, line 132: I have a feeling the `library_name` concept is a bit too low-level for such a high-level concept. 
I'd expect the `fhe`/`enc_linalg`/`enc_arith` level dialect's `generic` to have this, but not `secret` (if we are indeed adding it "on top" of the encrypted arith abstraction)

include/Dialect/Secret/Transforms/Passes.td
```diff
@@ -0,0 +1,15 @@
+#ifndef INCLUDE_DIALECT_SECRET_TRANSFORMS_PASSES_TD_
+#define INCLUDE_DIALECT_SECRET_TRANSFORMS_PASSES_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def SecretForgetSecrets : Pass<"secret-forget-secrets"> {
```
AlexanderViand-Intel, line 6: "Forgetting" secrets to me sounds like something that removes secret values, rather than the secret annotation. How about naming this "drop-secret-status" or something along those lines?

AlexanderViand-Intel (approved):
Only minor comments

include/Dialect/Secret/Transforms/Passes.td
j2kun: I was thinking of the ["forgetful functor"](https://en.wikipedia.org/wiki/Forgetful_functor) but maybe that is too big of a stretch.

I like the slightly more specific "drop-secret-to-cleartext" or "interpret-secret-as-cleartext", will update it on Monday.

j2kun:
Something got messed up with the internal integration of this PR, so I'm going to cut an internal patch which will show up as a PR with some weird changes in it (including the changes for Alex's two comments)

include/Dialect/Secret/IR/SecretOps.td
asraa: I would have thought that `secret` is actually a higher level than `fhe`?

j2kun: I think what Alex is saying makes sense, in that a program using `secret` would be wrapping a named op already (e.g., `secret` wrapping `tosa.matmul`) and that in the lowering to lower-level dialects you'd take the wrapped op and use it for `library_name` on a region of ops in `fhe_math` that implements something manually.

asraa: AH, I see. I'm in agreement, that's right.

j2kun:
Manually merged as https://github.com/google/heir/commit/69db9f04991fdc98c12ab39d08e3480adfc6629c

PR #84, opened 2023-08-03
add CI check for a single commit

j2kun:


.github/workflows/pr_require_single_commit.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # pin@v3
+
+      - name: Assert single commit
+        run: |
+          commits=${{ github.event.pull_request.commits }}
+          if (( ${commits} > 1 )); then
+            echo "Expected PR to have a single commit, but found ${commits}; google/heir requires a single commit per PR because Google's internal/external synchronization tooling does not support squashing, and instead would rebase all commits into the main branch."
```
asraa, line 18: Try this for workflow errors! https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error

asraa (approved):

.github/workflows/pr_require_single_commit.yml
asraa: oh the `core.setFailed` is from their typescript toolkit, in the workflow bash you can just do `::error::expected PR...`

j2kun: Just went with `error`.

asraa (approved):

Issue #76, opened 2023-07-28
Add pre-commit hook configuration to enforce style guide


j2kun:
https://pre-commit.com/hooks.html

e.g., under [hooks](https://pre-commit.com/hooks.html) it supports stuff like bazel-buildifier, clang-format, etc.

asraa:
We probably also want to make a CI check for this too, just in case pre-commit hooks are faulty or something else happens

j2kun:
According to the docs on the pre-commit site, all you have to do is register a workflow with a canned action and it will automatically pull in the precommit config and check them.

PR #87, opened 2023-08-04
chore: add pre-commit hooks and presubmit check

asraa:
Depends on https://github.com/google/heir/pull/86

* Adds pre-commit hook with clang-tidy, starlark, and basic checks to start.
* Added developer documentation to run and setup
* Add github action to run the pre-commit hooks as well

Fixes https://github.com/google/heir/issues/76


.github/workflows/docs.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     # disk.
     - name: "Build markdown files from tblgen sources"
       run: |
-        bazel build $(bazel query 'siblings(kind("gentbl_rule", @heir//...))' | grep "_filegroup$")
+        bazel build "$(bazel query 'siblings(kind("gentbl_rule", @heir//...))' | grep "_filegroup$")"
```
j2kun, line 29: This looks like it's broken because of the nested double quotes. It also seems to break https://github.com/google/heir/actions/runs/5764646709/job/15629048762?pr=87

.github/workflows/pr_style.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # pin@v3
+
+      - name: Assert single commit
+        env:
+          commits: ${{ github.event.pull_request.commits }}
+        run: |
+          if (( commits > 1 )); then
```
j2kun, line 18: Does `commits` need a `$` here?

j2kun (approved):
LGTM! Just a few nits

.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+    -   id: check-yaml
+    -   id: check-added-large-files
+        args: ['--maxkb=500000'] # 500MB
+
+# clang-format
+-   repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: "v16.0.6"
+    hooks:
+    -   id: clang-format
+        args: ["--style=Google"]
```
j2kun, line 20: I feel like if we're hoping to upstream some of the stuff in here, we may want to make it --style=LLVM? Or just have this default to using a .clang-format config file in the project root (see https://github.com/google/heir/pull/89 for a proposed initial config)

asraa: Right! I wanted to but copybara doesn't apply custom clangtidy rules (that I know of) so this might cause a breakage in the exported files.

.github/workflows/docs.yml
asraa: Ah, I see. I just rewrote the grep to be a bazel query filter and piped those into `bazel build`. shellcheck wasn't happy with the original being unquoted.

.pre-commit-config.yaml
asraa: Let's follow up with this on your #89

PR #88, opened 2023-08-07
bgv: add a bgv dialect definition

asraa:
This is a first iteration of a BGV dialect, with a few follow-ups posted as issues (that would benefit from individual PR discussion).

The BGV ciphertext includes:
* The number of polynomials, or "dim" of the ciphertext. (this may change after an operation like multiplication or relinearize). default is 2.
* The **list** of rings of the scheme that the BGV ciphertext was instantiated with. An operation that changes these rings, like `modulus_switch` will reference the index from which and to which the operation moves the ciphertext.

Noteworthy things:
* The `key_basis` attribute describes the key basis relative to which a ciphertext is encrypted. A basis of `(1, s)` is represented with `[0, 1]` - these correspond to the exponents of the secret key. After a rotation, it may be `[0, 2`] representing `(1, s^2)`. Or it may be `[0, 1, 2]` representing `(1, s, s^2)` after a `mul`.
* `bgv.relinearize` generalizes relinearization from galois automorphisms and from multiplications, the `from_basis` and `to_basis` describe the key basis change.
* With this particular approach, parameter selection would need to occur while lowering from FHE. It is possible that we could generalize and unspecialize the rings if needed. But that can be a follow-up. Or perhaps an parameter tuning pass can simply just make the params better later.

There are follow-up issues and `TODO`s in the code posted for:
* Adding an attribute for underlying plaintext bits to help with analysis
* Adding a `Plaintext` type and ops
* Adding verifiers / validations for ops.

### DIFF
--- a/include/Dialect/BGV/IR/BGVOps.td
+++ b/include/Dialect/BGV/IR/BGVOps.td
@@ -1,0 +1,136 @@
+#ifndef HEIR_INCLUDE_DIALECT_BGV_IR_BGVOPS_TD_
+#define HEIR_INCLUDE_DIALECT_BGV_IR_BGVOPS_TD_
+
+include "BGVDialect.td"
+include "BGVTypes.td"
+
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/InferTypeOpInterface.td"
+
+// TODO(https://github.com/google/heir/issues/101): Add verifiers for BGV ops.
+
+class BGV_Op<string mnemonic, list<Trait> traits = []> :
+        Op<BGV_Dialect, mnemonic, traits> {
+
+  // See https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-format
+  // TODO(https://github.com/google/heir/issues/106): Simplify this format by adding type constraints.
+  let assemblyFormat = [{
+    `(` operands `)` attr-dict `:`  `(` type(operands) `)` `->` type(results)
+  }];
+  let cppNamespace = "::mlir::heir::bgv";
+}
+
+// TODO(https://github.com/google/heir/issues/100): Add plaintext-ciphertext operations.
+
+def BGV_AddOp : BGV_Op<"add", [Commutative, SameOperandsAndResultType]> {
+  let summary = "Addition operation between ciphertexts.";
+
+  let arguments = (ins
+    Ciphertext:$x,
+    Ciphertext:$y
+  );
+
+  let results = (outs
+    Ciphertext:$output
+  );
+
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
+}
+
+def BGV_SubOp : BGV_Op<"sub", [SameOperandsAndResultType]> {
+  let summary = "Subtraction operation between ciphertexts.";
+
+  let arguments = (ins
+    Ciphertext:$x,
+    Ciphertext:$y
+  );
+
+  let results = (outs
+    Ciphertext:$output
+  );
+
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
+}
+
+def BGV_MulOp : BGV_Op<"mul", [Commutative]> {
+  let summary = "Multiplication operation between ciphertexts.";
+
+  let arguments = (ins
+    Ciphertext:$x,
+    Ciphertext:$y
+  );
+
+  let results = (outs
+    Ciphertext:$output
+  );
+
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type($x)`,` type($y) `->` type($output)" ;
+}
+
+def BGV_Rotate : BGV_Op<"rotate"> {
+  let summary = "Rotate the coefficients of the ciphertext using a Galois automorphism.";
+
+  let arguments = (ins
+    Ciphertext:$x,
+    I64Attr:$offset
+  );
+
+  let results = (outs
+    Ciphertext:$output
+  );
+}
+
+def BGV_Negate : BGV_Op<"negate", [SameOperandsAndResultType]> {
+  let summary = "Negate the coefficients of the ciphertext.";
+
+  let arguments = (ins
+    Ciphertext:$x
+  );
+
+  let results = (outs
+    Ciphertext:$output
+  );
+}
+
+def BGV_Relinearize : BGV_Op<"relinearize"> {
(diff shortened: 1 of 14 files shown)


AlexanderViand-Intel:
> * The ciphertext itself IMO only needs to have the degree (e.g. is it encrypted with `(1, s, s^2)` or `(1, s)`) and the modulus level (which can be an index in the range of the chosen leveled moduli of the scheme). The rest of the ring parameters can be module level.

Just some super quick feedback (on a plane with poor wifi, so didn't look at the code yet) on the naming: I think "degree" might be confusing, as I usually understand that to mean the polymod degree _n_ (4096, 8k, 16k, etc). Of course, the way you're using it also makes sense, and I think calling the number of polynomials in a Ciphertext it's "size" is equally confusing, just in the opposite direction (assuming batching where the number of slots is equal to _n_). Sorry for the not-so-helpful feedback ;)

AlexanderViand-Intel:
> * The ciphertext itself IMO only needs to have the degree (e.g. is it encrypted with `(1, s, s^2)` or `(1, s)`) and the modulus level (which can be an index in the range of the chosen leveled moduli of the scheme). The rest of the ring parameters can be module level.

Just some super quick feedback (on a plane with poor wifi, so didn't look at the code yet) on the naming: I think "degree" might be confusing, as I usually understand that to mean the polymod degree _n_ (4096, 8k, 16k, etc). Of course, the way you're using it also makes sense, and I think calling the number of polynomials in a Ciphertext it's "size" is equally confusing, just in the opposite direction (assuming batching where the number of slots is equal to _n_). Sorry for the not-so-helpful feedback ;)

j2kun:
If there are only the three options, I feel like an enum attr would be better than an integer. It could be called EncryptionType or RelinearizationPhase, or something like that.

include/Dialect/BGV/IR/BGVOps.td
```diff
@@ -0,0 +1,166 @@
...
+  );
+
+  let results = (outs
+    Ciphertext:$output
+  );
+
+   let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
+}
+
+def BGV_MulOp : BGV_Op<"mul", [Commutative, SameOperandsAndResultType]> {
```
j2kun, line 59: Quick note: if you have the (1, s) vs (1, s, s^2) attached to the type, then you can't use SameOperandsAndResultType. The different attributes, I believe, make the types different in the context of this trait.

asraa: Ah good point! I think it's useful then to drop the attribute - mul would change the ciphertext attributes.

asraa:
Yes! Completely agree with this feedback about the terminology "degree".

To me, `size` would also be confusing in the case of the galois automorphism / rotation applied, since the "Size" is still 2 there, but the power of s is distinct. "Phase" directly would seem like the encoded plaintext result. 

> If there are only the three options, I feel like an enum attr would be better than an integer. It could be called EncryptionType or RelinearizationPhase, or something like that.

There's not only three options, but rather any amount, we could multiply twice without relinearization and get `(1, s, s^2, s^3)` for e.g., and can get any `(1, s^i)` depending on the rotation applied.

> EncryptionType or RelinearizationPhase

These are good options, the canonical form is always `(1, s)`. We could then have an optional attr (a tensor of monomials)  called `KeyPhase`?  OR we could consider it as a "basis" for the ciphertext - so it could be `RelinearizationBasis` or `KeyBasis`?

AlexanderViand-Intel:
> the canonical form is always `(1, s)`. We could then have an optional attr (a tensor of monomials) called `KeyPhase`? OR we could consider it as a "basis" for the ciphertext - so it could be `RelinearizationBasis` or `KeyBasis`?

I really like the idea of a "KeyBasis", as this could also be used in multi-key settings! I also agree it should be optional, so we don't clutter up the textual representation and API.

asraa:
Some quick updates:
* I've placed the ciphertext "dimension" (number of polys / size) as a type attribute: this is because it can't really be inferred from ops, e.g. when it's a block argument, and it seems necessary to have when lowering to a specific sized tensor of polynomials

* Perhaps a ring specialization also needs to occur for the same reason that poly requires a ring on its type? originally i thought it would be fine as a module level attribute, but it may need to be on the type to lower the type to the tensor of polys in that ring @j2kun @AlexanderViand-Intel 

* Relinearization now takes a `from_basis` and `to_basis` attributes describing the key basis transformation. I'm not totally convinced they need to be type attr's. It may if it follows the same block argument logic, but at the same time we can always assume it's relative to the original args (the block args are encrypted canonically)

include/Dialect/BGV/IR/BGVOps.td
```diff
@@ -0,0 +1,142 @@
...
+    Ciphertext:$output
+  );
+}
+
+def BGV_Relinearize : BGV_Op<"relinearize"> {
+  let summary = "Relineraize the ciphertext.";
+
+  let arguments = (ins
+    Ciphertext:$x,
+    DenseI32ArrayAttr:$from_basis,
```
j2kun, line 112: I feel like the semantics of the basis could use a bit more documentation, at least to explain how `(1, s, s^2)` corresponds to the attribute `[0, 1, 2]` while a `(1, s^2)` would correspond to `[0, 2]`

j2kun (approved):
Only nits, otherwise LGTM!

include/Dialect/BGV/IR/BGVOps.td
```diff
@@ -0,0 +1,142 @@
...
+
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type($x)`,` type($y) `->` type($output)" ;
+}
+
+//===----------------------------------------------------------------------===//
+// Ring operations.
+//===----------------------------------------------------------------------===//
+
+def BGV_Rotate : BGV_Op<"rotate"> {
+  let summary = "Rotate the coefficients of the ciphertext using the Galois Key.";
```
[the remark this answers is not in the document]
asraa: Ah! An artifact from earlier. Galois keys are just special instances of relinearization keys, so this is addressed by a relinearize op.

include/Dialect/BGV/IR/BGVOps.td
asraa: added a description body to this op

asraa:
Comments addressed!

PR #89, opened 2023-08-08
Add clang-tidy and clang-format configs

j2kun:
I pulled these from LLVM/MLIR's configs.


.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: LLVM
```
asraa, line 1: Hey, I left this on another note: the reason I didn't include this is because copybara from internal can't apply LLVM clang-tidy, so this will cause a mismatch on what's exported from internal.

PR #94, opened 2023-08-14
Improve poly types and add a starter lowering pass

j2kun:
This PR makes slight improvements to the poly types by adding a ring attribute, makes small modifications to the poly ops by changing them to binary ops instead of variadic, and then implements the shell of a `lower-poly` pass that will lower to linalg + standard MLIR, with an eye towards lowering fully to LLVM.

I want to get this incomplete shell checked in to ease the review burden, and start collaborating with the details of the pass with other contributors.

### DIFF
--- a/include/Conversion/PolyToStandard/PolyToStandard.td
+++ b/include/Conversion/PolyToStandard/PolyToStandard.td
@@ -1,0 +1,19 @@
+#ifndef INCLUDE_CONVERSION_POLYTOSTANDARD_POLYTOSTANDARD_TD_
+#define INCLUDE_CONVERSION_POLYTOSTANDARD_POLYTOSTANDARD_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def PolyToStandard : Pass<"lower-poly"> {
+  let summary = "Lower `poly` to standard MLIR dialects.";
+
+  let description = [{
+    This pass lowers the `poly` dialect to standard MLIR, a mixture of affine,
+    tensor, and arith.
+  }];
+  let dependentDialects = [
+    "mlir::arith::ArithDialect",
+    "mlir::heir::poly::PolyDialect",
+  ];
+}
+
+#endif  // INCLUDE_CONVERSION_POLYTOSTANDARD_POLYTOSTANDARD_TD_
--- a/include/Dialect/Poly/IR/PolyOps.td
+++ b/include/Dialect/Poly/IR/PolyOps.td
@@ -5,6 +5,7 @@ include "PolyAttributes.td"
 include "PolyDialect.td"
 include "PolyTypes.td"
 include "mlir/Interfaces/InferTypeOpInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
 
 //===----------------------------------------------------------------------===//
 // Poly Operation definitions.
@@ -12,7 +13,6 @@ include "mlir/Interfaces/InferTypeOpInterface.td"
 
 class Poly_Op<string mnemonic, list<Trait> traits = []> :
         Op<Poly_Dialect, mnemonic, traits> {
-
   // See https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-format
   // TODO(b/291601096): Simplify this format by adding type constraints.
   let assemblyFormat = [{
@@ -25,60 +25,35 @@ class Poly_Op<string mnemonic, list<Trait> traits = []> :
 // Ring operations.
 //===----------------------------------------------------------------------===//
 
-def Poly_AddOp : Poly_Op<"add", [SameOperandsAndResultType]> {
+class Poly_BinOp<string mnemonic, list<Trait> traits = [Pure, SameOperandsAndResultType, Commutative]> :
+        Poly_Op<mnemonic, traits> {
+  let arguments = (ins Polynomial:$lhs, Polynomial:$rhs);
+  let results = (outs Polynomial:$output);
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
+}
+
+def Poly_AddOp : Poly_BinOp<"add"> {
   let summary = "Addition operation between polynomials.";
-
-  let arguments = (ins
-    Variadic<Poly>:$x
-  );
-
-  let results = (outs
-    Poly:$output
-  );
-
-   let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
 }
 
-def Poly_SubOp : Poly_Op<"sub", [SameOperandsAndResultType]> {
+def Poly_SubOp : Poly_BinOp<"sub"> {
   let summary = "Subtraction operation between polynomials.";
-
-  let arguments = (ins
-    Variadic<Poly>:$x
-  );
-
-  let results = (outs
-    Poly:$output
-  );
-
-  let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
 }
 
-def Poly_MulOp : Poly_Op<"mul", [SameOperandsAndResultType]> {
+def Poly_MulOp : Poly_BinOp<"mul"> {
   let summary = "Multiplication operation between polynomials.";
-
-  let arguments = (ins
-    Poly:$lhs,
-    Poly:$rhs,
-    Ring_Attr:$ring
-  );
-
-  let results = (outs
-    Poly:$output
-  );
-
-  let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
 }
 
 def Poly_MulConstantOp : Poly_Op<"mul_constant"> {
   let summary = "Multiplication by a constant of the field.";
 
   let arguments = (ins
-    Poly:$x,
+    Polynomial:$x,
     AnyInteger:$constant
   );
 
   let results = (outs
-    Poly:$output
+    Polynomial:$output
   );
 }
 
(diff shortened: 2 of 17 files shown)


j2kun:
`poly.constant` is a good idea, but I believe we'll need the conversion ops to/from tensor for dialect conversion materialization, assuming we lower `poly.poly` to `tensor.tensor`.

j2kun:
@AlexanderViand-Intel filed https://github.com/google/heir/issues/97 for your suggestion about poly.constant

include/Conversion/PolyToStandard/PolyToStandard.td
```diff
@@ -0,0 +1,21 @@
...
+  let summary = "Lower `poly` to standard MLIR dialects.";
+
+  let description = [{
+    This pass lowers the `poly` dialect to standard MLIR, a mixture of affine,
+    tensor, and arith.
+  }];
+  let dependentDialects = [
+    "mlir::arith::ArithDialect",
+    "mlir::heir::poly::PolyDialect",
+    "mlir::linalg::LinalgDialect",
```
asraa, line 16: do we expect linalg to surface in the result? is dependent dialects necessary dependents or signaling the dialects that will be surfaced from the pass?

asraa (approved):

include/Conversion/PolyToStandard/PolyToStandard.td
j2kun: Removed the non-arith dialects from this list just in case we don't end up using them

include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -101,17 +75,32 @@ def Poly_GetCoeffOp : Poly_Op<"get_coeff"> {
...
   );
 
   let results = (outs
     TensorOf<[AnyInteger]>:$output
   );
+
+  let builders = [
+    // Builder to infer output type from the input type
+    OpBuilder<(ins "::mlir::Value":$input, "::mlir::Value":$i, "::mlir::Value":$j)>
+  ];
```
AlexanderViand-Intel, line 103: What about using InferTypeOpInterface instead? That way, you only have to implement the actual type inference, with the builder(s) still being auto generated? See https://github.com/MarbleHE/HECO/blob/main/include/heco/IR/FHE/FHE.td#L135-L138

lib/Dialect/Poly/IR/PolyOps.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+void PolyFromCoeffsOp::build(OpBuilder &builder, OperationState &result,
+                             Value input, RingAttr ring) {
+  TensorType tensorType = dyn_cast<TensorType>(input.getType());
+  APInt cmod(APINT_BIT_WIDTH, 1);
+  cmod = cmod << tensorType.getElementTypeBitWidth();
+  Type resultType = PolynomialType::get(builder.getContext(), ring);
+  build(builder, result, resultType, input);
+}
```
AlexanderViand-Intel, line 25: What about using InferTypeOpInterface instead? That way, you only have to implement the actual type inference, with the builder(s) still being auto generated? See https://github.com/MarbleHE/HECO/blob/main/include/heco/IR/FHE/FHE.td#L135-L138

tests/poly/lower_poly.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: heir-opt --lower-poly %s > %t
+// RUN: FileCheck %s < %t
+
+#cycl_2048 = #poly.polynomial<1 + x**1024>
+#ring = #poly.ring<cmod=4294967296, ideal=#cycl_2048>
+module {
+  func.func @test_lower_from_coeffs() {
+    %c0 = arith.constant 0 : index
+    // 2 + 2x + 5x^2
+    %coeffs = "arith.constant"() <{value = dense<[2, 2, 5]> : tensor<3xi32>}> : () -> tensor<3xi32>
```
AlexanderViand-Intel, line 10: Why the generic form of arith.constant here?

AlexanderViand-Intel (changes requested):
Generally looks good, just a lot of nitpicking comments from my side.

The only real concern I have is deleting an op that was just DialectConvert'ed - I'm not sure that's always safe to do?

AlexanderViand-Intel:
Reading the [MLIR newsletter](https://discourse.llvm.org/t/mlir-news-53th-edition-16th-august-2023/72805) made me aware of `ConvertToLLVMPatternInterface` and I'd recommend we add it to the `poly` dialect :)

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -0,0 +1,156 @@
...
+      : OpConversionPattern<PolyFromCoeffsOp>(context) {}
+
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      PolyFromCoeffsOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    // the input tensor replaces the op
+    rewriter.replaceAllUsesWith(op.getResult(), adaptor.getOperands()[0]);
+    rewriter.eraseOp(op);
```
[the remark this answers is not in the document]
j2kun: Your intuition is right, and I've been wrestling with this for a few days now. See https://reviews.llvm.org/D158107

The general advice in the docs is: any mutations to the IR just need to go through the ConversionPatternRerwiter, but not all the rewriter methods are handled properly (hence the bug with the diff above). It should just be rewriter.replaceOp().

The need for these conversions is I suppose a matter of minute details. If you do a full conversion, then the pass will fail on the first op it can't legalize (and if you call the entire dialect illegal then it would fail on these ops). If you do a partial conversion you can get by it, but then it will insert `unrealized_conversion_cast` ops which, at the end of the pass, are replaced with custom materializations if they can't be folded.

I admit I'm still learning this whole framework, but I think the most straightforward way to do it is to have a converter for every op and not include any custom materializations and do a full conversion. You would only worry about custom materializations if the pass gets very complicated (like bufferization) and you can then split it into multiple passes, each a partial conversion, with custom materializations to keep the IR consistent across the passes, which is the advice recommended in [this talk](https://drive.google.com/file/d/1VfVajitgf8ZPnd-HRkJvaJiFLhBsluXN/view)

I will go ahead and update the pattern above to use the correct rewriter method (I only learned about this discrepancy yesterday)

j2kun:
> Reading the [MLIR newsletter](https://discourse.llvm.org/t/mlir-news-53th-edition-16th-august-2023/72805) made me aware of `ConvertToLLVMPatternInterface` and I'd recommend we add it to the `poly` dialect :)

Filed https://github.com/google/heir/issues/103

tests/poly/lower_poly.mlir
j2kun: My LSP recommended it as a replacement for the sequence of ops I had there before (arith.constant for each individual value + tensor.from_elements). I wasn't sure what the right syntax was for an `arith.constant` with an attribute, since there's no argument there. Will simplify it.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -0,0 +1,156 @@
...
+#define GEN_PASS_DEF_POLYTOSTANDARD
+#include "include/Conversion/PolyToStandard/PolyToStandard.h.inc"
+
+static Value materializeSource(OpBuilder &builder, Type type, ValueRange inputs,
+                               Location loc) {
+  assert(inputs.size() == 1);
+  auto inputType = inputs[0].getType();
+  if (isa<PolynomialType>(inputType))
+    llvm_unreachable(
+        "A polynomial should never be the input to a materializeSource.");
```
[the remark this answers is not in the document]
j2kun: After learning more about the dialect conversion stuff, I think I can actually remove all these custom materializations, are they're only needed if we expect unrealized materializations to persist past the end of the pass. So I will go ahead and remove them and add a comment.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  auto polyType = dyn_cast<PolynomialType>(inputType);
+  uint32_t idealDegree = polyType.getRing().ideal().getDegree();
+  auto startOp =
+      builder.create<arith::ConstantIntOp>(loc, 0, builder.getIndexType());
+  auto endOp = builder.create<arith::ConstantIntOp>(loc, idealDegree,
+                                                    builder.getIndexType());
+  return builder.create<ExtractSliceOp>(loc, inputs[0], startOp.getResult(),
+                                        endOp.getResult());
```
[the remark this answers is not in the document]
j2kun: As mentioned above, I'll remove the materializations anyway, so that will be unrelated to this PR.

But while I'm not opposed to adding poly.to_coeffs, I guess I'm not clear on the best practices around having ops that are specializations or other ops. Wouldn't we want to canonicalize to_coeffs as an extract_slice, and write all the patterns in terms of extract_slice as the most general case? Or have additional patterns for both of the ops separately, so that it's easier to write patterns on `to_coeffs` as you mentioned.

lib/Dialect/Poly/IR/PolyOps.cpp
j2kun: Just learning about this now. Will add it.

j2kun: On closer inspection, PolyFromCoeffsOp needs the special builder because the ring cannot be inferred from the tensor alone. I migrated ExtractSliceOp, though.

PR #119, opened 2023-08-21
Add a CGGI dialect

j2kun:
This PR implements an initial CGGI dialect.

Blocked by https://github.com/google/heir/pull/165


j2kun:
Update: reimplementing this dialect using the LWE and RLWE attributes on tensors, rather than custom types and attributes.

Issue #97, opened 2023-08-15
Add poly.constant


j2kun:
From https://github.com/google/heir/pull/94#issuecomment-1679029556

> we should add a poly.constant op to directly convert a #poly.polynomial to a !poly.poly without the need to go via tensors

The idea is to have an op that is analogous to `arith.constant`, which, by the use of an array attribute, can populate a tensor of constants, not just a single value, as in

```
%coeffs1 = "arith.constant"() <{value = dense<[2, 2, 5]> : tensor<3xi32>}> : () -> tensor<3xi32>
```

We could have something similar like

```
#cycl_2048 = #poly.polynomial<1 + x**1024>
#ring = #poly.ring<cmod=4294967296, ideal=#cycl_2048>
%poly1 = "poly.constant"() <{coefficients = dense<[2, 2, 5]> : !poly.poly<#ring> : () -> !poly.poly<#ring>
```

The attributes are already implemented, and so we would just need the op and lowerings.

ljhsiun2:
`%poly1 = "poly.constant"() <{coefficients = dense<[2, 2, 5]> : !poly.poly<#ring> : () -> !poly.poly<#ring>`

Noob question, any particular reason the attribute type should be this? I get an error along these lines for a little prototype.

`elements literal must be a shaped type`

j2kun:
The form I gave above is the "generic" form of the IR, and in that case I think you would need to annotate the `dense` with the type (`dense<[2, 2, 5]> : tensor<3xi32>`) since the same `dense` construction is used to define multi-dimensional arrays as attributes.

See, e.g., https://github.com/llvm/llvm-project/blob/4d1d8a885df292b23160b8998061c9fe17a0d677/mlir/test/Target/Cpp/common-cpp.mlir#L11

I think `arith.constant` allows you to infer that from the overall op's result type, meaning you would only have to write

```
%coeffs1 = arith.constant <dense<[2, 2, 5]>> : tensor<3xi32>
```

But in our case it might be clearer to annotate the attribute to make it explicit. If the output polynomial's cmod is not a power of two, the inference to determine what type the input coefficients are would require some logic (find the next power of two, say) which could be surprising.

j2kun:
Oh, I see my original comment also had a syntax error. I think it should be

```
%poly1 = "poly.constant"() <{coefficients = dense<[2, 2, 5]> : tensor<3xi32>}> : !poly.poly<#ring> : () -> !poly.poly<#ring>
```

[...]

PR #123, opened 2023-08-23
Initial glue for poly.constant (iss #97)

ljhsiun2:
Test currently doesn't work (as proposed in #97, or I'm writing it wrong), need to figure out if a custom parser is needed.

Glue is currently just a placeholder to get things to compile. 

CC @j2kun

### DIFF
--- a/include/Dialect/Poly/IR/PolyOps.td
+++ b/include/Dialect/Poly/IR/PolyOps.td
@@ -108,4 +108,20 @@ def Poly_PolyFromCoeffsOp : Poly_Op<"from_coeffs", [Pure]> {
   ];
 }
 
+def Poly_PolyFromConst : Poly_Op<"constant"> {
+  let summary = "Creates a Polynomial directly from coefficients without tensor storing.";
+  let arguments = (ins DenseI32ArrayAttr:$coefficients);
+  let results = (outs Polynomial:$output);
+
+  // let properties = (ins
+  //   DenseI32ArrayAttr:$coefficients
+  // );
+  let builders = [
+    // Builder that infers coefficient modulus from tensor bit width,
+    // and uses whatever input ring is provided by the caller.
+    OpBuilder<(ins "::mlir::Value":$coefficients, "RingAttr":$ring)>
+  ];
+
+}
+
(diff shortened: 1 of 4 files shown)


j2kun:
I put a comment somehwere but now I can't find it so I'll repeat it here: the following patch shows the "right" syntax.

```diff
-    %poly3 = "poly.constant"() <{ coefficients = dense<[2, 2, 5]> : tensor<3xi32> }> : !poly.poly<#ring1> : () -> !poly.poly<#ring1>
+    %poly3 = "poly.constant"() {coefficients = array<i32: 2, 2, 5>} : () -> !poly.poly<#ring1>
```

I had to learn about the difference between [properties (which is new as of 2023)](https://mlir.llvm.org/OpenMeetings/2023-02-09-Properties.pdf) and attributes. I didn't intend to use properites, which is the `<{ }>` notation, while `{ }` is the attribute notation.

I also found that when I switch from properties to attributes as above, but left the value as `dense<[2, 2, 5]>` I got an error that it claimed the value wasn't a dense I32 Array, which is silly. I looked up the source for the DenseI32ArrayAttr and saw that this `array<i32: 2, 2, 5>` was the "generic" form of the attribute, and it looks like other ops that have a simpler format all use a custom parser/printer. We could do that too, if the generic form is too cumbersome.

j2kun:
Also see the very timely PR https://github.com/google/heir/pull/124 which seems to opt us out of a migration to use properties as the default. If you're interested in digging into how to use properties instead of attributes, feel free to flip that flag back after it lands.

ljhsiun2:
Btw work is a bit hectic currently but I am still looking at this in my downtime.

Another noob question-- how can one look at the generated verilog here? Trying to run `--emit-verilog` works for the poly dialect.

j2kun:
Right now the verilog generator is a pretty special purpose pass that we used for some prototype lowerings. To get it to work with poly, we'd need to lower poly to upstream MLIR dialects, which is ongoing work in, e.g., https://github.com/google/heir/pull/134.

j2kun:
This is moving upstream to MLIR, so closing.

Issue #130, opened 2023-08-29
poly: extract_slice `InferTypeOpAdaptor` is wrong


mr0re1:
```mlir
error: 'poly.extract_slice' op inferred type(s) 'tensor<1024xui64, #poly.ring<cmod=2837465, ideal=#poly.polynomial<1 + x**1024>>>' are incompatible with return type(s) of operation 'tensor<2xi32>'
    %slice = poly.extract_slice(%p0, %c0, %c2) : (!poly.poly<#ring1>, index, index) -> tensor<2xi32>
```

It uses `poly.ring.ideal.degree` instead of difference of indexes to deduce shape.

Code: https://github.com/google/heir/blob/main/lib/Dialect/Poly/IR/PolyOps.cpp#L7

mr0re1:
@j2kun  could you please confirm if my assessment is correct?

j2kun:
That sounds right. An output slice should be the same size as the difference of the start/stop indices.

j2kun:
I'll go ahead and fix this one.

j2kun:
Hmmm an interesting twist: in this case extract_slice is given two index SSA values as arguments, which may not have statically known values. Would it make sense then, to use an unranked tensor type?

I'm also considering, for simplicity, changing them to static attributes for now, unless you think we need the variable-slice version for a WIP lowering.

j2kun:
I almost think that it would be easier to just achieve extract_slice via to_tensor and then tensor.extract_slice.

mr0re1:
> I almost think that it would be easier to just achieve extract_slice via to_tensor and then tensor.extract_slice.

+1 , let's avoid complexity and temptation to add more "tensor-like" ops, when it can be delegated to the `tensor` dialect.

j2kun:
I also saw the op was not yet used anywhere, so I'll just remove it.

mr0re1:
We could consider removing `GetCoeffOp` as well, as it can replaced with `to_tensor` + tensor op. Also it doesn't seem to have an intrinsic significance for the `poly` dialect.

j2kun:
I will include this in a PR that adds to_tensor

PR #131, opened 2023-08-29
Add `ElementwiseMappable` trait to Poly ops.

mr0re1:
Add the trait to all ops, except:
* `extract_slice` - need to fix `InferTypeOpAdaptor` first, #130 ;
* `from_coeffs` - current signature brakes `ElementwiseMappable` constraint on normal case:

```mlir
%poly1 = poly.from_coeffs(%coeffs1) : (tensor<3xi32>) -> !poly.poly<#ring1>

'poly.from_coeffs' op if an operand is non-scalar, then there must be at least one non-scalar result
```


include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -65,13 +65,13 @@ def Poly_GetCoeffOp : Poly_Op<"get_coeff"> {
   let summary = "Gets the coefficient of X^{index} from a polynomial.";
 
   let arguments = (ins
-    Polynomial:$x,
-    Index:$index
+    PolynomialLike:$x,
+    IndexLike:$index
```
asraa, line 69: My feeling here is that this might not need `*Like` operands. I can think of cases where we would want to extract a coefficient of a single polynomial, or perhaps the `i`th coefficient of a list, but I can't think of one where we'd want to extract element-wise. (I'm not sure if this was intended to have the trait?)

The cases I can think of `get_coeff` being useful are things like "sample extraction", or conversion from RLWE to LWE ciphertext. In this case if you had some RLWE ciphertext (let's say of dimension 2) of the form `(a, b)` with `a = a_0 + a_1 * x + ... + a_{n-1} * x^{n-1}` and  `b = b_0 + b_1 * x + .. + b_{n-1} * x^{n-1}`, then your LWE ciphertext is `(a_0, -a_{n-1}, ... , -a_1, b_0)`. 

The first coefficients from `a` could probably be made by extracting a slice, negating, and reversing, then appending the first coefficient. Then you'd insert the the extracted 0th coefficient of `b`. So, I could see a `0`th coefficient extraction of the Tensor of `(a, b)`, but I'm not certain that's a big gain over individual extractions of them.

WDYT for simplifying until we have a need for the `*Like` types? @AlexanderViand-Intel let me know if something else comes to mind for extraction

include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -44,16 +44,16 @@ def Poly_MulOp : Poly_BinOp<"mul"> {
...
 
-def Poly_MulConstantOp : Poly_Op<"mul_constant"> {
+def Poly_MulConstantOp : Poly_Op<"mul_constant", [ElementwiseMappable]> {
   let summary = "Multiplication by a constant of the field.";
 
   let arguments = (ins
-    Polynomial:$x,
-    AnyInteger:$constant
+    PolynomialLike:$x,
+    AnyIntegerLike:$constant
```
asraa, line 52: We may want both of these, but this one I'd also have expected a single constant to be multiplied across a `PolynomialLike` - e.g. Take `(a, b)` as a TensorOf<Poly> and multiply by the constant `q`, to retrieve `(q*a, q*b)`.

asraa (commented):
Thank you! I'm also not sure that `extract_slice` requires ElementwiseMappable until we realize we need it in lowerings (for a similar reason to `get_coeff` review comment). @AlexanderViand-Intel  would appreciate a review to check me on these!

include/Dialect/Poly/IR/PolyOps.td
mr0re1: Simpler is better, done.

asraa (approved):
LGTM from my end! @AlexanderViand-Intel would appreciate a quick peek at the `PolyTypes.td` file changes.

(In particular, I wonder if it can just be "AnyType"? `poly.get_coeff` might need to return a ring element...)

include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -65,13 +65,13 @@ def Poly_GetCoeffOp : Poly_Op<"get_coeff"> {
...
   let arguments = (ins
-    Polynomial:$x,
+    PolynomialLike:$x,
     Index:$index
   );
 
   // NOTE: This assumes the underlying type is an integer.
   let results = (outs
-    AnyInteger:$output
+    AnyIntegerLike:$output
```
AlexanderViand-Intel, line 74: I think I agree with @asraa here, having `poly.get_coeff` over tensors of polynomials might not be needed at the moment but adds a bunch of complexity both for dialects lowering to poly and for dialects lowering from poly.

AlexanderViand-Intel: If we did define it over tensors, wouldn't it then also be `ElementwiseMappable`? Assuming that's the semantics we wanted from that! If passing in tensors of polynomials, one might also reasonably replace the index with a shape and simply apply the first k-1 dimensions to the container/tensor and the last one to the polynomial itself....

I'm not suggesting we do that, just highlighting how having tensor inputs here might be a bit confusing as it's not clear what the semantics should be.

AlexanderViand-Intel (changes requested):
Thanks for looking into `ElementwiseMappable`!

include/Dialect/Poly/IR/PolyTypes.td
```diff
@@ -24,4 +24,9 @@ def Polynomial : Poly_Type<"Polynomial", "poly"> {
   let assemblyFormat = "`<` $ring `>`";
 }
 
+def PolynomialLike: TypeOrContainer<Polynomial, "polynomial-like">;
```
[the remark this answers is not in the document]
mr0re1: `ElementwiseMappable` doesn't change the signature but simply impose constraints on the Op:
```
/// An `ElementwiseMappable` op must satisfy the following properties:
///
/// 1. If any result is a vector (resp. tensor), then at least one operand must
/// be a vector (resp. tensor).
/// 2. If any operand is a vector (resp. tensor), then there must be at least
/// one result, and all results must be vectors (resp. tensors).
/// 3. The static types of all vector (resp. tensor) operands and results must
/// have the same shape.
/// 4. In the case of tensor operands, the dynamic shapes of all tensor operands
/// must be the same, otherwise the op has undefined behavior.
/// 5. ("systematic scalarization" property) If an op has vector/tensor
/// operands/results, then the same op, with the operand/result types changed to
/// their corresponding element type, shall be a verifier-valid op.
/// 6. The semantics of the op on vectors (resp. tensors) shall be the same as
/// applying the scalarized version of the op for each corresponding element of
/// the vector (resp. tensor) operands in parallel.
/// 7. ("systematic vectorization/tensorization" property) If an op has
/// scalar operands/results, the op shall remain verifier-valid if all scalar
/// operands are replaced with vectors/tensors of the same shape and
/// corresponding element types.
```

We could remove the trait from the ops, but tests will still pass.

The value of adding it is to hint generalized "tools" about the behavior of the Op, e.g. `Linalg/Transforms/Vectorization`.

include/Dialect/Poly/IR/PolyOps.td
mr0re1: Happy to take a simple road, we can make it `ElementwiseMappable` in the future.

AlexanderViand-Intel (approved):
LGTM

include/Dialect/Poly/IR/PolyTypes.td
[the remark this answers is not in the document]
asraa: > but I've been fighting weird bazel issues for an hour now and decided to give up for the day...

@AlexanderViand-Intel need any help from our side? Feel free to chat me or file an issue

Issue #105, opened 2023-08-17
poly: add function materialization patterns to lower-poly


j2kun:
I am planning to do this, just wanted to split it off from the initial https://github.com/google/heir/pull/94 in light of https://reviews.llvm.org/D158107

j2kun:
Picking this one up

j2kun:
Fixed by https://github.com/google/heir/pull/154

Issue #104, opened 2023-08-17
poly: lower poly.add and poly.mul


j2kun:
Initial poly lowering pass shell added in https://github.com/google/heir/pull/94

asraa:
For the sake of testing and making sure Poly is workable, I'm giving this a stab, started with naive implementation of `poly.add`.

j2kun:
I'll take poly.mul

j2kun:
Fixed by https://github.com/google/heir/pull/180, and we're just waiting for internal checks to pass before merging.

PR #134, opened 2023-08-30
poly: add lowering of poly.add

asraa:
Adds lowering of `poly.add`: which is polynomial ring addition (coefficient-wise addition modulo the coefficient modulus of the ring)

Part of https://github.com/google/heir/issues/104
Fixes https://github.com/google/heir/issues/105

### DIFF
--- a/lib/Conversion/PolyToStandard/PolyToStandard.cpp
+++ b/lib/Conversion/PolyToStandard/PolyToStandard.cpp
@@ -128,18 +128,22 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
   //    c0 = c1 % cmod
   // Otherwise, compute the adjusted result:
   //    c0 = ((c1 % cmod) + (n1 * 2^N % cmod)) % cmod
+  // TODO(https://github.com/google/heir/issues/159): Add proof to docs.
   LogicalResult matchAndRewrite(
       AddOp op, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
-    auto type = adaptor.getLhs().getType();
-    APInt mod = op.getType().cast<PolyType>().getRing().coefficientModulus();
-    auto cmod = rewriter.create<arith::ConstantOp>(
-        op.getLoc(), DenseElementsAttr::get(cast<ShapedType>(type), {mod}));
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
 
-    auto addExtendedOp = rewriter.create<arith::AddUIExtendedOp>(
-        op.getLoc(), adaptor.getLhs(), adaptor.getRhs());
-    auto c1ModOp = rewriter.create<arith::RemUIOp>(
-        op.getLoc(), addExtendedOp->getResult(0), cmod);
+    auto type = adaptor.getLhs().getType();
+
+    APInt mod =
+        cast<PolyType>(op.getResult().getType()).getRing().coefficientModulus();
+    auto cmod = b.create<arith::ConstantOp>(
+        DenseElementsAttr::get(cast<ShapedType>(type), {mod}));
+
+    auto addExtendedOp =
+        b.create<arith::AddUIExtendedOp>(adaptor.getLhs(), adaptor.getRhs());
+    auto c1ModOp = b.create<arith::RemUIOp>(addExtendedOp->getResult(0), cmod);
     // If mod divides 2^N, c1modOp is our result.
     if (mod.isPowerOf2()) {
       rewriter.replaceOp(op, c1ModOp.getResult());
@@ -151,17 +155,14 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
     APInt::udivrem(bigMod, mod.zext(bigMod.getBitWidth()), quotient, remainder);
     remainder = remainder.trunc(mod.getBitWidth());
 
-    auto bitwidth = rewriter.create<arith::ConstantOp>(
-        op.getLoc(),
+    auto bitwidth = b.create<arith::ConstantOp>(
         DenseElementsAttr::get(cast<ShapedType>(type), {remainder}));
-    auto adjustOp =
-        rewriter.create<arith::AddIOp>(op.getLoc(), c1ModOp, bitwidth);
+    auto adjustOp = b.create<arith::AddIOp>(c1ModOp, bitwidth);
 
-    auto selectOp = rewriter.create<arith::SelectOp>(
-        op.getLoc(), addExtendedOp.getResult(1), c1ModOp, adjustOp);
+    auto selectOp = b.create<arith::SelectOp>(addExtendedOp.getResult(1),
+                                              c1ModOp, adjustOp);
     // Mod the final result.
-    rewriter.replaceOp(
-        op, rewriter.create<arith::RemUIOp>(op.getLoc(), selectOp, cmod));
+    rewriter.replaceOp(op, b.create<arith::RemUIOp>(selectOp, cmod));
 
     return success();
   }
--- a/lib/Conversion/PolyToStandard/PolyToStandard.cpp
+++ b/lib/Conversion/PolyToStandard/PolyToStandard.cpp
@@ -128,7 +128,15 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
   //    c0 = c1 % cmod
   // Otherwise, compute the adjusted result:
   //    c0 = ((c1 % cmod) + (n1 * 2^N % cmod)) % cmod
-  // TODO(https://github.com/google/heir/issues/159): Add proof to docs.
+  //
+  // Note that `(c1 % cmod) + (n1 * 2^N % cmod)` will not overflow mod `2^N`.
+  // If it did, then it would require that `cmod > (2^N) / 2`.
+  // This would imply that `2^N % cmod = 2^N - cmod`.
+  // If the sum overflowed, then we would have
+  //    ((c1 % cmod) + (2^N % cmod)) > 2^N
+  //    ((c1 % cmod) + (2^N - cmod)) > 2^N
+  //    ((c1 % cmod) > cmod
+  // Which is a contradiction.
   LogicalResult matchAndRewrite(
       AddOp op, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
@@ -199,7 +207,8 @@ struct PolyToStandard : impl::PolyToStandardBase<PolyToStandard> {
     // target.addIllegalOp<MulOp>();
 
     RewritePatternSet patterns(context);
-    patterns.add<ConvertFromTensor, ConvertToTensor, ConvertAdd>(typeConverter, context);
+    patterns.add<ConvertFromTensor, ConvertToTensor, ConvertAdd>(typeConverter,
+                                                                 context);
     addStructuralConversionPatterns(typeConverter, patterns, target);
 
     // TODO(https://github.com/google/heir/issues/143): Handle tensor of polys.
--- a/lib/Conversion/PolyToStandard/PolyToStandard.cpp
+++ b/lib/Conversion/PolyToStandard/PolyToStandard.cpp
@@ -221,6 +221,7 @@ struct PolyToStandard : impl::PolyToStandardBase<PolyToStandard> {
              isLegalForReturnOpTypeConversionPattern(op, typeConverter);
     });
 
+    // TODO(https://github.com/google/heir/issues/143): Handle tensor of polys.
     if (failed(applyPartialConversion(module, target, std::move(patterns)))) {
       signalPassFailure();
     }
--- a/lib/Conversion/PolyToStandard/PolyToStandard.cpp
+++ b/lib/Conversion/PolyToStandard/PolyToStandard.cpp
@@ -118,44 +118,50 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
   // element-wise as elements of the ring, so they are performed modulo the
   // coefficient modulus.
   //
-  // Modular addition is performed by sign extending both arguments and
-  // performing (N+1)-bit addition. Then, if z = (x + y) is the (N+1)-bit sum,
-  // we return
-  //   truncate(z >= mod ? z - mod : z)
+  // To perform modular addition, assume that `cmod` is the coefficient modulus
+  // of the ring, and that `N` is the bitwidth used to store the ring elements.
+  // This may be much larger than `log_2(cmod)`.
+  //
+  // Let `x` and `y` be the inputs to modular addition, then:
+  //    c1, n1 = addui_extended(x, y)
+  // If the coefficient modulus divides `2^N`, then return
+  //    c0 = c1 % cmod
+  // Otherwise, compute the adjusted result:
+  //    c0 = ((c1 % cmod) + (n1 * 2^N % cmod)) % cmod
   LogicalResult matchAndRewrite(
       AddOp op, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
     auto type = adaptor.getLhs().getType();
-    auto elementTy =
-        type.cast<RankedTensorType>().getElementType().cast<IntegerType>();
-    auto extType = IntegerType::get(getContext(), elementTy.getWidth() + 1,
-                                    elementTy.getSignedness());
-    auto extShapeType = type.cast<ShapedType>().clone(extType);
+    APInt mod = op.getType().cast<PolyType>().getRing().coefficientModulus();
+    auto cmod = rewriter.create<arith::ConstantOp>(
+        op.getLoc(), DenseElementsAttr::get(cast<ShapedType>(type), {mod}));
 
-    mlir::Value extLhsOp = rewriter.create<arith::ExtUIOp>(
-        op.getLoc(), extShapeType, adaptor.getLhs());
-    mlir::Value extRhsOp = rewriter.create<arith::ExtUIOp>(
-        op.getLoc(), extShapeType, adaptor.getRhs());
+    auto addExtendedOp = rewriter.create<arith::AddUIExtendedOp>(
+        op.getLoc(), adaptor.getLhs(), adaptor.getRhs());
+    auto c1ModOp = rewriter.create<arith::RemUIOp>(
+        op.getLoc(), addExtendedOp->getResult(0), cmod);
+    // If mod divides 2^N, c1modOp is our result.
+    if (mod.isPowerOf2()) {
+      rewriter.replaceOp(op, c1ModOp.getResult());
+      return success();
+    }
+    // Otherwise, add (n1 * 2^N % cmod)
+    APInt quotient, remainder;
+    APInt bigMod = APInt(mod.getBitWidth() + 1, 2) << (mod.getBitWidth() - 1);
+    APInt::udivrem(bigMod, mod.zext(bigMod.getBitWidth()), quotient, remainder);
+    remainder = remainder.trunc(mod.getBitWidth());
 
-    auto addOp = rewriter.create<arith::AddIOp>(op.getLoc(), extShapeType,
-                                                extLhsOp, extRhsOp);
+    auto bitwidth = rewriter.create<arith::ConstantOp>(
+        op.getLoc(),
+        DenseElementsAttr::get(cast<ShapedType>(type), {remainder}));
+    auto adjustOp =
+        rewriter.create<arith::AddIOp>(op.getLoc(), c1ModOp, bitwidth);
 
-    APInt mod =
-        op.getType().cast<PolynomialType>().getRing().coefficientModulus();
-    assert(mod != 0 && "coefficient modulus must not be zero");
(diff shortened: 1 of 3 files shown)


lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -91,16 +109,38 @@ struct PolyToStandard : impl::PolyToStandardBase<PolyToStandard> {
     ConversionTarget target(*context);
     PolyToStandardTypeConverter typeConverter(context);
 
+    target.addLegalDialect<arith::ArithDialect>();
+
     // target.addIllegalDialect<PolyDialect>();
     target.addIllegalOp<PolyFromCoeffsOp>();
```
mr0re1, line 115: nit.  To be aligned with `patterns.add`:
```cpp
 target.addIllegalOp<PolyFromCoeffsOp, AddOp>();
```

asraa:
OK, am marking this as ready to review! I'll tackle `poly.mul` after or hand off :)

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -60,11 +66,27 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
+    auto type = adaptor.getLhs().getType();
+    auto addOp = rewriter.create<arith::AddIOp>(
+        op.getLoc(), type, adaptor.getLhs(), adaptor.getRhs());
+
+    APInt mod = op.getType().getRing().coefficientModulus();
+    assert(mod != 0 && "coefficient modulus must not be zero");
+    auto modConstOp = rewriter.create<arith::ConstantOp>(
+        addOp.getLoc(), DenseElementsAttr::get(type.cast<ShapedType>(), {mod}));
+
+    auto modOp = rewriter.create<arith::RemUIOp>(
```
asraa, line 85: Will update, self review, this won't handle overflow properly

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -60,11 +66,27 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
-    // TODO(https://github.com/google/heir/issues/104): implement
-    return failure();
+    auto type = adaptor.getLhs().getType();
+    auto addOp = rewriter.create<arith::AddIOp>(
+        op.getLoc(), type, adaptor.getLhs(), adaptor.getRhs());
+
+    APInt mod = op.getType().getRing().coefficientModulus();
+    assert(mod != 0 && "coefficient modulus must not be zero");
+    auto modConstOp = rewriter.create<arith::ConstantOp>(
+        addOp.getLoc(), DenseElementsAttr::get(type.cast<ShapedType>(), {mod}));
```
code-perspective, line 83: > addOp.getLoc()

Nitpicking and please feel free to ignore: how about adding unrelated ops to the main "op" location. (Mostly for readability and avoid confusion)

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -60,11 +66,27 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
+  // operation is defined within the polynomial ring. Coefficients are added
+  // element-wise as elements of the ring, so they are performed modulo the
+  // coefficient modulus.
   LogicalResult matchAndRewrite(
       AddOp op, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
-    // TODO(https://github.com/google/heir/issues/104): implement
-    return failure();
+    auto type = adaptor.getLhs().getType();
+    auto addOp = rewriter.create<arith::AddIOp>(
```
code-perspective, line 77: > AddIOp

Need to handle interger overflow with [`AddUIExtendedOp`](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithaddui_extended-arithadduiextendedop). @asraa found this when we were doing a walk through. Adding a reminder here.

code-perspective (approved):

asraa:
Will squash when review finished

tests/poly/lower_poly.mlir
```diff
@@ -4,12 +4,25 @@
 #cycl_2048 = #poly.polynomial<1 + x**1024>
 #ring = #poly.ring<cmod=4294967296, ideal=#cycl_2048>
 module {
-  func.func @test_lower_from_coeffs() {
-    %c0 = arith.constant 0 : index
-    // 2 + 2x + 5x^2
-    %coeffs = arith.constant dense<[2, 2, 5]> : tensor<3xi32>
+  func.func @test_lower_from_coeffs() -> !poly.poly<#ring> {
```
mr0re1, line 7: could you please add a test case for addition on `tensor<Nxpoly>` ?

asraa: Updated with a TODO - it's really unfortunate tensors of tensors are not a thing - I filed an issue to discuss what we can do (either detensorize or flatten the tensor)

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -60,11 +66,51 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
+    auto extType = IntegerType::get(getContext(), elementTy.getWidth() + 1,
+                                    elementTy.getSignedness());
+    auto extShapeType = type.cast<ShapedType>().clone(extType);
+
+    mlir::Value extLhsOp = rewriter.create<arith::ExtUIOp>(
+        op.getLoc(), extShapeType, adaptor.getLhs());
+    mlir::Value extRhsOp = rewriter.create<arith::ExtUIOp>(
+        op.getLoc(), extShapeType, adaptor.getRhs());
+
+    auto addOp = rewriter.create<arith::AddIOp>(op.getLoc(), extShapeType,
```
j2kun, line 93: Coefficients in poly are by definition unsigned. Or rather, the coefficient ring is always Z/nZ, which I'm assuming means we're picking the representative in [0, n-1] for all our computations. So shouldn't all these ops use the unsigned variant? Also, if the semantics were signed then you'd want signed extension, not zero-extension. I think it's just easier to do everything unsigned.

Ditto for the tests that create polys with i64s. Maybe this means we should have a verifier that the coefficients used to construct a poly are always unsigned?

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -91,17 +137,36 @@ struct PolyToStandard : impl::PolyToStandardBase<PolyToStandard> {
...
     // target.addIllegalOp<MulOp>();
 
-    // TODO(https://github.com/google/heir/issues/105): add upstream type
-    // converters for func/call/return
-
     RewritePatternSet patterns(context);
-    patterns.add<ConvertPolyFromCoeffs>(typeConverter, context);
+    patterns.add<ConvertPolyFromCoeffs, ConvertAdd>(typeConverter, context);
+
+    populateFunctionOpInterfaceTypeConversionPattern<func::FuncOp>(
```
j2kun, line 149: Possible to put this in a utility function so it can be shared with the other lowerings that use the same boilerplate? Something like

```
void populateStructuralConversionPatterns(RewritePatternSet& patterns, TypeConverter& typeConverter) {
    populateFunctionOpInterfaceTypeConversionPattern<func::FuncOp>(
          patterns, typeConverter);
    target.addDynamicallyLegalOp<func::FuncOp>([&](func::FuncOp op) {
      return typeConverter.isSignatureLegal(op.getFunctionType()) &&
             typeConverter.isLegal(&op.getBody());
    });
    populateCallOpTypeConversionPattern(patterns, typeConverter);
    target.addDynamicallyLegalOp<func::CallOp>(
        [&](func::CallOp op) { return typeConverter.isLegal(op); });
    populateReturnOpTypeConversionPattern(patterns, typeConverter);
    target.addDynamicallyLegalOp<func::ReturnOp>(
        [&](func::ReturnOp op) { return typeConverter.isLegal(op); });
    populateBranchOpInterfaceTypeConversionPattern(patterns, typeConverter);
    target.markUnknownOpDynamicallyLegal([&](Operation *op) {
      return isNotBranchOpInterfaceOrReturnLikeOp(op) ||
             isLegalForBranchOpInterfaceTypeConversionPattern(op,
                                                              typeConverter) ||
             isLegalForReturnOpTypeConversionPattern(op, typeConverter);
    });
}
```

j2kun: I know BGV already uses this, right?

tests/poly/lower_poly.mlir
```diff
@@ -4,12 +4,25 @@
...
+    // CHECK: [[X:%.+]] = arith.constant dense<2> : [[T:tensor<1024.*]]
+    %coeffs1 = arith.constant dense<2> : tensor<1024xi64>
+    // CHECK: [[Y:%.+]] = arith.constant dense<3> : [[T]]
+    %coeffs2 = arith.constant dense<3> : tensor<1024xi64>
     // CHECK-NOT: poly.from_coeffs
-    %poly = poly.from_coeffs(%coeffs) : (tensor<3xi32>) -> !poly.poly<#ring>
-    return
+    %poly0 = poly.from_coeffs(%coeffs1) : (tensor<1024xi64>) -> !poly.poly<#ring>
+    %poly1 = poly.from_coeffs(%coeffs2) : (tensor<1024xi64>) -> !poly.poly<#ring>
+    // CHECK: [[XEXT:%.+]] = arith.extui [[X]] : [[T]] to [[TEXT:tensor<1024.*]]
```
j2kun, line 16: The width of the element type is computed here and should be asserted. The poly cmod is `2**32` and so while I believe APInt doubles the width to 64, you add one so it may be 65?

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -60,11 +66,51 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
 
   using OpConversionPattern::OpConversionPattern;
 
+  // Convert add lowers a poly.add operation to arith operations. A poly.add
+  // operation is defined within the polynomial ring. Coefficients are added
+  // element-wise as elements of the ring, so they are performed modulo the
+  // coefficient modulus.
+  //
+  // Modular addition is performed by sign extending both arguments and
+  // performing (N+1)-bit addition. Then, if z = (x + y) is the (N+1)-bit sum,
```
j2kun, line 75: I can't help but wonder if it would make more sense to do this optimization in a separate pass/pattern to keep this lowering simpler. E.g,. could we later match on patterns that look like addui_extended followed by remui by a constant, and fold those greedily? Shouldn't it stand to reason that there are already such optimizations in MLIR? Or do we think that later passes/canonicalizations would muck that pattern up so it would be hard to match? I also wonder what existing canonicalizations might apply to a more straightforward lowering (addui + remui) vs this one---or maybe vice versa that eixsting canonicalizations would apply better to this form! I'm just unaware and would hence bias toward the simplest option.

In other words, I feel that if we can avoid mixing optimizations and conversions in the same pass, we should. Unless we have concrete evidence that this is the best place to do the optimization.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -60,11 +66,51 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
-    // TODO(https://github.com/google/heir/issues/104): implement
-    return failure();
+    auto type = adaptor.getLhs().getType();
+    auto elementTy =
+        type.cast<RankedTensorType>().getElementType().cast<IntegerType>();
+    auto extType = IntegerType::get(getContext(), elementTy.getWidth() + 1,
+                                    elementTy.getSignedness());
+    auto extShapeType = type.cast<ShapedType>().clone(extType);
+
+    mlir::Value extLhsOp = rewriter.create<arith::ExtUIOp>(
```
j2kun, line 88: Assuming optimizations will stay here: since you know the modulus and the element type bit width statically, you could check in advance whether there is a risk of overflow, and if not skip the sign extensions.

tests/poly/lower_poly.mlir
j2kun, line 1: If the optimizations stay in this pass, then I think it would be worthwhile to run the mlir-cpu-runner on the result (maybe extracting a single coefficient at the end as the return value) to ensure the lowering has the semantics you expect.

j2kun (changes requested):

lib/Conversion/PolyToStandard/PolyToStandard.cpp
asraa: > addui_extended

The problem is that `addui_extended` returns an overflow bit, not the result - so I don't really see how to get the remainder unless I applied that first. Perhaps I could do something clever by doing `a + (-mod + b)` when the overflow bit is 1. But that seems just the same as the optimization here to avoid remui.

> Shouldn't it stand to reason that there are already such optimizations in MLIR?

I would expect so only if they handled general modular arithmetic, but I'm not sure they do expect for power of 2 optimizations.

asraa: > Unless we have concrete evidence that this is the best place to do the optimization.

I guess I can understand that - I thought that lowerings are generally where we would want to implement optimized versions, but perhaps the patterns that can be recognized should be optimized in a rewriter. I'm not super fussy on this, just wanted to get some implementation down to start with. The addui_extended route didn't seem to work, so I could naively do a remui and trunci, that works for me. Rewriting that pattern (we can do like "barret reduction rewrite" for muls too) works

j2kun: Cleared this up in a chat:

case 1: no assumptions, compute `c1, n1 = addui_extended` then `c0 = ((c1 % cmod) + (n1 2^N % cmod)) % cmod`
case 2: cmod divides 2^N, compute `c1 = addui` (allowing overflow) then `c0 = c1 % cmod`

lib/Conversion/PolyToStandard/PolyToStandard.cpp
asraa: On point 1: yep, agreed to use addui and unsigned arith ops.

> Maybe this means we should have a verifier that the coefficients used to construct a poly are always unsigned?

I'm not sure, if we want our type conversion to handle `from_coeff` by just dropping the op and assigning to the original tensor, then I don't think we can use unsigned ints. Most of the arith dialect ops want signless integers, despite being for unsigned ops. So for example, `arith.addui_extended` requires operands to be `signless-integer-like` - which explicitly disallows `ui64`. 

Maybe this is why you originally used `i64`?

j2kun: Yeah I ran into the same issue. Signless will have to be the way to go, and I think i32 counts as signless?

lib/Conversion/PolyToStandard/PolyToStandard.cpp
j2kun: Actually I already added this in lib/Conversion/Utils.cpp, so you can reuse that.

tests/poly/lower_poly.mlir
asraa: Filing a follow-up issue for this, it's pretty complicated. It needs:

(1) I implemented conversion for `to_tensor` - I'll add it to a follow-up.
(2) It also needs some tricky handling of doing arith on tensors - lowering arith when the operands are tensors is surprisingly non-trivial.

asraa: https://github.com/google/heir/issues/156

asraa:
@j2kun This commit contains the update to the new algorithm: https://github.com/google/heir/pull/134/commits/2fb325791f97325b9cd0a71ce567b87e153e9081

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -54,8 +54,8 @@ struct ConvertFromTensor : public OpConversionPattern<FromTensorOp> {
   LogicalResult matchAndRewrite(
       FromTensorOp op, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
-    // FIXME: If the coeffs (the underlying tensor) has a smaller dimension and
-    // we intend to embed into the ring, we must handle the types here.
+    // Note: This currently requires that the underlying tensor that holds the
+    // coeffs has the same dimension as the ring.
```
j2kun, line 58: BTW, I changed this behavior in https://github.com/google/heir/pull/152/files#diff-439000080211781bbe504b2ce568ebf0aa9b638f1ff3d82549da7347b1da6ddfR19

Now the input tensor may have smaller shape than the polynomial degree. So I think that would imply you need an extra op to zero-extend the tensor... is this doable with a `tensor.pad`? Not sure... If you agree, let's file an issue for it and keep going.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -72,44 +72,50 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
-    auto modConstOp = rewriter.create<arith::ConstantOp>(
-        op.getLoc(), DenseElementsAttr::get(extShapeType, {mod}));
-
-    auto geCmp = rewriter.create<arith::CmpIOp>(
-        addOp.getLoc(), arith::CmpIPredicate::uge, addOp, modConstOp);
-    auto selectOp = rewriter.create<arith::SelectOp>(
-        geCmp->getLoc(), geCmp,
-        rewriter.create<arith::SubIOp>(geCmp->getLoc(), addOp, modConstOp),
-        addOp);
+    APInt mod = op.getType().cast<PolyType>().getRing().coefficientModulus();
```
j2kun, line 89: Nit: I suspect you can do op.getResult().getType() to avoid the cast.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -72,44 +72,50 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
+    auto cmod = rewriter.create<arith::ConstantOp>(
+        op.getLoc(), DenseElementsAttr::get(cast<ShapedType>(type), {mod}));
```
j2kun, line 91: Since you're generating a bunch of sequential statements, consider using ImplicitLocOpBuilder https://github.com/llvm/llvm-project/blob/d671126ad097ffd1bb19322005902676945862c2/mlir/include/mlir/IR/ImplicitLocOpBuilder.h#L23

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -72,44 +72,50 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
+    APInt quotient, remainder;
+    APInt bigMod = APInt(mod.getBitWidth() + 1, 2) << (mod.getBitWidth() - 1);
+    APInt::udivrem(bigMod, mod.zext(bigMod.getBitWidth()), quotient, remainder);
+    remainder = remainder.trunc(mod.getBitWidth());
+
+    auto bitwidth = rewriter.create<arith::ConstantOp>(
+        op.getLoc(),
+        DenseElementsAttr::get(cast<ShapedType>(type), {remainder}));
+    auto adjustOp =
+        rewriter.create<arith::AddIOp>(op.getLoc(), c1ModOp, bitwidth);
```
j2kun, line 112: I wonder if this could also overflow...

j2kun (approved):

lib/Conversion/PolyToStandard/PolyToStandard.cpp
asraa: Fixed in https://github.com/google/heir/pull/158, I'll rebase this PR once that's submitted first.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
asraa: It seems like the return value of `op.getResult().getType()` is an `mlir::Type`, same with `op->getResultTypes()[0]`.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
asraa: I don't think it can. I think I should write up a formal proof, so I'll file an issue on that.

asraa:
NOTE: This PR now requires the fixes from https://github.com/google/heir/pull/158

lib/Dialect/Poly/IR/PolyOps.cpp
```diff
@@ -32,7 +32,7 @@ LogicalResult FromTensorOp::verify() {
   }
 
   APInt coefficientModulus = ring.coefficientModulus();
-  unsigned cmodBitWidth = coefficientModulus.logBase2();
+  unsigned cmodBitWidth = coefficientModulus.ceilLogBase2();
```
asraa, line 35: fyi: new bug fix. after incorporating the verifier PR, and with the new modulus I was testing for the change, I hit a bug that a size `32` integer coeff array was not valid, `logBase2` returns floor of the log. @j2kun

Issue #127, opened 2023-08-29
process: Improve process around PR merges


asraa:
Right now, it is challenging for an external contributor to note when a PR should get merged. I think we should do the following changes:

To enforce external review:
(1) Require an external LGTM for each PR
(2) Enable a process that indicates whether the mirrored internal PR has that external LGTM, and only merge if that condition is satisfied

JAX does something similar, where once a PR gets an LGTM from GitHub, they update the reviewers to something like `jax-mergeready`.

To help external contributors to understand the process, we should document the merge / contribution process.

This can also be used to solve https://github.com/google/heir/issues/107; we can gate the `-mergeready` review on having the commits squashed.

j2kun:
I'd say just add the label manually for now. I like easy things.

asraa:
Agree for now as well. I'll update the workflows!

asraa:
Going to close for now, as the docs are updated. Feel free to open another one for any other process improvements

PR #136, opened 2023-08-30
ci: add pull_ready label on PR approval for internal migration

asraa:
Part of https://github.com/google/heir/issues/127

This PR adds a workflow that auto adds the `pull_ready` label on PR approval. The workflow will indicate failure if the PR is approved and not squashed into a single commit - this way, we don't check single commit until after the review is complete (I'm not sure we need the `squash ready` label since CI failure indicates this, so I'll update the docs once we test this process)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
 # This is used by .github/workflows/pr_review.yml to apply the `pull_ready`
 # label to any root change when it has approval.
 pull_ready:
-- '*'
+- '**'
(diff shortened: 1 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,4 @@
+# This is used by .github/workflows/pr_review.yml to apply the `pull_ready`
+# label to any root change when it has approval.
+pull_ready:
+- '*'
```
j2kun, line 4: You probably want `'**'` instead. From https://github.com/actions/labeler, a single star appears to only match files in the root directory of the repo when we want it to match any changed path.

Also note that this requires the `dot=true` option in the job configuration.

```
apply-pull-ready-label:
  ...
  steps:
    - name: ...
      uses: actions/labeler@...
      with:
        dot: true
```

j2kun (changes requested):
One small fix, otherwise LGTM!

j2kun (approved):

Issue #120, opened 2023-08-22
website: add mathjax support


j2kun:
In some docs we have a need for LaTeX, see https://google.github.io/heir/docs/dialects/polyattributes/

mr0re1:
Docsy [supports](https://www.docsy.dev/docs/adding-content/diagrams-and-formulae/
) KaTeX out of the box, sadly KaTeX doesn't work with `$inline$` math. 
Though it works with `$$ block $$` math, and special inline [syntax](https://www.docsy.dev/docs/adding-content/diagrams-and-formulae/), that doesn't look as nice as `$one$`.

The matter of enabling it is just adding `math: true` to page front matter.

Will keep looking on how to enable mathjax.

mr0re1:
Uploading screenshot for relevant PR.
<img width="897" alt="Screenshot 2023-09-01 at 1 35 36 PM" src="https://github.com/google/heir/assets/1387442/eb1157da-e500-4d5b-95e6-44986bf8dfe0">

PR #140, opened 2023-09-01
Add MathJax to docs

mr0re1:
* Use Docsy hook to include MathJax3 script and configure it to handle `$inline$` math.
* Explicitly disable KaTeX rendering to avoid mixing both.

Fixes: #120 

Docsy [supports](https://www.docsy.dev/docs/adding-content/diagrams-and-formulae/ ) KaTeX out of the box. The matter of enabling it is just adding `math: true` to page front matter or enabling it site-wise.

Screenshot 2023-09-01 at 1 35 36 PM

### DIFF
--- a/docs/hugo.toml
+++ b/docs/hugo.toml
@@ -167,6 +167,11 @@ enable = false
   icon = "fab fa-github"
   desc = "Development takes place here!"
 
+
+[params.katex]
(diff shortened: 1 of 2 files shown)


j2kun (approved):
Love it! Thanks for this!

PR #141, opened 2023-09-05
Fix ordering of Monomial

danlark1:


asraa (approved):
Thanks for this! I assume this supersedes https://github.com/google/heir/pull/142

PR #152, opened 2023-09-11
Poly: replace get_coeff and extract_slice with to_tensor

j2kun:
Also:

- add verifiers to to/from_tensor
- add more specific semantics to the docs
- simplify the assembly format
- rename PolynomialType to PolyType for more consistency (less name conflict with the Polynomial storage class in PolyAttr).


j2kun:
Fixes https://github.com/google/heir/issues/130

mr0re1 (approved):

Issue #148, opened 2023-09-06
poly: write verifiers to enforce coefficient sizes below coefficient modulus


asraa:
cc @j2kun 

The coefficient modulus may be much smaller than the APInt size that contains it. Because of this, an integer may validly be represented as a value much larger than cmod.

However, if we can verify that all poly's that are constructed are constructed with coeffs smaller than cmod, and all ops between poly's keep this invariant, than ops can use certain optimizations (e.g. barret reduction, or select and subtract) rather handling more remainder operations.

PR #153, opened 2023-09-11
Add a verifier that poly.from_tensor has properly normalized inputs

j2kun:
Fixes https://github.com/google/heir/issues/148

### DIFF
--- a/lib/Dialect/Poly/IR/PolyOps.cpp
+++ b/lib/Dialect/Poly/IR/PolyOps.cpp
@@ -1,5 +1,7 @@
 #include "include/Dialect/Poly/IR/PolyOps.h"
 
+#include "llvm/include/llvm/ADT/APInt.h"  // from @llvm-project
+
 namespace mlir {
 namespace heir {
 namespace poly {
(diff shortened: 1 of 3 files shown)


lib/Dialect/Poly/IR/PolyOps.cpp
```diff
@@ -17,14 +19,30 @@ LogicalResult FromTensorOp::verify() {
   auto tensorShape = getInput().getType().getShape();
   auto polyDegree = getOutput().getType().getRing().getIdeal().getDegree();
```
mr0re1, line 20: nit. add `ring` variable, to avoid lengthy declaration of `polyDegree` & `coeffModulus`.

mr0re1 (approved):

lib/Dialect/Poly/IR/PolyOps.cpp
j2kun: Done!

lib/Dialect/Poly/IR/PolyOps.cpp
```diff
@@ -15,16 +17,32 @@ void FromTensorOp::build(OpBuilder &builder, OperationState &result,
...
+    return emitOpError()
+           << "input type " << getInput().getType()
+           << " does not match output type " << getOutput().getType()
+           << ". The input type must be a tensor of shape [d] where d "
+              "is at most the degree of the polynomial generator of "
+              "the output ring's ideal.";
+  }
+
+  APInt coefficientModulus = ring.coefficientModulus();
+  unsigned cmodBitWidth = coefficientModulus.logBase2();
```
asraa, line 33: @j2kun shouldn't this be the `APInt` container? What if the coeffs use the same container, but the values are less than the coefficient modulus? 

I'm wondering for the case that the coeffs are generated as APInts somewhere in code. This would then fail if they were `i64`'s, but the values were smaller. Should we force the code generating these to rescale to the true bit width? I guess that can be a yes.

j2kun: I'm not sure I totally understand your question. If the coefficient values of a `from_tensor` op are generated in code, the generating code still has to pick an element type i32, i64, etc., and that is what is checked on line 34, not the actual coefficient values.

If the coefficient modulus is, say, 2^16, but contained in an APInt that has 64 bits of space, we'd still want to enforce that the element type fits in 16 bits. So checking the APInt container size on line 33 would be too permissive.

asraa: > If the coefficient modulus is, say, 2^16, but contained in an APInt that has 64 bits of space, we'd still want to enforce that the element type fits in 16 bits.

Got it - I was wondering for a sampling function, we'd likely give the actual bit width of the coefficient modulus, not the APInt container size - so that makes sense.

Issue #121, opened 2023-08-22
website: reorganize dialect docs


j2kun:
The dialect docs are organized in kind of a haphazard way, we should reorganize it by dialect

![image](https://github.com/google/heir/assets/2467754/b9887958-1f8b-444d-8a71-8ed3a5755c05)

j2kun:
This file copies the markdown files to the right places: https://github.com/google/heir/blob/main/.github/workflows/copy_tblgen_files.sh

PR #155, opened 2023-09-12
Group generated type/op/attr docs by dialect

j2kun:
Fixes https://github.com/google/heir/issues/121

Also added op docs for poly/bgv which didn't have them already


j2kun:
Example screenshot of the result: 
image

include/Dialect/BGV/IR/BUILD
```diff
@@ -47,6 +47,12 @@ gentbl_cc_library(
             ],
             "BGVDialect.cpp.inc",
         ),
+        (
```
asraa, line 50: Thanks! This actually reminds me to make a cleanup to create a bazel rule that does all the default creation for types, ops, and dialect. Similar to [`add_mlir_dialect` ](https://mlir.llvm.org/docs/Tutorials/CreatingADialect/#tablegen-targets)

include/Dialect/Secret/IR/SecretDialect.td
```diff
@@ -2,11 +2,13 @@
 #define HEIR_INCLUDE_DIALECT_SECRET_IR_SECRETDIALECT_TD_
 
 include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
 
 def Secret_Dialect : Dialect {
   let name = "secret";
-  let summary = "A dialect for computations that operate on encrypted secrets.";
```
asraa, line 8: Should we note somewhere in a style guide not to use `summary`? Was it making docs look bad?

asraa (approved):

include/Dialect/Secret/IR/SecretDialect.td
j2kun: I don't think it's bad, but it just emits the summary as an extra line above the description with no extra line break. So if you expect that then you can make it look right.

PR #157, opened 2023-09-12
Add verifier "type of x == type of output" to `Poly_MulConstantOp`

mr0re1:


include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -47,6 +47,8 @@ def Poly_MulConstantOp : Poly_Op<"mul_constant", [ElementwiseMappable]> {
   let results = (outs
     PolynomialLike:$output
   );
+
+  let hasVerifier = 1;
```
j2kun, line 51: Consider instead using the trait `AllTypesMatch<["x", "output"]>` which avoids the need for any custom C++

j2kun (approved):
One minor suggestion otherwise LGTM

include/Dialect/Poly/IR/PolyOps.td
mr0re1: Thank you! I was almost sure that `AllTypesMatch` exists )

mr0re1: I will update assembly format as well.

j2kun (approved):

Issue #161, opened 2023-09-13
Compile information about the different kinds of encodings used across FHE schemes


j2kun:

j2kun:
For one data point, the OpenFHE `pke` library supports BGV, BFV, and CKKS, and they use the same plaintext base and encoding parameters base for all schemes: https://github.com/openfheorg/openfhe-development/blob/4ebb28ea7bdd894a73bc5b73e59fcfbc78253308/src/pke/include/encoding/plaintext.h#L65

j2kun:
@AlexanderViand-Intel do you have any references to those more exotic encodings?

asraa:
CKKS in microsoft SEAL encoding is implemented separate from BG/FV - (some example here: https://github.com/microsoft/SEAL/blob/82b07db635132e297282649e2ab5908999089ad2/native/examples/5_ckks_basics.cpp#L77C44-L79C54) but per the scaling, it seems like that could be achieved with the common parameter base you linked.

j2kun:
So it seems like the three approaches are:

- CGGI: encode a small integer inside a larger integer segmented as `padding | cleartext | noise`, with `padding=1` usually.
- BGV/BFV with "DCRT": encode a vector of integers as evaluations of the plaintext polynomial at roots of unity, use INTT to get the actual polynomial plaintext (with some ring theory to show why INTT corresponds to a CRT reconstruction). If the input cleartext is too large for the polynomial ring coefficient space, you can precede this encoding with another CRT (just on plain integers) to decompose the cleartext into a vector of small-enough integers. Hence, they call this the "double CRT" representation.
- CKKS with RNS: encode a vector of real or complex numbers by inverting the "canonical embedding" (evaluate at half the roots of unity, ignoring complex conjugates since they can be derived from the first half), discretizing halfway through because arbitrary complex numbers won't land in the image of the canonical embedding. Apparently there are multiple discretization options (including component-wise randomized rounding!) but they all first scale the input by a scaling factor to preserve the MSBs.

One thing that I'm still not clear on is how BGV and BFV differ in their encodings. I have heard that BFV uses the LSBs and BGV the MSBs, but I don't see that difference materializing in SEAL or OpenFHE directly.

asraa:
>  I have heard that BFV uses the LSBs and BGV the MSBs, but I don't see that difference materializing in SEAL or OpenFHE directly.

[...]

PR #163, opened 2023-09-14
Add LWE toolbox dialect and CGGI-style lwe bitfield encoding attribute

j2kun:
Partial progress on https://github.com/google/heir/issues/161

This uses the `VerifiableTensorEncoding` interface to verify the encoding is compatible with the underlying tensor element type.

I also tweaked the lit config slightly to make it easier to check for expected error messages.

Future PRs will add RLWE encoding attributes as per https://github.com/google/heir/issues/161

### DIFF
--- a/lib/Dialect/LWE/IR/LWEDialect.cpp
+++ b/lib/Dialect/LWE/IR/LWEDialect.cpp
@@ -1,0 +1,54 @@
+#include "include/Dialect/LWE/IR/LWEDialect.h"
+
+#include "include/Dialect/LWE/IR/LWEAttributes.h"
+#include "llvm/include/llvm/ADT/TypeSwitch.h"            // from @llvm-project
+#include "mlir/include/mlir/IR/DialectImplementation.h"  // from @llvm-project
+
+// Generated definitions
+#include "include/Dialect/LWE/IR/LWEDialect.cpp.inc"
+#define GET_ATTRDEF_CLASSES
+#include "include/Dialect/LWE/IR/LWEAttributes.cpp.inc"
+
+namespace mlir {
+namespace heir {
+namespace lwe {
+
+void LWEDialect::initialize() {
+  addAttributes<
+#define GET_ATTRDEF_LIST
+#include "include/Dialect/LWE/IR/LWEAttributes.cpp.inc"
+      >();
+}
+
+LogicalResult BitFieldEncodingAttr::verifyEncoding(
+    ArrayRef<int64_t> shape, Type elementType,
+    ::llvm::function_ref<::mlir::InFlightDiagnostic()> emitError) const {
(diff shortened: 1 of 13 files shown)


lib/Dialect/LWE/IR/LWEDialect.cpp
```diff
@@ -0,0 +1,54 @@
...
+  addAttributes<
+#define GET_ATTRDEF_LIST
+#include "include/Dialect/LWE/IR/LWEAttributes.cpp.inc"
+      >();
+}
+
+LogicalResult BitFieldEncodingAttr::verifyEncoding(
+    ArrayRef<int64_t> shape, Type elementType,
+    ::llvm::function_ref<::mlir::InFlightDiagnostic()> emitError) const {
+  if (!elementType.isa<IntegerType>()) {
```
asraa, line 26: nit: should they also be signless? I don't think I'd understand why a signed integer would be encoded like this.

lib/Dialect/LWE/IR/LWEDialect.cpp
```diff
@@ -0,0 +1,54 @@
...
+  unsigned plaintextBitwidth = elementType.getIntOrFloatBitWidth();
+  unsigned cleartextBitwidth = getCleartextBitwidth();
+  if (plaintextBitwidth < cleartextBitwidth)
+    return emitError() << "The tensor element type's bitwidth "
+                       << plaintextBitwidth
+                       << " is too small to store the cleartext, "
+                       << "which has bit width " << cleartextBitwidth << "";
+
+  auto cleartextStart = getCleartextStart();
+  if (cleartextStart < 0 || cleartextStart >= plaintextBitwidth)
```
asraa, line 41: nice - this implies `plaintextBitwidth` is greater than zero too.

asraa (approved):
LGTM!

lib/Dialect/LWE/IR/LWEDialect.cpp
j2kun: Done

j2kun:
@asraa found some prior art here, a pre-commit hook that lets you specify arbitrary search/replace regex. I have it limited to running only on cpp/h files, but if we want to add more hooks that include non-cpp/h files, we can probably work around that.

E.g., we could do something like check for // FIXME and // TODO without an attached github issue. Thoughts?

asraa (approved):
RE the // FIXME, // TODO - that would be great! Feel free to tackle in another PR.

j2kun:
One more small change to manage internal testing: removing the changes to lit config (pipefail=false) and instead using LLVM's `not` binary to invert status codes on failure. The Google-internal lit tester sadly forces pipefail, and ignores the external lit config anyway (sigh). But it's not too bad.

Issue #110, opened 2023-08-18
feat: dialect to represent combinational circuits


asraa:
Some programs may benefit from translation into a combinational circuit with logic gates that can then be optimized by tools (like Yosys). 

These programs can eventually target FHE schemes that implement boolean logic / LUTs (such as TFHE).

In order to integrate combinational lgoic optimizations, we need a dialect or IR that can represent these programs, and then create passes that may invoke programs like ABC through Yosys to optimize gates.

The ops involved would be LUTs (perhaps with rewrite patterns that would be able to optimize these LUTs (related: https://github.com/google/fully-homomorphic-encryption/issues/62 @johannesmono) or fix the LUTs to certain sizes) and boolean gates.

@j2kun

asraa:
CIRCT has a comb dialect: https://circt.llvm.org/docs/Dialects/Comb/

asraa:
WDYT about CIRCT as a dependency? IMO yes it's large, but I think it'd be great to re-use the existing comb dialect.

j2kun:
I recall they were missing a LUT operation, but they seem to have added it recently: [comb.truth_table (::circt::comb::TruthTableOp)](https://circt.llvm.org/docs/Dialects/Comb/#combtruth_table-circtcombtruthtableop)

It is a big dependency... I'm just taking quick notes here to see what else they provide that would make the dependency worthwhile

- CombToArith and CombToLLVM passes, which don't seem particularly useful for us.
- A [LowerComb pass](https://circt.llvm.org/docs/Passes/#-lower-comb) with the single pattern `TruthTableToMuxTree` that seems designed to help with their `ExportVerilog` codegen that doesn't support LUTs (not sure why... we do it with shift operators).
- An [analysis](https://github.com/llvm/circt/blob/fb1ad396f25d19ab1a11842295ce151181a7f4c2/lib/Dialect/Comb/CombAnalysis.cpp#L22) function that computes information about which bits are known to have specific values, which is currently only used in the `icmp` folder [here](https://github.com/llvm/circt/blob/16359bade7f76558b88c585e382f8061d05dbfb9/lib/Dialect/Comb/CombFolds.cpp#L3082)
- [Folders](https://github.com/llvm/circt/blob/fb1ad396f25d19ab1a11842295ce151181a7f4c2/lib/Dialect/Comb/CombFolds.cpp) which seems to have the majority of the value as a bunch of canonicalizers, including:
  - Variadic operand flattening
  - Width narrowing
  - Constant folding
[...]

PR #164, opened 2023-09-14
feat: add support for CIRCT's comb dialect

asraa:
* Adds a fork of CIRCT's Comb dialect (which also requires HW and some Support modules).

Fixes https://github.com/google/heir/issues/110

### DIFF
--- a/.pre-commit-search-and-replace.yaml
+++ b/.pre-commit-search-and-replace.yaml
@@ -3,3 +3,13 @@
 # filesystem.
 - search: '/^#include "mlir\/(?!include\/mlir\/)/'
   replacement: '#include "mlir/include/mlir/'
+# Same for llvm paths.
+- search: '/^#include "llvm\/(?!include\/llvm\/)/'
+  replacement: '#include "llvm/include/llvm/'
+# Ensure that all C++ mlir include paths include a "// from @llvm-project"
+# comment import into Google's internal filesystem.
+- search: '/^#include ("mlir\/.*")$/'
+  replacement: '#include \1 // from @llvm-project'
+# Same for llvm paths.
+- search: '/^#include ("llvm\/.*")$/'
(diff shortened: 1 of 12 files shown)


j2kun:
Still in a draft state, but you assigned me so I assume you want at least a pre-review. It LGTM! This seems like tedious work and so thank you for doing it.

j2kun (approved):

asraa:
Ready now! Note that the last change was that I excluded `third_party` from pre-commit formatting. I'm staging a PR that auto-updates the third_party folder weekly :)

asraa:
@j2kun this is ready for re-review! Slimmed down - plus added a few regex's to ensure that we're adding `// from @llvm-project` on MLIR/LLVM includes.

include/Dialect/Comb/BUILD
j2kun, line 1: Can you add the docs calls as well? Otherwise the website deployment will fail

To the dialect: 

```
        (
            ["-gen-dialect-doc"],
            "CombDialect.md",
        ),
```

To the types:

```
        (
            ["-gen-typedef-doc"],
            "CombTypes.md",
        ),
```

To the ops

```

        (
            ["-gen-op-doc"],
            "CombOps.md",
        ),
```

(Is there an enum docs? If so, add it and the website build script should pick it up automatically)

j2kun, line 1: This should be in include/Dialect/Comb/IR for consistency (add the IR path to the end)

j2kun (approved):
Two small nits, otherwise LGTM! I will do a second round of cleanup after this goes in.

include/Dialect/Comb/BUILD
asraa: No enum :/ it looks like there's types, op, dialect, and attrs.

asraa:
all set! feel free to let me know what else to clean-up

PR #165, opened 2023-09-14
Add RLWE encoding attributes

j2kun:
Adds three more encodings:

 - PolyCoefficientEncoding, for CGGI's bootstrapping rotation polynomial encoding
 - PolyEvaluationEnvoding, for standard BGV/BFV encodings
 - PolyRoundedEmbeddingEncoding, for CKKS's standard encoding

### DIFF
--- a/include/Dialect/LWE/IR/LWEAttributes.td
+++ b/include/Dialect/LWE/IR/LWEAttributes.td
@@ -19,7 +19,19 @@ class LWE_EncodingAttr<string attrName, string attrMnemonic, list<Trait> traits 
   let assemblyFormat = "`<` struct(params) `>`";
 }
 
-def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encoding"> {
+class LWE_EncodingAttrWithScalingFactor<string attrName, string attrMnemonic, list<Trait> traits = []>
+    : LWE_EncodingAttr<attrName, attrMnemonic, traits> {
+  // These parameters represent the base-2 logarithm of the scaling factor to
+  // scale cleartexts in preparation for noise growth of FHE schemes. This
+  // representation restricts the scaling factors to being powers of two.
+  let parameters = (ins
+    "unsigned":$cleartext_start,
+    "unsigned":$cleartext_bitwidth
+  );
+}
+
+def LWE_BitFieldEncoding
+  : LWE_EncodingAttrWithScalingFactor<"BitFieldEncoding", "bit_field_encoding"> {
   let summary = "An attribute describing encoded LWE plaintexts using bit fields.";
   let description = [{
     A bit field encoding of an integer describes which contiguous region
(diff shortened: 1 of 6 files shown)


asraa:
cc @b5li just wanted to CC you on the work Jeremy's doing here on encodings for LWE/RLWE schemes

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,143 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+    This attribute can be used in multiple ways:
+
+    - On a `poly.poly`, it asserts that the polynomial has been transformed
+      from a coefficient list using the canonical embedding.
+    - On a tensor of `poly.poly`, it asserts that the tensor is an RLWE
+      ciphertext for some RLWE scheme that supports the approximate embedding
+      encoding.
+
+    A typical flow for the CKKS scheme using this encoding would be to apply an
+    INTT operation to invert the canonical embedding to be a polynomial with
```
b5li, line 188: INTT -> inverse FFT

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,143 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+    %evals = arith.constant <[1, 2, 4, 5]> : tensor<4xi16>
+    // Note no `intt` operation exists in poly yet.
+    %poly1 = poly.intt %evals : tensor<4xi16> -> !poly.poly<#ring, #eval_encoding>
+    %poly2 = poly.intt %evals : tensor<4xi16> -> !poly.poly<#ring, #eval_encoding>
+    %rlwe_ciphertext = tensor.from_elements %poly1, %poly2 : tensor<2x!poly.poly<#ring, #eval_encoding>>
+    ```
+  }];
 }
+
+def RLWE_PolyRoundedEmbeddingEncoding
```
b5li, line 141: This name seems a bit unclear, as embedding could mean many things. Maybe use something else? In the literature it is usually called just "CKKS encoding" .

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,143 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+
+    This map is called the "canonical embedding" $\tau$, though conjugates are
+    discarded for efficiency because $p(\overline{z}) = \overline{p(z)}$.
+
+    Then the complete decoding operation is
+    $\textup{Decode}(p) = (1/\Delta)\tau(p)$, where $\Delta$ is a scaling
+    parameter.
+
+    The encoding operation is the inverse of the decoding operation, but
+    because inputs may be any complex numbers, a rounding step is required
```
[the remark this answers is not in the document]
j2kun: When I say "any complex numbers" here, I specifically meant to emphasize "any" meaning they may not lie in the (discrete lattice) image of tau. I can clarify that.

include/Dialect/LWE/IR/LWEAttributes.td
j2kun: Are you saying that because it actually does a (real/complex) FFT in the classical sense? Or to make the prose more general because there are multiple possible implementations and "FFT" can be viewed as generically capturing all options (algorithmically)

include/Dialect/LWE/IR/LWEAttributes.td
j2kun: I'm trying to give them names that suggest their semantic meaning, rather than tie them to a (not future proof) black-box label like CKKS. If a new scheme not invented by the same authors uses the same encoding, then the name ceases to be sensible. Maybe "InverseCanonicalEmbeddingEncoding" ?

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,143 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+    Map a polynomial $p \in \mathbb{Z}[x] / (x^N + 1) \to \mathbb{C}^{N/2}$
+    by evaluating it at the following $N/2$ points, where
+    $\omega = e^{2 \pi i / 2N}$ is the primitive $2N$th root of unity:
+
+    \[
+      \omega, \omega^3, \omega^5, \dots, \omega^{N-1}
+    \]
+
+    This map is called the "canonical embedding" $\tau$, though conjugates are
+    discarded for efficiency because $p(\overline{z}) = \overline{p(z)}$.
```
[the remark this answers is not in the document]
j2kun: z has an imaginary part if and only if conj(z) does, so throwing out conjugates is not enough to ensure the inverse is real.

I think what you're saying is that the cleartext space would not be allowed to be all of C^{N} (if you hypothetically used the full canonical embedding), but rather the subspace consisting of vectors (x_1, \dots, x_N) for which x_i = conj(x_{N-i}). That's the property that makes the inverse real, and those conditions reduce the dimension by half when N is a power of 2 (another mistake in the above is not assuming a power of 2 N; in general I seem to remember the new dimension is euler's totient of N?)

I need to recall the proof of that property ensuring the inverse is real. Do you have a reference to a proof?

If that's right, then the reason is still kind of efficiency: we don't represent the entire N-dimensional codomain with conjugates included because the second half is duplicate information enforced by those constraints (and the constraints "come first" semantically).

include/Dialect/LWE/IR/LWEAttributes.td
b5li: Yes, I meant the classical complex FFT. See for example the openFHE implementation: https://github.com/openfheorg/openfhe-development/blob/4ebb28ea7bdd894a73bc5b73e59fcfbc78253308/src/pke/lib/encoding/ckkspackedencoding.cpp#L265

include/Dialect/LWE/IR/LWEAttributes.td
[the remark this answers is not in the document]
j2kun: Er-- wait, based on what I said throwing out conjugates IS enough to make it real :)

I think I can reword it for clarity anyway.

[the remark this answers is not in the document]
b5li: It follows from the fundamental theorem of algebra https://en.wikipedia.org/wiki/Fundamental_theorem_of_algebra and https://en.wikipedia.org/wiki/Complex_conjugate_root_theorem.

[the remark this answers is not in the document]
j2kun: Let's chat about it tomorrow.

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,143 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+      ciphertext for some RLWE scheme that supports the evaluation encoding.
+
+    A typical workflow for the BFV/BGV schemes using this encoding would be
+    to apply a INTT operation to the input list of cleartexts to convert from
+    evaluation form to coefficient form, then encrypt the resulting polynomial
+    in coefficient form, then apply NTT back to the evaluation form for faster
+    multiplication of ciphertexts.
+
+    The points chosen are fixed to be the powers of a primitive root of unity
+    in the relevant coefficient domain, which allows one to use NTT/INTT to
```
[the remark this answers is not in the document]
j2kun: Done.

j2kun:
Note the website build is breaking due to https://github.com/llvm/llvm-project/commit/d339d8fede356f0acfad6d6c1bb960269f2467a9#commitcomment-128747641

Added the custom shortcode needed to this PR.

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -19,7 +19,19 @@ class LWE_EncodingAttr<string attrName, string attrMnemonic, list<Trait> traits
   let assemblyFormat = "`<` struct(params) `>`";
 }
 
-def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encoding"> {
+class LWE_EncodingAttrWithScalingFactor<string attrName, string attrMnemonic, list<Trait> traits = []>
+    : LWE_EncodingAttr<attrName, attrMnemonic, traits> {
+  let parameters = (ins
+    // These two parameters identify a scaling factor to scale cleartexts in
+    // preparation for noise growth of FHE schemes. In this format they are
+    // restricted to being powers of two.
```
asraa, line 27: This makes it seem as if `cleartext_start` and `cleartext_bitwidth` are powers of two (but instead they represent the log of the scaling factor or the bit index of the cleartext) - maybe a slight reword like "The parameters are the log of the scaling factor, which restricts the scaling factor to powers of two"

include/Dialect/Poly/IR/PolyTypes.td
```diff
@@ -20,8 +20,24 @@ def Polynomial : Poly_Type<"Poly", "poly"> {
...
-  let parameters = (ins Ring_Attr:$ring);
-  let assemblyFormat = "`<` $ring `>`";
+  let parameters = (ins
+    Ring_Attr:$ring,
+    OptionalParameter<"Attribute">:$encoding
+  );
+
+  // Skip the default builder since the custom builder enables the default
+  // value for the encoding parameter, And I couldn't get DefaultValueParameter
+  // to auto-generate a builder with the appropriate default value.
```
asraa, line 30: if you file a follow-up issue for this i'm happy to take a look - i was messing around with default values at some point.

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,165 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+
+    Example:
+
+    ```
+    #generator = #poly.polynomial<1 + x**1024>
+    #ring = #poly.ring<cmod=65536, ideal=#generator>
+    #lwe_encoding = #lwe.poly_evaluation_encoding<cleartext_start=30, cleartext_bitwidth=3>
+
+    %evals = arith.constant <[1, 2, 4, 5]> : tensor<4xi16>
+    // Note no `intt` operation exists in poly yet.
```
asraa, line 223: You may want to label this with a TODO(issue) just so we can make sure to fixup docs when it is implemented.

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -55,10 +67,165 @@ def LWE_BitFieldEncoding: LWE_EncodingAttr<"BitFieldEncoding", "bit_field_encodi
...
+    %poly2 = poly.intt %evals : tensor<4xi16> -> !poly.poly<#ring, #eval_encoding>
+    %rlwe_ciphertext = tensor.from_elements %poly1, %poly2 : tensor<2x!poly.poly<#ring, #eval_encoding>>
+    ```
+  }];
+}
+
+// TODO: does it make sense to use the bit-field parameters here? It may
+// not work sensibly because the rounding happens after scaling-post-INTT,
+// though you can still use the parameters in this attribute to infer the
+// right power of two to use.
```
asraa, line 146: I'm not sure I totally understand why it's bad to use bit-field parameters here - just because it's defining the `Delta` scaling that occurs after doesn't seem to preclude its use here (I think the semantics of bit-field encoding still applies here, and it's part of "inverse_canonical_embedding_encoding" that it occurs after the inverse canonical embedding operation). Maybe I'm misunderstanding this comment.

asraa (approved):
LGTM - some nits about comments / leftover TODOs, feel free to resolve

include/Dialect/LWE/IR/LWEAttributes.td
j2kun: I think the restriction to powers of 2 may not make sense here, since the scaling factor is not happening via bit shifting at this stage. I believe the scaling happens before rounding to integers. If there is a need for non-power-of-two scaling, then we would want different parameters. I'll file an issue

include/Dialect/LWE/IR/LWEAttributes.td
j2kun: Done

include/Dialect/Poly/IR/PolyTypes.td
j2kun: Done

include/Dialect/LWE/IR/LWEAttributes.td
j2kun: Done

Issue #108, opened 2023-08-17
secret: Add basic folders and verifiers for secret ops


j2kun:
See TODOs in code linking to this issue created in https://github.com/google/heir/pull/78

j2kun:
Another verification to add:  all ops within a secret.generic can only use as operands the values in the args of the containing basic block, and those args have to correspond to the cleartext versions of the secret.generic's operands in the same order.

j2kun:
I think this is done, but we just have to clean up the TODOs and note that instead of folders they are canonicalizers. I imagine we could put the same logic in a folder too, but I'm not sure what's the benefit of doing that.

PR #169, opened 2023-09-20
Add DistributeGeneric pass

j2kun:
First part of https://github.com/google/heir/issues/170

Next step is to figure out how to handle distributing through ops that have nested regions.

### DIFF
--- a/lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
+++ b/lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
@@ -1,0 +1,208 @@
+#include "include/Dialect/Secret/Transforms/DistributeGeneric.h"
+
+#include "include/Dialect/Secret/IR/SecretOps.h"
+#include "mlir/include/mlir/IR/PatternMatch.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace secret {
+
+#define GEN_PASS_DEF_SECRETDISTRIBUTEGENERIC
+#include "include/Dialect/Secret/Transforms/Passes.h.inc"
+
+// Inline the inner block of a secret.generic that has no secret operands.
+//
+// E.g.,
+//
+//    %res = secret.generic ins(%value : i32) {
+//     ^bb0(%clear_value: i32):
+//       %c7 = arith.constant 7 : i32
+//       %0 = arith.muli %clear_value, %c7 : i32
+//       secret.yield %0 : i32
+//    } -> (!secret.secret<i32>)
+//
+// is transformed to
+//
+//    %0 = arith.constant 0 : i32
+//    %res = arith.muli %value, %0 : i32
+//
+struct CollapseSecretlessGeneric : public OpRewritePattern<GenericOp> {
+  CollapseSecretlessGeneric(mlir::MLIRContext *context)
+      : OpRewritePattern<GenericOp>(context, /*benefit=*/2) {}
+
+  LogicalResult matchAndRewrite(GenericOp op,
+                                PatternRewriter &rewriter) const override {
+    for (Type ty : op.getOperandTypes()) {
+      if (dyn_cast<SecretType>(ty)) {
+        return failure();
+      }
+    }
+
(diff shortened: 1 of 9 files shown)


lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
```diff
@@ -0,0 +1,203 @@
...
+      : OpRewritePattern<GenericOp>(context, /*benefit=*/1) {}
+
+  LogicalResult matchAndRewrite(GenericOp op,
+                                PatternRewriter &rewriter) const override {
+    bool hasUnusedOps = false;
+    Block *body = op.getBody();
+    for (BlockArgument arg : body->getArguments()) {
+      if (arg.use_empty()) {
+        hasUnusedOps = true;
+        rewriter.updateRootInPlace(op, [&]() {
```
mr0re1, line 82: Is it safe to remove by index? Would next iteration `arg.getArgNumber()` be decremented "magically" ?

mr0re1 (approved):

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
j2kun: It looks like the implementation does indeed update argNumber for all arguments: https://github.com/llvm/llvm-project/blob/9c5b38b49fa2a1a81dac78faa0a8432eff495615/mlir/lib/IR/Block.cpp#L192

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    Operation &firstOp = body->getOperations().front();
+    // The inputs to the op are generic op's block arguments (cleartext values),
+    // and they need to change to be the corresponding generic op's normal
+    // operands (maybe secret values).
+    SmallVector<Value> newInputs;
+    // The indices of the new inputs in the original block argument list
+    SmallVector<unsigned> newInputIndices;
+    for (Value val : firstOp.getOperands()) {
```
asraa, line 146: do you need to check that `firstOp`'s operands don't use the secret inputs? (In the example case - that the firstOp does not use `clear_value`.)

j2kun: IMO this is something that should be enforced by the verifier: all ops within a secret.generic can only use as operands the values in the args of the containing basic block, and those args have to correspond to the cleartext versions of the secret.generic's operands in the same order. Added to https://github.com/google/heir/issues/108

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
j2kun: Update, you were right :) It passes tests, but hits an assertion which apparently only runs in debug mode internally

Updating to

```
    for (int i = 0; i < body->getArguments().size(); ++i) {
      BlockArgument arg = body->getArguments()[i];
      if (arg.use_empty()) { 
        hasUnusedOps = true;  
        rewriter.updateRootInPlace(op, [&]() {
          body->eraseArgument(i);
          op.getOperation()->eraseOperand(i);
        });
        // Ensure the next iteration uses the right arg number
        --i;
      }
    }
```

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
asraa: Oh, that part I understand - but my question is that it seems like from this description
```
// Split a secret.generic containing multiple ops into its underlying op if no
// secret values are included in its input
```

that this intends to peel apart the first op only when the op doesnt use the block arguments.

Now reading the code, it seems like maybe that comment is incorrect, and actually this is just breaking apart the secret.generic ops?

j2kun: Ah, yeah I think I had originally intended that, and then decided to split that into its own separate rewrite rule for simplicity. Will update the comment.

asraa (approved):

j2kun:
Sigh, due to copybara import issues we had to merge this in a special commit https://github.com/google/heir/commit/dda03f5d4f8bea3597bf44e26b7994f36f8c2cb5

Closing

PR #171, opened 2023-09-22
Add `PolyExt` dialect for "external" operations on `Poly`

mr0re1:


include/Dialect/PolyExt/IR/PolyExtOps.td
```diff
@@ -0,0 +1,43 @@
...
+class PolyExt_Op<string mnemonic, list<Trait> traits = []> :
+        Op<PolyExt_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::poly_ext";
+}
+
+def PolyExt_CModSwitchOp : PolyExt_Op<"cmod_switch", traits = [Pure, ElementwiseMappable]> {
+  let summary = "Move polynomial into a ring with different `cmod` by scaling polynomial coefficients proportionally";
+
+  let description = [{
+    `poly_ext.cmod_switch` "moves" polinomial $x$ from ring $Z/QZ[X]/(f(x))$ to the ring $Z/qZ[X]/(f(x))$.
```
asraa, line 21:
```suggestion
    `poly_ext.cmod_switch` "moves" polynomial $x$ from ring $Z/QZ[X]/(f(x))$ to the ring $Z/qZ[X]/(f(x))$.
```

lib/Dialect/PolyExt/IR/PolyExtDialect.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+// Generated definitions.
+#include "include/Dialect/PolyExt/IR/PolyExtDialect.cpp.inc"
+#define GET_OP_CLASSES
+#include "include/Dialect/PolyExt/IR/PolyExtOps.cpp.inc"
+
+namespace mlir::heir::poly_ext {
+
+//===----------------------------------------------------------------------===//
+// EncryptedArith dialect.
```
asraa, line 18:
```suggestion
// PolyExt dialect.
```

Or possible just remove the ASCII style comment!

include/Dialect/PolyExt/IR/PolyExtDialect.td
```diff
@@ -0,0 +1,25 @@
...
+// PolyExt dialect definitions
+//===----------------------------------------------------------------------===//
+
+def PolyExt_Dialect : Dialect {
+  // The namespace of the dialect.
+  // This corresponds to the string provided in `getDialectNamespace`.
+  let name = "poly_ext";
+
+  let description = [{
+    The PolyExt dialect is a place for "external" operations on
```
asraa, line 18: nit: possibly add some more description here -are external operations that operate on the polynomial ring rather than between elements of the same polynomial ring?

I'm just drive-by reviewing right now, so I think I may be missing some context about why this is a separate dialect, and ops within the poly dialect can't transform the types attribute. i will find some time to catch up.

j2kun: I think the main reasons to have a separate dialect are (1) separate the crypto-specific ops from the parts of poly we would be realistically able to upstream (would have use beyond HEIR) and (2) because the ops here are not 
 ring operations (morphisms) in any of the usual senses, and hence fall under the category of "external" https://en.wikipedia.org/wiki/External_(mathematics).

include/Dialect/PolyExt/IR/PolyExtOps.td
```diff
@@ -0,0 +1,43 @@
...
+    The move is done by scaling the polynomial coefficients by $q/Q$ and "rounding" so input and output coufficients are congruent modulo `congruence_modulus` $t$.
+    The resulting polynomial $\tilde{x} \approx (q/Q)x$ and $\tilde{x} \equiv x \mod t$.
+
+    To skip the "rounding" step, set `congruence_modulus` to 1.
+  }];
+  let arguments = (ins
+    PolynomialLike:$x,
+    // TODO: make congruence_modulus optional with default value 1
+    APIntAttr:$congruence_modulus
```
j2kun, line 30: Perhaps:

`DefaultValuedAttr<I64Attr, "1">` ? I don't think you can get away with an APInt attribute unless you define the storage class for it, and 64 bits is probably enough...

lib/Dialect/PolyExt/IR/PolyExtDialect.cpp
j2kun: +1 the ascii style comments don't do anything for me

lib/Dialect/PolyExt/IR/PolyExtDialect.cpp
```diff
@@ -0,0 +1,63 @@
...
+// Dialect construction: there is one instance per context and it registers its
+// operations, types, and interfaces here.
+void PolyExtDialect::initialize() {
+  addOperations<
+#define GET_OP_LIST
+#include "include/Dialect/PolyExt/IR/PolyExtOps.cpp.inc"
+      >();
+}
+
+namespace {
```
j2kun, line 30: nit: I would think this inner namespace is not necessary, since this cpp class won't be part of any other compilation unit.

include/Dialect/PolyExt/IR/PolyExtOps.td
```diff
@@ -0,0 +1,43 @@
...
+  let cppNamespace = "::mlir::heir::poly_ext";
+}
+
+def PolyExt_CModSwitchOp : PolyExt_Op<"cmod_switch", traits = [Pure, ElementwiseMappable]> {
+  let summary = "Move polynomial into a ring with different `cmod` by scaling polynomial coefficients proportionally";
+
+  let description = [{
+    `poly_ext.cmod_switch` "moves" polinomial $x$ from ring $Z/QZ[X]/(f(x))$ to the ring $Z/qZ[X]/(f(x))$.
+    The move is done by scaling the polynomial coefficients by $q/Q$ and "rounding" so input and output coufficients are congruent modulo `congruence_modulus` $t$.
+    The resulting polynomial $\tilde{x} \approx (q/Q)x$ and $\tilde{x} \equiv x \mod t$.
```
j2kun, line 23: > `\tilde{x} \equiv x \mod t`

Not sure what you mean here. Why not just write the formula out exactly?

```
The resulting polynomial $\tilde{x}(t) = \sum_{k=0}^{\text{degree}(x)} \lfloor q x_k /Q \rceil t^k$
```

include/Dialect/PolyExt/IR/PolyExtOps.td
```diff
@@ -0,0 +1,43 @@
...
+        Op<PolyExt_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::poly_ext";
+}
+
+def PolyExt_CModSwitchOp : PolyExt_Op<"cmod_switch", traits = [Pure, ElementwiseMappable]> {
+  let summary = "Move polynomial into a ring with different `cmod` by scaling polynomial coefficients proportionally";
+
+  let description = [{
+    `poly_ext.cmod_switch` "moves" polinomial $x$ from ring $Z/QZ[X]/(f(x))$ to the ring $Z/qZ[X]/(f(x))$.
+    The move is done by scaling the polynomial coefficients by $q/Q$ and "rounding" so input and output coufficients are congruent modulo `congruence_modulus` $t$.
```
j2kun, line 22: Remind me what the purpose of the second modulus $t$ is? cc @asraa

include/Dialect/PolyExt/IR/PolyExtOps.td
```diff
@@ -0,0 +1,43 @@
...
+    The resulting polynomial $\tilde{x} \approx (q/Q)x$ and $\tilde{x} \equiv x \mod t$.
+
+    To skip the "rounding" step, set `congruence_modulus` to 1.
```
j2kun, line 25: Wouldn't it still round, just rounding to the nearest integer?

tests/poly_ext/ops.mlir
```diff
@@ -0,0 +1,22 @@
...
+#ring1 = #poly.ring<cmod=7917, ideal=#my_poly>
+#ring2 = #poly.ring<cmod=3123, ideal=#my_poly>
+
+module {
+  func.func @test_ops(%p0 : !poly.poly<#ring1>) {
+    %cmod_switch = poly_ext.cmod_switch %p0 {congruence_modulus=117} : !poly.poly<#ring1> -> !poly.poly<#ring2>
+    return
+  }
+
+  func.func @test_elementwise_ops(%p0 : !poly.poly<#ring1>, %p1: !poly.poly<#ring1>) {
```
j2kun, line 15: While I have no problem with keeping this test, in general I think it's not necessary to write tests to exercise purely-upstream capabilities. Unless you did something specific in this PR to enable this beyond adding the ElementwiseMappable trait.

j2kun (approved):
All my comments are nitpicks, so LGTM!

include/Dialect/PolyExt/IR/PolyExtOps.td
asraa: `t` is the actual plaintext modulus, while `Q` and `q` are the ring moduli - for these schemes you can project onto a smaller ring and thus scale the noise by the ratio of `Q` and `q`. For correctness you require that this operation keeps the values equivalent modulo the plaintext modulus. Often, the moduli are chosen to be the same mod `t`, but otherwise you need to add a small delta (the rounding) to correct for that.

include/Dialect/PolyExt/IR/PolyExtDialect.td
asraa: Ah, that makes a lot of sense. I like the idea anyway of keeping `poly` as neat as possible. Thanks!

include/Dialect/PolyExt/IR/PolyExtOps.td
mr0re1: That's correct, it will still do some rounding to the nearest integer, but it's technical detail what kind of rounding is used.
I guess I better simply remove this line )

tests/poly_ext/ops.mlir
mr0re1: I agree that there aren't much "testing" value here. There main reason is a "documentation" to what operation actually looks like, e.g. I frequently refer to `tests/poly/poly_ops.mlir` to copy-paste correct syntax instead of comprehending `assemblyFormat`.

include/Dialect/PolyExt/IR/PolyExtOps.td
mr0re1: There is a potential bug in ODS. I would work on repro of this in upstream but it may take some time. I would rather push it "as is" and address the TODO after the bug is fixed/denounced.

j2kun: I also ran into some difficulties trying to make defaults work, but never figured it out. OK to submit as is.

Issue #172, opened 2023-09-25
"Optional" bison/flex dependency potentially not optional?


AlexanderViand-Intel:
I recently had to set up my environment again, and I realized that building HEIR (current main) with `bazel build @heir//...` as described in the getting started document fails with bison/flex not found if either of them isn't installed. 

However, the documentation specifies that they are optional: https://github.com/google/heir/blob/a59ec085b3aea604348a33f1b8e2f9343ba7141e/docs/content/en/docs/getting_started.md?plain=1#L16C1-L18

AlexanderViand-Intel:
Good point! Until #118 is resolved, we should adjust the documentation to either drop the "(optional)" or swap out the `bazel build @heir//...` command with something more targeted.

j2kun:
No better time to fix it than now, I'll pick it up :)

j2kun:
My impression here is that we're going to have to add flex/bison as full dependencies for yosys, or else tweak the yosys BUILD file to not include any of its lexers or parsers for the various frontends (verilog, rltil, etc.) which might not even be possible and at best would make the build brittle.

We have the right build files internally, I'd just have to port them.

asraa:
@j2kun I also have added Yosys as a library in parts of my proof of concepts.

The easiest way, IMO is actually using https://github.com/hdl/bazel_rules_hdl, since you'll need to port all the 3P dependencies as well.

asraa:
You can actually strip away a lot of these if you use the slimmest Yosys build needed (I experimented with only adding RTLIL parser / lexer), but it was helpful to just use this repo for the bison/flex dependencies:

https://github.com/asraa/heir/blob/b44349b05534b896b934736a7cbd0cab3c3cbf95/WORKSPACE#L176-L194

Issue #118, opened 2023-08-21
chore: Make the Yosys dependency optional


j2kun:
We had some initial prototypes involving yosys for optimization, and this adds flex and bison as user-facing dependencies since there's a yosys CLI invocation in some tests.

If we migrate to using Yosys as a library we can avoid those deps, and in the mean time we can make the yosys dep optional in the workspace file, so lower the barrier to entry.

PR #173, opened 2023-09-25
Add flex/bison and other yosys deps as explicit dependencies

j2kun:
Fixes https://github.com/google/heir/issues/172

Fixes https://github.com/google/heir/issues/118


j2kun:
@asraa This slightly adapts your PR, and it will also take some tweaking internally with the copybara rules. Will do that after pull_ready.

docs/content/en/docs/getting_started.md
```diff
@@ -12,16 +12,12 @@ weight: 1
     `>=5.5`
 -   A C compiler (like [gcc](https://gcc.gnu.org/) or
     [clang](https://clang.llvm.org/))
--   [Python](https://www.python.org/) (optional, for some tests)
--   GNU [flex](https://github.com/westes/flex) and
-    [bison](https://www.gnu.org/software/bison/), for the Yosys dependency
-    (optional)
```
asraa, line 18: flex and byson are required for building the yosys lib as well, right? if so, mark as `optional, for some tests`? (it might be possible to only require them for the verilog running tests by making yosys an additional dep instead of bundling in the test utilities)

asraa (approved):

asraa (approved):
Thanks!

j2kun:
Migrating to an internal change, since the yosys dependency changes conflict with internal copybara transforms.

PR #174, opened 2023-09-26
Add verifiers for secret.generic and yield

j2kun:
Note one of the verifiers is implemented via a new trait `HasParent<"GenericOp">`

### DIFF
--- a/lib/Dialect/Secret/IR/SecretOps.cpp
+++ b/lib/Dialect/Secret/IR/SecretOps.cpp
@@ -37,6 +37,25 @@ ParseResult YieldOp::parse(OpAsmParser &parser, OperationState &result) {
                  parser.resolveOperands(opInfo, types, loc, result.operands));
 }
 
+LogicalResult YieldOp::verify() {
+  // Trait verifier ensures parent is a GenericOp
+  auto parent = llvm::cast<GenericOp>(getParentOp());
+
+  for (int i = 0; i < getValues().size(); ++i) {
+    auto yieldSecretType = SecretType::get(getValues().getTypes()[i]);
+    if (yieldSecretType != parent.getResultTypes()[i]) {
+      return emitOpError()
+             << "If a yield op returns types T, S, ..., then the enclosing "
+                "generic op must have result types secret.secret<T>, "
+                "secret.secret<S>, ... But this yield op has operand types: "
+             << getValues().getTypes()
+             << "; while the enclosing generic op has result types: "
+             << parent.getResultTypes();
+    }
+  }
+  return success();
+}
+
 void GenericOp::print(OpAsmPrinter &p) {
   ValueRange inputs = getInputs();
   if (!inputs.empty())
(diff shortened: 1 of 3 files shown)


j2kun:
Oops, looks like my verifiers found some bad test setup in some other `secret` tests. Will patch this tomorrow morning

j2kun:
Fixed the issues, should be good now

lib/Dialect/Secret/IR/SecretOps.cpp
```diff
@@ -94,6 +112,59 @@ ParseResult GenericOp::parse(OpAsmParser &parser, OperationState &result) {
...
+             << " and generic operand argument of type " << operand.getType()
+             << ". If the operand is not secret, it must be the same type as "
+                "the block argument";
+    }
+
+    if (operandType && arg.getType() != operandType.getValueType()) {
+      return emitOpError() << "Type mismatch between block argument "
+                           << arg.getArgNumber() << " of type " << arg.getType()
+                           << " and generic operand argument of type "
+                           << operand.getType()
```
code-perspective, line 136: Im guessing Operand.getType() will print out the ValueType within Secret.secret.

Im slightly confused by the error message on this one being similar to the one above for non-secret oeprand. Perhaps update the message to "type mismatch between block argument <> of type <> and **ValueType** of generic operand argument of type <>"

lib/Dialect/Secret/IR/SecretOps.cpp
```diff
@@ -20,6 +20,24 @@ ParseResult YieldOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  // Trait verifier ensures parent is a GenericOp
+  auto parent = llvm::cast<GenericOp>(getParentOp());
+
+  for (int i = 0; i < getValues().size(); ++i) {
+    auto yieldSecretType = SecretType::get(getValues().getTypes()[i]);
+    if (yieldSecretType != parent.getResultTypes()[i]) {
+      return emitOpError()
+             << "Result types of yield op must match result types of enclosing "
+                "generic op. Yield op result types: "
+             << getValues().getTypes()
```
code-perspective, line 33: should this be yieldSecretType ?

(Im not entirely familiar with how MLIR error messages work, but I would assume it will throw some form of array indexing error)

lib/Dialect/Secret/IR/SecretOps.cpp
```diff
@@ -20,6 +20,24 @@ ParseResult YieldOp::parse(OpAsmParser &parser, OperationState &result) {
...
+             << "; enclosing generic op result types: "
+             << parent.getResultTypes();
```
code-perspective, line 35: Should this be parent.getResultTypes()[i] ?

(Im not entirely familiar with how MLIR error messages work, but I would assume it will throw some form of array indexing error)

code-perspective (commented):
Thanks!

Just a couple of nits on error messages.

lib/Dialect/Secret/IR/SecretOps.cpp
j2kun: I could just print one type, but I think it would be more helpful to show all the result types here. e.g., if the yield takes as input i32, i32, and the generic has results secret.secret<i32> and secret.secret<i64> , it might be confusing somehow if you're looking at the first argument and not realizing it's reporting an error about the second argument. But if you see the whole thing then it might be easier to read.

lib/Dialect/Secret/IR/SecretOps.cpp
j2kun: Clarified

lib/Dialect/Secret/IR/SecretOps.cpp
j2kun: Clarified

lib/Dialect/Secret/IR/SecretOps.cpp
code-perspective: Ah got it! Yes whole thing is easier.

My bad! i assumed the iterator would throw some form of error, but I tried it and it prints just fine (getting used to MLIR)

code-perspective (approved):
Thanks!

Issue #111, opened 2023-08-18
feat: Integrate tools to optimize circuits


asraa:
Related: https://github.com/google/heir/issues/110

We can take a high-level arithmetic program and create optimization passes that can invoke tools like Yosys and ABC to optimize and output a combinational circuit.

The combinational circuit can then be lowered to boolean FHE schemes.

asraa:
See this other tool here: https://github.com/lsils/mockturtle

And some discussion about this here: https://discourse.llvm.org/t/integration-with-abc-optimizer/65264/2

asraa:
I think it would be nice to go

Arith / other high-level -> Combinational logic block -> [ABC Transform] -> Optimized combinational logic block

Combinational logic can then be lowered to  schemes that support boolean gates / luts.

asraa:
I also noticed this `ArithToCombPass` from CIRCT: 
https://github.com/llvm/circt/blob/main/include/circt/Transforms/Passes.td#L56C5-L71

j2kun:
I think we can mark this one as fixed now.

j2kun:
Reopening so we can migrate the TODOs to new issues

PR #184, opened 2023-10-02
feat: convert a Yosys's RTLIL module to MLIR (Arith + Comb)

asraa:
Part of https://github.com/google/heir/issues/111

### DIFF
--- a/lib/Transforms/YosysOptimizer/LUTImporterTest.cpp
+++ b/lib/Transforms/YosysOptimizer/LUTImporterTest.cpp
@@ -1,0 +1,187 @@
+#include <cstddef>
+#include <memory>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "gmock/gmock.h"  // from @googletest
+#include "gtest/gtest.h"  // from @googletest
+#include "include/Dialect/Comb/IR/CombDialect.h"
+#include "include/Dialect/Comb/IR/CombOps.h"
+#include "kernel/yosys.h"  // from @at_clifford_yosys
+#include "lib/Transforms/YosysOptimizer/LUTImporter.h"
+#include "llvm/include/llvm/ADT/STLExtras.h"            // from @llvm-project
+#include "llvm/include/llvm/ADT/SmallVector.h"          // from @llvm-project
+#include "mlir/include/mlir//IR/Location.h"             // from @llvm-project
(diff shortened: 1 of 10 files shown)


lib/Transforms/YosysOptimizer/RTLILImporter.cpp
```diff
@@ -0,0 +1,162 @@
...
+      auto concatOp = b.create<comb::ConcatOp>(retBits);
+      returnValues.push_back(concatOp.getResult());
+    } else {
+      returnValues.push_back(getWireValue(resultWire));
+    }
+  }
+  b.create<func::ReturnOp>(returnValues);
+
+  // OperationFolder folder(context);
+  // folder.tryToFold(function);  // TODO: add failed(), LogicalResult
```
asraa, line 157: @j2kun: QQ here: the resulting function has redundant constants as I created the arguments for each op - I expected this folder to de-dupe them, but this didn't end up working. Have you ever come across anything else? The builder also uses `b.createOrFold<arith::ConstantOp>` where necessary.

asraa: Maybe I should be using this for constant insertion https://mlir.llvm.org/doxygen/classmlir_1_1OperationFolder.html#a3573a3dcbb3cbb1b3f48b57c358851c1 - I'll give it a shot

lib/Transforms/YosysOptimizer/LUTImporterTest.cpp
```diff
@@ -0,0 +1,164 @@
...
+using ::testing::Test;
+
+// Returns a list of cell names that are topologically ordered using the Yosys
+// toder output. This is extracted from the lines containing cells in the
+// output:
+// -- Running command `torder -stop * P*;' --
+
+// 14. Executing TORDER pass (print cells in topological order).
+// module test_add
+//   cell $abc$167$auto$blifparse.cc:525:parse_blif$168
```
j2kun, line 22: You don't end up using these names for anything besides equality comparison, do you? Well just in case, if you end up needing to do additional random string processing on them, you can also consider running `rename -enumerate`

https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/rename.html

Which will automatically rename all the cells to have simpler increasing integer names.

j2kun (approved):
LGTM! This looks like it was a slog. Very well done.

PR #185, opened 2023-10-02
Add more poly ops needed for lowering polymul

j2kun:
Adding poly.degree, poly.monomial, which will be useful for lowering poly.mul since I can implement the poly remainder routine using poly ops before the lowering starts.

Also encountered a bug in the bit width of the type conversion, and this led me down a long path that ended with a simpler poly.add lowering, thanks to Asra and Shruthi for pair programming on that.

### DIFF
--- a/lib/Conversion/PolyToStandard/PolyToStandard.cpp
+++ b/lib/Conversion/PolyToStandard/PolyToStandard.cpp
@@ -7,6 +7,7 @@
 #include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
 #include "mlir/include/mlir/Dialect/Func/Transforms/FuncConversions.h"  // from @llvm-project
 #include "mlir/include/mlir/Dialect/Linalg/IR/Linalg.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/SCF/IR/SCF.h"        // from @llvm-project
 #include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
 #include "mlir/include/mlir/Dialect/Utils/StructuredOpsUtils.h"  // from @llvm-project
 #include "mlir/include/mlir/IR/PatternMatch.h"  // from @llvm-project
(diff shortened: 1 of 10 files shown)


lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -1,12 +1,15 @@
 #include "include/Conversion/PolyToStandard/PolyToStandard.h"
 
+#include <iostream>
```
asraa, line 3: nit: remove debugging include

tests/poly/lower_monomial.mlir
```diff
@@ -0,0 +1,16 @@
...
+#cycl_2048 = #poly.polynomial<1 + x**1024>
+#ring = #poly.ring<cmod=4294967296, ideal=#cycl_2048>
+
+func.func @test_monomial() {
+  // CHECK: arith.constant 1023
+  %deg = arith.constant 1023 : index
+  // CHECK: arith.constant 5
+  %five = arith.constant 5 : i32
+  // CHECK: arith.constant
+  // CHECK: tensor.insert
```
asraa, line 13: nit: could you assert that the check inserts at index 5 for correctness?

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -107,6 +115,71 @@ struct ConvertToTensor : public OpConversionPattern<ToTensorOp> {
...
+        op,
+        /*resultTypes=*/TypeRange{b.getIndexType()},
+        /*operands=*/ValueRange{initIndex.getResult()},
+        /*beforeBuilder=*/
+        [&](OpBuilder &nestedBuilder, Location nestedLoc, ValueRange args) {
+          Value index = args[0];
+          ImplicitLocOpBuilder b(nestedLoc, nestedBuilder);
+          auto coeff = b.create<tensor::ExtractOp>(coeffs, ValueRange{index});
+          auto cmpOp =
+              b.create<arith::CmpIOp>(arith::CmpIPredicate::sgt, coeff, c0);
```
asraa, line 168: Small question for me: By using "sgt", do you interpret the unsigned coefficient as signed? Would that be an issue for large coeffs?

It's possible here I think to also use a "not equal to" op.

asraa (commented):
Nice! Looks really good. I'll add pull_ready, but just have one question on degree lowering.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
j2kun: Done

tests/poly/lower_monomial.mlir
j2kun: Done.

lib/Conversion/PolyToStandard/PolyToStandard.cpp
j2kun: Fixed.

asraa (approved):

PR #180, opened 2023-09-27
Lower poly mul to standard

j2kun:

PR #188, opened 2023-10-05
Add poly.leading_term and poly.monomial_mul

j2kun:


include/Dialect/Poly/IR/PolyOps.td
```diff
@@ -48,25 +48,36 @@ def Poly_MulConstantOp : Poly_Op<"mul_constant", [ElementwiseMappable, AllTypesM
...
   let summary = "Create a polynomial that consists of a single monomial.";
   let arguments = (ins AnyInteger:$coefficient, Index:$degree);
   let results = (outs Polynomial:$output);
-  let assemblyFormat = "operands attr-dict `:` `(` type($coefficient) `,` type($degree) `)` `->` qualified(type($output))";
+}
+
+def Poly_RotateOp: Poly_Op<"rotate", [AllTypesMatch<["input", "output"]>]> {
+  let summary = "Multiply a polynomial by a monomial $x^k$";
+  let description = [{
+    This can be interpreted as rotating the coefficients by $k$ positions to
```
mr0re1, line 75: n00b question. Does "rotation" explanation only works with specific types of ring ideals? IIUC in general multiplication by monomial will cause transformation that is hard to call "rotation".
We can either skip "multiple by monomial" part or don't add this op and use `mul` instead.

j2kun: Yes, I have a verifier in https://github.com/google/heir/pull/188/commits/da95c8deedf7447b78b9060d17fdca140bb94398 that ensures this. Mainly I'm adding this so that the lowering of poly.mul can be more incremental: first lower poly.mul to a sequence of simpler poly ops: a naive polymul followed by a remainder op. The remainder op requires multiplication of a polynomial by a monomial without any modulus operation, and that is the same as picking a sufficiently large `n` and using `x^n - 1` as the modulus.

It's not entirely pretty, but once I get this damn polymul lowering implemented somehow, we can revisit it more deliberately.

And moreover, this is probably not as relevant for FHE because most of our rings will be x^n + 1, and polymul will be lowered completely differently in that case. The polymul lowering that needs this is mainly to be general enough that it can work for "any" polynomial ring (I have some comments in https://github.com/google/heir/pull/180 about why it's still not "any" ring even in general, because the leading coefficient of the ideal's generator needs to have a multiplicative inverse in the coefficient group).

j2kun: Updating the docs is a good idea though.

j2kun: Now that I think about it, how bad would it be to support rotation in general, even if it doesn't have a nice interpretation as multiplication by a monomial? Not sure, but with my latest commit we can enable that just by deleting the verifier.

asraa: IMO since the op is called `rotate` on a polynomial's coefficients, it doesn't seem to necessitate having anything to do with monomial multiplication. I would expect that higher level ops would be aware that monomial multiplication is lowering to a slot rotation when the ring satisfies the property.

(Also: for TFHE and other libs, don't we use the anticyclic part, `x^n + 1`, so mul by monomial would result in some sign flipping?)

lib/Conversion/PolyToStandard/PolyToStandard.cpp
```diff
@@ -113,6 +113,37 @@ struct ConvertToTensor : public OpConversionPattern<ToTensorOp> {
...
+    unsigned numTerms = tensorType.getShape()[0];
+    coeffs.reserve(numTerms);
+    // This is inefficient for large-degree polys, but as of this writing we
+    // don't have a lowering that uses a sparse representation.
+    for (int i = 0; i < numTerms; ++i) {
+      coeffs.push_back(rewriter.getIntegerAttr(eltTy, 0));
+    }
+    for (const auto &term : attr.getPolynomial().getTerms()) {
+      coeffs[term.exponent.getZExtValue()] = rewriter.getIntegerAttr(
+          eltTy, term.coefficient.trunc(eltTy.getIntOrFloatBitWidth()));
```
asraa, line 139: Does this need a verifier that the coefficient terms have size less than the modulus bit width? or else we can APInt::mod them here? (also, are they potentially negative? test case just uses small positive int)

include/Dialect/Poly/IR/PolyOps.td
j2kun: The thing that bothers me slightly is that rotation is not a ring operation in any sense I can think of, and the sense that we need it for TFHE is indeed multiplication by a monomial, not plain rotation. Should I rename the op `monomial_mul` and then just only support it for the rings we need it for?

lib/Conversion/PolyToStandard/PolyToStandard.cpp
j2kun: Yes, I think I should add a verifier for poly.constant to make sure the input polynomial agrees with the ring in all senses (low enough degree, coeffs, etc.) adding this as issue https://github.com/google/heir/issues/190

include/Dialect/Poly/IR/PolyOps.td
j2kun: Renamed to `monomial_mul`

asraa (approved):

PR #191, opened 2023-10-06
feat: add yosys optimizer pass

asraa:
Part of https://github.com/google/heir/issues/111

This adds a global yosys optimization on a program - it booleanizes the arithmetic circuit, and then converts it into LUTs from the comb dialect.

### DIFF
--- a/lib/Transforms/YosysOptimizer/RTLILImporter.cpp
+++ b/lib/Transforms/YosysOptimizer/RTLILImporter.cpp
@@ -1,6 +1,7 @@
 #include "lib/Transforms/YosysOptimizer/RTLILImporter.h"
 
 #include <cassert>
+#include <sstream>
 #include <string>
 #include <utility>
 
@@ -23,6 +24,19 @@ using ::Yosys::RTLIL::Module;
 using ::Yosys::RTLIL::SigSpec;
 using ::Yosys::RTLIL::Wire;
 
+llvm::SmallVector<std::string, 10> getTopologicalOrder(
+    std::stringstream &torderOutput) {
+  llvm::SmallVector<std::string, 10> cells;
+  std::string line;
+  while (std::getline(torderOutput, line)) {
+    auto lineCell = line.find("cell ");
+    if (lineCell != std::string::npos) {
+      cells.push_back(line.substr(lineCell + 5, std::string::npos));
+    }
+  }
+  return cells;
+}
+
 void RTLILImporter::addWireValue(Wire *wire, Value value) {
   wireNameToValue.insert(std::make_pair(wire->name.str(), value));
 }
@@ -33,7 +47,9 @@ Value RTLILImporter::getWireValue(Wire *wire) {
   return wireNameToValue.at(wireName);
 }
 
-Value RTLILImporter::getBit(const SigSpec &conn, ImplicitLocOpBuilder &b) {
+Value RTLILImporter::getBit(
+    const SigSpec &conn, ImplicitLocOpBuilder &b,
+    llvm::MapVector<Wire *, SmallVector<Value>> &retBitValues) {
   // Because the cells are in topological order, and Yosys should have
   // removed redundant wire-wire mappings, the cell's inputs must be a bit
   // of an input wire, in the map of already defined wires (which are
@@ -50,9 +66,13 @@ Value RTLILImporter::getBit(const SigSpec &conn, ImplicitLocOpBuilder &b) {
         b.getIntegerAttr(b.getIntegerType(1), bit.as_int()));
     return constantOp;
   }
-  // Extract the bit of the multi-bit input wire.
+  // Extract the bit of the multi-bit input or output wire.
   assert(conn.as_bit().is_wire());
   auto bit = conn.as_bit();
+  if (retBitValues.contains(bit.wire)) {
+    auto offset = retBitValues[bit.wire].size() - bit.offset - 1;
+    return retBitValues[bit.wire][offset];
+  }
   auto argA = getWireValue(bit.wire);
   auto extractOp =
       b.createOrFold<comb::ExtractOp>(b.getI1Type(), argA, bit.offset);
@@ -124,7 +144,7 @@ func::FuncOp RTLILImporter::importModule(
 
     SmallVector<Value, 4> inputValues;
     for (const auto &conn : getInputs(cell)) {
-      inputValues.push_back(getBit(conn, b));
+      inputValues.push_back(getBit(conn, b, retBitValues));
     }
     auto *op = createOp(cell, inputValues, b);
     auto resultConn = getOutput(cell);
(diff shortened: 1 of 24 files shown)


tools/heir-opt.cpp
```diff
@@ -102,6 +104,15 @@ int main(int argc, char **argv) {
...
   mlir::heir::poly::registerPolyToStandardPasses();
   mlir::heir::bgv::registerBGVToPolyPasses();
+  // Register yosys optimizer pipeline
+  const char *abcEnvPath = std::getenv("HEIR_ABC_PATH");
+  const char *yosysRunfilesEnvPath = std::getenv("HEIR_YOSYS_LIBS");
+  if (abcEnvPath == nullptr || yosysRunfilesEnvPath == nullptr) {
+    llvm::errs() << "external files for yosys optimizer not found, please set "
+                    "HEIR_ABC_PATH and HEIR_YOSYS_LIBS\n";
+    return EXIT_FAILURE;
+  }
```
asraa, line 114: I debated placing this inside of the reigstration function and having the register func instead return a logical result...

asraa (commented):
Ready for review now! 

I'm not 10000% sold on the env vars, only because it makes running as a bazel-bin binary one step harder, but that's not super high priority.

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    auto topologicalOrder = getTopologicalOrder(cellOrder);
+
+    // Insert the optimized MLIR.
+    LUTImporter lutImporter = LUTImporter(&getContext());
+    Yosys::RTLIL::Design *design = Yosys::yosys_get_design();
+    func::FuncOp func =
+        lutImporter.importModule(design->top_module(), topologicalOrder);
+    op.getBody().takeBody(func.getBody());
+
+    func.dump();
```
j2kun, line 92: I like the idea of printing, but let's hide it behind an LLVM_DEBUG. See, e.g., this little PR I upstreamed to MLIR just adding debug messages: https://github.com/llvm/llvm-project/pull/69476

It will get printed when you run with `--debug`, and maybe let's dress it up a little with `llvm::outs() << "Converted & optimized func op via yosys. Input func:\n" << op << "\n\nOutput func:\n" << func << std::endl`

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    pm.addPass(createYosysOptimizer(runfiles, abcPath, options.abcFast));
+    pm.addPass(mlir::createCanonicalizerPass());
+  };
+}
+
+void registerYosysOptimizerPipeline(std::string runfiles, std::string abcPath) {
+  PassPipelineRegistration<YosysOptimizerPipelineOptions>(
+      "yosys-optimizer", "The yosys optimizer pipeline.",
+      buildYosysPipelineLambda(runfiles, abcPath));
+}
```
j2kun, line 119: Would the following simplify the lambda? I'm just not quite sure why it needs the extra function to generate the lambda.

```
void registerYosysOptimizerPipeline(std::string runfiles, std::string abcPath) {
  PassPipelineRegistration<YosysOptimizerPipelineOptions>(
      "yosys-optimizer", "The yosys optimizer pipeline.",
       [&](OpPassManager &pm, const YosysOptimizerPipelineOptions &options) {
         pm.addPass(createYosysOptimizer(runfiles, abcPath, options.abcFast));
         pm.addPass(mlir::createCanonicalizerPass());
       }));
}
```

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -0,0 +1,122 @@
...
+abc -exe {3} -lut 3 {4};
+opt_clean -purge;
+rename -hide */c:*; rename -enumerate */c:*;
+techmap -map {2}/map_lut_to_lut3.v; opt_clean -purge;
+hierarchy -generate * o:Y i:*; opt; opt_clean -purge;
+clean;
+)";
+
+struct YosysOptimizer
+    : public PassWrapper<YosysOptimizer, mlir::OperationPass<mlir::ModuleOp>> {
```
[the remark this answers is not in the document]
j2kun: Actually, just put it in a github issue and assign it to me. I think I can make it work.

j2kun (changes requested):
I know I left a bunch of comments, but I think this PR is very straightforward and easy to understand. Very well done!

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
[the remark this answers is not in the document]
asraa: Oh yes! The initial reason was because of the custom init method which eventually left over the defunct tablegen - I think the issue was I also have custom registration methods, so would probably need to remove the directives to generate the pass reg methods that conflicted with what I wanted. If it's not as trivial as that, I'll create a GH issue!

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
asraa: hah! this was leftover from some debugging, but this is great to know. This will be useful as this pass adds the local optimization option.

tests/BUILD
```diff
@@ -23,6 +23,8 @@ filegroup(
         "@llvm-project//mlir:libmlir_runner_utils.so",
         "@llvm-project//mlir:mlir-cpu-runner",
         "@llvm-project//mlir:mlir-opt",
+        "@edu_berkeley_abc//:abc",
```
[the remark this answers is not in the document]
asraa: Sooo agree - one thing I was thinking of was that if we can skip registration based on an env var in case the yosys deps need to be skipped - this goes back to this earlier thing: https://github.com/google/heir/pull/191#discussion_r1364256334

I think I'll go ahead and implement that - I'll add some documentation saying that if you'd like to skip the Yosys build, add this env var and we'll build with a selective configuration.

lib/Transforms/YosysOptimizer/RTLILImporter.cpp
```diff
@@ -140,12 +159,15 @@ func::FuncOp RTLILImporter::importModule(
     assert(retBitValues.contains(output.as_wire()) ||
            retBitValues.contains(output.as_bit().wire));
     if (conn.second.chunks().size() == 1) {
-      Value connValue = getBit(conn.second, b);
+      // This may be a single bit, or it may be a whole wire.
+      if (conn.second.is_wire()) {
```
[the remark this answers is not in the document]
asraa: ah, thanks for the catch! the switch on whether it's a wire is now inside `getBit`, so this is a defunct branch.

tools/heir-opt.cpp
asraa: Added some docs about disabling the build/test yosys dependent targets, and also included added documentation on the yosys-optimizer pass env var requirements and how to disable the pass.

asraa: This also fixes https://github.com/google/heir/issues/118 with a test tag filter & build setting now.

j2kun (approved):
LGTM!

j2kun:
Hooray!! 🎉

asraa:
I know! one step closer :P

PR #196, opened 2023-10-13
Add Polynomial::toIdentifier

j2kun:
This is needed to define the name of the remainder func when lowering poly.mul in https://github.com/google/heir/pull/180


lib/Dialect/Poly/IR/Polynomial.cpp
```diff
@@ -66,6 +67,39 @@ void Polynomial::print(raw_ostream &os) const {
   }
 }
 
+std::string Polynomial::toIdentifier() const {
```
asraa, line 70: just for simplicity could you reuse the print logic with a private hidden printer that takes the os and the separator? ` + ` or `_`?

It looks like you have different ways of handling the exponent multiplicatin though (`**`) is that for simplicity or because of some reason not to have asteriks?

j2kun: Done.

asraa (approved):

Issue #200, opened 2023-10-16
poly: make the dense elements attr more robust


j2kun:
In `ConvertMul::matchAndRewrite` there are a few places in which we create a `arith.constant` op using a dense elements attr. The number type for this op has caused me some pain, in that if you get it wrong by having a tensor element type that disagrees with the value type passed in, you get a nondescript assertion failure that is hard to debug.

The current code kind of punts because the tests are all 64-bit, and so it passes in something like `0L` and it fits. But in general we'd want to inspect the types and set up the dense elements attr to be large enough to fit. For example, if the two input polynomials have coefficients that fit in `i64`, we'd need `i128`, if they're `i16` we could use `i32`. Setting up tests that simply lower `poly.mul` with arguments using such coefficients should demonstrate the problem.

PR #203, opened 2023-10-17
MLIR cpu runner tests for lowering poly mul

j2kun:


include/Dialect/Polynomial/IR/PolynomialAttributes.td
```diff
@@ -39,18 +39,26 @@ def Ring_Attr : Polynomial_Attr<"Ring", "ring"> {
     An attribute specifying a polynomial quotient ring with integer
     coefficients, $\mathbb{Z}/n\mathbb{Z}[x] / (p(x))$.
 
+    `cwidth` is the bit width of the coefficient storage type,
     `cmod` is the coefficient modulus $n$, and `ideal` is the ring ideal
     $(p(x))$. Because all ideals in a single-variable polynomial ring are
     principal, the ideal is defined by a single polynomial.
 
       #ring = #polynomial.ring<cmod=1234, ideal=#polynomial.polynomial<x**1024 + 1>>
```
asraa, line 47: nit: add `cwidth` here since it's required

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -145,7 +145,7 @@ struct ConvertConstant : public OpConversionPattern<ConstantOp> {
     }
     for (const auto &term : attr.getPolynomial().getTerms()) {
       coeffs[term.exponent.getZExtValue()] = rewriter.getIntegerAttr(
-          eltTy, term.coefficient.trunc(eltTy.getIntOrFloatBitWidth()));
+          eltTy, term.coefficient.zextOrTrunc(eltTy.getIntOrFloatBitWidth()));
```
asraa, line 148: hmm i'd definitely expect there to be tests that would detect if something weird is happening here, but is there an issue with zero extending the elements that are otherwise using signed arithmetic? (is the behavior different if you use `sextOrTrunc`?) I wonder if nothing is being detected because the coeff width is usually just large enough for the coefficients?

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -439,7 +444,10 @@ struct ConvertMul : public OpConversionPattern<MulOp> {
     // set the dense elements attr properly based on that (i.e., what if it's an
     // i16 or i128?)
     auto polymulOutput = b.create<arith::ConstantOp>(
-        polymulTensorType, DenseElementsAttr::get(polymulTensorType, 0L));
+        polymulTensorType,
+        DenseElementsAttr::get(
+            polymulTensorType,
+            APInt(polymulTensorType.getElementTypeBitWidth(), 0L)));
```
asraa, line 450: does this fix the TODO comment above since you're using the APInt with the element bit width now?

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -682,18 +690,19 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
   auto toTensorOp = builder.create<ToTensorOp>(inputType, whileOp.getResult(0));
-  auto remainderModOp = builder.create<arith::ConstantOp>(
-      inputType,
-      DenseElementsAttr::get(inputType, ring.getCmod().getZExtValue()));
-  auto remainderCoeffsModded = builder.create<arith::RemUIOp>(
-      toTensorOp.getResult(), remainderModOp.getResult());
+  auto remainderModArg = builder.create<arith::ConstantOp>(
+      inputType, DenseElementsAttr::get(
+                     inputType, APInt(inputType.getElementTypeBitWidth(),
+                                      ring.getCmod().getZExtValue())));
```
asraa, line 696: here again, correct me if i'm wrong but zero extending the cmod seems like it would change its value if you're using it for a signed remainder below

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -724,13 +733,7 @@ void PolynomialToStandard::runOnOperation() {
...
   PolynomialToStandardTypeConverter typeConverter(context);
 
-  target.addLegalDialect<arith::ArithDialect>();
-
-  // target.addIllegalDialect<PolynomialDialect>();
-  target.addIllegalOp<FromTensorOp, ToTensorOp, AddOp, SubOp, LeadingTermOp,
-                      MonomialOp, MulScalarOp, MonomialMulOp, ConstantOp,
-                      MulOp>();
-
+  target.addIllegalDialect<PolynomialDialect>();
```
asraa, line 736: YAY!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

include/Dialect/Polynomial/IR/PolynomialAttributes.td
j2kun: This is actually a bit of a trick. The custom parser/printer for this attribute doesn't include the `cwidth`, and the builder infers it from the `cmod` attribute. So it's not needed in the docs anywhere, really. I will just remove that line 42.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: Let me use `sextOrTrunc` and then I was going to follow up this PR with more tests, and I will try to exercise this one as well.

j2kun:
Fixes https://github.com/google/heir/issues/200

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: Done.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: Done.

PR #206, opened 2023-10-24
poly: sub is not commutative

AlexanderViand-Intel:
While reviewing `poly` ahead of the meeting, I noticed that the `Poly_BinOp` always included the `Commutative` trait, which marked `poly.sub` as commutative, too. This removes `Commutative` from `BinOp` and instead sets it manually on `poly.add` and `poly.mul`.


asraa (approved):

j2kun:
Yay Alex's first PR!

PR #210, opened 2023-10-24
Add a means to generate poly mul lowering tests

j2kun:
This PR creates a simple test harness to generate poly-mul lowering tests from a static config file, using sympy to compute the coefficients of the expected result


tests/poly/runner/lower_mul_0.mlir
```diff
@@ -0,0 +1,39 @@
+// WARNING: this file is autogenerated. Do not edit manually, instead see
+// tests/poly/runner/generate_test_cases.py
+
+//-------------------------------------------------------
+// entry and check_prefix are re-set per test execution
+// DEFINE: %{entry} =
```
asraa, line 6: very cool! did not know about DEFINEs

tests/poly/runner/generate_test_cases.py
```diff
@@ -0,0 +1,222 @@
...
+  func.call @printMemrefI32(%U) : (memref<*xi32>) -> ()
+  return
+}}}}
+// expected_result: {{expected}}
+// CHECK_TEST_{{test_number}}: {{coefficient_list}}
+"""
+
+GEN_TENSOR_OP_TEMPLATE_WITH_TRUNC = """
+  %3 = poly.to_tensor %2 : !poly_ty_{{test_number}} -> tensor<{{degree}}x{{container_type}}>
+  %tensor = arith.{trunc_ext_op} %3 : tensor<{{degree}}x{{container_type}}> to tensor<{{degree}}xi32>
```
asraa, line 65: interesting, I'm guessing this is because of the limitation of the print call being i32 that we have to check the result mod 2^32? I think that'll end up fine- if something was wrong, then chances are there's a case where it'll show up wrong mod 2^32.

could you add a quick comment for this template that we check equality mod 2^32 because of the function call limitation and expect to see issues in lowerings arise below i32? we can reevaluate if there's something funky we do for larger bit widths.

tests/poly/runner/lower_mul_tests.toml
```diff
@@ -0,0 +1,6 @@
+[[test]]
+ideal = "1 + x**12"
+cmod = 4294967296  # 2**32
+p0 = "1 + x**10"
+p1 = "1 + x**11"
+container_type = "i32"
```
asraa, line 6: will there be some error if container_type is larger than bit size of cmod? (https://github.com/google/heir/blob/8ca202760fed9f966a3c10ac6ff990f6af40f5d4/lib/Dialect/Poly/IR/PolyOps.cpp#L43-L50) or are the TOML files expected to be generated so it's always true? (is the reason this arg isn't explicit because you want to test fromtensor op's extension? if so, i feel like it's not necessarily useful, since these are testing the poly op lowerings correctness)

asraa (approved):
WOW python is super easy to review!

Thank you!

tests/poly/runner/lower_mul_tests.toml
j2kun: The container_type is meant to be the element type of the tensor output by the lowered poly.mul. I felt it was too hard to compute that value at test generation time, or rather, it duplicates too much of the logic of the thing it's testing. So it's assumed to be set correctly by the user. If it doesn't exactly match the output of the lowered poly mul, it will be a verifier error at parse time, because the templated ops will be doing a "trunci" or "extsi" with a type signature that doesn't match the type of the input. So it seems pretty hard to silently get wrong.

Issue #216, opened 2023-10-26
[tfhe-rs] Add tfhe-rs dialect


j2kun:
This is an exit dialect for the tfhe-rs rust API. Only planning to support the shortint API

PR #217, opened 2023-10-26
tfhe_rs dialect part 1: basic types and ops

j2kun:
This PR adds a tfhe_rs exit dialect, part of https://github.com/google/heir/issues/216

Next up:
- Encrypt/decrypt ops
- heir-translate codegen
- more ops as needed

### DIFF
--- a/include/Dialect/BGV/IR/BUILD
+++ b/include/Dialect/BGV/IR/BUILD
@@ -102,7 +102,7 @@ gentbl_cc_library(
         ),
         (
             ["-gen-op-doc"],
-            "SecretOps.md",
+            "BGVOps.md",
         ),
     ],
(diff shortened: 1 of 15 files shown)


asraa (approved):
Thanks for starting this!

I'm guessing from this PR that to start, the tfhe-rs parameters are implicit and we'll used some fixed set when lowering tfhe-rs to external llvm calls?

j2kun:
> Thanks for starting this!
> 
> I'm guessing from this PR that to start, the tfhe-rs parameters are implicit and we'll used some fixed set when lowering tfhe-rs to external llvm calls?

Yeah, I'm just trying to get something basic and inefficient, end to end, so we can iterate.

PR #218, opened 2023-10-31
CGGI Dialect and LWE ciphertexts

asraa:
* Moves LWECiphertext type to the LWE dialect
* Adds an LWEPlaintext type and encode operation to the LWE dialect
* Adds a CGGI trivial encryption into a ring to get an LWECiphertext type (it's possible this could also go to LWE?)

This gives sufficient functionality to implement a LUT based lowering from comb dialect.

### DIFF
--- a/include/Dialect/CGGI/IR/CGGIOps.td
+++ b/include/Dialect/CGGI/IR/CGGIOps.td
@@ -1,0 +1,75 @@
+#ifndef HEIR_INCLUDE_DIALECT_CGGI_IR_CGGIOPS_TD_
+#define HEIR_INCLUDE_DIALECT_CGGI_IR_CGGIOPS_TD_
+
+include "CGGIDialect.td"
+
+include "include/Dialect/Polynomial/IR/PolynomialAttributes.td"
+include "include/Dialect/LWE/IR/LWETypes.td"
+
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/InferTypeOpInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+class CGGI_Op<string mnemonic, list<Trait> traits = []> :
+        Op<CGGI_Dialect, mnemonic, traits> {
+  let assemblyFormat = [{
+    `(` operands `)` attr-dict `:`  `(` qualified(type(operands)) `)` `->` qualified(type(results))
+  }];
+  let cppNamespace = "::mlir::heir::cggi";
+}
+
+// --- Operations for a gate-bootstrapping API of a CGGI library ---
+
+class CGGI_BinaryGateOp<string mnemonic>
+    : CGGI_Op<mnemonic, [Pure, Commutative, SameOperandsAndResultType]> {
+  let arguments = (ins LWECiphertext:$lhs, LWECiphertext:$rhs);
+  let results = (outs LWECiphertext:$output);
+  // Note: error: type of result #0, named 'output', is not buildable and a buildable type cannot be inferred
+  // LWECiphertext is not buildable?
+  let assemblyFormat = "operands attr-dict `->` type($output)" ;
+}
+
+def CGGI_AndOp : CGGI_BinaryGateOp<"and"> { let summary = "Logical AND of two ciphertexts."; }
+def CGGI_OrOp  : CGGI_BinaryGateOp<"or">  { let summary = "Logical OR of two ciphertexts."; }
+def CGGI_XorOp : CGGI_BinaryGateOp<"xor"> { let summary = "Logical XOR of two ciphertexts."; }
+
+class CGGI_NotOp : CGGI_Op<"not", [Pure, SameOperandsAndResultType, Involution]> {
+  let arguments = (ins LWECiphertext:$input);
+  let results = (outs LWECiphertext:$output);
+  let assemblyFormat = "operands attr-dict" ;
+  let summary = "Logical NOT of two ciphertexts";
+}
+
+class CGGI_LutOp<string mnemonic> : CGGI_Op<mnemonic, [Pure, Commutative, SameOperandsAndResultType]> {
+  let results = (outs LWECiphertext:$output);
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type($output)" ;
+
+  let description = [{
+    An op representing a lookup table applied to some number `n` of ciphertexts
+    encrypting boolean input bits.
+
+    Over cleartext bits `a, b, c`, using `n = 3` for example, the operation
+    computed by this function can be interpreted as
+
+    ```
+      truth_table >> {c, b, a}
+    ```
+
+    where `{c, b, a}` is the unsigned 3-bit integer with bits `c, b, a` from most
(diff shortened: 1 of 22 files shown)


tests/cggi/syntax.mlir
```diff
@@ -0,0 +1,21 @@
...
+!ciphertext = !lwe.lwe_ciphertext<encoding = #encoding, ring = #ring>
+
+module {
+  func.func @test_syntax(%arg0 : !ciphertext) -> !ciphertext {
+    %0 = arith.constant 0 : i1
+    %1 = arith.constant 1 : i1
+    %2 = lwe.encode %0 { encoding = #encoding }: i1 to !plaintext
+    %3 = lwe.encode %1 { encoding = #encoding }: i1 to !plaintext
+    %4 = cggi.trivial_encrypt %2 { ring = #ring } : !plaintext to !ciphertext
+    %5 = cggi.trivial_encrypt %3 { ring = #ring } : !plaintext to !ciphertext
```
asraa, line 17: note: i'm playing around here with the assembly format to see if it's possible to remove the `!plaintext`.

include/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -0,0 +1,90 @@
...
+  let summary = "A lookup table on two inputs.";
+  let arguments = (ins LWECiphertext:$b, LWECiphertext:$a, IndexAttr:$lookup_table);
+}
+
+def CGGI_Lut3Op : CGGI_LutOp<"lut3"> {
+  let summary = "A lookup table on three inputs.";
+  let arguments = (ins LWECiphertext:$c, LWECiphertext:$b, LWECiphertext:$a, IndexAttr:$lookup_table);
+}
+
+def CGGI_TrivialEncryptOp: CGGI_Op<"trivial_encrypt", [Pure]> {
```
j2kun, line 75: Worth having in LWE dialect? It's not CGGI specific.

include/Dialect/LWE/IR/LWETypes.td
```diff
@@ -0,0 +1,58 @@
...
+
+    This type keeps track of the plaintext integer encoding for the LWE
+    Ciphertext to ensure proper decoding after decryption. It also keeps track
+    of the ring where the LWE ciphertext is defined, which provides information
+    on the ciphertext shape and the ring operations used in LWE operations.
+  }];
+
+  let parameters = (ins
+    LWE_BitFieldEncoding:$encoding,
+    Ring_Attr:$ring
```
j2kun, line 34: Here and in CGGI, why exactly do we need the ring? You want this type to support both LWE and RLWE? I feel they should be different types, but if we keep them as one type, we should really have the ring attribute be optional or constrained somehow.

tests/cggi/syntax.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: heir-opt -verify-diagnostics %s
```
j2kun, line 1: `-verify-diagnostics` does not do anything unless you have specific IR comments that assert messages are written to stderr. https://mlir.llvm.org/getting_started/TestingGuide/#diagnostic-verification-tests

Ditto for lwe/types.mlir

include/Dialect/LWE/IR/LWETypes.td
```diff
@@ -0,0 +1,58 @@
...
+// A base class for all types in this dialect
+class LWE_Type<string name, string typeMnemonic>
+    : TypeDef<LWE_Dialect, name> {
+  let mnemonic = typeMnemonic;
+}
+
+// LWE Ciphertexts are ranked tensors of integers representing the LWE samples
+// and the bias.
+def LWECiphertext : LWE_Type<"LWECiphertext", "lwe_ciphertext"> {
+  let summary = "A type for an LWE ciphertexts";
```
j2kun, line 21: `A type for an LWE ciphertexts` -> `A type for LWE ciphertexts` or `A type for an LWE ciphertext`

j2kun (changes requested):
Nits, except for the question about whether LWECiphertext will be both RLWE and LWE. I think we should have RLWE as its own type.

include/Dialect/LWE/IR/LWETypes.td
asraa: Oops! Ah! Yes, RLWE/LWE should be separate - this should instead have the cmod and dimension N of the lwe ciphertexts

include/Dialect/CGGI/IR/CGGIOps.td
asraa: Yep! Noted as well, moved.

tests/cggi/syntax.mlir
asraa: Done

include/Dialect/LWE/IR/LWETypes.td
asraa: Done

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -230,4 +230,13 @@ def RLWE_InverseCanonicalEmbeddingEncoding
   }];
 }
 
+def LWE_LWEParams : AttrDef<LWE_Dialect, "LWEParams"> {
+  let mnemonic = "lwe_params";
+
+  let parameters = (ins "APInt": $cmod, "unsigned":$dimension);
```
j2kun, line 236: This may need the same trick we had as for Poly, where MLIR will try to compare APInts in the uniquer, and fail if they have different bit widths.

I wonder if we should bring up this to the MLIR folks more generally.

lib/Dialect/LWE/IR/LWEAttributes.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "mlir/include/mlir/Support/LogicalResult.h"  // from @llvm-project
+
+#define GET_ATTRDEF_CLASSES
+#include "include/Dialect/LWE/IR/LWEAttributes.cpp.inc"
+
+namespace mlir {
+namespace heir {
+namespace lwe {
+
+void LWEParamsAttr::print(AsmPrinter &p) const {
```
j2kun, line 17: I recently learned about

` let assemblyFormat = "`<` struct(params) `>`";`

which you might find is sufficient to remove the custom parser/printer

j2kun (approved):

include/Dialect/LWE/IR/LWEAttributes.td
asraa: Done

lib/Dialect/LWE/IR/LWEAttributes.cpp
asraa: I think the issue was that the default assembly format doesn't allow for parsing/printing APInts: 
```
bazel-out/k8-dbg/bin/include/Dialect/LWE/IR/LWEAttributes.cpp.inc:256:52: error: incomplete type 'mlir::FieldParser<llvm::APInt>' used in nested name specifier
  256 |         _result_cmod = ::mlir::FieldParser<APInt>::parse(odsParser);
  ```

Issue #221, opened 2023-11-02
[cleanup] Migrate APInt's in attributes to APIntAttr


asraa:
We use this in polynomial cmod and also LWE cmod

See  https://discord.com/channels/636084430946959380/642426447167881246/1169678505064276028

Issue #204, opened 2023-10-17
poly: sometimes PolyToStandard pass crashes due to APInts of different widths being compared


j2kun:
Sometimes (nondeterministically) I'll get a crash when running `--poly-to-standard` on `lower_mul.mlir`. Stack trace below. It appears to be happening when two RingAttrs are constructed with cmod's with different bit widths, and the comparison test in the AttributeUniquer complains because two APInt's with different bit widths cannot be compared. Maybe changing the ringattr to take an element type container for the coefficients would fix this, since the APInt can always have that same container type, and two RingAttrs' cmods wouldn't be compared unless the first test for equality of the container type passes.

```
# | Stack dump:
# | 0.  Program arguments: heir-opt --poly-to-standard /home/j2kun/.cache/bazel/_bazel_j2kun/b07a48cf9da8c73a462d3608e5c508dd/sandbox/linux-sandbox/9014/execroot/heir/bazel-out/k8-dbg/bin/tests/poly/lower_mul.mlir.test.runfiles/heir/tests/poly/lower_mul.mlir
# |  #0 0x000055c0c1fc4dd8 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:723:22
# |  #1 0x000055c0c1fc51b5 PrintStackTraceSignalHandler(void*) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:798:1
# |  #2 0x000055c0c1fc26b7 llvm::sys::RunSignalHandlers() /proc/self/cwd/external/llvm-project/llvm/lib/Support/Signals.cpp:105:20
# |  #3 0x000055c0c1fc46b4 SignalHandler(int) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:413:1
# |  #4 0x00007f2123591520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
# |  #5 0x00007f21235e5a7c __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
# |  #6 0x00007f21235e5a7c __pthread_kill_internal ./nptl/pthread_kill.c:78:10
# |  #7 0x00007f21235e5a7c pthread_kill ./nptl/pthread_kill.c:89:10
# |  #8 0x00007f2123591476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
[...]

PR #222, opened 2023-11-02
Migrate APInt attribute arguments to Builtin_IntegerAttr

j2kun:
Fixes https://github.com/google/heir/issues/221

### DIFF
--- a/include/Dialect/Polynomial/IR/PolynomialAttributes.td
+++ b/include/Dialect/Polynomial/IR/PolynomialAttributes.td
@@ -3,8 +3,9 @@
 
 include "include/Dialect/Polynomial/IR/PolynomialDialect.td"
 
-include "mlir/IR/DialectBase.td"
 include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/BuiltinAttributes.td"
+include "mlir/IR/DialectBase.td"
 
 class Polynomial_Attr<string name, string attrMnemonic, list<Trait> traits = []>
     : AttrDef<Polynomial_Dialect, name, traits> {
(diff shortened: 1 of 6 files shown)


j2kun:
Fixes https://github.com/google/heir/issues/204

asraa (approved):

include/Dialect/Polynomial/IR/PolynomialAttributes.td
```diff
@@ -46,26 +47,29 @@ def Ring_Attr : Polynomial_Attr<"Ring", "ring"> {
...
-  // parser/printer ensures that cwidth is automatically inferred from cmod, and
-  // not printed or parsed, so that the cwidth parameter need not be known to the
-  // user.
-  let parameters = (ins "int64_t": $cwidth, "APInt": $cmod, "Polynomial":$ideal);
+  let parameters = (ins "IntegerAttr": $cmod, "Polynomial":$ideal);
 
   let builders = [
-    AttrBuilderWithInferredContext<(ins "APInt": $cmod, "Polynomial":$ideal), [{
-      return $_get(ideal.getContext(), cmod.getBitWidth(), cmod, ideal);
+    AttrBuilderWithInferredContext<
```
asraa, line 53: do you end up using this builder? i was slightly concerned about the use of `zextOrTrunc` here, so if it's unused (I don't see it in the diff, and it appears different) maybe best to remove it until tested or needed

asraa (approved):
Thanks!

include/Dialect/Polynomial/IR/PolynomialAttributes.td
asraa: Oh, maybe this is used when the parser sees the `: i<width>` specified?

asraa: Ah, yes when fixing my PR just now, I found out the default generated parser uses a builder (with IntegerAttr type param) like 
```
LWEParamsAttr::get(odsParser.getContext(),
      IntegerAttr((*_result_cmod)),
      unsigned((*_result_dimension)));
```

and it ensured that the `_result_cmod` was a parser that can gather the type attribute on the int.

> This is supposed to ensure that the integer type and the APInt's bit width always match.

I guess what I'm asking is - do you need it right now if you have no callers?

j2kun: Fair enough, I thought I was using it. Removed.

Issue #199, opened 2023-10-16
poly: add a test for lower_mul with a divisor that has a non-1 leading coefficient


j2kun:
The initial lowering for poly.mul involves computing a remainder of the product modulo the ring's polynomial modulus. This in turn computes an inverse of the leading coefficient of the divisor. E.g., for a poly mod of 3x^1024 + 1, this would be the inverse of 3 modulo the coefficient modulus of the ring.

There are two more cases to test: that the division uses the correct inverse when an inverse exists (e.g., 3 mod 1024), and that the lowering fails with a sensible error message when the leading coefficient has no inverse (e.g., 2 mod 1024).

j2kun:
This can be added via the testing framework introduced by https://github.com/google/heir/pull/210

PR #223, opened 2023-11-03
test non-1 coefficients in mul reduction

j2kun:
Fixes https://github.com/google/heir/issues/199


asraa (approved):
Nice! Very easy code :) It's so weird there's no int coeff option :/

PR #225, opened 2023-11-03
Minimal codegen for tfhe_rust dialect

j2kun:
Blocked by https://github.com/google/heir/pull/217

Outline of the process:

- Write an analysis pass to generate variable names for each SSA value
- Code gen minimal ops to support composing two encrypted bits into a bitmask, and applying a lookup table.

Next:
- Op to create a lookup table from a static constant
- Re-use `scf.forall` op to represent the parallel-executed regions of the circuit (pass that precedes this codegen step will handle toposorting the circuit)
- Figure out how we're gonna do e2e testing of this.

It seems like an e2e test that compiles the result with rustc and tfhe-rs is not suitable for living in this project. But we should have some e2e test flow, even if manual, that asserts the generated code works against a specific tfhe-rs API version.


lib/Analysis/SelectVariableNames/SelectVariableNames.cpp
```diff
@@ -0,0 +1,40 @@
+#include "include/Analysis/SelectVariableNames/SelectVariableNames.h"
+
+#include <iostream>
```
asraa, line 3: nit: i always keep this in after debugging by accident, do you need this?

lib/Target/TfheRust/TfheRustEmitter.cpp
```diff
@@ -0,0 +1,234 @@
...
+    if (failed(translate(op))) {
+      return failure();
+    }
+  }
+
+  return success();
+}
+
+LogicalResult TfheRustEmitter::printOperation(func::FuncOp funcOp) {
+  os << "pub fn " << funcOp.getName() << "(\n";
```
asraa, line 105: kinda silly, but should the func private visibility attribute be used to determine whether the rust fn is `pub`?

asraa (approved):
Nice!

lib/Target/TfheRust/TfheRustEmitter.cpp
j2kun: I suppose we could.... I'm not entirely sure what that would do for us though. I'll think about it and add it to my TODO list.

lib/Analysis/SelectVariableNames/SelectVariableNames.cpp
j2kun: Done, thanks!

PR #226, opened 2023-11-06
[comb] Update comb dialect lookup table attr to an IntegerAttr

asraa:
This is in prep for Comb to CGGI lowering.


j2kun (approved):
LGTM!

Issue #227, opened 2023-11-06
Lower comb to CGGI dialect


asraa:
Requires:

- [x] Update comb lookup table attr to integer attr
- [x]  Make CGGI encoding and LWE params optional / unspecified on initial lowering
- [x]  Define lowering patterns for (1) truth table op (2) plain integers to trivial encrypted ciphertexts and (3) secret.secret data types to ciphertexts

asraa:
This is done

PR #231, opened 2023-11-06
lwe: add unspecified lwe encoding and encryption params

asraa:
Part of #227

### DIFF
--- a/include/Dialect/LWE/IR/LWEAttributes.td
+++ b/include/Dialect/LWE/IR/LWEAttributes.td
@@ -63,11 +63,42 @@ def LWE_BitFieldEncoding
     Example:
 
     ```
-    #lwe_encoding = #lwe.polynomial_coefficient_encoding<cleartext_start=30, cleartext_bitwidth=3>
+    #lwe_encoding = #lwe.bit_field_encoding<cleartext_start=30, cleartext_bitwidth=3>
     %lwe_ciphertext = arith.constant <[1,2,3,4]> : tensor<4xi32, #lwe_encoding>
     ```
   }];
 }
+
+def LWE_UnspecifiedBitFieldEncoding
+  : LWE_EncodingAttr<"UnspecifiedBitFieldEncoding",
+      "unspecified_bit_field_encoding"> {
+  let summary = "An attribute describing unspecified bit field encodings.";
+  let description = [{
+    See LWE_BitFieldEncoding for a description of bit field encodings.
+
+    This attribute describes an unspecified bit field encoding; this is where
+    the starting bit position of the cleartext bit field is unspecified, but its
+    width is fixed. A noise growth analysis should be performed to determine the
+    optimal amount of bits needed for noise and padding to specify the bit field
+    encodings starting bit position.
+
+    Example:
+
+    ```
+    #lwe_encoding = #lwe.unspecified_bit_field_encoding<cleartext_bitwidth=3>
+    %lwe_ciphertext = arith.constant <[1,2,3,4]> : tensor<4xi32, #lwe_encoding>
+    ```
+  }];
+
+  // These parameters represent unspecified encodings and simply hold the
+  // cleartext bit width in preparation for specifying the encoding scaling
+  // factors after noise growth analysis.
+  let parameters = (ins
+    "unsigned":$cleartext_bitwidth
+  );
+}
+
+def AnyLWEEncodingAttr : AnyAttrOf<[LWE_BitFieldEncoding, LWE_UnspecifiedBitFieldEncoding]>;
 
 def RLWE_PolynomialCoefficientEncoding
   : LWE_EncodingAttrWithScalingFactor<"PolynomialCoefficientEncoding", "polynomial_coefficient_encoding"> {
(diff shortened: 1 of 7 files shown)


include/Dialect/LWE/IR/LWETypes.td
```diff
@@ -28,8 +28,10 @@ def LWECiphertext : LWE_Type<"LWECiphertext", "lwe_ciphertext"> {
   }];
 
   let parameters = (ins
-    LWE_BitFieldEncoding:$encoding,
-    LWE_LWEParams:$lwe_params
+  // FIXME: Encoding attrs belong on ops and types, should the encoding
+  //   attributes be a different class?
+    "::mlir::Attribute":$encoding,
```
asraa, line 33: @j2kun curious if you there's an obvious answer here. I had created an attribute with a "AnyOf" constraint (any one of the LWE encoding attributes), but these attr's buildable types are "::mlir::Attribute" - I *think* in order to use them as type parameters they need a c++ buildable type.

I was looking into `AttrOrTypeParameters` but had some troubles using it, although I think I *should*. Does that seem like the right approach?

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -63,12 +73,36 @@ def LWE_BitFieldEncoding
...
 
     ```
-    #lwe_encoding = #lwe.polynomial_coefficient_encoding<cleartext_start=30, cleartext_bitwidth=3>
+    #lwe_encoding = #lwe.bit_field_encoding<cleartext_start=30, cleartext_bitwidth=3>
+    %lwe_ciphertext = arith.constant <[1,2,3,4]> : tensor<4xi32, #lwe_encoding>
+    ```
+  }];
+}
+
+def LWE_UnspecifiedBitFieldEncoding
```
j2kun, line 82: Why have a sep

include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -30,6 +30,16 @@ class LWE_EncodingAttrWithScalingFactor<string attrName, string attrMnemonic, li
   );
 }
 
+class LWE_UnspecifiedEncodingAttr<string attrName, string attrMnemonic, list<Trait> traits = []>
```
j2kun, line 33: Why have a parent class here?

include/Dialect/LWE/IR/LWEAttributes.td
asraa: I assume the comment got caught off but was along the lines of "why not just make the cleartext_start optional/default to 31?" It seemed a little odd to check the presence of this value in determining if encoding params were set rather than separating the attributes, but having a c++ helper would help here.

include/Dialect/LWE/IR/LWETypes.td
j2kun: As best I can tell the attribute constraints only work with ops. IMO it would be fine to have a generic Attribute parameter here, and dyn_cast it when we read it, since we will need to do that anyway if there are multiple options.

I guess the alternative is to have a base class with the width field somehow, though I'm not sure how to do that in tablegen, and not sure it's worth the effort since we'll need to cast either way.

tests/lwe/ops.mlir
```diff
@@ -0,0 +1,30 @@
+// RUN: heir-opt --split-input-file %s 2>&1 | FileCheck %s
```
j2kun, line 1: Consider adding an invalid case, using the diagnostic infrastructure to assert the error message:

See, e.g., https://github.com/google/heir/blob/53124b671d1152b23fd44818de00231dbae13273/tests/secret/verifier.mlir#L1-L4

`--verify-diagnostics` enables the checks, and ` // expected-error@+1 {{<regex>}}` checks for the error emitted at the op below the comment.

Not sure if it needs to be in a separate file due to FileCheck complaining about no input after the diagnostic check passes...

j2kun (approved):
LGTM! Added some optional suggestions.

include/Dialect/LWE/IR/LWEAttributes.td
j2kun: Sorry, just fat fingered enter :) Ignore this one.

include/Dialect/LWE/IR/LWEAttributes.td
asraa: This was to anticipate an RLWE unspecified encoding, but i'll defer to when its needed :) Removed.

tests/lwe/ops.mlir
asraa: Added some invalid cases! Also lead me to write some verifiers ensuring that the encoding and lwe params op attrs match the outputs' attributes.

include/Dialect/LWE/IR/LWETypes.td
asraa: Yeah, doing the base class attribute would "make sense". I think Zama roughly achieved it for their unspecified secret key with custom C++ code, but that did have to use casting as well.

PR #233, opened 2023-11-07
End-to-end test for generated tfhe-rs code

j2kun:
This PR adds the necessary components for running end-to-end tests of generated tfhe-rs code in HEIR.

The actual test itself is marked `manual`, so it will not be run by CI, which is intended because of the extra dependency on cargo, `tfhe` crate, and all its dependencies. Future work on the tfhe-rs codegen should come with the expectation of running these tests manually on a machine that has cargo installed.

Will follow up this PR with more e2e tests, along with more codegen support.

### DIFF
--- a/tests/tfhe_rust/end_to_end/src/main.rs
+++ b/tests/tfhe_rust/end_to_end/src/main.rs
@@ -1,0 +1,34 @@
+use clap::Parser;
+use tfhe::shortint::parameters::get_parameters_from_message_and_carry;
+
+mod fn_under_test;
+
+// TODO(https://github.com/google/heir/issues/235): improve generality
+#[derive(Parser, Debug)]
+struct Args {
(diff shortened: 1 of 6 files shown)


tests/tfhe_rust/end_to_end/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+    #[arg(id = "input_1", index = 1)]
+    input1: u8,
+
+    #[arg(id = "input_2", index = 2)]
+    input2: u8,
+}
+
+fn main() {
+    let flags = Args::parse();
+    let parameters = get_parameters_from_message_and_carry((1 << flags.message_bits) - 1, flags.carry_bits);
```
asraa, line 25: nice find!

asraa (approved):

PR #240, opened 2023-11-08
comb: lower comb to CGGI dialect

asraa:
This just sets up the infra and the secret to ciphertext type conversion.

I'm working on the lowerings through secret.generic.

Small part of https://github.com/google/heir/issues/227

### DIFF
--- a/tests/comb_to_cggi/types.mlir
+++ b/tests/comb_to_cggi/types.mlir
@@ -1,0 +1,9 @@
+// RUN: heir-opt --comb-to-cggi %s | FileCheck %s
+
+module {
+  // CHECK: func.func @types([[ARG:%.*]]: [[T:!lwe.lwe_ciphertext<.*>]]) -> [[T]]
+  func.func @types(%arg0: !secret.secret<i1>) -> !secret.secret<i1> {
+    // CHECK: return [[ARG]] : [[T]]
+    return %arg0 : !secret.secret<i1>
+  }
+}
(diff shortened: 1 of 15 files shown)


include/Dialect/LWE/IR/LWEAttributes.td
```diff
@@ -267,6 +267,8 @@ def LWE_LWEParams : AttrDef<LWE_Dialect, "LWEParams"> {
   let parameters = (ins "IntegerAttr": $cmod, "unsigned":$dimension);
 
   let assemblyFormat = "`<` struct(params) `>`";
+
+  let isOptional = 1;
```
j2kun, line 271: The docs on this say

```
  // Whether the attribute is optional. Typically requires a custom
  // convertFromStorage method to handle the case where the attribute is
  // not present.
```

Do we need that?

asraa: Oops, nope - I was playing around with when exploring the codegen for optional attrs, but nope - the op that uses that (trivial encrypt) where it may be optional wraps it in `OptionalAttr`.

j2kun (approved):

tests/comb_to_cggi/types.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: heir-opt --comb-to-cggi %s | FileCheck %s
+
+// FIXME: Need a pass that converts yosys optimizer extract to splitting the inputs.
```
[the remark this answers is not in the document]
asraa: I'll file an issue, but this was unrelated to this test (I was scratching a note) so I'll remove.

j2kun (approved):

PR #259, opened 2023-11-16
cggi to tfhe_rust conversion pass

j2kun:
First step toward https://github.com/google/heir/issues/258

still needs to handle the Lut2Op, as well as the other binary gate ops.

### DIFF
--- a/lib/Conversion/CGGIToTfheRust/CGGIToTfheRust.cpp
+++ b/lib/Conversion/CGGIToTfheRust/CGGIToTfheRust.cpp
@@ -1,0 +1,278 @@
+#include "include/Conversion/CGGIToTfheRust/CGGIToTfheRust.h"
+
+#include "include/Dialect/CGGI/IR/CGGIDialect.h"
+#include "include/Dialect/CGGI/IR/CGGIOps.h"
+#include "include/Dialect/LWE/IR/LWEAttributes.h"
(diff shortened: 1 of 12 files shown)


lib/Conversion/CGGIToTfheRust/CGGIToTfheRust.cpp
```diff
@@ -0,0 +1,278 @@
...
+    case 256:
+      return tfhe_rust::EncryptedUInt256Type::get(ctx);
+    default:
+      llvm_unreachable("Unsupported bitwidth");
+  }
+}
+
+// Seems like this would be better as a method on the
+// LWE_EncodingAttrWithScalingFactor class, but I still have the problem of the
+// type returned by getEncoding being a vanilla Attribute. Probably we need a
```
asraa, line 59: I think this was because of the any class that allowed for any LWE encodings. 

This was really annoying honestly :/ Yeah, like you say, adding some c++ code for a common interface or code to handle variadic types might end up better

asraa (approved):

lib/Conversion/CGGIToTfheRust/CGGIToTfheRust.cpp
j2kun: I think the right solution is to use MLIR interfaces/traits, but I just haven't figured out how to make a custom interface yet.

Issue #260, opened 2023-11-16
Yosys Optimizer crashes on input


asraa:
```
  func.func @for_6440052803555699788(%arg0: i8, %arg1: i8) -> i32 {
    %c-128_i16 = arith.constant -128 : i16
    %c0_i32 = arith.constant 0 : i32
    %0 = arith.extsi %arg0 : i8 to i16
    %1 = arith.subi %0, %c-128_i16 : i16
    %2 = arith.extsi %1 : i16 to i32
    %3 = arith.extsi %arg1 : i8 to i32
    %4 = arith.muli %2, %3 : i32
    %5 = arith.addi %c0_i32, %4 : i32
    return %5 : i32
  }
```

In working on the booleanizatin / splitting inputs, I ran into this crashing case. Investigating.

asraa:
It seems to be crashing on an assert in the Yosys RTLIL library, that 
```
ERROR: Assert `width_ == 1' failed in external/at_clifford_yosys/kernel/rtlil.cc:4649.
```

I suspect that it's because of the multi-bit constants

asraa:
This is a smaller Yosys module that fails

```
attribute \top 1
attribute \src "/tmp/file4v9Ejg:1.1-11.10"
module \test_multibit_const

  attribute \src "/tmp/file4v9Ejg:3.28-3.33"
  wire width 8 output 2 signed \_out_

  attribute \src "/tmp/file4v9Ejg:2.27-2.31"
  wire width 8 input 1 signed \arg1

  attribute \LUT 2'01
  attribute \module_not_derived 1
  attribute \src "lib/Transforms/YosysOptimizer/yosys/map_lut_to_lut3.v:72.10-81.19"
  cell \lut3 \_0_
    connect \A \arg1 [7]
    connect \B 0
    connect \C 0
    connect \P0 1'1
    connect \P1 1'0
    connect \P2 0
    connect \P3 0
    connect \P4 0
    connect \P5 0
    connect \P6 0
    connect \P7 0
    connect \Y \_out_ [7]
  end

  connect \_out_ [6:0] \arg1 [6:0]
end
```

PR #262, opened 2023-11-16
yosys optimizer: fix bug with chunk connections

asraa:
Fixes https://github.com/google/heir/issues/260

This fixes a bug in the RTLIL importer, where a connection was made between a wire and another chunked wire. The code did not support the RHS chunked wire. A chunked wire is a part of a multi-bit wire. For e.g

```
module \test_multibit_const

  attribute \src "/tmp/file4v9Ejg:3.28-3.33"
  wire width 8 output 2 signed \_out_

  attribute \src "/tmp/file4v9Ejg:2.27-2.31"
  wire width 8 input 1 signed \arg1

  attribute \LUT 2'01
  attribute \module_not_derived 1
  attribute \src "lib/Transforms/YosysOptimizer/yosys/map_lut_to_lut3.v:72.10-81.19"
  cell \lut3 \_0_
    connect \A \arg1 [7]
    connect \B 0
    connect \C 0
    connect \P0 1'1
    connect \P1 1'0
    connect \P2 0
    connect \P3 0
    connect \P4 0
    connect \P5 0
    connect \P6 0
    connect \P7 0
    connect \Y \_out_ [7]
  end

  connect \_out_ [6:0] \arg1 [6:0]
end
```

In the line

```
  connect \_out_ [6:0] \arg1 [6:0]
```

we connect a chunk of `\out` with a chunk of `\arg1`. The code currently only supported chunked LHS by iterating through each bit of the connection. Now, we also do that same code if RHS is chunked.

### DIFF
--- a/lib/Transforms/YosysOptimizer/RTLILImporter.cpp
+++ b/lib/Transforms/YosysOptimizer/RTLILImporter.cpp
@@ -1,6 +1,7 @@
 #include "lib/Transforms/YosysOptimizer/RTLILImporter.h"
 
 #include <cassert>
+#include <iostream>
 #include <sstream>
 #include <string>
 #include <utility>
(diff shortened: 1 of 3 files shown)


j2kun (approved):

asraa:
Fixing a broken test - it turns out this does some un-needed expansion for one of the cases.

j2kun:
Looks like there's still an assertion failure

```
LUTImporterTest: lib/Transforms/YosysOptimizer/RTLILImporter.cpp:58: mlir::Value mlir::heir::RTLILImporter::getBit(const Yosys::RTLIL::SigSpec&, mlir::ImplicitLocOpBuilder&, llvm::MapVector<Yosys::RTLIL::Wire*, llvm::SmallVector<mlir::Value> >&): Assertion `conn.is_wire() || conn.is_fully_const() || conn.is_bit()' failed.
```

asraa:
Yeah, I fixed it with an updated comment, just rebuilding to make sure the fix doesn't break the MLIR tests.

j2kun (approved):

Issue #248, opened 2023-11-09
tfhe_rust: add end-to-end testing to Github CI


j2kun:
We'd need to install the rust toolchain as part of the CI setup, but otherwise it should work.

For Google reasons I don't think we can add it by default to the entire bazel build, but we could use the following script to run the rust tests as a last step in the GitHub workflow.

```bash
bazel query "filter('.mlir.test$', //tests/tfhe_rust/end_to_end/...)" \
  | xargs bazel test --sandbox_writable_path=$HOME/.cargo "$@"
```

PR #265, opened 2023-11-17
Run tfhe-rs tests in CI

j2kun:
Fixes https://github.com/google/heir/issues/248


asraa (approved):
Not totally sure why  rust-toolchain action is mising the toolchain when it should be optional... maybe it needs a version bump

j2kun:
> Not totally sure why rust-toolchain action is mising the toolchain when it should be optional... maybe it needs a version bump

It's because I pinned to a commit hash. If you use `@stable` it will read that and infer the rust toolchain version.

Issue #268, opened 2023-11-17
Support yosys-optimizer on a secret-ified op


asraa:
Yosys Optimizer doesn't currently support a module that contains a secret-ified op. This is because the verilog translator doesn't know how to convert secret types.

One thing that could be done is to translate the "forget-secrets" converted module, and then re-introduce the secret types.

(I also expect subsequent lowerings & conversions will choke on secret-wrapped types)

```
/usr/local/google/home/asraa/git/heir/tests/arith_to_tfhe_rust/add_one.mlir:5:3: error: 'func.func' op Failed to translate op func.func
  func.func @add_one(%in: !secret.secret<i8>) -> (!secret.secret<i8>) {
  ^
/usr/local/google/home/asraa/git/heir/tests/arith_to_tfhe_rust/add_one.mlir:5:3: note: see current operation: 
"func.func"() <{function_type = (!secret.secret<i8>) -> !secret.secret<i8>, sym_name = "add_one"}> ({
^bb0(%arg0: !secret.secret<i8>):
  %0 = "arith.constant"() <{value = 1 : i8}> : () -> i8
  %1 = "secret.generic"(%arg0, %0) ({
  ^bb0(%arg1: i8, %arg2: i8):
    %2 = "arith.addi"(%arg1, %arg2) : (i8, i8) -> i8
    "secret.yield"(%2) : (i8) -> ()
  }) : (!secret.secret<i8>, i8) -> !secret.secret<i8>
  "func.return"(%1) : (!secret.secret<i8>) -> ()
}) : () -> ()
```

j2kun:
Oh, this is a good example of what I meant by having the yosys optimizer apply more generically: we could just apply the yosys optimizer to the body of the generic (excluding the yield), follow that with a secret-distribute-generic, and then add support in comb-to-cggi for secret types and generic ops (which makes a lot more sense because comb is semantically not operating on encrypted data).

j2kun:
I think we could even do this with a bit of wizardry by marking secret.generic as isolated from above and anchoring the entire yosys-optimizer pass on it, so that it will automatically apply to each secret.generic op in parallel... I will try this out.

Issue #257, opened 2023-11-16
Optimize how much to unroll a loop body vs how much yosys can optimize it


j2kun:
On the side of gate-bootstrapped CGGI, we had a path where we fully unrolled all loops before running a circuit optimizer, and for large IRs that was infeasibly slow. We hacked together a different approach in lib/Conversion/MemrefToArith/ExtractLoopBody.cpp that results in only the inner-most body of a loop being optimized via yosys, and retaining the rest of the loop structure.

There is likely an intermediate approach where we can unroll the loop body to a different factor depending on how much yosys is able to optimize it. A decent initial approach might be greedy: starting from no unrolling to a factor of N—say, based on a fixed time budget—and compare the size of the input circuit to the size of the output circuit (this will be a bit fuzzy because the input circuit is not techmapped, but the output circuit is, but I'm hoping relative sizes will be sufficient to compare), and stop once the incremental benefit of unrolling more is plateauing, decreasing, or the time budget is exhausted.

As an outline of the work required for this:

- Write a pass that doesn't unroll at all, but extracts the innermost body of a loop-next, optimizes it with yosys, and inserts it back into the loop body, replacing the old body. The hard part will likely be the booleanization interface, but @asraa is working on that in https://github.com/google/heir/issues/244 and one could extract the loop body first to a separate function in order to use that, then inline the result.
- Upgrade the pass to unroll-and-optimize progressively, and write the results to the terminal so we can plot a few different functions and plot the results look like on a curve to confirm that a greedy approach would actually be appropriate
- Implement the greedy selection.

I'm hoping that, once we incorporate arithmetic circuit optimization, the same pass could be re-used for that and apply to non-CGGI contexts.

j2kun:
The first step of this is done: optimizing the inner-most body of a loop-nest.

j2kun:
Last step: add a helper script that runs the loop unrolling with lots of different unroll factors, and prints the relative statistics.

PR #270, opened 2023-11-18
Generalize yosys optimizer to operate on secret.generic

j2kun:
See https://github.com/google/heir/issues/268 for context


j2kun:
Hmmm, I was trying to make this nicer by refactoring the part that invokes Yosys into a RewritePattern, but it seems like the Yosys API is global and doesn't support being run in that context. I end up getting a bunch of errors like `malloc_consolidate(): invalid chunk size`

j2kun:
OK so this is a preliminary idea and I wanted to have the chance to talk to you about it, but that will have to wait for a week due to holidays. Let me just write up my idea here in the mean time.

tl;dr: you can try this:

```
bazel run //tools:heir-opt -- --pass-pipeline='builtin.module(func.func(secret.generic(yosys-optimizer)))' $PWD/tests/yosys_optimizer/add_one_generic.mlir
```

[result in a gist](https://gist.github.com/j2kun/146b190ba53ed6bf5aa2afa63f39c6c8)

And chain it with `secret-distribute-generic`

```
bazel run //tools:heir-opt -- --pass-pipeline='builtin.module(func.func(secret.generic(yosys-optimizer)), secret-distribute-generic)' $PWD/tests/yosys_optimizer/add_one_generic.mlir
```

[result in a gist](https://gist.github.com/j2kun/46ee5fdd21ee6d660df1f4ab83dcdfa7)

Longer explanation:

My general idea is that we should maintain the `secret.secret` as long as we don't have any lower-level FHE dialect involved in the computation. The way we can achieve this in the `tosa-to-tfhe-rs` flow is as follows:

- Convert a model to arith (affine loops + arith loop bodies)
- Wrap the computational parts of each loop body in a single `secret.generic`, as well as the computational parts between each loop.  (This requires somehow ensuring there are no intermediate stores/loads, or pushing them to the start/end of the loop body; I don't remember what we did for this before, but I imagine we just relied on the loads/stores just so happening to be at the start/end of the loop body)
- Run (a cleaned up version of) this yosys optimizer pass in this PR on each `secret.generic` op, possibly with loop-unrolling optimizations as discussed in https://github.com/google/heir/issues/257. The output of that are `secret.generic` ops, one for each block form the previous step, and each containing many `comb` ops.
- Run `secret-distribute-generic` on the result, which splits the generics down to single comb ops.
- Convert each single op to a cggi op (at which point the `i1`s and `tensor<?xi1>`s get converted to `lwe.ciphertext`).
- Run the lwe to cggi pass and onward as usual.

This ensures that `comb` ops never have overloaded semantics: they are only ever allowed to operate on cleartext types, and it's the conversion of a secret.generic op _containing_ a lut op that is allowed to become an `lwe` lut op.

j2kun:
> should there be asserts that there's only one op per generic body...?

As a precondition for the passes that require this, yes.

> whether the change to tensor<?xi1> from #264 is still compatible.

I suspect we can make the outline in this work regardless of which type we choose, but I agree tensor is the better choice.

j2kun:
@asraa Updated this PR and it should be ready at least for us to discuss:

- Added "manual" type conversions before and after the `secret.generic` being converted. This is essentially the `arith` version of `comb.extract_bit` and the inverse.
- Updated the tests to use `secret.generic` instead of func.

j2kun:
Still todo:

- Replace the manual type conversions with a custom op (`secret.cast`) which will be cleaned up during the dialect conversion.

j2kun:
@asraa ready when you are

include/Dialect/Secret/IR/SecretOps.td
```diff
@@ -154,5 +154,35 @@ def Secret_GenericOp : Secret_Op<"generic", [
...
+    A `cast` operation represents a type cast from one secret type to another,
+    that is used to enable the intermixing of various equivalent secret types
+    before a lower-level FHE scheme has been chosen.
+
+    For example, `secret.cast` can be used to convert a `secret<i8>` to a
+    `secret<tensor<8xi1>>` as a compatibility layer between boolean and
+    non-boolean parts of a program. The pass that later lowers the IR to
+    specific FHE schemes would need to replace these casts with appropriate
+    scheme-specific operations, and it is left to those later passes to
+    determine which casts are considered valid.
```
asraa, line 169: Hmm.. I think I agree that no verifiers should check that there's equality in the # of bits / no loss of information. I think this could end up representing information loss or casting to a larger type, depending on user constraints for how much leeway there is anyway (maybe we cast to some representation with less precision)

lib/Target/Verilog/VerilogEmitter.cpp
```diff
@@ -651,11 +679,32 @@ LogicalResult VerilogEmitter::printOperation(math::CountLeadingZerosOp op) {
...
+    mlir::heir::secret::GenericOp op,
+    std::optional<llvm::StringRef> moduleName) {
+  llvm::StringRef name;
+  if (moduleName.has_value())
+    name = moduleName.value();
+  else {
+    // I wanted something more unique here, but an op.getLoc() doesn't print
+    // as a valid verilog identifier. Maybe with enough string massaging it
+    // could.
+    name = "generic_body";
```
asraa, line 692: Since this isn't unique, do you ever expect emitting Verilog for an IR with more than one generic? There's a few cases now of verilog translation, right?
(1) No secret.generic, full IR translation (current support)
(2) A full IR with one secret.generic body (one unique generic_body?)
(3) A full IR with multiple secret.generic bodies

I think (3) isn't actually supported properly by `--emit-verilog` because of this fixed name, right? I think we want YosysOptimizer to run the `translateToVerilog` func on the `secret.generic` ops, but I'm not sure what happens when you run the full VerilogEmitter on (2) and (3) - maybe (2) results in an internal module? and (3) maybe results in duplicate module names?

Could you add a test of one or multiple secret.generic bodies in `tests/verilog/secret_generic.mlir`?

I ended up hashing the for loop stored value for the for loop [names](https://github.com/google/heir/blob/649b4f689e97a0109215b4949f455d8115b1f4e7/lib/Conversion/MemrefToArith/ExtractLoopBody.cpp#L151C51-L151C76), I think op.getLoc() wasn't reliable while I was mutating the IR in that case... but I assume you could hash it here

include/Target/Verilog/VerilogEmitter.h
```diff
@@ -38,8 +46,28 @@ class VerilogEmitter {
...
 
+  // A helper to generalize the work of emitting a func.func and a
+  // secret.generic
+  LogicalResult printFunctionLikeOp(Operation *op,
+                                    llvm::StringRef verilogModuleName,
+                                    ArrayRef<BlockArgument> arguments,
+                                    TypeRange resultTypes,
+                                    // FIXME: what is the right type for this?
+                                    // would like to pass something like
+                                    // ArrayRef<Block>
```
asraa, line 57: nit: still relevant? I figure since this matched `getBlocks()` return value this is probably easiest.

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -79,63 +82,234 @@ struct YosysOptimizer : public impl::YosysOptimizerBase<YosysOptimizer> {
...
+        return !castOps.contains(operand.getOwner());
+      });
+  return success();
+}
+
+// Optimize the body of a secret.generic op.
+// FIXME: consider utilizing
+// https://mlir.llvm.org/docs/PassManagement/#dynamic-pass-pipelines
+void YosysOptimizer::runOnOperation() {
+  auto result = getOperation()->walk([&](secret::GenericOp op) {
```
asraa, line 292: Took me a while to realize, but right now we only end up caring about secret.generic bodies (since all out args are ciphertext args). later, lets say we have some plaintext value arg and there's some circuit logic on that arg outside of a secret.generic - in that case, we still operate on the plaintext value in normal plaintext arithmetic, but it may need to be converted to a booleanized input depending on the case - we tackle that later.

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -79,63 +82,234 @@ struct YosysOptimizer : public impl::YosysOptimizerBase<YosysOptimizer> {
...
+  return success();
+}
+
+/// Convert a secret.generic's results from secret.secret<tensor<3xi1>>
+/// to secret.secret<i3>.
+LogicalResult convertOpResults(secret::GenericOp op,
+                               DenseSet<Operation *> &castOps,
+                               SmallVector<Value> &typeConvertedResults) {
+  for (Value opResult : op.getResults()) {
+    if (!opResult.getType().isa<secret::SecretType>()) {
```
asraa, line 149: could you add a testcase for this case of a non-secret yield? I'm having trouble understanding what happens (why doesn't the tensor.from_elements result need to be concatted back to an int?)

asraa (commented):
Thank you! This was tough...

lib/Target/Verilog/VerilogEmitter.cpp
j2kun: I agree that not having a unique name is probably bad in general, but I think in this case it's actually OK.

For one, the YosysOptimizer is calling this once per generic op (walking the IR) and then deleting the resulting func op after putting it back in line in the original IR. So there's no name conflicts there.

The only other way this could break is if someone uses `emit-verilog` directly on an IR with secret.generic in it. But that would be invalid, right? Because the generic would be inside a func, that would end up creating two nested verilog modules, and that would be invalid verilog. Or, put another way, anyone who would have a reason to use `emit-verilog` directly should have already lowered their IR to remove the secret dialect entirely.

Perhaps we could just protect this better by having two entry points to this emitter class? Or have the `emit-verilog` pass pre-verify there's no secret involved?

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
j2kun: Actually, I think this code is unreachable. `secret.generic` always has a secret result. This is enforced here: https://github.com/google/heir/blob/6e58f1d4686193262b1b52c572a64e57fddb972a/lib/Dialect/Secret/IR/SecretOps.cpp#L47-L53

I will remove this check and add a comment.

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
j2kun: So this pass still does have the `arith` ops required to booleanify plaintext inputs to the generic. That's in the branch on line 120 calling to `convertIntegerValue`. I opted for that instead of using something like `unrealized_conversion_cast`, but let me know if you think I should tweak it.

```c++
    if (!opOperand.get().getType().isa<secret::SecretType>()) {
      // The type is not secret, but still must be booleanized
      OpBuilder builder(op);
      auto fromElementsOp = convertIntegerValue(opOperand.get(), convertedType,
                                                builder, op.getLoc());
      typeConvertedArgs.push_back(fromElementsOp.getResult());
      continue;
    }
```

This implies that the circuit logic has to be inside the generic, and it seems that we would be better to fold it to occur _outside_ the generic (filed https://github.com/google/heir/issues/299 for a canonicalization pattern to hoist ops that can be hoisted)

include/Target/Verilog/VerilogEmitter.h
j2kun: I was just waiting for a response to my question about it: https://discord.com/channels/636084430946959380/636732535434510338/1175228413070561322

Will fix.

j2kun:
Fixed all but that last remaining comment.

lib/Target/Verilog/VerilogEmitter.cpp
asraa: Yeah, I agree that the cases (2) and (3) that have a secret.generic shouldn't likely be allowed by the `emit-verilog` command for now.

>  Because the generic would be inside a func, that would end up creating two nested verilog modules, and that would be invalid verilog. 

Yeah, it would need to unnest the two modules which we can do like the for loop cases, but I think it's completely unnecessary.

> Or have the emit-verilog pass pre-verify there's no secret involved?

Yeah! I was looking to see if we can just disallow the VerilogEmitter from working on secret types!

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
asraa: Actually the case I was thinking of was something like this:

```
func @main(%plaintext: i8, %secret_data: i8) {
   %0 = arith.addi 100, %plaintext
   secret.generic(%0, %secret_data) { ... }
} 
```

so that first instruction happens before the booleanization for secret.generic.

In this case, we skip booleanizing that first instruction at all, but I think that's fine for now (or maybe that's what you mean by " it seems that we would be better to fold it to occur outside the generic"?).

asraa (approved):
LGTM! It'd be nice to add a test error out on verilog emitter with secret.generic, leaving that comment open.

lib/Target/Verilog/VerilogEmitter.cpp
j2kun: Fixed and tested!

PR #272, opened 2023-11-20
add or-tools dependency to WORKSPACE

j2kun:
This is so that we have access to a MIP solver (SCIP) and a linear solver (eigen) for use in future optimization passes. Also see https://jeremykun.com/2023/11/15/mlir-a-global-optimization-and-dataflow-analysis/

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -4,6 +4,7 @@ workspace(name = "heir")
 
 load(
     "@bazel_tools//tools/build_defs/repo:git.bzl",
+    "git_repository",
     "new_git_repository",
 )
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -193,3 +194,124 @@ org_gnu_gnulib()
...
+
+git_repository(
+    name = "highs",
+    branch = "bazel",
+    remote = "https://github.com/ERGO-Code/HiGHS.git",
+)
+
+# ## Swig support
+# # pcre source code repository
+# new_git_repository(
```
asraa, line 305: was this one ever needed since we're already using c/c++, or is there something internal to or-tools that might need it?

asraa (approved):

PR #275, opened 2023-11-21
Add an LWE noise propagation model

j2kun:


j2kun:
Reminding myself what's left before vacation:

- Add a new pass called `validate-noise` that runs the noise propagation analysis and validates the noise does not exceed a max allowable noise (I think the way I've set it up the max is inferrable from the LWE params alone)
- Test the pass (tests/noise/validate_noise.mlir)

j2kun:
TODO

- split into smaller PRs
- fix FIXMEs
- infer max noise as best I can
- attach GH issue to known deficiencies in the concrete noise model
- test

j2kun:
Putting this on hold for a few more days while I work on other priorities. Notes to self:

- Last thing I did was replace the "optional<int>" version of the variance lattice with a three-state variant: uninitialized (not a ciphertext value), set (a known value), and unbounded (a pessimistic fixedpoint).
- The tests aren't quite raising errors how I expect, for example in `validate_noise_errors.mlir` it's not raising an error because the function's block argument isn't being set to the right value. It seems it should either be unbounded or lwe.add should have special handling for uninitialized error (currently says "unknown" when it should be more specific).
- I did not end up overriding `visitNonControlFlowArguments` because I _think_ it's not necessary, but I may be wrong on that.

PR #286, opened 2023-11-28
docs: add docs on how to use the templates on getting started

asraa:


docs/content/en/docs/getting_started.md
```diff
@@ -64,3 +64,45 @@ To run them manually, run
 ```bash
 pre-commit run --all-files
 ```
+
+## Creating a New Pass
+
+The `templates` folder contains Python scripts to create boilerplate for new conversion or (dialect-specific) transform passes.
```
j2kun, line 70: Maybe add `python templates/template.py --help` which should display a nice help string (or maybe we need to add that).

j2kun (approved):

docs/content/en/docs/getting_started.md
asraa: Currently doesn't work, but I'll fix in a follow up issue!

PR #292, opened 2023-11-28
Add LWE/CGGI parameters and passes to set defaults

j2kun:
After our chat, I had to add two more things:

- convert function signatures in the LWE default setting pass
- support adding the default LWE `params` attribute on `lwe.trivial_encrypt` ops


include/Dialect/CGGI/Transforms/Passes.td
```diff
@@ -0,0 +1,19 @@
+#ifndef INCLUDE_DIALECT_CGGI_TRANSFORMS_PASSES_TD_
+#define INCLUDE_DIALECT_CGGI_TRANSFORMS_PASSES_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def SetDefaultParameters : Pass<"cggi-set-default-parameters"> {
+  let summary = "Set default parameters for CGGI ops";
+  let description = [{
+  This pass adds default parameters to all CGGI ops as `cggi_params` named
```
asraa, line 9: nit: it would be nice to mention the default parameters somewhere in the docs for here and the LWE pass, either inline or with a link to the source code (it seems like the jaxite params are used)

include/Dialect/LWE/Transforms/Passes.td
```diff
@@ -0,0 +1,20 @@
+#ifndef INCLUDE_DIALECT_LWE_TRANSFORMS_PASSES_TD_
+#define INCLUDE_DIALECT_LWE_TRANSFORMS_PASSES_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def SetDefaultParameters : Pass<"lwe-set-default-parameters"> {
+  let summary = "Set default parameters for LWE ops";
+  let description = [{
+  This pass adds default parameters to all `lwe` types as the `lwe_params`
+  attribute, and for `lwe` ops ops as the `params` attribute, overriding any
```
asraa, line 10:
```suggestion
  attribute, and for `lwe` ops as the `params` attribute, overriding any
```

tests/cggi/set_default_parameters.mlir
```diff
@@ -0,0 +1,24 @@
...
+  %1 = arith.constant 1 : i1
+  %2 = lwe.encode %0 { encoding = #encoding }: i1 to !plaintext
+  %3 = lwe.encode %1 { encoding = #encoding }: i1 to !plaintext
+  // CHECK: lwe.trivial_encrypt
+  %4 = lwe.trivial_encrypt %2 { params = #params } : !plaintext to !ciphertext
+  // CHECK: lwe.trivial_encrypt
+  %5 = lwe.trivial_encrypt %3 { params = #params } : !plaintext to !ciphertext
+  // CHECK cggi.lut3
+  // Testing that the pass adds the right kind of attribute to the op
+  // CHECK cggi_params = #cggi.cggi_params
```
asraa, line 21: nit: could you please fix this syntax for the CGGI op?

asraa (approved):
Thanks! May end up as a future problem, but I also expect any `lwe.unspecified_bit_field_encodings` should also have a default. The unspecified encoding only specifies the bit length, and then the cleartext_start would need specification. I assume that needs to be done to calculate the invariant :/

include/Dialect/CGGI/Transforms/Passes.td
j2kun: Done!

j2kun:
Due to an issue with copybara integration, this was merged squashed with https://github.com/google/heir/commit/5b025b1b2d8786a15867ea8ff7ca7894108ee279

Issue #287, opened 2023-11-28
bgv : Add a verifier for Rotate


Maokami:
I believe there should be a check ensuring that the dimension of the ciphertext, which serves as the operand of rotation, is two.

Justification:
- SEAL throw an error when the above condition is violated:
https://github.com/microsoft/SEAL/blob/206648d0e4634e5c61dcf9370676630268290b59/native/src/seal/evaluator.h#L1046
- There's a bug in OpenFHE when the above condition is violated:
https://github.com/openfheorg/openfhe-development/issues/575

However, due to my lack of HE knowledge, I don't know why the rotation doesn't work when the dimension of ciphertext is greater than 2. Let me know if anyone knows!

j2kun:
I will summarize the work needed to do this here, in case you are looking to dip your feet into the coding. A high-level explanation of how verifiers work is [here](https://jeremykun.com/2023/09/13/mlir-verifiers/).

- Add `let hasVerifier = 1;` to the BGV rotate op in [BGVOps.td](https://github.com/google/heir/blob/1a2ccfe22f2038638460f09442634e1768c8781d/include/Dialect/BGV/IR/BGVOps.td#L84)
- Add a new implementation of a function with signature `LogicalResult RotateOp::verify() { ... }` to [BGVDialect.cpp](https://github.com/google/heir/blob/1a2ccfe22f2038638460f09442634e1768c8781d/lib/Dialect/BGV/IR/BGVDialect.cpp)
- Add a lit test in [tests/bgv](https://github.com/google/heir/tree/1a2ccfe22f2038638460f09442634e1768c8781d/tests/bgv) that uses `--verify-diagnostics` to assert the appropriate error message is reported when verification fails. A good example is [here](https://github.com/google/heir/blob/1a2ccfe22f2038638460f09442634e1768c8781d/tests/secret/verifier.mlir#L1-L4)

asraa:
> Let me know if anyone knows!

While I don't think it's theoretically impossible to apply a galois rotation on a length 3 ciphertext, I think the reason both implementations avoid doing so is because the Galois Keys are computed to switch the ciphertext from being encrypted from a key `(1, s^i)` to a `(1, s)`. If the ciphertext has length 3, then I'm not sure why there aren't static size checks, but the key switching matrix would only apply to the first two polys of the ciphertext and then probably leave an incorrect result.

Maokami:
@j2kun @asraa Thank you for the pointers and explanations!

[...]

PR #295, opened 2023-11-29
Add a verifier for bgv.rotate and corresponding test (#287)

Maokami:
- add verifier for bgv.rotate
- add simple expected-error-test for bgv.rotate

### DIFF
--- a/lib/Dialect/BGV/IR/BGVDialect.cpp
+++ b/lib/Dialect/BGV/IR/BGVDialect.cpp
@@ -42,45 +42,56 @@ void BGVDialect::initialize() {
 }
 
 LogicalResult MulOp::verify() {
-  auto x = this->getX().getType();
-  auto y = this->getY().getType();
+  auto x = getX().getType();
+  auto y = getY().getType();
   if (x.getDim() != y.getDim()) {
-    return this->emitOpError() << "input dimensions do not match";
+    return emitOpError() << "input dimensions do not match";
   }
-  auto out = this->getOutput().getType();
+  auto out = getOutput().getType();
   if (out.getDim() != 1 + x.getDim()) {
-    return this->emitOpError() << "output.dim == x.dim + 1 does not hold";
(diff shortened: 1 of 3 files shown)


tests/bgv/verifier.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: heir-opt --verify-diagnostics --split-input-file %s
```
j2kun, line 1: Nit: `--split-input-file` is only necessary if you have multiple tests in the same file that must be parsed as if they were separate files (it's a weird convention), and the separator to mark when to split the file is `// -----`

lib/Dialect/BGV/IR/BGVDialect.cpp
```diff
@@ -53,6 +53,17 @@ LogicalResult MulOp::verify() {
   }
   return success();
 }
+LogicalResult Rotate::verify() {
+  auto x = this->getX().getType();
```
j2kun, line 57: nit: I know the rest of this file uses `this->`, but it should not be necessary (here or in the rest of the file). I think you can just delete `this->` everywhere and it should run the same.

Context: Generated Op classes have this strange property that they are treated both as their own class (with named arguments to the fields created via tablegen) and simultaneously as a pointer to `Operation`, which has methods that generically apply to all ops, and can do things like mutate the operation in place.

j2kun (approved):
Looks great, thanks for the contribution! I added two comments for optional improvements. If you don't have time for those I can go in and fix them after it merges.

tests/bgv/verifier.mlir
Maokami: Applied!

lib/Dialect/BGV/IR/BGVDialect.cpp
Maokami: Applied!

j2kun (approved):

Issue #282, opened 2023-11-28
"Secret-ize" data in an MLIR module


asraa:
For the tf2fhe pipeline, we need to convert inputs to secret - here we can utilize an attribute on the function, so I'll need to create two things:
* Augment tosa-to-arith to add an attribute on the func we want to "Secret"-ize
* Secretize pass

j2kun:
This is done right? OK to close?

asraa:
Yep! both are done now.

PR #298, opened 2023-11-29
feat: add secretize pass to annotate func args with secret

asraa:
Initial part of https://github.com/google/heir/issues/282

The top-level function can be specified the same way the cpu runner is:

```
heir-opt -secretize=entry-function=hello_world %s
``` 

Resulting in an IR whose entry-function arguments are annotated with a `secret` attribute:

```
module {
    func.func @hello_world(%arg0: i1 {secret.secret}, %arg1: i1 {secret.secret}) -> () {
        %0 = comb.truth_table %a, %b -> 6 : ui4
        return
    }
}

```

### DIFF
--- a/include/Dialect/Secret/IR/SecretDialect.td
+++ b/include/Dialect/Secret/IR/SecretDialect.td
@@ -15,6 +15,13 @@ def Secret_Dialect : Dialect {
     custom types for arithmetic on secret integers of various bit widths.
   }];
 
+  let extraClassDeclaration = [{
+    /// Name of the attribute indicate whether an argument of a function is a
+    //secret.
+    constexpr const static ::llvm::StringLiteral
+        kArgSecretAttrName = "secret.secret";
+  }];
+
   let cppNamespace = "::mlir::heir::secret";
   let useDefaultTypePrinterParser = 1;
 }
--- a/include/Transforms/Secretize/Secretize.td
+++ b/include/Transforms/Secretize/Secretize.td
@@ -1,0 +1,25 @@
+#ifndef INCLUDE_TRANSFORMS_SECRETIZE_SECRETIZE_TD_
+#define INCLUDE_TRANSFORMS_SECRETIZE_SECRETIZE_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def Secretize : Pass<"secretize", "ModuleOp"> {
+  let summary = "Adds secret argument attributes to entry function";
+
+  let description = [{
+    Adds a secret.secret attribute argument to each argument in the entry
+    function of an MLIR module. By default, the function is `main`. This may be
+    overridden with the option -entry-function=top_level_func.
+  }];
+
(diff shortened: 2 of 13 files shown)


include/Dialect/Secret/IR/SecretDialect.td
```diff
@@ -15,6 +15,13 @@ def Secret_Dialect : Dialect {
     custom types for arithmetic on secret integers of various bit widths.
   }];
 
+  let extraClassDeclaration = [{
+    /// Name of the attribute indicate whether an argument of a function is a
+    //secret.
+    constexpr const static ::llvm::StringLiteral
+        kArgSecretAttrName = "secret.secret";
```
j2kun, line 22: nit: any reason to call it `secret.secret` as opposed to just `secret`?

include/Transforms/Secretize/Secretize.td
```diff
@@ -0,0 +1,25 @@
...
+
+include "mlir/Pass/PassBase.td"
+
+def Secretize : Pass<"secretize", "ModuleOp"> {
+  let summary = "Adds secret argument attributes to entry function";
+
+  let description = [{
+    Adds a secret.secret attribute argument to each argument in the entry
+    function of an MLIR module. By default, the function is `main`. This may be
+    overridden with the option -entry-function=top_level_func.
```
j2kun, line 12:
```suggestion
    overridden with the option `entry-function=top_level_func`.
```

j2kun (approved):
Nice one!

include/Dialect/Secret/IR/SecretDialect.td
asraa: Good question! It's because function arg attributes must be part of a dialect namespace, and so `secret` is interpreted as a 'builtin' / non-dialect attribute. I tried that first :)

```
heir/bazel-out/k8-dbg/bin/tests/secretize/multiple.mlir.test.runfiles/heir/tests/secretize/multiple.mlir:11:5: error: 'func.func' op arguments may only have dialect attributes
# |     func.func @main(%a: i1, %b: i1) -> () {
# |     ^
# | /usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/sandbox/linux-sandbox/5670/execroot/heir/bazel-out/k8-dbg/bin/tests/secretize/multiple.mlir.test.runfiles/heir/tests/secretize/multiple.mlir:11:5: note: see current operation: 
# | "func.func"() <{arg_attrs = [{secret}, {secret}], function_type = (i1, i1) -> (), sym_name = "main"}> ({
# | ^bb0(%arg0: i1, %arg1: i1):
# |   "func.call"(%arg0, %arg1) <{callee = @inner}> : (i1, i1) -> ()
# |   "func.return"() : () -> ()
# | }) : () -> ()
# `-----------------
```

Issue #306, opened 2023-12-01
Factor a secret.generic down to loop bodies


j2kun:
As a step toward tf2fhe, we start with a pass that encloses an entire function's body in secret.generic, and a later step converts a secret.generic to an optimized boolean circuit. We want that second step to apply only to the bodies of `affine.for` loop nests, as well as to the sections of ops that occur between loop nests.

This issue is to write a pattern similar to `distribute-generic`, but which stops short of converting everything to single-op generics, according to some configuration that specifies what the boundary is.

Because `affine.for` is intimately related to `affine.load` and `affine.store`, we should also hoist those ops outside of a secret.generic.

j2kun:
Remaining TODO: split a secret.generic when the op that can be factored through is mid-way through the generic's body.

asraa:
> I think we'd need to do this anyway for our immediate plan, because we can't include the loads/stores in the yosys optimizer. 

Why not? In your current example, I agree, since we can't translate the generic as a self-contained module unless the allocated memref is an argument to secret. In general though, we can translate loads and stores to read and writes to arrays in yosys. 

I think... actually... how do we get around the removal of IsolatedFromAbove in general with yosys optimizer?

To me, the original testcase should have hoisted an allocated memref to the following, but I know this breaks the usual type conversion to secret types as `secret<memref<10x...>>`, which means you can't read/write/allocate as normal memrefs anymore.
```
%alloc = memref.alloc() : memref<10xsecret<i8>>
```

j2kun:
Just added https://github.com/google/heir/pull/324, which should handle the issue about isolation: greedily apply that pattern at the start of the yosys optimizer pass, and it will capture all the ambient scope as (non-secret) arguments to the generic.

PR #308, opened 2023-12-01
Factor secret.generic through loops

j2kun:
Part of https://github.com/google/heir/issues/306

Last remaining work here: factor through an affine loop that isn't the first op in the IR.


lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
```diff
@@ -62,6 +76,7 @@ struct CollapseSecretlessGeneric : public OpRewritePattern<GenericOp> {
   }
 };
 
+// FIXME: move this pattern so it can be reused in canonicalization.
```
j2kun, line 79: These FIXMEs are addressed in https://github.com/j2kun/heir/pull/7, which is in my fork but I will re-open it against google/heir after this PR is merged.

asraa (approved):

PR #309, opened 2023-12-01
feat: add pass to convert func arguments to secret types & secret.generic body

asraa:


lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -0,0 +1,93 @@
...
+          IRMapping mp;
+          for (unsigned i = 0; i < blockArguments.size(); ++i) {
+            mp.map(opEntryBlock.getArgument(i), blockArguments[i]);
+          }
+          for (auto &entryOp : opEntryBlock.getOperations()) {
+            b.clone(entryOp, mp);
+          }
+          auto *returnOp = b.getBlock()->getTerminator();
+          b.create<secret::YieldOp>(loc, returnOp->getOperands());
+          returnOp->erase();
```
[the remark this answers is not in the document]
asraa: I played around with `block.moveBefore`, but seemed to have trouble getting it to work

lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -0,0 +1,93 @@
...
+        newInputs.push_back(newTy);
+      }
+    }
+
+    if (!hasSecrets) {
+      // Match failure, no secret inputs.
+      return failure();
+    }
+
+    llvm::SmallVector<Type, 4> newOutputs;
```
j2kun, line 44: nit, you might appreciate `llvm::map_range` here:

```suggestion
    auto newOutputs = llvm::map_range(
        op.getResultTypes(), [](Type t) { return secret::SecretType::get(t); });

    op.setFunctionType(
        FunctionType::get(getContext(), {newInputs}, TypeRange{newOutputs}));
```

lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -0,0 +1,93 @@
...
+    for (unsigned i = 0; i < op.getNumResults(); i++) {
+      newOutputs.push_back(secret::SecretType::get(op.getResultTypes()[i]));
+    }
```
j2kun, line 47:
```suggestion
    auto newOutputs = llvm::to_vector<6>(llvm::map_range(
        op.getResultTypes(), [](Type t) -> Type { return secret::SecretType::get(t); }));
```

lib/Transforms/Secretize/WrapGeneric.cpp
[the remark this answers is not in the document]
j2kun: I think your way is actually the best way because the generic hasn't been constructed yet at the time you want to do the moving. For example, I tried this, in which I tried giving access to the region pointer being constructed so that I could clone into the region. But it failed because the `cloneInto` was expecting the region input to have a terminator.

```
diff --git a/include/Dialect/Secret/IR/SecretOps.td b/include/Dialect/Secret/IR/SecretOps.td
index 6d7cced..a9fe08a 100644
--- a/include/Dialect/Secret/IR/SecretOps.td
+++ b/include/Dialect/Secret/IR/SecretOps.td
@@ -134,12 +134,18 @@ def Secret_GenericOp : Secret_Op<"generic", [
       CArg<"ValueRange">:$inputs,
       CArg<"TypeRange">:$outputTypes,
       CArg<"function_ref<void(OpBuilder &, Location, ValueRange)>">:$bodyBuilder
+    )>,
+    OpBuilder<(ins
+      CArg<"ValueRange">:$inputs,
+      CArg<"TypeRange">:$outputTypes,
+      CArg<"function_ref<void(OpBuilder &, Location, ValueRange, Region *)>">:$bodyBuilder
     )>
   ];
 
   let extraClassDeclaration = [{
     /// Defining the function type we use for building the body of affine.for.
     using BodyBuilderFn = function_ref<void(OpBuilder &, Location, ValueRange)>;
+    using BodyBuilderWithRegionFn = function_ref<void(OpBuilder &, Location, ValueRange, Region *region)>;
   }];
 
   // let hasCanonicalizer = 1;
diff --git a/lib/Dialect/Secret/IR/SecretOps.cpp b/lib/Dialect/Secret/IR/SecretOps.cpp
index f66c302..57073ce 100644
--- a/lib/Dialect/Secret/IR/SecretOps.cpp
+++ b/lib/Dialect/Secret/IR/SecretOps.cpp
@@ -217,6 +217,17 @@ void RevealOp::build(OpBuilder &builder, OperationState &result,
 void GenericOp::build(OpBuilder &builder, OperationState &result,
                       ValueRange inputs, TypeRange outputTypes,
                       GenericOp::BodyBuilderFn bodyBuilder) {
+  build(builder, result, inputs, outputTypes,
+        [bodyBuilder](OpBuilder &b, Location loc, ValueRange blockArguments,
+                      Region *region) {
+          // just ignore the region arg
+          bodyBuilder(b, loc, blockArguments);
+        });
+}
+
+void GenericOp::build(OpBuilder &builder, OperationState &result,
+                      ValueRange inputs, TypeRange outputTypes,
+                      GenericOp::BodyBuilderWithRegionFn bodyBuilder) {
   for (Type ty : outputTypes) result.addTypes(ty);
   result.addOperands(inputs);
 
@@ -231,7 +242,7 @@ void GenericOp::build(OpBuilder &builder, OperationState &result,
 
   OpBuilder::InsertionGuard guard(builder);
   builder.setInsertionPointToStart(&bodyBlock);
-  bodyBuilder(builder, result.location, bodyBlock.getArguments());
+  bodyBuilder(builder, result.location, bodyBlock.getArguments(), bodyRegion);
 }
 
 }  // namespace secret
diff --git a/lib/Transforms/Secretize/WrapGeneric.cpp b/lib/Transforms/Secretize/WrapGeneric.cpp
index 721976b..8ddeec1 100644
--- a/lib/Transforms/Secretize/WrapGeneric.cpp
+++ b/lib/Transforms/Secretize/WrapGeneric.cpp
@@ -54,15 +54,13 @@ struct WrapWithGeneric : public OpRewritePattern<func::FuncOp> {
     rewriter.setInsertionPointToStart(&opEntryBlock);
     auto newGeneric = rewriter.create<secret::GenericOp>(
         op.getLoc(), op.getArguments(), newOutputs,
-        [&](OpBuilder &b, Location loc, ValueRange blockArguments) {
+        [&](OpBuilder &b, Location loc, ValueRange blockArguments, Region *region) {
           //  Map the input values to the block arguments.
           IRMapping mp;
           for (unsigned i = 0; i < blockArguments.size(); ++i) {
             mp.map(opEntryBlock.getArgument(i), blockArguments[i]);
           }
-          for (auto &entryOp : opEntryBlock.getOperations()) {
-            b.clone(entryOp, mp);
-          }
+          op.getRegion().cloneInto(region, mp);
           auto *returnOp = b.getBlock()->getTerminator();
           b.create<secret::YieldOp>(loc, returnOp->getOperands());
           returnOp->erase();
```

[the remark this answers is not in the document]
asraa: > I think your way is actually the best way because the generic hasn't been constructed yet at the time you want to do the moving. 

ah! when I tried doing it after too, `moveBefore` doesn't allow for a nice IR mapping to the new block arguments, so that was pretty annoying as well.

Thank you!

Maybe one day we'll look back at it and realize there was a one-liner. For no, it works :)

lib/Transforms/Secretize/WrapGeneric.cpp
asraa: Woah!! I love it. Very functional.

include/Transforms/Secretize/Passes.td
```diff
@@ -22,4 +22,19 @@ def Secretize : Pass<"secretize", "ModuleOp"> {
...
 }
 
-#endif  // INCLUDE_TRANSFORMS_SECRETIZE_SECRETIZE_TD_
+def WrapGeneric : Pass<"wrap-generic", "ModuleOp"> {
+  let summary = "Wraps regions using secret args in secret.generic bodies";
+
+  let description = [{
+    This pass wraps regions that use secret arguments in secret.generic bodies.
+
+    Secret arguments are annotated using a `secret.secret` argument attribute.
```
j2kun, line 31: As I understand it, the pass currently only supports `func.func`, and has special treatment of that op due to how the type signature of the func comes with attributes on each type.

But this description makes it sound like the op is intended to support any region-bearing operation. Is that what you intended? Maybe `func.func` is just the initial limited support? I'm don't think block arguments can have arbitrary attributes attached to them, so I don't see a way to do it generically. But maybe we could define an interface for an op like "annotated secret" that would return a list of `OpOperand` corresponding to which arguments should be made secret? And somehow we would implement that interface for `func`... though I'm not sure how that would work if we can't change `func`'s tablegen.

If not, maybe just update the docs here to say it's just supporting `func.func`.

lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -0,0 +1,92 @@
...
+      // Match failure, no secret inputs.
+      return failure();
+    }
+
+    auto newOutputs = llvm::to_vector<6>(llvm::map_range(
+        op.getResultTypes(),
+        [](Type t) -> Type { return secret::SecretType::get(t); }));
+
+    op.setFunctionType(
+        FunctionType::get(getContext(), {newInputs}, {newOutputs}));
```
j2kun, line 49: Wouldn't this drop any function arguments that don't have the secret annotation? There's no "else" block to the if statement on line 28. This works for our case where we assume all inputs are secret, but in general I can imagine cases where we want to have a function with only some inputs secret.

I think it would be best to keep those arguments by appending to `newInputs` in the else case. The changes in https://github.com/google/heir/pull/308 mean you don't have to pass the non-secret args to the generic anymore, but you can still do it (meaning no other changes are needed in this file), and then after https://github.com/j2kun/heir/pull/7 the canonicalization pass will automatically remove them (see `RemoveNonSecretGenericArgs`).

j2kun (changes requested):
A few minor changes requested, otherwise LGTM!

include/Transforms/Secretize/Passes.td
j2kun: I asked this question about the possibility of implementing an out-of-tree interface for an upstream op: https://discord.com/channels/636084430946959380/642426447167881246/1180722773165871124

lib/Transforms/Secretize/WrapGeneric.cpp
asraa: argh! yes, i originally had that in, until I had at some point switched APIs to the line above ` op.getArgument(i).setType(newTy)` where I wouldn't need to consider the non-secrets. 

it turns out, there isn't a good way of setting the output type without `setFunctionType`, and so yes, I need to use the correct `newInputs` here.

asraa: fixed, added a testcase!

include/Transforms/Secretize/Passes.td
asraa: Updated the docs! Without being able to annotate block arguments generally, this pass won't work in other cases. I think it is possible to annotate arguments in `llvm.func` but I'm not even sure that is possible.

j2kun (approved):
LGTM!

Issue #316, opened 2023-12-06
wrap-generic: assertion failure when wrapping for loops


asraa:
I think the ordering of copying operations and then adding to the erasure list creates a segfault in wrap-generic.

This testcase
```
module attributes {tf_saved_model.semantics} {
  memref.global "private" constant @__constant_16xi32 : memref<16xi32> = dense<[0, 0, -5438, -5515, -1352, -1500, -4152, -84, 3396, 0, 1981, -5581, 0, -6964, 3407, -7217]>
  memref.global "private" constant @__constant_16x1xi8 : memref<16x1xi8> = dense<[[-9], [-54], [57], [71], [104], [115], [98], [99], [64], [-26], [127], [25], [-82], [68], [95], [86]]>
  func.func @main(%arg0: memref<1x1xi8> {iree.identifier = "serving_default_dense_input:0", secret.secret, tf_saved_model.index_path = ["dense_input"]}) -> (memref<1x16xi32> {iree.identifier = "StatefulPartitionedCall:0", tf_saved_model.index_path = ["dense_2"]}) attributes {tf_saved_model.exported_names = ["serving_default"]} {
    %c-128_i16 = arith.constant -128 : i16
    %0 = memref.get_global @__constant_16x1xi8 : memref<16x1xi8>
    %1 = memref.get_global @__constant_16xi32 : memref<16xi32>
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<1x16xi32>
    affine.for %arg1 = 0 to 1 {
      affine.for %arg2 = 0 to 16 {
        affine.for %arg3 = 0 to 1 {
          %2 = affine.load %arg0[%arg1, %arg3] : memref<1x1xi8>
          %3 = affine.load %0[%arg2, %arg3] : memref<16x1xi8>
          %4 = affine.load %1[%arg2] : memref<16xi32>
          %5 = arith.extsi %2 : i8 to i16
          %6 = arith.subi %5, %c-128_i16 : i16
          %7 = arith.extsi %6 : i16 to i32
          %8 = arith.extsi %3 : i8 to i32
          %9 = arith.muli %7, %8 : i32
          %10 = arith.addi %4, %9 : i32
          affine.store %10, %alloc[%arg1, %arg2] : memref<1x16xi32>
        }
      }
    }
    return %alloc : memref<1x16xi32>
  }
}
```

results in 
```
[...]

PR #317, opened 2023-12-06
fix: fix wrap-generic handling of affine.for

asraa:
Fixes https://github.com/google/heir/issues/316

I now replaced the op by op cloning to moving blocks, which avoids the issue of recursively walking the ops multiple times.

Added a testcase with the intermediate hello world IR!

### DIFF
--- a/lib/Transforms/Secretize/WrapGeneric.cpp
+++ b/lib/Transforms/Secretize/WrapGeneric.cpp
@@ -51,11 +51,14 @@ struct WrapWithGeneric : public OpRewritePattern<func::FuncOp> {
     op.setFunctionType(
         FunctionType::get(getContext(), {newInputs}, {newOutputs}));
 
-    // Create a secret.generic op and pull the original function block in.
+    // Create a new block where we will insert the new secret.generic and move
+    // the function ops into.
     Block &opEntryBlock = op.getRegion().front();
-    rewriter.setInsertionPointToStart(&opEntryBlock);
+    auto newBlock = rewriter.createBlock(
+        &opEntryBlock, opEntryBlock.getArgumentTypes(),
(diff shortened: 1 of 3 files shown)


lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -65,27 +68,19 @@ struct WrapWithGeneric : public OpRewritePattern<func::FuncOp> {
...
 
-    rewriter.replaceOp(
-        opEntryBlock.getTerminator(),
-        rewriter.create<func::ReturnOp>(op.getLoc(), newGeneric.getResults()));
-
-    for (auto *erase : llvm::reverse(opsToErase)) {
-      rewriter.eraseOp(erase);
-    }
+    Block &genericBlock = newGeneric.getRegion().front();
+    rewriter.inlineBlockBefore(&opEntryBlock,
```
j2kun, line 80: I'm surprised that this doesn't also need an IR mapping. Maybe it's not a problem because in the IR tested, the block being inlined never uses the original function arguments?

I added https://github.com/google/heir/issues/318 to investigate more later.

j2kun (approved):

lib/Transforms/Secretize/WrapGeneric.cpp
asraa: The call for `inlineBlockBefore`'s third argument is the arg values it's using to replace the original block args with, which in this case is the `  genericBlock.getArguments());`

I can add a test case that's smaller but hello world does use the func arguments.

asraa: Updated the test to include using the func arg.

PR #321, opened 2023-12-07
Generic split halfway

j2kun:
Fixes https://github.com/google/heir/issues/306


lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
```diff
@@ -229,66 +237,228 @@ struct SplitGeneric : public OpRewritePattern<GenericOp> {
...
+      GenericOp genericOp, ValueRange newValuesToYield,
+      PatternRewriter &rewriter) const {
+    YieldOp yieldOp = genericOp.getYieldOp();
+    LLVM_DEBUG(yieldOp.emitRemark() << "Before adding new values\n");
+    yieldOp.getValuesMutable().append(newValuesToYield);
+    // Can't print here because yield will fail verification
+    auto newTypes = llvm::to_vector<4>(
+        llvm::map_range(yieldOp.getValues().getTypes(), [](Type t) -> Type {
+          SecretType newTy = secret::SecretType::get(t);
+          LLVM_DEBUG(llvm::dbgs() << "Adding new type: " << newTy << "\n");
```
asraa, line 256: I'm surprised this works for a yield that already has some secret values yielded. wouldn't it double-secret the yielded type? (it's not clear to me that it's tested - both testcases are empty yields)

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
```diff
@@ -229,66 +237,228 @@ struct SplitGeneric : public OpRewritePattern<GenericOp> {
...
+  // generic op and cloning all its ops, since I can't find a way to modify the
+  // op's return type in-place.
+  //
+  // Returns the new op and the value range corresponding to the new result
+  // values of the generic.
+  std::pair<GenericOp, ValueRange> addNewYieldedValues(
+      GenericOp genericOp, ValueRange newValuesToYield,
+      PatternRewriter &rewriter) const {
+    YieldOp yieldOp = genericOp.getYieldOp();
+    LLVM_DEBUG(yieldOp.emitRemark() << "Before adding new values\n");
```
asraa, line 250: Since you don't print just the originals, should you remove this statement? It seems like you the next statements print the resulting yields with the new types (not just the new ones?)

asraa (approved):
As always, really nice!

include/Dialect/Secret/IR/SecretOps.td
```diff
@@ -156,6 +156,14 @@ def Secret_GenericOp : Secret_Op<"generic", [
   let extraClassDeclaration = [{
     /// Defining the function type we use for building the body of affine.for.
     using BodyBuilderFn = function_ref<void(OpBuilder &, Location, ValueRange)>;
+
+    // Returns the unique YieldOp in the body of this generic
+    YieldOp getYieldOp();
```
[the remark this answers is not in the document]
j2kun: This tablegen adds a new header declaration, and in this PR the implementation is added in `SecretOps.cpp`.

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
j2kun: The `yield` operand is always plaintext, and this `newTypes` variable is passed to `rewriter.create` as the result `TypeRange` of the overall `generic` op.

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
j2kun: Happy to remove it. I just found it really helpful to have a before/after dump for every modification I made to the IR, since it was really tricky to debug issues.

lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
asraa: ah right! thank you!

PR #324, opened 2023-12-07
Add `CaptureAmbientScope` pattern and shim pass

j2kun:
Added a shim pass so as to have a proper lit test for it. I don't think the pass itself is useful, and we can consider removing it once it's tested within the context of YosysOptimizer.

### DIFF
--- a/include/Dialect/Secret/Transforms/Passes.td
+++ b/include/Dialect/Secret/Transforms/Passes.td
@@ -40,5 +40,14 @@ def SecretDistributeGeneric : Pass<"secret-distribute-generic"> {
   ];
 }
 
+def SecretCaptureGenericAmbientScope : Pass<"secret-capture-generic-ambient-scope"> {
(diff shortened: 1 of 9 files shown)


include/Dialect/Secret/Transforms/Passes.td
```diff
@@ -40,5 +40,14 @@ def SecretDistributeGeneric : Pass<"secret-distribute-generic"> {
   ];
 }
 
+def SecretCaptureGenericAmbientScope : Pass<"secret-capture-generic-ambient-scope"> {
+  let summary = "Capture the ambient scope used in a secret.generic";
+  let description = [{
+    For each value used in the body of a `secret.generic` op, which is defined
+    in the ambient scope outside the `generic`, add it to the argument list of
+    the `generic.
```
asraa, line 48:
```suggestion
    the `generic`.
```

asraa (approved):
Thank you!

PR #326, opened 2023-12-07
fix: Make comb-to-cggi work after yosys-optimizer on secret.generics with multi-bit inputs

asraa:
This adds the glue necessary to the `comb-to-cggi` pass in order to operate correctly with `secret.generic` and `secret` types. This PR does:
* Add a `secret.cast` pattern to reconcile the conversion casts between multi-bit secret integers and secret tensors of single-bits. This pass converts assumes multi-bit secret integers are encoded into tensors of single-bit ciphertexts, so these `secret.cast` operations are not necessary to convert between these two types.
* Updates the LWE ciphertext type to use a minimum of 3 cleartext bits because we use LUT 3's. There is an attached TODO to generalize this.
* We had to update all operand types inside the secret.generic body to be promoted to their secret-ified types. (not just truth table ops, but also extract ops)

Background:
 
Assuming we can get Yosys optimized results out of a secret.generic body, we will then need to convert the comb operations inside a secret.generic to CGGI ciphertexts. There is an example IR in `tests/comb_to_cggi/add_one.mlir` that is produced from the yosys optimizer. 

The yosys optimizer splits multi-bit inputs into tensors of single-bits. These are contained within secret.generic bodies.

When the types and ops are hoisted out of the body, we need to convert the secret multi-bit tensors into ciphertext tensors.

This becomes problematic, because the internal ops like `tensor.extract` do not operate nicely with the `secret<tensor<8xi1>>`. To fix this, we promote the types to the tensors of LWE ciphertext.


asraa:
Still thinking about this. If we end up wanting to lower SIMD ciphertexts, then I don't think it would matter whether we are using `secret<tensor<1024xi8>>`, rather than `tensor<1024xsecret<i8>>`. Inside generic bodies, we would still be operating on the tensors and using tensor arithmetic.

When lowering in the equivalent arith-to-ckks or other RLWE types, then we would still probably hoist these types outside as tensors of RLWE ciphertexts that encode i8's in their coefficients, and then run a pass that can SIMD-ify the computation as desired and transform the tensors of ciphertexts into a single ciphertext (thereby changing the RLWE encoding on the types)

j2kun:
I'm a little confused from the IR you posted. I understand tensor can't take a secret as input, but the input to the `tensor.extract` is the result of `%1 = "secret.cast"(<<UNKNOWN SSA VALUE>>) : (!secret.secret<i8>) -> !secret.secret<tensor<8xi1>>`, and that unknown input was presumably the type-casted block argument, `%arg0: !lwe.lwe_ciphertext<encoding = #lwe.unspecified_bit_field_encoding<cleartext_bitwidth = 8>>`

Does that block argument represent a single i8 LWE ciphertext? The problem here is that we haven't chosen lwe encoding params yet, so we can't lower it to an appropriate tensor of ciphertexts?

It seems like going through the YosysOptimizer pass inherently forces a particular encoding choice: we have a minimum `cleartext_bitwidth` enforced by the ABC lut size, and we require the ciphertexts semantically represent single bits (so we know we have to lower the arg to a tensor of size 8).

Can't we just force the parameter selection in this lowering, say, by recognizing the use of comb inside a secret.generic? If we do that, then we should be able to resolve all the cast ops.

asraa:
> Does that block argument represent a single i8 LWE ciphertext?

Yes, this is the original `i8` argument of the function that's been type converted. The type converter was only converting an int of bit width N to an LWE ciphertext with N bitwidths. 

> The problem here is that we haven't chosen lwe encoding params yet, so we can't lower it to an appropriate tensor of ciphertexts?

I think I see what you mean - in our case, we want comb-to-cggi to lower the argument to a tensor of single bit LWE encodings, in which case presumably the casts would go away, I think (I'll try!)

However, I don't really understand if this is the right behavior in generality. Shouldn't we pick the encoding at the scheme level? It's still possible to extract out bits of an encoded ciphertext and operate on them (should the circuit require that), rather than breaking up the ciphertext into the most minimally sized parts (i1's here).

> we have a minimum cleartext_bitwidth enforced by the ABC lut size, and we require the ciphertexts semantically represent single bits (so we know we have to lower the arg to a tensor of size 8).

Definitely, yes, LUT # -> cleartext_bitwidth makes sense. However, somehow we also have to capture the expected range is single bits for the inputs to the original ciphertext, but I'm assuming that'll come through with the noise interface.

> Can't we just force the parameter selection in this lowering, say, by recognizing the use of comb inside a secret.generic? If we do that, then we should be able to resolve all the cast ops.

Let me try that! Thanks. In the meantime I tried working through updating the types, but there's so many cast assumptions on the types that it's proving pretty tedious.

j2kun:
I think this belongs in the comb to CGGI pass, not the yosys optimizer
pass. And even if it's not technically forced, I think starting from that
default assumption is the right approach for now.

asraa:
> I think this belongs in the comb to CGGI pass, 

I agree! The last commit started the changes there. I think the decision should be made at the scheme choice.

asraa:
@j2kun I finally (finally!) figured out how to get this pass to work. I had to add in a `secret.cast` pattern as well here, to reconcile the casts Yosys Optimizer places that should become no-ops in this lowering (since `secret.secret<i8>` is reconciled with `secret.secret<tensor<8xi1>>`.

lib/Conversion/CombToCGGI/CombToCGGI.cpp
```diff
@@ -151,6 +180,55 @@ struct ConvertTruthTableOp : public OpConversionPattern<TruthTableOp> {
...
+
+    if (auto inputIntTy = dyn_cast<IntegerType>(inputTy)) {
+      if (auto outputTensorTy = dyn_cast<RankedTensorType>(outputTy)) {
+        auto outputTensorElt =
+            dyn_cast<IntegerType>(outputTensorTy.getElementType());
+        assert(outputTensorElt &&
+               "expected underlying integer type of secret tensor");
+        if (inputIntTy.getWidth() ==
+            outputTensorTy.getNumElements() * outputTensorElt.getWidth()) {
+          rewriter.replaceOp(op, op.getInput());
```
j2kun, line 209: Shouldn't this be the adaptor's input? You want the value after it's been type-converted, right?

asraa: Yeah - It still works out (I wonder if the pattern applicator ends up converting it again...)

tests/comb_to_cggi/add_one.mlir
```diff
@@ -1,38 +1,95 @@
...
+    %c64_i8 = arith.constant 64 : i8
+    %19 = arith.andi %c1_i8, %c64_i8 : i8
+    %20 = arith.shrsi %19, %c6_i8 : i8
+    %21 = arith.trunci %20 : i8 to i1
+    %c7_i8 = arith.constant 7 : i8
+    %c-128_i8 = arith.constant -128 : i8
+    %22 = arith.andi %c1_i8, %c-128_i8 : i8
+    %23 = arith.shrsi %22, %c7_i8 : i8
+    %24 = arith.trunci %23 : i8 to i1
+    %from_elements = tensor.from_elements %3, %6, %9, %12, %15, %18, %21, %24 : tensor<8xi1>
```
j2kun, line 46: nit: I think if you add `--canonicalize` to the pass (or the lit test run command) it will fold this op away, since the output ends up re-extracting the individual bits to trivial_encrypt them.

lib/Conversion/CombToCGGI/CombToCGGI.cpp
asraa: Updated

tests/comb_to_cggi/add_one.mlir
asraa: True! Updated the testcase.

lib/Conversion/CombToCGGI/CombToCGGI.cpp
```diff
@@ -151,23 +196,53 @@ struct ConvertTruthTableOp : public OpConversionPattern<TruthTableOp> {
...
 struct CombToCGGI : public impl::CombToCGGIBase<CombToCGGI> {
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     auto *module = getOperation();
-    SecretTypeConverter typeConverter(context);
+
+    // TODO(https://github.com/google/heir/issues/250): The lutSize here is
+    // fixed on the assumption that the comb dialect is using ternary LUTs.
+    // Generalize lutSize by doing an analysis pass on the input combinational
+    // operations and integers.
```
j2kun, line 235: Optional: You might not need an analysis pass. I imagine you could just walk the IR here, type switch to match on comb ops, and use a max to produce either 2 or 3 based on whether there are any lut3's in the IR.

lib/Conversion/CombToCGGI/CombToCGGI.cpp
```diff
@@ -115,24 +164,20 @@ struct ConvertTruthTableOp : public OpConversionPattern<TruthTableOp> {
...
     SmallVector<mlir::Value, 4> lutInputs;
     for (Value val : adaptor.getOperands()) {
       auto integerTy = dyn_cast<IntegerType>(val.getType());
       // If any of the arguments to the truth table are ciphertexts, we must
       // encode and trivially encrypt the plaintext integers arguments.
       if (ciphertextArg && integerTy) {
-        auto encoding = lwe::UnspecifiedBitFieldEncodingAttr::get(
-            ctx, integerTy.getWidth());
+        assert(integerTy.getWidth() == 1 && "LUT inputs should be single-bit");
+        auto ctxtTy = getLWECiphertextForInt(ctx, integerTy, 3);
```
j2kun, line 178: Does this 3 also need updating as per the TODO below? If so, maybe add another comment here so we don't miss it later. Or, it can access the type converter (which is stored on the OpConversionPattern) and just reference the bit width field there.

tests/comb_to_cggi/add_one.mlir
```diff
@@ -1,38 +1,57 @@
...
+      %13 = comb.truth_table %extracted, %true, %false -> 6 : ui8
+      %14 = comb.truth_table %5, %extracted_1, %false -> 105 : ui8
+      %15 = comb.truth_table %6, %extracted_2, %false -> 105 : ui8
+      %16 = comb.truth_table %7, %extracted_3, %false -> 105 : ui8
+      %17 = comb.truth_table %8, %extracted_4, %false -> 105 : ui8
+      %from_elements = tensor.from_elements %12, %10, %17, %16, %15, %14, %4, %13 : tensor<8xi1>
+      secret.yield %from_elements : tensor<8xi1>
+    } -> !secret.secret<tensor<8xi1>>
+    // CHECK: [[CONCAT:%.*]] = tensor.from_elements
+    %26 = secret.cast %25 : !secret.secret<tensor<8xi1>> to !secret.secret<i8>
```
j2kun, line 53:
```suggestion
    %2 = secret.cast %1 : !secret.secret<tensor<8xi1>> to !secret.secret<i8>
```

j2kun:
I think the test failures may be related to this being deployed https://github.com/bazelbuild/bazel/issues/18958

Now I see this: 

> WARNING: --enable_bzlmod is set, but no MODULE.bazel file was found at the workspace root. Bazel will create an empty MODULE.bazel file. Please consider migrating your external dependencies from WORKSPACE to MODULE.bazel. For more details, please refer to https://github.com/bazelbuild/bazel/issues/18958.

But I don't see that warning in https://github.com/google/heir/actions/runs/7169640934

j2kun:
Confirmed, it may not be because of bazelmod, but it is definitely due to the one-hour-ago release of bazel version 7.0.0. I will push a patch to force an earlier bazel version for now

j2kun:
See https://github.com/google/heir/pull/331 for the fix

lib/Conversion/CombToCGGI/CombToCGGI.cpp
asraa: Yes! I was considering walking this as well. For now, I'm going to keep this as hard-coded, since the only comb operation we have supported for Yosys to output is LUT 3's.

The next change here would be if we do builtin Yosys gates and have a backup bitwidth size of 1 in case there's not LUTs found.

lib/Conversion/CombToCGGI/CombToCGGI.cpp
asraa: Right! Actually I can clean this up. It can't access the type converter's class here without a cast, but I know how to fix that and I'd prefer just calling the type converter directly and hiding this detail too :)

tests/comb_to_cggi/add_one.mlir
asraa: Ohhh nice thank you for catching this really crazy copy/paste error I made (I copy pasted the CHECKs from the previous IR but didn't update this line between the CHECKs)

lib/Conversion/CombToCGGI/CombToCGGI.cpp
asraa: Updated.

j2kun (approved):

PR #327, opened 2023-12-07
Properly handle yields when distributing generic through loops

j2kun:
Fixes https://github.com/google/heir/issues/306

The hello world test case is handled now, along with memrefs being preserved (conservatively, with no dataflow analysis) in secret generics.

### DIFF
--- a/lib/Dialect/Secret/IR/SecretPatterns.cpp
+++ b/lib/Dialect/Secret/IR/SecretPatterns.cpp
@@ -2,7 +2,9 @@
 
 #include "include/Dialect/Secret/IR/SecretOps.h"
 #include "include/Dialect/Secret/IR/SecretTypes.h"
-#include "mlir/include/mlir/IR/PatternMatch.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/MemRef/IR/MemRef.h"  // from @llvm-project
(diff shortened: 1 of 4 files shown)


lib/Dialect/Secret/IR/SecretPatterns.cpp
```diff
@@ -16,7 +18,19 @@ LogicalResult CollapseSecretlessGeneric::matchAndRewrite(
...
   }
 
-  YieldOp yieldOp = dyn_cast<YieldOp>(op.getBody()->getOperations().back());
+  // Allocation ops cannot be collapsed because they may be used in later
+  // generics. We should eventually do a proper dataflow analysis to ensure
+  // they can be collapsed when no secret data is added to them.
+  //
+  // There is no good way to identify an allocation op in general. Maybe we can
+  // upstream a trait for this?
+  for (auto &op : op.getBody()->getOperations()) {
```
asraa, line 27: I think you could also use `getOps<memref::AllocOp>()` and that way skip checking other ops

asraa (approved):
Thank you for the many comments! This was handy for review and no doubt will come in use later.

lib/Dialect/Secret/IR/SecretPatterns.cpp
j2kun: Fixed! TIL about `[[maybe_unused]]`

Issue #328, opened 2023-12-07
Add an OpenFHE exit dialect for arithmetic schemes


j2kun:

Maokami:
Hi, @j2kun!

I'm interested in this topic, so I created an initial version based on the tfhe-rs dialect. There may be design issues since I simply mimicked the tfhe-rs dialect, but I would appreciate it if you could take a look at it when you have a chance.

In my opinion, when compared to the tfhe-rs dialect, there are additional considerations that should be taken into account in the OpenFHE dialect.

1. How to represent plaintext in the dialect (OpenFHE primarily uses PackedEncoding for BGV/BFV and CKKSPackedEncoding for CKKS).
2. Since it supports multiple schemes, a way to represent which scheme to use in the dialect.
3. How to represent encryption parameters (N, the bitsizes of q_i's, the number of q_i's, etc.) in the dialect.

j2kun:
Thank you for this! I will take a closer look on Monday.

In the mean time, I wanted to point you to an encoding attribute I defined in https://github.com/google/heir/blob/153eaf6622101375ba9e687dc17fbe076c916bb5/include/Dialect/LWE/IR/LWEAttributes.td. There you can find

- `RLWE_PolynomialCoefficientEncoding`
- `RLWE_PolynomialEvaluationEncoding`
- `RLWE_InverseCanonicalEmbeddingEncoding`

We haven't figured out exactly how we want to use these attributes yet, but they may provide you with something to put on a plaintext or ciphertext to represent the chosen encoding.

Maokami:
Aha, I see, I'll take a little more time to look at it.

Maokami:
Hi @AlexanderViand-Intel 😸!

I'm just getting started with catching up on the project, but due to my limited experience, it's a bit challenging for me to figure out what needs to be done and where to begin. As a rough idea, I'm thinking about:

1. Adding OpenFHE dialect ops (relinearization, rescale/modswith, bootstrapping, ...)
2. Creating passes for OpenFHE dialect (from BGV dialect?)
3. OpenFHE code generation (If I understand correctly, this is what @j2kun is planning to do next, right?)

Do you have any additional tasks or priorities I should consider? Your guidance would be much appreciated! 😊

j2kun:
The weather has messed up my week a bit, but I can sketch out a plan later today for an end-to-end BGV path.

j2kun:
## Ops

[...]

PR #329, opened 2023-12-10
Draft: Initial version of OpenFHE dialect

Maokami:
This PR contains initial draft for OpenFHE exit dialect(#328)

- Added primitive types (RLWEPlaintext, RLWECiphertext, cryptocontext, publickey)
- Added primitive operators (add, sub, negate, mult, rotate)
- Added corresponding simple tests

### DIFF
--- a/include/Dialect/LWE/IR/LWETypes.td
+++ b/include/Dialect/LWE/IR/LWETypes.td
@@ -37,6 +37,16 @@ def LWECiphertext : LWE_Type<"LWECiphertext", "lwe_ciphertext"> {
   let assemblyFormat = "`<` struct(params) `>`";
 }
 
+def RLWECiphertext : LWE_Type<"RLWECiphertext", "rlwe_ciphertext"> {
+  let summary = "A type for RLWE ciphertexts";
+
+  let parameters = (ins
+    "::mlir::Attribute":$encoding,
+    OptionalParameter<"RLWEParamsAttr">:$rlwe_params
+  );
+
+  let assemblyFormat = "`<` struct(params) `>`";
+}
 
 def LWEPlaintext : LWE_Type<"LWEPlaintext", "lwe_plaintext"> {
   let summary = "A type for LWE plaintexts";
@@ -47,6 +57,16 @@ def LWEPlaintext : LWE_Type<"LWEPlaintext", "lwe_plaintext"> {
     This type keeps track of the plaintext integer encoding for the LWE
     plaintext before it is encrypted.
   }];
+
+  let parameters = (ins
+    "::mlir::Attribute":$encoding
+  );
+
+  let assemblyFormat = "`<` struct(params) `>`";
+}
+
+def RLWEPlaintext : LWE_Type<"RLWEPlaintext", "rlwe_plaintext"> {
+  let summary = "A type for RLWE plaintexts";
 
   let parameters = (ins
     "::mlir::Attribute":$encoding
--- a/include/Dialect/Openfhe/IR/OpenfheOps.td
+++ b/include/Dialect/Openfhe/IR/OpenfheOps.td
@@ -1,0 +1,104 @@
+#ifndef INCLUDE_DIALECT_OPENFHE_IR_OPENFHEOPS_TD_
+#define INCLUDE_DIALECT_OPENFHE_IR_OPENFHEOPS_TD_
+
+include "OpenfheDialect.td"
+include "OpenfheTypes.td"
+
+include "include/Dialect/LWE/IR/LWETypes.td"
+include "mlir/IR/BuiltinAttributes.td"
+include "mlir/IR/CommonTypeConstraints.td"
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+class Openfhe_Op<string mnemonic, list<Trait> traits = []> :
+        Op<Openfhe_Dialect, mnemonic, traits> {
+  let assemblyFormat = [{
+    operands attr-dict `:` `(` type(operands) `)` `->` type(results)
+  }];
+  let cppNamespace = "::mlir::heir::openfhe";
+}
+
+def EncryptOp : Openfhe_Op<"encrypt", [Pure]> {
+  let arguments = (ins Openfhe_CryptoContext:$cryptoContext, RLWEPlaintext:$plaintext, Openfhe_PublicKey:$publicKey);
+  let results = (outs RLWECiphertext:$output);
+}
+
+def NegateOp : Openfhe_Op<"negate", [
+    Pure,
+    AllTypesMatch<["ciphertext", "output"]>
+]> {
+  let arguments = (ins Openfhe_CryptoContext:$cryptoContext, RLWECiphertext:$ciphertext);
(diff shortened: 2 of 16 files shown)


include/Dialect/Openfhe/IR/OpenfheTypes.td
```diff
@@ -0,0 +1,55 @@
...
+  : Openfhe_Type< "Plaintext" # scheme, "plaintext" # scheme> {
+  let summary = "A plaintext corresponding to OpenFHE's ciphertext type in " # scheme # " scheme";
+}
+
+foreach scheme = ["BFV", "BGV", "CKKS"] in {
+  def Openfhe_Plaintext # scheme : Openfhe_Plaintext<scheme>;
+}
+
+def Openfhe_PlaintextType :
+  AnyTypeOf<[
```
j2kun, line 28: After looking over the OpenFHE pke API, it does appear that they have a [single plaintext type](https://github.com/openfheorg/openfhe-development/blob/4ebb28ea7bdd894a73bc5b73e59fcfbc78253308/src/pke/include/encoding/plaintext-fwd.h#L44 and [ciphertext type](https://github.com/openfheorg/openfhe-development/blob/4ebb28ea7bdd894a73bc5b73e59fcfbc78253308/src/pke/include/ciphertext-fwd.h#L45) for all three arithmetic schemes, and their ciphertext type is (in every example I can see from their docs) using the "double-CRT" representation. I think it would be a reasonable starting point to reuse the `lwe` types here, and effectively hard-code the double-CRT encoding choice when generating code. We can generalize it later as needed.

include/Dialect/Openfhe/IR/OpenfheTypes.td
```diff
@@ -0,0 +1,55 @@
...
+}
+
+def Openfhe_CiphertextType :
+  AnyTypeOf<[
+    Openfhe_CiphertextBFV,
+    Openfhe_CiphertextBGV,
+    Openfhe_CiphertextCKKS,
+  ]>;
+
+def Openfhe_ServerKey : Openfhe_Type<"ServerKey", "server_key"> {
```
j2kun, line 51: In the OpenFHE API they seem to call this type `CryptoContext`. Would you mind re-naming it to `Openfhe_CryptoContext`? I think it would help the user of this dialect understand the intended purpose of this type. In a perfect world, we can make this exit dialect a 1-1 mirror of the OpenFHE API.

include/Dialect/Openfhe/IR/OpenfheOps.td
```diff
@@ -0,0 +1,54 @@
...
+    AllTypesMatch<["lhs", "rhs", "output"]>
+]> {
+  let arguments = (ins
+    Openfhe_ServerKey:$serverKey,
+    Openfhe_CiphertextType:$lhs,
+    Openfhe_CiphertextType:$rhs
+  );
+  let results = (outs Openfhe_CiphertextType:$output);
+}
+
```
j2kun, line 53: Please add a `MulOp` corresponding to [`EvalMult`](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/include/cryptocontext.h#L1697)

I imagine we would also need [`RotateOp` ](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/include/cryptocontext.h#L2083) here, which would lower from [`bgv.rotate`](https://github.com/google/heir/blob/5cdfcea42de3a872922919eebeda44a26424538d/include/Dialect/BGV/IR/BGVOps.td#L74)

The `CryptoContext` API has a lot of other higher level ops (squaring, sin/cos, evaluating a polynomial, etc.). I think we should not add those yet, but consider adding them in the future if they are more than a convenient wrapper.

j2kun: You may also want to consider adding ops that incorporate cleartexts (looks like they have overloaded their API for this case, e.g., [for EvalMult](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/include/cryptocontext.h#L1942))

But I would be fine to wait until the need for it shows up in code generation.

j2kun (changes requested):
Looks great! I added a few comments that I think are necessary for the first implementation of the code generation. Otherwise I'm happy to merge it in.

include/Dialect/Openfhe/IR/OpenfheTypes.td
Maokami: Do you mean changing `Openfhe_ServerKey` to `Openfhe_CryptoContext`?

j2kun: Exactly, just a name change to make it closer to the OpenFHE API.

include/Dialect/Openfhe/IR/OpenfheTypes.td
asraa: Agree to put rlwe ciphertext types from the `lwe` dialect - but we currently don't have that implemented (we only have lwe ciphertexts there). So it may be worth putting the unified ciphertexts types there and then add their attributes (https://github.com/openfheorg/openfhe-development/blob/4ebb28ea7bdd894a73bc5b73e59fcfbc78253308/src/pke/include/ciphertext.h#L101) there when needed

Maokami: Do you mean that we need to add new `RLWECiphertextType` in the `lwe` dialect to represent the "unified ciphertexts types?

asraa: I think so! Lattigo also has a unified ciphertext type so it would also use a generic `RLWECiphertextType` (https://pkg.go.dev/github.com/ldsec/lattigo/v2/rlwe)

Maokami: Applied!

include/Dialect/Openfhe/IR/OpenfheTypes.td
Maokami: Applied!

include/Dialect/Openfhe/IR/OpenfheOps.td
Maokami: Applied!

include/Dialect/LWE/IR/LWETypes.td
```diff
@@ -37,6 +37,17 @@ def LWECiphertext : LWE_Type<"LWECiphertext", "lwe_ciphertext"> {
   let assemblyFormat = "`<` struct(params) `>`";
 }
 
+// TODO: Generalize RLWECiphertext
```
j2kun, line 40: No need for the TODO here or below on RLWEPlaintext. If you have some specific action you want to leave a note about, please attach it to a github issue (new or existing) such as

```
// TODO(https://github.com/google/heir/issues/337): add support for foo
```

j2kun (approved):
LGTM! Thanks for this contribution.

include/Dialect/LWE/IR/LWETypes.td
j2kun: I will mark the PR as `pull_ready` after this is resolved.

Maokami: Aha, Thanks!

Issue #320, opened 2023-12-06
Add `secret.generic` canonicalization pattern removing an unused yielded value


j2kun:
Be sure to add the pattern to `DistributeGeneric`

PR #334, opened 2023-12-12
Add a canonicalizer to remove unused secret.yield'ed values

j2kun:
Includes a bit of cleanup putting the helpers to add/remove yielded values on the GenericOp class.


j2kun:
Fixes https://github.com/google/heir/issues/320

lib/Dialect/Secret/IR/SecretPatterns.cpp
```diff
@@ -43,6 +43,29 @@ LogicalResult RemoveUnusedGenericArgs::matchAndRewrite(
   return hasUnusedOps ? success() : failure();
 }
 
+LogicalResult RemoveUnusedYieldedValues::matchAndRewrite(
+    GenericOp op, PatternRewriter &rewriter) const {
+  bool foundUnused = false;
```
asraa, line 48: small nit: you can remove this bool and just check `!valuesToRemove.empty()`

asraa (approved):

lib/Dialect/Secret/IR/SecretPatterns.cpp
j2kun: Fixed!

asraa (approved):

PR #347, opened 2023-12-15
Unroll and optimize

j2kun:
Main part of https://github.com/google/heir/issues/257

### DIFF
--- a/include/Transforms/YosysOptimizer/YosysOptimizer.td
+++ b/include/Transforms/YosysOptimizer/YosysOptimizer.td
@@ -13,13 +13,43 @@ def YosysOptimizer : Pass<"yosys-optimizer"> {
     Note that booleanization changes the function signature: multi-bit integers
     are transformed to a tensor of booleans, for example, an `i8` is converted
     to `tensor<8xi1>`.
+
+    The optimizer will be applied to each `secret.generic` op containing
+    arithmetic ops that can be optimized.
+
+    Optional parameters:
+
+    - `abc-fast`: Run the abc optimizer in "fast" mode, getting faster compile
+      time at the expense of a possibly larger output circuit.
+    - `unroll-factor`: Before optimizing the circuit, unroll loops by a given
+      factor. If unset, this pass will not unroll any loops.
   }];
+  // TODO(https://github.com/google/heir/issues/257): add option for the pass to select
+  // the unroll factor automatically.
 
   let dependentDialects = [
     "mlir::arith::ArithDialect",
     "mlir::heir::comb::CombDialect",
     "mlir::heir::secret::SecretDialect",
     "mlir::memref::MemRefDialect"
+  ];
+}
+
+def UnrollAndOptimizeAnalysis : Pass<"unroll-and-optimize-analysis"> {
+  let summary = "Iteratively unroll and optimize an IR, printing optimization stats.";
+
+  let description = [{
+    This pass invokes the `--yosys-optimizer` pass while iteratively applying
+    loop-unrolling. Along the way, it prints statistics about the optimized
+    circuits, which can be used to determine an optimal loop-unrolling factor
+    for a given program.
+  }];
+
+  let dependentDialects = [
+    "mlir::arith::ArithDialect",
+    "mlir::heir::comb::CombDialect",
+    "mlir::heir::secret::SecretDialect",
+    "mlir::tensor::TensorDialect"
   ];
 }
 
--- a/lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
+++ b/lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
@@ -17,16 +17,20 @@
 #include "include/Target/Verilog/VerilogEmitter.h"
 #include "lib/Transforms/YosysOptimizer/LUTImporter.h"
 #include "lib/Transforms/YosysOptimizer/RTLILImporter.h"
-#include "llvm/include/llvm/ADT/SmallVector.h"           // from @llvm-project
-#include "llvm/include/llvm/Support/Debug.h"             // from @llvm-project
-#include "llvm/include/llvm/Support/FormatVariadic.h"    // from @llvm-project
-#include "llvm/include/llvm/Support/raw_ostream.h"       // from @llvm-project
+#include "llvm/include/llvm/ADT/SmallVector.h"         // from @llvm-project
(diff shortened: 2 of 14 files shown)


j2kun:
Still TODO:

- Get an end-to-end example working with a loop nest that is unrolled and optimized
- Print stats to the terminal
- Figure out a heuristic for stopping

j2kun:
The merging of adjacent generics fails because the loop unrolling inserts `affine.apply` instructions between the generics

```mlir
affine.for %arg2 = 0 to 10 step 2 {
  secret.generic ins(%arg0, %0 : !secret.secret<memref<10x10xi8>>, !secret.secret<memref<1xi8>>) {
  ^bb0(%arg3: memref<10x10xi8>, %arg4: memref<1xi8>):
    %2 = affine.load %arg3[%arg1, %arg2] : memref<10x10xi8>
    %3 = affine.load %arg4[0] : memref<1xi8>
    %4 = arith.addi %2, %3 : i8
    affine.store %4, %arg4[0] : memref<1xi8>
    secret.yield
  }
  %1 = affine.apply affine_map<(d0) -> (d0 + 1)>(%arg2)
  secret.generic ins(%arg0, %0 : !secret.secret<memref<10x10xi8>>, !secret.secret<memref<1xi8>>) {
  ^bb0(%arg3: memref<10x10xi8>, %arg4: memref<1xi8>):
    %2 = affine.load %arg3[%arg1, %1] : memref<10x10xi8>
    %3 = affine.load %arg4[0] : memref<1xi8>
    %4 = arith.addi %2, %3 : i8
    affine.store %4, %arg4[0] : memref<1xi8>
    secret.yield
  }
}
```

j2kun:
After rebasing, now it's complaining that `Only functions with a single return type are supported`, and so it can't emit metadata. Tomorrow I will have to think about how to bypass this, since in our case the arguments are memrefs that are written to as outparams.

j2kun:
Making incremental progress here:

- Supporting zero return types in VerilogEmitter and EmitMetadata, because some of the generic bodies of the affine loops don't return anything (they just write to memrefs which are inputs).


Next step seems to be updating the post-yosys step `convertOpOperands` that adds a compatibility layer between the original input type and the yosys-converted input type. I think I may just expand the scope of `secret.cast` to cover memref conversions.

j2kun:
I ended up removing an assert that now results in the "compatibility layer" simply being `secret.cast` a `secret<memref<5x4xi8>>` to a `secret<tensor<160xi1>>` and back again, since I'm expecting @asraa to switch from tensor to memref in this code path anyway.

I also decided to expand affine.apply (which is inserted during loop unrolling) and replace it with standard arith ops, but now it runs into the problem that the inputs to the `secret.generic` are `index` type, and that is not supported in the VerilogEmitter. I was trying to remember how we handled this before... just pick a bit width and insert cast ops?

j2kun:
After a bit more progress, now I'm blocked on the fact that the design output by Yosys/ABC has this cell whose output write is the same as one of its input wires

```
  attribute \LUT 4'0110
  attribute \module_not_derived 1
  attribute \src "lib/Transforms/YosysOptimizer/yosys/map_lut_to_lut3.v:47.10-56.19"
  cell \lut3 \_1686_
    connect \A $abc$85582$new_n1096_
    connect \B \arg2 [0]
    connect \C 0
    connect \P0 1'0
    connect \P1 1'1
    connect \P2 1'1
    connect \P3 1'0
    connect \P4 0
    connect \P5 0
    connect \P6 0
    connect \P7 0
    connect \Y \arg2 [0]
  end
```

??? How is this possible?

j2kun:
OK, I've taken the approach of lifting the loads and stores out of the generic to optimize. Still TODO:

- Support for > 1 return value in yosys optimizer
- Finish the test I'm currently using to drive the feature work
- Add new tests:
  - Make a loop for which, when you unroll it you can actually optimize better than if you don't
  - Make a loop which has per-iteration dependent values, and add a forward store to load pattern

j2kun:
It works!!! My god, I'm free.

@asraa this is currently branched off https://github.com/google/heir/pull/413, and after that goes in I'll rebase this and clean up the commits a bit.

j2kun:
Added a more complicated test that computes a cumulative sum of an array, however, the central generic op seems to be sometimes split into multiple, bypassing the joint optimization. So I will have to pick this up tomorrow on that thread.

j2kun:
This PR still needs to be rebased over https://github.com/google/heir/pull/413

j2kun:
Rebased and squashed: @asraa it should be ready for review!

include/Transforms/YosysOptimizer/YosysOptimizer.td
```diff
@@ -23,4 +35,22 @@ def YosysOptimizer : Pass<"yosys-optimizer"> {
   ];
 }
 
+def UnrollAndOptimizeAnalysis : Pass<"unroll-and-optimize-analysis"> {
```
j2kun, line 38: To be fair, this part is not yet done. I will work on this in a followup PR.

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -121,8 +132,10 @@ Value convertIntegerValue(Value value, Type convertedType, OpBuilder &b,
   return allocOp.getResult();
 }
 
-/// Convert a secret.generic's operands secret.secret<i3>
-/// to secret.secret<memref<3xi1>>.
+/// Convert a secret.generic's operands:
+///
+///  - secret.secret<i3> -> secret.secret<tensor<3xi1>>
+///  - secret.secret<memref<Nxi3>> -> secret.secret<memref<3Nxi1>>
```
asraa, line 138: are these accurate? i think we generally are converting `i3` to `memref<3xi1>` and `memref<Nxi3>` to `memref<Nx3xi1>`

asraa (approved):

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
j2kun: Will revert. I'm not sure why I wrote this.

include/Transforms/YosysOptimizer/YosysOptimizer.td
j2kun: Added TODO

j2kun:
Merged via f6472d488a8fc9b81e9e2a6a8e7848eebabd3088

PR #349, opened 2023-12-15
fix: fix some bugs in secret patterns

asraa:
Hello world now completes. There were two bugs:
1. Operands were being erased inside a loop over the op operands, which caused issues in referencing deleted args.
2. There was an assertion that any arguments to an op that needs to be moved must be a block arg or defined "BeforeInBlock", but this fails when the blocks of the op to move and the arg's defining op are in different loops. Replaced with checking dominance.

### DIFF
--- a/lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
+++ b/lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
@@ -6,19 +6,21 @@
 #include "include/Dialect/Secret/IR/SecretOps.h"
 #include "include/Dialect/Secret/IR/SecretPatterns.h"
 #include "include/Dialect/Secret/IR/SecretTypes.h"
-#include "llvm/include/llvm/ADT/SmallVector.h"  // from @llvm-project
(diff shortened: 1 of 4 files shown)


lib/Dialect/Secret/Transforms/DistributeGeneric.cpp
```diff
@@ -298,6 +300,7 @@ struct SplitGeneric : public OpRewritePattern<GenericOp> {
            "source and target generics must be in the same block");
 
     IRMapping cloningMp;
+    mlir::DominanceInfo dom(sourceGeneric->getParentOfType<func::FuncOp>());
```
j2kun, line 303: Since this is only used for an assert, does this constructor incur any additional runtime penalty? I know computing dominance info can be expensive... If it does, consider putting it inside an `ifdef DEBUG` just before to the assert (see https://github.com/llvm/llvm-project/blob/e8a15eca92f1a10b3af4f4e52f54d9d2d7612bf5/libc/src/assert/assert.h#L17C1-L17C1)

j2kun (approved):

j2kun:
Hooray! 🎉

Issue #356, opened 2023-12-19
secret-capture-generic-ambient-scope: error in replacing the captured variables.


asraa:
This small test case errors out because the index that's added to the secret generic is replaced even outside the secret.generic body.

Error:
```
/usr/local/google/home/asraa/git/heir/tests/secret/capture_hello_world.mlir:38:15: error: 'affine.load' op index must be a valid dimension or symbol identifier
        %20 = affine.load %0[%arg2, %arg1] : memref<16x1xi8>
              ^
/usr/local/google/home/asraa/git/heir/tests/secret/capture_hello_world.mlir:38:15: note: see current operation: %2 = "affine.load"(%0, %arg6, %arg5) <{map = affine_map<(d0, d1) -> (d0, d1)>}> : (memref<16x1xi8>, index, index) -> i8
```

```
module attributes {tf_saved_model.semantics} {
  memref.global "private" constant @__constant_1x16xi8 : memref<1x16xi8> = dense<[[-39, 59, 39, 21, 28, -32, -34, -35, 15, 27, -59, -41, 18, -35, -7, 127]]>
  memref.global "private" constant @__constant_16xi32_0 : memref<16xi32> = dense<[-729, 1954, 610, 0, 241, -471, -35, -867, 571, 581, 4260, 3943, 591, 0, -889, -5103]>
  memref.global "private" constant @__constant_16x16xi8 : memref<16x16xi8> = dense<"0xF41AED091921F424E021EFBCF7F5FA1903DCD20206F9F402FFFAEFF1EFD327E1FB27DDEBDBE4051A17FC241215EF1EE410FE14DA1CF8F3F1EFE2F309E3E9EDE3E415070B041B1AFEEB01DE21E60BEC03230A22241E2703E60324FFC011F8FCF1110CF5E0F30717E5E8EDFADCE823FB07DDFBFD0014261117E7F111EA0226040425211D0ADB1DDC2001FAE3370BF11A16EF1CE703E01602032118092ED9E5140BEA1AFCD81300C4D8ECD9FE0D1920D8D6E21FE9D7CAE2DDC613E7043E000114C7DBE71515F506D61ADC0922FE080213EF191EE209FDF314DDDA20D90FE3F9F7EEE924E629000716E21E0D23D3DDF714FA0822262109080F0BE012F47FDC58E526">
[...]

Issue #355, opened 2023-12-19
secret-capture-generic-ambient-scope hangs on input


asraa:
For hello world, this pass seems to take >5 min, or a long time...

```
module attributes {tf_saved_model.semantics} {
  memref.global "private" constant @__constant_1x16xi8 : memref<1x16xi8> = dense<[[-39, 59, 39, 21, 28, -32, -34, -35, 15, 27, -59, -41, 18, -35, -7, 127]]>
  memref.global "private" constant @__constant_16xi32_0 : memref<16xi32> = dense<[-729, 1954, 610, 0, 241, -471, -35, -867, 571, 581, 4260, 3943, 591, 0, -889, -5103]>
  memref.global "private" constant @__constant_16x16xi8 : memref<16x16xi8> = dense<"0xF41AED091921F424E021EFBCF7F5FA1903DCD20206F9F402FFFAEFF1EFD327E1FB27DDEBDBE4051A17FC241215EF1EE410FE14DA1CF8F3F1EFE2F309E3E9EDE3E415070B041B1AFEEB01DE21E60BEC03230A22241E2703E60324FFC011F8FCF1110CF5E0F30717E5E8EDFADCE823FB07DDFBFD0014261117E7F111EA0226040425211D0ADB1DDC2001FAE3370BF11A16EF1CE703E01602032118092ED9E5140BEA1AFCD81300C4D8ECD9FE0D1920D8D6E21FE9D7CAE2DDC613E7043E000114C7DBE71515F506D61ADC0922FE080213EF191EE209FDF314DDDA20D90FE3F9F7EEE924E629000716E21E0D23D3DDF714FA0822262109080F0BE012F47FDC58E526">
  memref.global "private" constant @__constant_16xi32 : memref<16xi32> = dense<[0, 0, -5438, -5515, -1352, -1500, -4152, -84, 3396, 0, 1981, -5581, 0, -6964, 3407, -7217]>
  memref.global "private" constant @__constant_16x1xi8 : memref<16x1xi8> = dense<[[-9], [-54], [57], [71], [104], [115], [98], [99], [64], [-26], [127], [25], [-82], [68], [95], [86]]>
[...]

PR #357, opened 2023-12-19
fix: fix capturing affine indices in secret generic

asraa:
Fixes https://github.com/google/heir/issues/356
Fixes https://github.com/google/heir/issues/355

https://github.com/llvm/llvm-project/blob/315a5cce89d8f15da5c47d85abbc8155b9c0f0b0/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L298-L304

* `affine.store` and `affine.load` indices must be block arguments of `affine` ops, not of `secret.generic`, so the only way to handle this is to convert them to corresponding memref ops.
* `affine` read and write ops are also able to use inline constants like `affine.load %0[0]` while memref ops cannot. So this also builds the indices needed when constant affine access indices are used

### DIFF
--- a/lib/Dialect/Secret/IR/BUILD
+++ b/lib/Dialect/Secret/IR/BUILD
@@ -42,6 +42,8 @@ cc_library(
         "@heir//include/Dialect/Secret/IR:ops_inc_gen",
         "@heir//include/Dialect/Secret/IR:types_inc_gen",
         "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:AffineAnalysis",
+        "@llvm-project//mlir:AffineDialect",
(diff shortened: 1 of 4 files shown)


j2kun:
LGTM!

I do wonder what, if anything, we lose from doing this, but I imagine any affine-level optimizations would occur before this step.

j2kun (approved):

asraa:
>  I imagine any affine-level optimizations would occur before this step.

Right, this step is actually only a "prep" step for Yosys Optimizer - since secret.generic can reference out of scope, if we wanted to play with affine-level optimizations we would likely not use this pass until we're ready for Yosys Optimization.

PR #364, opened 2024-01-02
Adding the Tfhe-rs-bool dialect 

WoutLegiest:
The previous tfhe-rs boolean operations and attributes are removed from the TfheRust dialect and putted in a separate dialect. All the tests (mlir + e2e) are changed accordingly.

### DIFF
--- a/bazel/import_llvm.bzl
+++ b/bazel/import_llvm.bzl
@@ -7,7 +7,7 @@ load(
 
 def import_llvm(name):
     """Imports LLVM."""
-    LLVM_COMMIT = "c9e10034c4f82724dd8ba6f31c1546bf581c90ba"
+    LLVM_COMMIT = "12f6d556d22e2b3a170d92f648b231d2a1cb99fd"
 
     new_git_repository(
         name = name,


lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
j2kun, line 1: Would you be open to refactoring this so that the shared logic with TfheRustEmitter.cpp are unified in one place? OK to file as a followup ticket if you're not sure how to go about it.

tests/tfhe_rust_bool/end_to_end/src/main.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    #[arg(id = "input_2", index = 2, action)]
+    input2: u8,
+}
+
+fn main() {
+    let flags = Args::parse();
+    let (client_key, server_key) = tfhe::boolean::gen_keys();
+
+    let ct_1 = client_key.encrypt(flags.input1 == 0u8);
```
j2kun, line 20: Seems like True should be equality with 1, not 0, right? I know it's a bug in the test harness, but maybe add another test to catch this since the existing test doesn't cover it (unlucky!).

tests/tfhe_rust_bool/types.mlir
j2kun, line 1: This file can probably be deleted since the same PR that introduces the type has other tests that exercise the type can be parsed correctly. We added this test in other cases where the ops were split into separate PRs from the types, and so this ensured the type-only PR was correct in the mean time.

j2kun (commented):
A few nits, otherwise LGTM!

include/Dialect/TfheRust/IR/TfheRustOps.td
```diff
@@ -17,26 +17,45 @@ class TfheRust_Op<string mnemonic, list<Trait> traits = []> :
...
+    TfheRust_ServerKey:$serverKey,
+    TfheRust_CiphertextType:$lhs,
+    TfheRust_CiphertextType:$rhs
+  );
+  let results = (outs TfheRust_CiphertextType:$output);
+}
 
-def CreateTrivialOp : TfheRust_Op<"create_trivial", [Pure]> {
-  let arguments = (ins TfheRust_ServerKey:$serverKey, AnyInteger:$value);
+def BitXorOp : TfheRust_Op<"bitxor", [
```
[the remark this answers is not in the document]
WoutLegiest: Now we rely on the tfhe api itself, making it somewhat more future proof? I think the maintenance this way is handled by zama

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
WoutLegiest: Definitely, that was on my planning as well. This is a lot of the same code

tests/tfhe_rust_bool/end_to_end/src/main.rs
WoutLegiest: I'll add it. It was done because the parser cannot (easily) process Boolean values

tests/tfhe_rust_bool/types.mlir
WoutLegiest: I'll remove it.

include/Dialect/TfheRust/IR/TfheRustOps.td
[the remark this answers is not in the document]
j2kun: It's true, but the way I see it we don't need to add the ops until they're actually used in the codebase somewhere (either by passes generating them or in tests). It's just the YAGNI principle: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

It also happens that someone comes along, sees those ops, and thinks they are used when they are not, and then gets confused by doing some unnecessary work based on that false assumption.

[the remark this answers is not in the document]
WoutLegiest: Okay, principle makes sense in such a large codebase. I changed the file. I left the bitand, since the emitter has the output code

j2kun (changes requested):
LGTM! As a final step, please manually squash the commits in this PR. Merging the imported CL inside Google will result in the commits preserved; there's sadly no way yet to automatically squash and merge.

PR #366, opened 2024-01-10
Add backlinks from issues to attached TODOs

j2kun:
The action will add/maintain comments on issues, and comment on an issue if it's closed with an outstanding TODO.

See an example comment here: https://github.com/j2kun/pmfp-code/issues/4

### DIFF
--- a/.github/workflows/todo.yaml
+++ b/.github/workflows/todo.yaml
@@ -1,0 +1,15 @@
+name: Update TODO backlinks
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # pin@v3
+    - name: todo-backlinks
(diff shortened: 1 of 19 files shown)


.github/workflows/todo.yaml
```diff
@@ -0,0 +1,15 @@
...
+  push:
+    branches:
+      - main
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: todo-backlinks
+      uses: j2kun/todo-backlinks@v0.0.2
```
asraa, line 12: This is such a cool action, I've always wanted something like this! Would you be OK / like me to share with other team I know who are using GitHub issues heavily in code comments?

minor nit: use the hash instead of version tagging since github versioning is mutable

asraa (approved):

.github/workflows/todo.yaml
j2kun: Of course! I just hacked it together over the holiday for fun, so any improvements or suggestions are welcome. Will fix the hash and then merge

Issue #371, opened 2024-01-12
micro_speech: fails from upstream migration of tosa.rescale


j2kun:
Internal guitar tests failed with

```
error: invalid properties {double_round = true, input_zp = 0 : i32, multiplier = array<i32: 1653229999, 1516545207, 2000799311, 1159928266, 1498403863, 1285645282, 2146175029, 1756589032>, output_zp = -128 : i32, per_channel = true, scale32 = true, shift = array<i32: 41, 43, 41, 41, 41, 41, 41, 41>} for op tosa.rescale: Invalid attribute `shift` in property conversion: array<i32: 41, 43, 41, 41, 41, 41, 41, 41>
    %20 = "tosa.rescale"(%19) <{double_round = true, input_zp = 0 : i32, multiplier = array<i32: 1653229999, 1516545207, 2000799311, 1159928266, 1498403863, 1285645282, 2146175029, 1756589032>, output_zp = -128 : i32, per_channel = true, scale32 = true, shift = array<i32: 41, 43, 41, 41, 41, 41, 41, 41>}> : (tensor<1x25x20x8xi32>) -> tensor<1x25x20x8xi8>
```

Note: upstream MLIR integration doesn't cover guitar tests since they are long-running. So we gotta migrate this one ourselves.

j2kun:
This appears to be the offending commit: https://github.com/llvm/llvm-project/commit/af78e5daf0791135485dbd7972ffedb927727a6b

Need to switch i32 to i8

PR #372, opened 2024-01-12
migrate tosa.rescale shift attr to i8

j2kun:
Fixes https://github.com/google/heir/issues/371

Found all offenders with 

```bash
rg 'shift = array<i32'
```

And to confirm:

```bash
rg -o 'shift = array<...'
```


asraa (approved):
Thank you!

PR #373, opened 2024-01-12
Temporary fix for GITHUB_BASE_REF

j2kun:
The links in the generated comments are broken because apparently the `github.base_ref` is empty when running on main. @asraa I misread that variable from https://docs.github.com/en/actions/learn-github-actions/contexts#github-context, and it looks like there's no variable that says what the default branch is, so I just hard-coded it for now. I will probably update the action to use a proper "input" instead of an environment variable, and call it something clearer than "BASE_REF". But this should fix the links in the short term, and the action will edit the comments in place to fix it.


asraa (approved):
Perhaps it's a github.ref? You can also use conditionals in these vars (github.base_ref || github.ref)

j2kun:
I think it might be `github.ref_name`, I'll try that

PR #377, opened 2024-01-16
support no return values in verilog emitter, handle multi-dim variable access

j2kun:
Incremental feature extracted from https://github.com/google/heir/pull/347


asraa (approved):
Thanks!

PR #378, opened 2024-01-16
add MergeAdjacentGenerics pattern

j2kun:
This pattern merges pairs of adjacent generic ops (with no ops in between them).

Incremental feature extracted from https://github.com/google/heir/pull/347

This could be enhanced later to support merging when extraneous ops (not adding a new dependency) are between two generics.

### DIFF
--- a/lib/Dialect/Secret/IR/SecretPatterns.cpp
+++ b/lib/Dialect/Secret/IR/SecretPatterns.cpp
@@ -14,6 +14,7 @@
 #include "mlir/include/mlir/IR/AffineExpr.h"             // from @llvm-project
 #include "mlir/include/mlir/IR/Block.h"                  // from @llvm-project
 #include "mlir/include/mlir/IR/BuiltinTypes.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/IRMapping.h"              // from @llvm-project
 #include "mlir/include/mlir/IR/PatternMatch.h"           // from @llvm-project
 #include "mlir/include/mlir/IR/Region.h"                 // from @llvm-project
 #include "mlir/include/mlir/IR/Types.h"                  // from @llvm-project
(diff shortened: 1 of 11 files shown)


lib/Dialect/Secret/IR/SecretPatterns.cpp
```diff
@@ -178,6 +181,138 @@ LogicalResult CaptureAmbientScope::matchAndRewrite(
...
+  GenericOp nextGenericOp = dyn_cast<GenericOp>(genericOp->getNextNode());
+  if (!nextGenericOp) {
+    return failure();
+  }
+
+  SmallVector<Type> newResultTypes;
+  newResultTypes.reserve(genericOp->getNumResults() +
+                         nextGenericOp->getNumResults());
+  newResultTypes.append(genericOp->getResultTypes().begin(),
+                        genericOp->getResultTypes().end());
```
asraa, line 195: In the case where the yielded value of the first is only used as an argument to the second, we will still end up yielding it, it seems, right? Even though it may not have any more uses after this (does canonicalize clean up unused yielded values? i feel like that was a thing)

asraa (approved):
Nice!

lib/Dialect/Secret/IR/SecretPatterns.cpp
j2kun: Yes, RemoveUnusedYieldedValues: https://github.com/google/heir/blob/ea8e2f305657e40ce2c14b6d9ed76fadcc0cbc27/include/Dialect/Secret/IR/SecretPatterns.h#L92 which is part of canonicalization: https://github.com/google/heir/blob/ea8e2f305657e40ce2c14b6d9ed76fadcc0cbc27/lib/Dialect/Secret/IR/SecretOps.cpp#L334

j2kun:
Merged manually in https://github.com/google/heir/pull/392

PR #383, opened 2024-01-17
add CaptureAmbientScope to yosys optimizer by default

j2kun:
For some reason, tests/yosys_optimizer/chunk_connections.mlir was timing out, so I added the fast flag


asraa (approved):
Nice! Was the added time from the scope capture? I noticed myself that capturing the ambient scope was slow as well before break, although I can't remember if I resolved that in some way. (maybe I ran cse before it for hello_world? but it shouldn't matter in this case)

j2kun:
No, it was just `abc` hanging and I couldn't tell why

PR #384, opened 2024-01-18
VerilogEmitter: Support index type in function signatures

j2kun:
The emitter scans the uses of the index type to find the largest memref that uses it, then chooses the smallest bit width for the index type to properly index that memref.


asraa (commented):
Nice!

PR #390, opened 2024-01-21
Add more OpenFHE dialect type and ops

Maokami:
Add more ops in OpenFHE dialect (#328)

#### KeySwitch
This operation explicitly needs a corresponding Evalkey for key switching operation.
#### EvalMultNoRelin / EvalMultAndRelin / EvalMult
After taking a closer look at the OpenFHE implementation, it seems that these three functions do the following:
```
EvalMult (c1, c2) : Mult(c1, c2) + Relinearize(s^2 -> s) // size(c1) = size(c2) = size(output) = 2
EvalMultNoRelin (c1, c2) : Mult(c1,c2) // size(c1) and size(c2) can vary.
EvalMultAndRelin (c1, c2) : Mult(c1, c2) + Relinearize(s^i-> s)  where i = 2, 3, ... //  size(c1) and size(c2) can vary, size(output) = 2
```
#### Square / Relinearlize
These operations (and EvalMult, EvalMultAndRelin) are implicitly use EvalKey.

#### ModReduce
a.k.a Rescale in CKKS (question : When is ModReduce used in BGV?)

#### LevelReduce
a.k.a Modulus switch

#### EvalAutomorphism / FindAutomorphismIndex/FindAutomorphismIndices
I'm not sure, but these operation is generalization of rotation.

EvalAutomorphism takes an index `i` and a map `EvalKeyMap` (mapping i to EvalKey), but internally only `EvalKeyMap(i)` (of type: `EvalKey`) is used. Therefore, I designed `openfhe.automorph` to require only a single `EvalKey`.

### DIFF
--- a/include/Dialect/Openfhe/IR/OpenfheOps.td
+++ b/include/Dialect/Openfhe/IR/OpenfheOps.td
@@ -18,58 +18,52 @@ class Openfhe_Op<string mnemonic, list<Trait> traits = []> :
   let cppNamespace = "::mlir::heir::openfhe";
 }
 
+class Openfhe_UnaryOp<string mnemonic, list<Trait> traits = []>
+  :  Openfhe_Op<mnemonic, traits # [
+    Pure,
+    AllTypesMatch<["ciphertext", "output"]>,
+]>{
+ let arguments = (ins
+    Openfhe_CryptoContext:$cryptoContext,
+    RLWECiphertext:$ciphertext
+  );
+  let results = (outs RLWECiphertext:$output);
+}
+
+class Openfhe_BinaryOp<string mnemonic, list<Trait> traits = []>
+  :  Openfhe_Op<mnemonic, traits # [
+    Pure,
+    AllTypesMatch<["lhs", "rhs", "output"]>,
+]>{
+ let arguments = (ins
+    Openfhe_CryptoContext:$cryptoContext,
+    RLWECiphertext:$lhs,
+    RLWECiphertext:$rhs
+  );
+  let results = (outs RLWECiphertext:$output);
+}
+
 def EncryptOp : Openfhe_Op<"encrypt", [Pure]> {
   let arguments = (ins Openfhe_CryptoContext:$cryptoContext, RLWEPlaintext:$plaintext, Openfhe_PublicKey:$publicKey);
   let results = (outs RLWECiphertext:$output);
 }
 
-def NegateOp : Openfhe_Op<"negate", [
+def AddOp : Openfhe_BinaryOp<"add"> { let summary = "OpenFHE add operation of two ciphertexts."; }
+def SubOp : Openfhe_BinaryOp<"sub"> { let summary = "OpenFHE sub operation of two ciphertexts."; }
+
+def MulOp : Openfhe_BinaryOp<"mul"> { let summary = "OpenFHE mul operation of two ciphertexts with relinearization."; }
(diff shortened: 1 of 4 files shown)


j2kun:
> When is ModReduce used in BGV?

So I actually don't know enough about the arithmetic schemes to say if this is useful for BGV or not. Since this API handles all of BGV/BFV/CKKS, do you think it would it be reasonable to have the ops support all three (adding any extra ops later as needed)? There's nothing currently in this file that is specific to BGV.

include/Dialect/Openfhe/IR/OpenfheOps.td
```diff
@@ -90,15 +91,120 @@ def MulConstOp : Openfhe_Op<"mulconst",[
...
+]> {
+  let arguments = (ins
+    Openfhe_CryptoContext:$cryptoContext,
+    RLWECiphertext:$lhs,
+    RLWECiphertext:$rhs
+  );
+  let results = (outs RLWECiphertext:$output);
+}
+
+def MulNoRelinOp : Openfhe_Op<"mul_no_relin",[
```
j2kun, line 106: Please extract this pattern as a base `class`, i.e. using the same organizational style as the CGGI Dialect: https://github.com/google/heir/blob/c21eb96dd5f33722682a20f88d6f844401fd2421/include/Dialect/CGGI/IR/CGGIOps.td#L25

It might look something like;

```
class Openfhe_BinaryOp<string mnemonic, list<Trait> traits = []>
  :  Openfhe_Op<mnemonic, traits # [
    Pure,
    AllTypesMatch<["lhs", "rhs", "output"]>,
]>{
 let arguments = (ins
    Openfhe_CryptoContext:$cryptoContext,
    RLWECiphertext:$lhs,
    RLWECiphertext:$rhs
  );
  let results = (outs RLWECiphertext:$output);
}
```

And each binary op can be defined then as

```
def MulNoRelinOp : Openfhe_BinaryOp<"mul_no_relin", [OpSpecificTrait1, ...]> { let summary = "..."; }
```

You could do a similar unary op base class as well. Up to you

j2kun (changes requested):
Added one suggested comment. Otherwise LGTM!

AlexanderViand-Intel:
> > When is ModReduce used in BGV?
> 
> So I actually don't know enough about the arithmetic schemes to say if this is useful for BGV or not. Since this API handles all of BGV/BFV/CKKS, do you think it would it be reasonable to have the ops support all three (adding any extra ops later as needed)? There's nothing currently in this file that is specific to BGV.

In BGV, modulo switching is used as a noise management technique and/or to reduce the size of the ciphertext (e.g., prior to sending a result back to the client). If possible, it'd be great to have support for the entire evaluator API, including the CKKS stuff!

j2kun:
Well there's ModReduce and LevelReduce. I'm guessing LevelReduce is a discretized ModReduce, while ModReduce is a general ModReduce that applies to CKKS only (in OpenFHE's way of doing things)?

Maokami:
Hmm, I'm not sure but it seems to me that the `ModReduce` API is used for modulus switching operation in BGV and for rescaling operation for CKKS?

Maokami:
>  If possible, it'd be great to have support for the entire evaluator API, including the CKKS stuff!

I agree!

include/Dialect/Openfhe/IR/OpenfheOps.td
Maokami: Applied!

j2kun (approved):

Issue #391, opened 2024-01-22
bgv: add lowering to openfhe


Maokami:
## BGV to Openfhe

The BGV dialect will be pretty close to Openfhe, but since it will lower to other library dialects, we would ideally want to know the least common denominator of BGV library APIs. I don't know exactly what that is, but I am also comfortable with that boundary evolving as we add more libraries.

_Originally posted by @j2kun in https://github.com/google/heir/issues/328#issuecomment-1898975618_

AlexanderViand-Intel:
The major challenge with different flavors of BGV will probably be the noise management, which is quite different between different implementations. HElib has some kind of "dynamic noise estimate" going on, and I think the BGV implementation in SEAL uses the same approach. OpenFHE, on the other hand, supports (potentially in addition to the dynamic approach) an easier-to-use version of BGV where all you need to know is depth + max number of additions/etc between multiplications. 

However, I guess this isn't actually a major concern for the BGV-> library lowerings, as we can just assume that if someone lowers something using them, they already ensured that the BGV instructions are "valid" w.r.t noise as implemented in the library.

Issue #99, opened 2023-08-16
bgv: Add a attribute to ciphertext to track plaintext bits


asraa:
In the last HEIR meeting, we noted that rather than holding onto the underlying encrypted type, that it would be more useful to keep track of the number of plaintext bits that are encoded (potentially in each slot) of a BGV ciphertext. This would make the error analysis better. Likewise, at the BGV layer, it really is no longer relevant whether something was SIMD or not - this can occur at a higher level pass.

It may also be useful to have some notions of which slots are being used? Although not sure. That can be a separate issue.

j2kun:
I think we could perhaps also migrate the `bgv::Ciphertext` type to the `lwe` dialect as an RLWE ciphertext type, which does not currently exist but we want to have it there.

j2kun:
Agreed with Asra. I intended the current attribute to be generic enough to
model both BFV and BGV (isn't it? For LSB, the messages start bits would be
a small index), and if it's not we should make it so.

j2kun:
Taking a second look, yes I see how the docs are confusing. I think for BGV the interpretation of

```mlir
#lwe_encoding = #lwe.bit_field_encoding<cleartext_start=4, cleartext_bitwidth=4>
```

Would be this: where `n` stands for noise and `b` for the message bits

```
       nn...n|bbbb
    MSB^         ^LSB
```

j2kun:
This is fixed now since we moved BGV to use LWE types and attributes.

PR #395, opened 2024-01-25
Draft : Add BGVToOpenfhe conversion

Maokami:
Add pass from BGV to OpenFHE dialect and corresponding simple syntax test. (#391)

Current Status:
- [X] bgv.add
- [X] bgv.sub
- [X] bgv.mul
- [X] bgv.negate
- [x] bgv.rotate
- [x] bgv.relinearize
- [x] bgv.modulus_switch

### DIFF
--- a/lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
+++ b/lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
@@ -1,0 +1,273 @@
+#include "include/Conversion/BGVToOpenfhe/BGVToOpenfhe.h"
+
+#include "include/Dialect/BGV/IR/BGVDialect.h"
+#include "include/Dialect/BGV/IR/BGVOps.h"
+#include "include/Dialect/BGV/IR/BGVTypes.h"
+#include "include/Dialect/LWE/IR/LWEAttributes.h"
+#include "include/Dialect/LWE/IR/LWEDialect.h"
+#include "include/Dialect/LWE/IR/LWEOps.h"
+#include "include/Dialect/LWE/IR/LWETypes.h"
+#include "include/Dialect/Openfhe/IR/OpenfheDialect.h"
+#include "include/Dialect/Openfhe/IR/OpenfheOps.h"
+#include "include/Dialect/Openfhe/IR/OpenfheTypes.h"
+#include "lib/Conversion/Utils.h"
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"   // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
+namespace mlir::heir::bgv {
+
+#define GEN_PASS_DEF_BGVTOOPENFHE
+#include "include/Conversion/BGVToOpenfhe/BGVToOpenfhe.h.inc"
+
+class ToLWECiphertextTypeConverter : public TypeConverter {
+ public:
+  // Convert ciphertext to RLWE ciphertext
+  ToLWECiphertextTypeConverter(MLIRContext *ctx) {
+    addConversion([](Type type) { return type; });
+    addConversion([ctx](CiphertextType type) -> Type {
+      assert(type.getLevel().has_value());
+      auto level = type.getLevel().value();
+      assert(level < type.getRings().getRings().size());
+
+      auto ring = type.getRings().getRings()[level];
+      auto cmod = ring.getCmod();
+      auto polyDegree = ring.getIdeal().getDegree();
+      auto dim = type.getDim();
+      return lwe::RLWECiphertextType::get(
+          // TODO(#99): Set a default encoding when the adding # of plaintext
+          // bits on BGV Ciphertext is done.
+          ctx, lwe::PolynomialEvaluationEncodingAttr::get(ctx, 0, 0),
+          lwe::RLWEParamsAttr::get(ctx, cmod, dim, polyDegree));
+    });
+  }
+};
+
+bool containsBGVOps(func::FuncOp func) {
+  auto walkResult = func.walk([&](Operation *op) {
+    if (llvm::isa<bgv::BGVDialect>(op->getDialect()))
+      return WalkResult::interrupt();
+    return WalkResult::advance();
+  });
+  return walkResult.wasInterrupted();
+}
+
+FailureOr<Value> getContextualCryptoContext(Operation *op) {
+  Value cryptoContext = op->getParentOfType<func::FuncOp>()
+                            .getBody()
+                            .getBlocks()
+                            .front()
+                            .getArguments()
+                            .front();
+  if (!cryptoContext.getType().isa<openfhe::CryptoContextType>()) {
+    return op->emitOpError()
+           << "Found BGV op in a function without a public "
+              "key argument. Did the AddCryptoContextArg pattern fail to run?";
+  }
+  return cryptoContext;
+}
+
+struct AddCryptoContextArg : public OpConversionPattern<func::FuncOp> {
+  AddCryptoContextArg(mlir::MLIRContext *context)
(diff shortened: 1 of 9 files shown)


lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
```diff
@@ -0,0 +1,178 @@
...
+      assert(type.getLevel().has_value());
+      auto level = type.getLevel().value();
+      assert(level < type.getRings().getRings().size());
+
+      auto ring = type.getRings().getRings()[level];
+      auto cmod = ring.getCmod();
+      auto polyDegree = ring.getIdeal().getDegree();
+      auto dim = type.getDim();
+      return lwe::RLWECiphertextType::get(
+          // FIXME: What should be the default encoding?
```
j2kun, line 37: I'm guessing this TODO needs to be resolved first to enable this: https://github.com/google/heir/blob/e83489c16703eb969757c59cd48785b388097bea/include/Dialect/BGV/IR/BGVTypes.td#L37

https://github.com/google/heir/issues/99

Given the parameter corresponding to the set of rings, I'm not so sure anymore about the comment

> I think we could perhaps also migrate the bgv::Ciphertext type to the lwe dialect as an RLWE ciphertext type

But if the BGVCiphertext type stored the number of bits of the underlying message, then I believe that would be sufficient information to choose a default encoding via `#lwe.polynomial_evaluation_encoding<cleartext_start=K, cleartext_bitwidth=B>` for some `K` and `B` indexing the coefficient space Z/qZ

Two notes:

- I know that BGV cane use either RLWE_PolynomialCoefficientEncoding or RLWE_PolynomialEvaluationEncoding, but IIUC RLWE_PolynomialEvaluationEncoding is the most commonly used one, so that should be the default.
- There is definitely a lot more discussion to be had about how to pick an encoding, and I think the simplest thing to do for now is to pick any sensible default we can, file an issue for further discussion, and (optional, perhaps in followup PR) allow the user to manually override the default choice as a pass option.

j2kun:
Just took a quick peek at the draft and it's looking really great so far! Happy to answer any questions you have along the way.

Maokami:
I believe everything is mostly finished except for the above-mentioned todo. I would appreciate it if you could review this PR!

## Changes
- Now, `bgv.MulOp` is converted to `openfhe.MulNoRelinOp`.
- Since the `Relinearize()` function in OpenFHE relinearizes a ciphertext to the lowest level (for (1,s)), the `to_basis` of `bgv.RelinOp` must be [0,1].
- Since the `ModReduce()` function in OpenFHE decreases the level of a ciphertext by 1, `fromLevel == toLevel + 1` holds for `bgv.ModulusSwitch`.

## Question
- Do we need to add more tests? 
- Can we leave the above TODO as is until the #99 resolved , or does it need to be modified?

tests/bgv/to_openfhe.mlir
```diff
@@ -0,0 +1,46 @@
+// RUN: heir-opt --bgv-to-openfhe %s > %t
+// RUN: FileCheck %s < %t
+
+// This simply tests for syntax.
```
j2kun, line 4:
```suggestion
```

Remove this comment, as it's not true of this file.

tests/bgv/to_openfhe.mlir
```diff
@@ -0,0 +1,46 @@
...
+module {
+  // CHECK: func.func @test_fn([[X:%.+]]: [[T:.*33538049.*]]) -> [[T]] {
+  func.func @test_fn(%x : !ct) -> !ct {
+    // CHECK: return [[X]] : [[T]]
+    return %x : !ct
+  }
+
+  // CHECK: func.func @test_ops([[C:%.+]]: [[S:.*crypto_context.*]], [[X:%.+]]: [[T:.*33538049.*]], [[Y:%.+]]: [[T]]) {
+  func.func @test_ops(%x : !ct, %y : !ct) {
+    %negate = bgv.negate(%x) : !ct
```
j2kun, line 25: Please add `// CHECK` directives for what ops these should lower to.

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    Value cryptoContext = result.value();
+    rewriter.replaceOp(op, rewriter.create<openfhe::RelinOp>(
+                               op.getLoc(), cryptoContext, adaptor.getX()));
+    return success();
+  }
+};
+
+bool checkModulusSwitchLevels(unsigned long fromLevel, unsigned long toLevel) {
+  return fromLevel == toLevel + 1;
```
j2kun, line 201: Ditto, a comment saying how this is required due to the semantics of the OpenFHE API would be helpful.

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
```diff
@@ -0,0 +1,264 @@
...
+    auto toLevel = adaptor.getToLevel();
+    if (!checkModulusSwitchLevels(fromLevel, toLevel)) {
+      op.emitError() << "fromLevel must be toLevel + 1, got "
+                     << "fromLevel: " << fromLevel << " and "
+                     << "toLevel: " << toLevel;
+      return failure();
+    }
+
+    Value cryptoContext = result.value();
+    rewriter.replaceOp(op, rewriter.create<openfhe::RelinOp>(
```
j2kun, line 225:
```suggestion
    rewriter.replaceOp(op, rewriter.create<openfhe::ModReduceOp>(
```

(or perhaps it's LevelReduce? I'm not sure the difference)

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+  LogicalResult matchAndRewrite(
+      Relinearize op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    FailureOr<Value> result = getContextualCryptoContext(op.getOperation());
+    if (failed(result)) return result;
+
+    auto toBasis = adaptor.getToBasis();
+    if (!checkRelinToBasis(toBasis)) {
+      op.emitError() << "toBasis must be [0, 1], got "
```
j2kun, line 188: Please add a code comment explaining why it must be [0,1]. You explained in the PR, but that is harder to find when someone hits this error.

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
```diff
@@ -0,0 +1,264 @@
...
+      assert(type.getLevel().has_value());
+      auto level = type.getLevel().value();
+      assert(level < type.getRings().getRings().size());
+
+      auto ring = type.getRings().getRings()[level];
+      auto cmod = ring.getCmod();
+      auto polyDegree = ring.getIdeal().getDegree();
+      auto dim = type.getDim();
+      return lwe::RLWECiphertextType::get(
+          // FIXME: What should be the default encoding?
```
j2kun, line 38: Let's just change this to a TODO(#99) to set a default encoding and we can submit the rest of this PR as-is.

j2kun (changes requested):
LGTM! Just a few small changes and I think it's ready to go

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
Maokami: Applied!

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
Maokami: Applied!

tests/bgv/to_openfhe.mlir
Maokami: Applied!

j2kun (approved):

Issue #385, opened 2024-01-18
Add basic OpenFHE codegen and end-to-end test harness


j2kun:

j2kun:
Remaining TODO above

PR #422, opened 2024-02-02
Add OpenFHE as a bazel-native build dependency

j2kun:
Tested with `bazel build @openfhe//...:all`

Part of https://github.com/google/heir/issues/385

This PR only adds the WORKSPACE dependency, so we don't yet have to worry about syncing the internal/external versions of OpenFHE via copybara.

New deps:

- OpenFHE: https://github.com/openfheorg/openfhe-development
- Cereal: https://github.com/USCiLab/cereal
- RapidJSON: https://github.com/Tencent/rapidjson/

# Hacks and workaround notes

## Repo patches

I had to remove the following line from
`bazel/openfhe/add_config_core.patch` to make the build work.  We will
probably want to fix this if to get 128-bit int support, or infer it
from the system somehow instead of relying on cmake to generate it.
(Why can't this be passed as `-DHAVE_INT128`?)

```
+#define HAVE_INT128 TRUE
```


tests/openfhe/end_to_end/README.md
```diff
@@ -0,0 +1,40 @@
+# End to end Rust codegen tests
```
asraa, line 1: nit: i think this was supposed to change to OpenFHE codegen and links to openfhe instead?

tests/openfhe/end_to_end/BUILD
```diff
@@ -0,0 +1,22 @@
...
+)
+
+glob_lit_tests(
+    name = "all_tests",
+    data = [
+        "src/main.cc",
+        "@heir//tests:test_utilities",
+    ],
+    default_tags = [
+        "manual",
```
asraa, line 17: ah i think `src/main.cc` isn't a thing yet - since you haven't added the full e2e test execution, can you remove the `manual` and data here so that when the e2e test functionality is added we can review that part there? (Since these don't depend on cargo, i assume we don't need to run them with manual)

asraa (commented):
LGTM, thank you! Adding new bazel deps is always painstaking

Was the `bazel/openfhe/BUILD` file necessary to mark a subpackage, or is it possible to remove the empty file? If not, can you mark that it's just a subpackage for openfhe deps' builds?

tests/openfhe/end_to_end/README.md
j2kun: oh, whoops, the end-to-end part was meant to be in a subsequent PR, not this one. So everything in this directory is just random partial work. Let me clean it up.

j2kun:
> Was the `bazel/openfhe/BUILD` file necessary to mark a subpackage, or is it possible to remove the empty file? If not, can you mark that it's just a subpackage for openfhe deps' builds?

Yes, and done.

PR #429, opened 2024-02-06
ci: check cache to determine runner size

asraa:
Uses a smaller runner (4 core) when there's a cache hit, and 16 core when there's no cache hit. maybe i should use 8 cores..

I tested the runner switching on my own fork, but (1) i don't have access to two different runners so all i tested was the use of a variable runner and (2) i couldn't finish the llvm build on the default runners with no cache, so i'm not entirely sure that the 4 core is enough for a cached build

Also adds fastbuild to the rust e2e tests so it doesn't discard the cache for that

TODO: Also enable caching when tests fail (https://github.com/actions/cache/tree/main/save#always-save-cache)

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -2,3 +2,4 @@
 self-hosted-runner:
   labels:
     - ubuntu-20.04-16core
+    - ubuntu-20.04-4core
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -11,9 +11,42 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 jobs:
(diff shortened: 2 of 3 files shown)


j2kun (approved):
Amazing, you're a wealth of knowledge.

asraa:
I found out why we're not reusing the caches even with the same keys! It's likely due to https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache - we can't access caches between our PRs (so you'll see on https://github.com/google/heir/actions/caches that the caches each have their own refs despite having the same key).

Next time I feel like going down this rabbit hole I'd like to figure out why the pushes to main caches aren't being shared with the PRs - they should have access to builds on `main`, and I'm not sure why our runs on merges to main aren't used.

Issue #396, opened 2024-01-26
Integrate Polygeist as a C++ frontend


j2kun:
After [additional reading](https://discourse.llvm.org/t/rfc-polygeist-llvm-incubator-proposal/60890), Polygeist seems like it is poised to be the primary bridge between ClangAST and MLIR, while ClangIR will continue to lower directly to LLVM. This implies we should probably focus on using it as our C++ front end.

https://github.com/heir-compiler/HEIR has an implementation that we should be able to adapt to our use cases.

We will have to decide, however, how to integrate this into HEIR since the two projects are probably compiled against different LLVM versions. This may result in supporting Polygeist in CI only, but not integrating it directly into the build (similar to the rust e2e tests).

j2kun:
I think we're realistically going to wait for ClangIR to lower to scf, and polygeist to be upstreamed (https://discourse.llvm.org/t/rfc-add-scf-to-affine-conversion-pass-in-mlir/88036)

PR #467, opened 2024-02-27
WIP: polygeist integration

j2kun:
Cf. https://github.com/google/heir/issues/396

This WIP PR is intended to explore running Polygeist on a variety of C++ input programs, and attempting to feed those as input to HEIR passes, so we can see what aspects are missing to start lowering to FHE dialects.


j2kun:
My rough plan here is to check in polygeist outputs for the transpiler repo
examples, and map out what changes are needed to lower them to cggi, and
then file those as issues with the tests checked in but disabled with a
TODO.

PR #471, opened 2024-02-29
Initial port of HECO auto-SIMD passes for BGV

j2kun:
Partial progress toward https://github.com/google/heir/issues/475

### DIFF
--- a/include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
+++ b/include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
@@ -1,0 +1,88 @@
+#ifndef INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTCANONICALIZATION_TD_
+#define INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTCANONICALIZATION_TD_
+
+include "TensorExtOps.td"
+include "mlir/Dialect/Arith/IR/ArithOps.td"
+include "mlir/Dialect/Tensor/IR/TensorOps.td"
+include "mlir/IR/PatternBase.td"
+
+// TODO(#515): refactor these helpers to a common file with InsertRotate.td
+defvar DefOverflow = ConstantEnumCase<Arith_IntegerOverflowAttr, "none">;
+
+def MakeSingleResultVariadic: NativeCodeCall<"{ $0 }">;
+
+def CreateIndexCastOp : NativeCodeCall<
+      "$_builder.create<arith::IndexCastOp>($0.getLoc(), $1.getType(), $0)">;
+
+def IsZero :
+    Constraint<
+      CPred<"llvm::cast<mlir::IntegerAttr>($0).getValue().isZero()">>;
+
+def OutOfBoundsOfTensorDim :
+  Constraint<
+    CPred<
+      "llvm::cast<mlir::IntegerAttr>($0).getValue().getSExtValue() < 0 "
+      "|| llvm::cast<mlir::IntegerAttr>($0).getValue().getSExtValue() > "
+      "llvm::cast<mlir::ShapedType>($1.getType()).getShape()[0]"
+    >
+  >;
+
+// rotate %t, 0 -> %t
+def DropZeroRotation : Pat<
+  (TensorExt_RotateOp $tensor, (ConstantLikeMatcher APIntAttr:$c0)),
+  (replaceWithValue $tensor),
+  [(IsZero $c0)]
+>;
+
+// rotate %t, x -> rotate %t, x mod size
+def NormalizeRotationIndex : Pat<
+  (TensorExt_RotateOp $tensor, (Arith_ConstantOp:$shiftOp APIntAttr:$shiftAmount)),
+  (TensorExt_RotateOp $tensor,
+      (Arith_RemUIOp
+        $shiftOp,
+        // Only works for 1D tensors: index is taken modulo the tensor length,
+        // i.e., dim 0
+        (CreateIndexCastOp
+          (Tensor_DimOp $tensor, (Arith_ConstantOp ConstantAttr<IndexAttr, "0">)),
+          $shiftOp))
+      ),
+  [(OutOfBoundsOfTensorDim $shiftAmount, $tensor)]
+>;
+
+// %0 = rotate %t, x
+// %1 = rotate %0, y
+// ---> rotate %t (x+y)
+def CombineSequentialRotates : Pat<
+  (TensorExt_RotateOp
+    (TensorExt_RotateOp $tensor, (Arith_ConstantOp:$xOp APIntAttr:$x)),
+    (Arith_ConstantOp:$yOp APIntAttr:$y)),
+  (TensorExt_RotateOp $tensor, (Arith_AddIOp $xOp, $yOp, DefOverflow)),
+  []
+>;
+
+// A rotation followed by extraction can be extracted directly from the
+// original tensor.
+def RotatePlusExtractToIndexedExtract : Pat<
+  (Tensor_ExtractOp
+    (TensorExt_RotateOp $tensor, $shift),
+    (variadic $index)),
+  (Tensor_ExtractOp
+    $tensor,
+    (MakeSingleResultVariadic (Arith_AddIOp $shift, $index, DefOverflow)))
+>;
+
+// Rotating two tensors by the same amount can be converted to a single
+// post-rotation. This can result in eliminating either the rotation (because
+// it can be combined with a later rotation) or the arith op itself, if it is
+// is identical to an existing arith op applied before the rotation.
+foreach ArithOp = [Arith_AddIOp, Arith_SubIOp, Arith_MulIOp] in {
+  def FactorParallelRotationsThroughOp_#ArithOp : Pat<
+    (ArithOp
+      (TensorExt_RotateOp $t1, $i),
+      (TensorExt_RotateOp $t2, $i),
+      $ovf),
+    (TensorExt_RotateOp (ArithOp $t1, $t2, $ovf), $i)
+  >;
+}
+
+#endif  // INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTCANONICALIZATION_TD_
--- a/lib/Dialect/TensorExt/Transforms/CollapseInsertionChains.cpp
+++ b/lib/Dialect/TensorExt/Transforms/CollapseInsertionChains.cpp
@@ -1,0 +1,175 @@
+#include "include/Dialect/TensorExt/Transforms/CollapseInsertionChains.h"
+
+#include "include/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "llvm/include/llvm/Support/Debug.h"             // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Matchers.h"               // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"           // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+
+#define DEBUG_TYPE "collapse-insertion-chains"
+
+namespace mlir {
+namespace heir {
+namespace tensor_ext {
+
+#define GEN_PASS_DEF_COLLAPSEINSERTIONCHAINS
+#include "include/Dialect/TensorExt/Transforms/Passes.h.inc"
+
+template <typename Op>
+FailureOr<int64_t> get1DExtractionIndex(Op op) {
+  auto insertIndices = op.getIndices();
+  if (insertIndices.size() != 1) return failure();
+
+  // Each index must be constant; this may require running --canonicalize or
+  // -sccp before this pass to apply folding rules (use -sccp if you need to
+  // fold constants through control flow).
+  Value insertIndex = *insertIndices.begin();
+  auto insertIndexConstOp = insertIndex.getDefiningOp<arith::ConstantIndexOp>();
+  if (!insertIndexConstOp) return failure();
+
+  auto insertOffsetAttr =
+      llvm::dyn_cast<IntegerAttr>(insertIndexConstOp.getValue());
+  if (!insertOffsetAttr) return failure();
+
+  return insertOffsetAttr.getInt();
+}
+
+/// A pattern that searches for sequences of extract + insert, where the
+/// indices extracted and inserted have the same offset, and replaced them with
+/// a single rotate operation.
+struct ConvertAlignedExtractInsertToRotate
+    : public OpRewritePattern<tensor::InsertOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  // Given an insert and extract op, compute the shift between their two
+  // access indices. Only works for 1D tensors.
+  FailureOr<int64_t> calculateShift(tensor::InsertOp insertOp,
+                                    tensor::ExtractOp extractOp) const {
+    auto insertIndexRes = get1DExtractionIndex(insertOp);
+    auto extractionIndexRes = get1DExtractionIndex(extractOp);
+    if (failed(insertIndexRes) || failed(extractionIndexRes)) return failure();
+
+    auto insertTensorType =
+        insertOp.getDest().getType().cast<RankedTensorType>();
+    auto extractTensorType =
+        extractOp.getTensor().getType().cast<RankedTensorType>();
+    if (insertTensorType.getShape() != extractTensorType.getShape())
+      return failure();
+
+    auto shift = (extractionIndexRes.value() - insertIndexRes.value());
+    if (shift < 0) shift += insertTensorType.getShape()[0];
+
+    LLVM_DEBUG({
+      llvm::dbgs() << "Insertion index: " << insertIndexRes.value() << "\n";
+      llvm::dbgs() << "Extraction index: " << extractionIndexRes.value()
+                   << "\n";
+      llvm::dbgs() << "Shift: " << shift << "\n";
+    });
+    return shift;
+  }
+
+  LogicalResult matchAndRewrite(tensor::InsertOp insertOp,
+                                PatternRewriter &rewriter) const override {
+    auto extractOp = insertOp.getScalar().getDefiningOp<tensor::ExtractOp>();
+    if (!extractOp) return failure();
+
+    auto shiftRes = calculateShift(insertOp, extractOp);
+    if (failed(shiftRes)) return failure();
+
+    int64_t shift = shiftRes.value();
+    DenseSet<int64_t> accessedIndices;
+    accessedIndices.insert(get1DExtractionIndex(insertOp).value());
+
+    // Check if there are corresponding insertions into all other indices,
+    // which are extracted from the same source tensor with the same shift.
+    //
+    // The problem is that because tensors have value semantics and not pointer
+    // semantics, we will see each new insertion use a different SSA value, like
+    // this:
+    //
+    // %extracted_1 = tensor.extract %original_source[%c1] : tensor<4096xi16>
+    // %inserted_1 = tensor.insert %extracted_1 into %original_dest[%c5] :
+    // tensor<4096xi16> %extracted_2 = tensor.extract %original_source[%c2] :
+    // tensor<4096xi16> %inserted_2 = tensor.insert %extracted_2 into
+    // %inserted_1[%c6] : tensor<4096xi16>
+    //
+    // Note hwo inserted_1 replaces original_dest for the subsequent insert,
+    // and inserted_2 will replace inserted_1 for the next one.
+    //
+    // So we need to traverse the insertions in order to follow the chain. Note
+    // that a more sophisticated pass might be able to support less
+    // well-structured DAGs of insertions and extractions, but we will improve
+    // this when that becomes necessary.
+
+    // Also note that the greedy pattern rewriter will start from the first op
+    // in the block order, so we can assume that if the pattern matches, it
+    // matches on the first insert op encountered.
+    auto extractionSource = extractOp.getTensor();
+    auto current = insertOp;
+    while (true) {
+      bool foundNext = false;
+
+      for (auto *user : current.getResult().getUsers()) {
+        if (auto nextInsert = dyn_cast<tensor::InsertOp>(user)) {
+          auto nextExtract =
+              nextInsert.getScalar().getDefiningOp<tensor::ExtractOp>();
+          if (!nextExtract) continue;
+
+          // We're inserting into this tensor from a different tensor than
+          // earlier insertions in the chain, so we can't continue.
+          if (nextExtract.getTensor() != extractionSource) return failure();
+
+          auto nextShiftRes = calculateShift(nextInsert, nextExtract);
+          if (failed(nextShiftRes)) return failure();
+
+          if (nextShiftRes.value() != shift) return failure();
+
+          accessedIndices.insert(get1DExtractionIndex(nextInsert).value());
+          current = nextInsert;
+          foundNext = true;
+        }
+      }
+
+      // This can either be because we reached the end of the chain, or else
+      // because the chain is incomplete.
+      if (!foundNext) break;
+    }
+
+    // We didn't cover the entire tensor, so the downstream user of this tensor
+    // may be depending on the original data in the untouched indices being in
+    // tact.
+    if (accessedIndices.size() !=
+        extractionSource.getType().cast<RankedTensorType>().getShape()[0])
+      return failure();
+
+    // The last insertion must be replaced because its user is the final end
+    // user
+    rewriter.replaceOpWithNewOp<tensor_ext::RotateOp>(
+        current, extractionSource,
+        rewriter.create<arith::ConstantIntOp>(current.getLoc(), shift,
+                                              /*width=*/32));
+
+    // The rest of the chain of insertions and extractions itself will be
+    // DCE'd by canonicalization if possible.
+    return success();
+  }
+};
+
+struct CollapseInsertionChains
+    : impl::CollapseInsertionChainsBase<CollapseInsertionChains> {
+  using CollapseInsertionChainsBase::CollapseInsertionChainsBase;
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    RewritePatternSet patterns(context);
+
+    patterns.add<ConvertAlignedExtractInsertToRotate>(context);
+    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
+  }
+};
+
+}  // namespace tensor_ext
+}  // namespace heir
+}  // namespace mlir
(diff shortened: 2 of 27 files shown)


j2kun:
Some notes:

With this PR as of b1d3eec75f902606e0229843de4017fb1b29def2, I can use the following pipeline to fully unroll and vectorize the box_blur program from HECO: 

```bash
bazel run //tools:heir-opt --  --secretize=entry-function=box_blur --wrap-generic --canonicalize --cse --full-loop-unroll --straight-line-vectorize $PWD/tests/simd/box_blur.mlir
```

However, after the full loop unroll, the program has some 500k lines, and if you try to `canonicalize` or `cse`, MLIR will hang (I let it run for ~3h and it didn't finish).

I also tried the following variant that, although it's specially crafted for `box_blur`, uses the existing MLIR passes (nor my custom `full-loop-unroll`) to do the same thing (ruling out that my loop unroll is buggy)

```bash
bazel run //tools:heir-opt --  --secretize=entry-function=box_blur --wrap-generic --canonicalize --cse --affine-loop-unroll=unroll-factor=2 --affine-loop-unroll=unroll-factor=2 --canonicalize --cse --affine-loop-unroll=unroll-factor=64 --canonicalize --cse --affine-loop-unroll=unroll-factor=64 --canonicalize --cse $PWD/tests/simd/box_blur.mlir
```

If you remove the last three flags, and run

```bash
bazel run //tools:heir-opt --  --secretize=entry-function=box_blur --wrap-generic --canonicalize --cse --affine-loop-unroll=unroll-factor=2 --affine-loop-unroll=unroll-factor=2 --canonicalize --cse --affine-loop-unroll=unroll-factor=64 --canonicalize --cse $PWD/tests/simd/box_blur.mlir
```

Then you get a more reasonable program: an affine loop of size 64 with 900-ish statements in the body: https://gist.github.com/j2kun/3d3bf0d41b1f503f42ecd54753b2bd31

The checked-in HECO MLIR files seem to be working with a smaller dimension than 64x64 (cf. https://github.com/MarbleHE/HECO/blob/main/test/filecheck/boxblur/boxblur_2_unrolled.mlir which has only ~1.1k ops compared to the ~100k that a full unroll gives when I run it). Perhaps @AlexanderViand-Intel can comment more on this discrepancy.

Either way, I think it is safe to say that if we want to test a BGV-to-OpenFHE path, it would make sense to stop short of a full unroll. But then we need some heuristic to determine when to stop unrolling.

After the above, applying `--straight-line-vectorize` gives this IR: https://gist.github.com/j2kun/a9d52c0bf0c5228dcb14f4571ed1f68c

Which, excluding all the tensor ops, reduces to:

```mlir
module {
  func.func @box_blur(%arg0: !secret.secret<tensor<4096xi16>>) -> !secret.secret<tensor<4096xi16>> {
    %0 = secret.generic ins(%arg0 : !secret.secret<tensor<4096xi16>>) {
    ^bb0(%arg1: tensor<4096xi16>):
      %1 = affine.for %arg2 = 0 to 64 iter_args(%arg3 = %arg1) -> (tensor<4096xi16>) {
        %2 = arith.addi %arg2, %c-1 : index
        %3 = arith.muli %from_elements, %cst_3 : tensor<2xindex>
        %4 = arith.addi %arg2, %c1 : index
        %5 = arith.muli %4, %c64 : index
        %6 = arith.addi %from_elements_5, %cst_2 : tensor<126xindex>
        %7 = arith.remui %from_elements_132, %cst_1 : tensor<128xindex>
        %8 = arith.addi %from_elements_292, %cst_0 : tensor<63xindex>
        %9 = arith.remui %from_elements_356, %cst : tensor<64xindex>
        %10 = arith.addi %extracted, %c63 : index
        %11 = arith.remui %10, %c4096 : index
        %12 = arith.addi %extracted_4, %c63 : index
        %13 = arith.remui %12, %c4096 : index
        %14 = arith.addi %5, %c63 : index
        %15 = arith.remui %14, %c4096 : index
        %16 = arith.addi %extracted, %c64 : index
        %17 = arith.remui %16, %c4096 : index
        %18 = arith.addi %extracted_4, %c64 : index
        %19 = arith.remui %18, %c4096 : index
        %20 = arith.addi %5, %c64 : index
        %21 = arith.remui %20, %c4096 : index
        %22 = arith.addi %from_elements_588, %from_elements_589 : tensor<64xi16>
        %23 = arith.addi %from_elements_660, %from_elements_661 : tensor<64xi16>
        %24 = arith.addi %from_elements_726, %from_elements_654 : tensor<64xi16>
        %25 = arith.addi %from_elements_791, %from_elements_655 : tensor<64xi16>
        %26 = arith.addi %from_elements_856, %from_elements_656 : tensor<64xi16>
        %27 = arith.addi %from_elements_921, %from_elements_657 : tensor<64xi16>
        %28 = arith.addi %from_elements_986, %from_elements_658 : tensor<64xi16>
        %29 = arith.addi %from_elements_1051, %from_elements_659 : tensor<64xi16>
        affine.yield %inserted_1178 : tensor<4096xi16>
      }
      secret.yield %1 : tensor<4096xi16>
    } -> !secret.secret<tensor<4096xi16>>
    return %0 : !secret.secret<tensor<4096xi16>>
  }
}
```

As you can see, most of the ops naturally batch into size 64, but at the beginning some don't, and we'd need to accommodate for that in the `--straight-line-vectorize` pass by giving the pass an allowed vector width. (Which itself raises the question of how to set the batch size at this stage, long before cryptographic parameters are decided)

AlexanderViand-Intel:
Mh, the unrolling issues are odd! HECO's evaluation contains this very similar 64*64 [roberts cross example](https://github.com/MarbleHE/HECO/blob/main/evaluation/compile_time/heco_input/robertscross_64x64.mlir) which compiles in a few seconds (definitively not hours). I'll try and look into that before our meeting this week, but I don't see any obvious differences in what the HEIR version and the HECO version are doing that would explain this.

Btw, it's probably worth noting that HECO's heuristics conceptually don't require the unrolling, that's just a shortcut to simplify the implementation. The relative offset/target slot logic should all work as affine loop analyses, too.

j2kun:
Update: 

I think we can do this entirely with DRR-generated patterns, but I found a bug in the way they handle variadic inputs, which is required to work with tensor extract ops. Jacques said he thinks he has a quick fix and will address it soon. https://discourse.llvm.org/t/compilation-failure-with-drr-generated-pattern/77385

In the mean time, the rotation alignment patterns seem to work, and while this PR needs a lot of cleanup, I think we can get something in before Thursday.

j2kun:
> Btw, it's probably worth noting that HECO's heuristics conceptually don't require the unrolling, that's just a shortcut to simplify the implementation. The relative offset/target slot logic should all work as affine loop analyses, too.

I would love to try to build a more general affine loop analysis instead of unrolling, I'm just not quite sure how to do that.

AlexanderViand-Intel:
>However, after the full loop unroll, the program has some 500k lines, and if you try to canonicalize or cse, MLIR will hang (I let it run for ~3h and it didn't finish). 

Do you still experience the time-out issue with unrolling the 64*64 boxblur? I just reinstalled HECO and ran `boxblur_1_ssa.mlir` through the full pipeline (`heco --full-pass boxblur_1_ssa.mlir`) and it took less than a second if LLVM is built in Release mode (admittedly on a server with many cores, but this should take no more than 30s even on a laptop) and 27s if LLVM is built in Debug mode (again, on a powerful server, but that should be at most a few minutes, not hours).  
I tried to have a look at what's going on in HEIR here, but the PR currently fails to build for me (and also in CI, so I guess it's not my machine).

>The checked-in HECO MLIR files seem to be working with a smaller dimension than 64x64 (cf. https://github.com/MarbleHE/HECO/blob/main/test/filecheck/boxblur/boxblur_2_unrolled.mlir which has only ~1.1k ops compared to the ~100k that a full unroll gives when I run it). Perhaps @AlexanderViand-Intel can comment more on this discrepancy. 

The boxblur example is indeed a 64x64 program: doing just (`--unroll-loops`) gives you a file with just over 300'000 lines, but `--canonicalize --cse` then simplifies back down to the ~1k lines in `boxblur_2_unrolled.mlir` (this really highlights how  wasteful a "hack" the loop unrolling is).

> > Btw, it's probably worth noting that HECO's heuristics conceptually don't require the unrolling, that's just a shortcut to simplify the implementation. The relative offset/target slot logic should all work as affine loop analyses, too.
> 
> I would love to try to build a more general affine loop analysis instead of unrolling, I'm just not quite sure how to do that.

It's definitively not something I can do off the top of my head, either, but it should be pretty easy to translate the ideas from working on concrete indices to considering everything relative to the iteration variable. Unfortunately, unless the existing unroll approach & heuristic start becoming a performance bottleneck, it's hard to justify re-inventing the wheel here. Especially given that we now also have Viaduct-HE and it's more general approach to finding batching opportunities.

j2kun:
Update: The DRR patterns now work, and I've narrowed down the slow canonicalization on the huge IR to our secret.generic canonicalizers. It looks like `HoistPlaintextOps` is slow: we're doing a linear scan to find an op that can be hoisted, then only hoisting one. So we should hoist them all in a loop instead.

AlexanderViand-Intel:
Following up from the working group meeting earlier today, here's all the notes on "ordering" of the passes I found in the HECO paper and the code:

Collecting potentially "rotate-and-reduce"-able  BinOps into variadic ops must happen before the main batching pass (this one's pretty obvious, not sure why this one made it into the paper but none of the other ones)
image

Then, there's a few comments in the pipeline, again [there's one](https://github.com/MarbleHE/HECO/blob/artifact/src/tools/heco/heco.cpp#L98) related to rotate-and-reduce that mentions that we can only detect that the operands in the variadic ops as being extracted from the same tensor after we CSE to remove redundant versions of those tensors.

Regarding the main batching pass, there's [this comment](https://github.com/MarbleHE/HECO/blob/artifact/src/tools/heco/heco.cpp#L87) in the pipeline builder that says we must canonicalize just beforehand because it only handles constant indices (but that might be an implementation detail that doesn't apply to your DDR version).

More interestingly, there's [this comment](https://github.com/MarbleHE/HECO/blob/artifact/src/Passes/hir2hir/Batching.cpp#L324) in the batching pass code itself that says "We must translate in order of appearance for this to work, so we walk manually". If this is correct, then the pattern rewriter approach might not work for this step (unless you can force the rewriter to do it top-down?). However, I'm not convinced that this comment is actually true...

j2kun:
At this point I have the box_blur (4x4 and 64x64) and hamming distance examples from HECO producing equivalent (but not identical) programs in HEIR.

For example, box-blur 64x64

```mlir
bazel-bin/tools/heir-opt '--secretize=entry-function=box_blur' --wrap-generic --canonicalize --cse --full-loop-unroll \
 --insert-rotate \
 --cse --canonicalize \
 --collapse-insertion-chains --canonicalize \
 $PWD/tests/simd/box_blur_64x64.mlir

module {
  func.func @box_blur(%arg0: !secret.secret<tensor<4096xi16>>) -> !secret.secret<tensor<4096xi16>> {
    %c65_i32 = arith.constant 65 : i32
    %c127 = arith.constant 127 : index
    %c4032 = arith.constant 4032 : index
    %c3968 = arith.constant 3968 : index
    %c63 = arith.constant 63 : index
    %0 = secret.generic ins(%arg0 : !secret.secret<tensor<4096xi16>>) {
    ^bb0(%arg1: tensor<4096xi16>):
      %1 = tensor_ext.rotate %arg1, %c3968 : tensor<4096xi16>, index
      %2 = tensor_ext.rotate %arg1, %c4032 : tensor<4096xi16>, index
      %3 = arith.addi %1, %2 : tensor<4096xi16>
      %4 = arith.addi %3, %arg1 : tensor<4096xi16>
      %5 = tensor_ext.rotate %4, %c63 : tensor<4096xi16>, index
      %6 = arith.addi %5, %2 : tensor<4096xi16>
      %7 = arith.addi %6, %arg1 : tensor<4096xi16>
      %8 = tensor_ext.rotate %7, %c63 : tensor<4096xi16>, index
      %9 = tensor_ext.rotate %arg1, %c127 : tensor<4096xi16>, index
      %10 = arith.addi %8, %9 : tensor<4096xi16>
      %11 = arith.addi %10, %arg1 : tensor<4096xi16>
      %12 = tensor_ext.rotate %11, %c3968 : tensor<4096xi16>, index
      %13 = arith.addi %12, %2 : tensor<4096xi16>
      %14 = arith.addi %13, %arg1 : tensor<4096xi16>
      %15 = tensor_ext.rotate %14, %c65_i32 : tensor<4096xi16>, i32
      secret.yield %15 : tensor<4096xi16>
    } -> !secret.secret<tensor<4096xi16>>
    return %0 : !secret.secret<tensor<4096xi16>>
  }
}
```

Notice the above differs from HECO in that not all rotation operations are done before the addition operations. I don't know if this is good or bad for the ultimate program runtime, but my basis of "equivalent" is merely in counting ops.

The pass breaks down into three components:

### InsertRotate

This is the bulk of the work, applying the patterns in `[TensorExtRotationAlignment.td](https://github.com/google/heir/pull/471/files#diff-9ce57e69f039834441a73d42cffc28ee59aad7e1551c21a2a69530a78b0b1036)` to insert rotations and align slots according to a subset of the same rules as HECO.

What is NOT included is aligning slots to non-arithmetic-op downstream users (like extract, see CollapseInsertionChains).

### CSE/canonicalize

InsertRotate alone does not put the IR into a form that is suitable for CSE/canonicalize to simplify, so I added patterns in `[TensorExtCanonicalization.td](https://github.com/google/heir/pull/471/files#diff-bdef9d4fd2c8e03b87ca0d94b1693029f7582ef9f4f45f3ef8712f543fe9b63a)` to account for this:

- `FactorParallelRotationsThroughOp`: push parallel rotations through arithmetic ops, which removes one rotation and better aligns duplicate arith ops and duplicate rotation ops for cse
- `DropZeroRotation`: eliminate a rotation by shift of zero
- `NormalizeRotationIndex`: ensure all rotation shifts are positive and within tensor dim bounds
- `SumOfConstantRotations`: combine sequences of rotations by constants

### CollapseInsertionChains

The above is enough to simplify the entire IR, and what is left may be a final "rotation" as implemented by a sequence of extract and insert ops into the final output.

I could have added a pattern to `InsertRotate` that accounted for this by aligning rotations to downstream extract ops, but I felt I could handle this case more generally by matching any well-structured chain of insertions and replacing it with a rotate.

### Performance

As far as performance, the 64x64 example finishes in ~1 minute. Here's the timing breakdown. I think there are likely some performance improvements I could make to the insertion chain pass, since it should "just" be doing a linear scan of the IR. But since it's fast enough for now, I will leave that as a followup task. It may become irrelevant if we replace this with an affine analysis that avoids unrolling the loop.

```
===-------------------------------------------------------------------------===
                         ... Execution time report ...
===-------------------------------------------------------------------------===
  Total Execution Time: 56.1786 seconds

  ----Wall Time----  ----Name----
    0.0042 (  0.0%)  Parser
    0.0004 (  0.0%)  Secretize
    0.0013 (  0.0%)  WrapGeneric
    0.0009 (  0.0%)  Canonicalizer
    0.0002 (  0.0%)  CSE
    0.0000 (  0.0%)    (A) DominanceInfo
    3.2923 (  5.9%)  FullLoopUnroll
   25.2879 ( 45.0%)  InsertRotate
    0.6241 (  1.1%)  CSE
    0.0000 (  0.0%)    (A) DominanceInfo
    0.2596 (  0.5%)  Canonicalizer
   26.7031 ( 47.5%)  CollapseInsertionChains
    0.0008 (  0.0%)  Canonicalizer
    0.0019 (  0.0%)  Output
    0.0018 (  0.0%)  Rest
   56.1786 (100.0%)  Total
```

AlexanderViand-Intel (changes requested):
First of all thanks for porting the HECO batching optimization! 🤩 Expressing this directly over `tensor` and expressing the rewrites in DDR instead of C++makes this re-implementation a lot cleaner than the original (as evidenced by the fact that my review comments are mostly nitpicking).

I did notice one minor and one major departure from the HECO solutions, though, which mean the optimization falls a bit short in terms of final program performance: the first one (lack of `combine` op, see comment below) is only really relevant for programs that don't perfectly batch, but the second one (issues with "target slot" selection, I  think) also affects things like `boxblur` and mean that the generated code would be significantly slower.

> Notice the above differs from HECO in that not all rotation operations are done before the addition operations. I don't know if this is good or bad for the ultimate program runtime, but my basis of "equivalent" is merely in counting ops.

What do you mean by equivalent when counting ops here? For example, this is what HECO produces for boxblur 
(`heco --fhe-pass evaluation/comparison/heco_input/boxblur_64x64.mlir`):

```llvm
func.func private @encryptedBoxBlur_64x64(%arg0: !fhe.batched_secret<4096 x i16>) -> !fhe.batched_secret<4096 x i16> {
    %0 = fhe.rotate(%arg0) by 64 : <4096 x i16>
    %1 = fhe.rotate(%arg0) by 65 : <4096 x i16>
    %2 = fhe.rotate(%arg0) by 1 : <4096 x i16>
    %3 = fhe.add(%arg0, %0, %1, %2) : (!fhe.batched_secret<4096 x i16>, !fhe.batched_secret<4096 x i16>, !fhe.batched_secret<4096 x i16>, !fhe.batched_secret<4096 x i16>) -> !fhe.batched_secret<4096 x i16>
    return %3 : !fhe.batched_secret<4096 x i16>
}
```
Note that there are only three rotations, as the kernel is `2x2` so we need to bring together four pixels (one of which we don't need to rotate).  In comparison, the HEIR version needs seven rotations which is obviously still a massive improvement over the baseline, but since rotations make up for virtually all of the runtime of this program, it's still around 2x slower.

> What is NOT included is aligning slots to non-arithmetic-op downstream users (like extract, see CollapseInsertionChains).

I think this might be the cause of these differences: without the lookahead ("target slot selection" in HECO parlance), you might not be going to get the full "magic". Alternatively, it might be because the additions aren't grouped into a single op? I'll try and stare at the IR produced in the intermediate steps a bit more since it would be nice to close the gap, but I don't think it's totally necessary, either.

> As far as [compile time] performance, the 64x64 example finishes in ~1 minute.  [...] It may become irrelevant if we replace this with an affine analysis that avoids unrolling the loop.

It's hard to do an apples-to-apples comparison of the compile time between HECO and HEIR, because HEIR obviously has the whole `secret.generic` handling in there, too, but both take <1 min on my (server) machine which seems good enough for now, especially given that a non-unrolled version should be significantly faster anyway.

include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
```diff
@@ -0,0 +1,99 @@
...
+        // i.e., dim 0
+        (CreateIndexCastOp
+          (Tensor_DimOp $tensor, (Arith_ConstantOp ConstantAttr<IndexAttr, "0">)),
+          $shiftOp))
+      ),
+  [(OutOfBoundsOfTensorDim $shiftAmount, $tensor)]
+>;
+
+// rotate %t, x + rotate %t y -> rotate %t (x+y)
+def SumOfConstantRotations : Pat<
```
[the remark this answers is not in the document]
j2kun: Done in 16926c40

lib/Dialect/TensorExt/Transforms/CollapseInsertionChains.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "include/Dialect/TensorExt/Transforms/Passes.h.inc"
+
+template <typename Op>
+FailureOr<int64_t> get1DExtractionIndex(Op op) {
+  auto insertIndices = op.getIndices();
+  if (insertIndices.size() != 1) return failure();
+
+  // Each index must be constant; this may require running --canonicalize or
+  // -sccp before this pass to apply folding rules (use -sccp if you need to
+  // fold constants through control flow).
```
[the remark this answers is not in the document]
j2kun: Done in e3d3c453

include/Dialect/TensorExt/Transforms/TensorExtRotationAlignment.td
```diff
@@ -0,0 +1,83 @@
+#ifndef INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTPATTERNS_TD_
+#define INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTPATTERNS_TD_
+
+include "include/Dialect/TensorExt/IR/TensorExtOps.td"
+include "mlir/Dialect/Arith/IR/ArithOps.td"
+include "mlir/Dialect/Tensor/IR/TensorOps.td"
+include "mlir/IR/PatternBase.td"
+
+// The patterns in this file are intended to align with the automatic-SIMD
+// batching heuristics from the HECO project. See section 4.4 of
```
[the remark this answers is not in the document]
j2kun: Fixed in 8409a273

j2kun:
> What do you mean by equivalent when counting ops here?

I was basing that claim on this [file](https://github.com/MarbleHE/HECO/blob/3e13744233ab0c09030a41ef98b4e061b6fa2eac/test/filecheck/boxblur/boxblur_5_batched.mlir) which contains eight rotations (but the last one is only necessary to extract the constant term) as do all subsequent snapshotted files in that directory. As I understand it, HECO is also applying the rotate-and-sum pattern to reduce that to log_2(8) = 3 rotations? So perhaps that was just omitted from those checked in files. I have a draft PR that I haven't yet pushed that implements a rotate-and-sum pattern in isolation, but now I see that this is why having rotations interspersed with addition ops is a problem: it will be harder to identify rotate-and-sum opportunities. However, if that's OK with you, I'd still like to check this PR in so I can have a good baseline to improve against.

> as you'd have to walk the def-use chain all the way until you find a tensor.insert to figure out the target slot.

I think the slot target selection is actually fine, but what is not included is target slot selection _when the downstream op is an insert_. I could add this (it would just be a slight variation on the patterns already in this PR), but I'd want to find a counterexample program that shows it's necessary first. I will file an issue and put a TODO in the code.

AlexanderViand-Intel:
Thanks for the updates - once the build error is fixed, this should be good to merge from my perspective 👍 

>I was basing that claim on this [file](https://github.com/MarbleHE/HECO/blob/3e13744233ab0c09030a41ef98b4e061b6fa2eac/test/filecheck/boxblur/boxblur_5_batched.mlir) which contains eight rotations (but the last one is only necessary to extract the constant term) as do all subsequent snapshotted files in that directory. As I understand it, HECO is also applying the rotate-and-sum pattern to reduce that to log_2(8) = 3 rotations? 

I don't think rotate-and-reduce gives you anything here, this is just an out-of-date file 🙈  I focused on `evaluation/` examples for the USENIX artifact evaluation so there's clearly some inconsistent/outdated stuff in `test/` (probably shouldn't have switched off CI after losing the university credits, even if it's a bit pricy 💸)

>I think the slot target selection is actually fine, but what is not included is target slot selection when the downstream op is an insert. 

Yes, in HECO the "target" slot is specifically the index at which a result is later inserted (or `-1` if nothing is found). Though, iirc, there's a tradeoff: if you look ahead too far, you might remove rotate-and-reduce opportunities. I expect some (maybe all?) of this to be side effects of the unroll-and-cse approach, though so it might not be worth investing too much effort to fix.

j2kun:
TODOs and new issues filed (along with minor cleanup and build fix)

- https://github.com/google/heir/issues/511: rotate-and-sum pattern
- https://github.com/google/heir/issues/512: tensor_ext.combine op and pattern
- https://github.com/google/heir/issues/513: slot target selection for insert ops
- https://github.com/google/heir/issues/514: determine if we need more sophisticated target slot selection patterns for arith ops
- https://github.com/google/heir/issues/515: clean up DRR helpers

lib/Dialect/TensorExt/Transforms/InsertRotate.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace mlir {
+namespace heir {
+namespace tensor_ext {
+
+#define GEN_PASS_DEF_INSERTROTATE
+#include "include/Dialect/TensorExt/Transforms/Passes.h.inc"
+
+namespace alignment {
+// In an inner namespace to avoid conflicts with canonicaliation patterns
```
asraa, line 18: nit: typo (curious why pre-submit didn't catch this!)

```suggestion
// In an inner namespace to avoid conflicts with canonicalization patterns
```

include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
```diff
@@ -0,0 +1,101 @@
...
+        "), 0)">;
+
+def CreateIndexCastOp : NativeCodeCall<
+      "$_builder.create<arith::IndexCastOp>($0.getLoc(), $1.getType(), $0)">;
+
+def IsZero :
+    Constraint<
+      CPred<"llvm::cast<mlir::IntegerAttr>($0).getValue().isZero()">>;
+
+def AreOpposites :
```
asraa, line 29: This and other helpers (IndexType, MakeMatchingTypedZero) seem unused - maybe they were needed for follow-ups?

(Mostly I was curious where`AreOpposites` is used - since the constraint seems more like `AreEqual`)

include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
```diff
@@ -0,0 +1,101 @@
...
+  [(IsZero $c0)]
+>;
+
+// rotate %t, x -> rotate %t, x mod size
+def NormalizeRotationIndex : Pat<
+  (TensorExt_RotateOp $tensor, (Arith_ConstantOp:$shiftOp APIntAttr:$shiftAmount)),
+  (TensorExt_RotateOp $tensor,
+      (Arith_RemUIOp
+        $shiftOp,
+        // Only works for 1D tensors: index is taken modulo the tensor length,
```
asraa, line 56: was this because rotation indices are variadic? (I expect a lot of the TF programs may have multi-dim tensors, for some reason inputs there are usually <1xDIM>
Can tablegen maybe do a for-loop over the dimensions, or would you just need to write a pattern for each dimension size to generalize it?

asraa (approved):

AlexanderViand-Intel (approved):

include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
j2kun: Mainly it's because I don't understand what a rotation would mean for a multi-dimensional tensor in this context. At the very least, you'd have to specify the axes of rotation (then this could select the appropriate dimension to rotate), but that would have to correspond to a meaningful operation on the underlying ciphertext. Otherwise, we might as well use the more general upstream tensor operations like scatter/gather, or even the permutation capabilities of `vector.transfer_read/write`. We only need `tensor_ext.rotate` because of how it matches the FHE computation model, after all, and that is closely tied to how the ciphertexts are packed as coefficient arrays.

j2kun: This probably implies I should have a verifier for tensor_ext.rotate to ensure it only takes 1D inputs. Will file an issue.

include/Dialect/TensorExt/IR/TensorExtCanonicalization.td
j2kun: Fixed. 

I replaced a pattern that combined rotate x & rotate -x with the input tensor, which was generalized to the addition pattern followed by dropping zero rotation. Just didn't remove the helpers.

lib/Dialect/TensorExt/Transforms/InsertRotate.cpp
j2kun: Fixed

PR #477, opened 2024-03-04
Add a full loop unroll pass

j2kun:


include/Transforms/FullLoopUnroll/FullLoopUnroll.td
```diff
@@ -0,0 +1,16 @@
...
+include "mlir/Pass/PassBase.td"
+
+def FullLoopUnroll : Pass<"full-loop-unroll"> {
+  let summary = "Fully unroll all loops";
+  let description = [{
+  Scan the IR for affine.for loops and unroll them all.
+  }];
+  let dependentDialects = [
+
+  ];
```
asraa, line 13: nit: did it not complain that the affine dialect wasn't registered for the pass? maybe it's because nothing was anchored on it..? i suppose remove this empty option then

asraa (approved):

include/Transforms/FullLoopUnroll/FullLoopUnroll.td
j2kun: I think you only need this when the pass generates new ops from the dependent dialect.

j2kun: Fixed!

PR #478, opened 2024-03-05
Add TensorExt dialect with rotate op

j2kun:
This is intended to be a generic rotate op that operates on any tensor type, mirroring the `fhe.rotate` from HECO, but operating in conjunction with `secret.generic`.

Later, a lowering from secret to BGV would have to lower an op like

```
secret.generic ins(%0: secret.secret<tensor<64xi16>>) {
^bb0(%arg1: tensor<16xi16>):
  %1 = tensor_ext.rotate %arg1, %c7
  secret.yield %1
}
```

where the rotation amount is a plaintext value defined in the ambient scope.

### DIFF
--- a/include/Dialect/TensorExt/IR/TensorExtOps.td
+++ b/include/Dialect/TensorExt/IR/TensorExtOps.td
@@ -1,0 +1,40 @@
+#ifndef INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTOPS_TD_
+#define INCLUDE_DIALECT_TENSOREXT_IR_TENSOREXTOPS_TD_
+
+include "include/Dialect/TensorExt/IR/TensorExtDialect.td"
+include "mlir/IR/BuiltinAttributes.td"
+include "mlir/IR/CommonTypeConstraints.td"
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/InferTypeOpInterface.td"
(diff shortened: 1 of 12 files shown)


include/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -0,0 +1,40 @@
...
+
+class TensorExt_Op<string mnemonic, list<Trait> traits = []> :
+        Op<TensorExt_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::tensor_ext";
+}
+
+def TensorExt_RotateOp : TensorExt_Op<"rotate", [Pure, AllTypesMatch<["tensor", "output"]>]> {
+  let summary = "Rotate a tensor some number of indices left.";
+  let description = [{
+    This op represents a left-rotation of a tensor by given number of indices.
```
asraa, line 20: This may be with wrap-around right? I was going to comment that I wonder if a canonicalizer or validator should enforce / re-write wrap around indices with ones within the range, but I think that may not be possible for variable rotations.

asraa (approved):

include/Dialect/TensorExt/IR/TensorExtOps.td
j2kun: It is with wrap-around, correct. I don't think we need to worry about this though, unless we get to a lowering that actually implements rotate in terms of a loop.

Issue #479, opened 2024-03-05
Speed up HoistPlaintextOps


j2kun:
The pattern currently does a linear scan of all the ops inside a generic, then only hoists one. For a large loop unroll inside a generic body, this slows `--canonicalize` to a crawl.

PR #485, opened 2024-03-05
Speed up HoistPlaintextOps

j2kun:
Fixes https://github.com/google/heir/issues/479

The two main problems:

1. `HoistPlaintextOps` used `std::find_if` to scan the entire generic body, only to extract a single op, leading to a quadratic scan when there were a linear number of ops to hoist. When unrolling an affine.for, each iteration gets an affine.apply op, and they can all be lifted, so it nearly always causes quadratic behavior.
2. In `extractOpBeforeGeneric`, the newly-created single-op generic was cloning all of the operands from the original generic op as its operands, even if it had zero secret operands. When lifting ~1k ops, this became a problem because the output of each single-op generic is added as an operand to the original `generic` it was lifted from, so we got into a situation like:
  - start: original generic has 1 arg
  - lift 1 op with 1 arg; original generic gets +1 arg
  - lift 1 op with 2 args; original generic gets +1 arg
  - lift 1 op with 3 args; original generic gets +1 arg
  - ... (x1000)
After that, we have a quadratic number of unused args, so we apply a quadratic number of EraseUnusedGenericArg patterns. By more finely tracking the subset of operands needed in the new generic, we can reduce this to (# ops to lift * # operands per lifted op), which is minimal. The only tricky bit is that, if the op being extracted is a loop nest, you have to walk the op's subtree to find all the generic arguments it depends on.

The test I added ran in 15s with the bug still in place, and after these fixes it runs in 0.5s.

### DIFF
--- a/lib/Dialect/Secret/IR/SecretPatterns.cpp
+++ b/lib/Dialect/Secret/IR/SecretPatterns.cpp
@@ -567,16 +567,30 @@ LogicalResult HoistPlaintextOps::matchAndRewrite(
     return true;
   };
 
-  auto it = std::find_if(opRange.begin(), opRange.end(),
-                         [&](Operation &op) { return canHoist(op); });
-  if (it == opRange.end()) {
-    return failure();
+  LLVM_DEBUG(
+      llvm::dbgs() << "Scanning generic body looking for ops to hoist...\n");
+
+  // We can't hoist them as they are detected because the process of hoisting
+  // alters the context generic op.
+  llvm::SmallVector<Operation *> opsToHoist;
+  bool hoistedAny = false;
+  for (Operation &op : opRange) {
+    if (canHoist(op)) {
+      opsToHoist.push_back(&op);
+      hoistedAny = true;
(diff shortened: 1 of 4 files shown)


lib/Dialect/Secret/IR/SecretPatterns.cpp
```diff
@@ -567,16 +567,30 @@ LogicalResult HoistPlaintextOps::matchAndRewrite(
...
+  // We can't hoist them as they are detected because the process of hoisting
+  // alters the context generic op.
+  llvm::SmallVector<Operation *> opsToHoist;
+  bool hoistedAny = false;
+  for (Operation &op : opRange) {
+    if (canHoist(op)) {
+      opsToHoist.push_back(&op);
+      hoistedAny = true;
+    }
   }
```
asraa, line 582: nit: you could consider `llvm::make_filter_range` here

asraa (approved):
Nice! 15 seconds to .5 seconds is great

PR #504, opened 2024-03-07
Demo elementwise lowering of tensor<poly> to loops

j2kun:
Adds the `MemRefElementTypeInterface` to the polynomial type, and demonstrates the lowering to affine loops

```
bazel run //tools:heir-opt -- --convert-elementwise-to-linalg --one-shot-bufferize --convert-linalg-to-affine-loops $PWD/tests/polynomial/elementwise.mlir
```

The strange part is that lowering linalg to loops requires de-tensorizing the inputs to linalg.generic, which inserts `bufferization.to_memref` and `bufferization.to_tensor` as materializations. To completely eliminate those requires converting everything to memrefs, see https://github.com/google/heir/blob/main/tools/heir-opt.cpp#L85-L99

I tried running `--polynomial-to-standard` after this, but ran into another issue 

```
/home/j2kun/fhe/heir/tests/polynomial/elementwise.mlir:8:8: 
error: failed to legalize unresolved materialization from 
'!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>' 
to 'tensor<1024xi25>' that remained live after conversion
  %0 = polynomial.add(%arg0, %arg1) : tensor<2x!poly>
       ^
/home/j2kun/fhe/heir/tests/polynomial/elementwise.mlir:8:8: 
note: see current operation: %7 = "builtin.unrealized_conversion_cast"(%6) :
 (!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>) -> tensor<1024xi25>
/home/j2kun/fhe/heir/tests/polynomial/elementwise.mlir:8:8: 
note: see existing live user here: %8 = arith.extsi %6 : tensor<1024xi25> to tensor<1024xi26>
```

PR #508, opened 2024-03-09
Add support for tensors in -polynomial-to-standard 

### LINKED ISSUE #143: PolyToStandard: handling tensors of poly?

We have a few options here:

AlexanderViand-Intel:
Fixes #143 (see discussion there) by adding support for dealing with `tensor<..x!polynomial.polynomial<..>>` types in the `-polynomial-to-standard` pass.

This PR assumes the `ElementwiseMappable` `polynomial` operations have somehow been "unrolled" (c.f. #504) and focuses on lowering the resulting IR, which no longer contains `polynomial` operations over tensors, but contains various other operations with `tensor<..x!polynomial.polynomial<..>>` types. This is necessary because MLIR does not alllow tensor-of-tensor (which the pass would produce) and (afaik) does not offer any built-in systems to handle lowering of `ElementwiseMappable` operations in a more generic fashion.

* Declares `tensor` dialect dynamically legal 
(using `typeConverter.isLegal`, ops are illegal if their operands require type conversion)
* Adds a `ConvertAny` conversion pattern that matches any operation and simply replaces the operation by a copy of itself with the types converted.
* Adds `ConvertExtract` and `ConvertInsert` patterns that match `tensor.extract` and `tensor.insert` operations and rewrites them to `tensor.extract_slice` and `tensor.insert_slice` operations, respectively.
* Extends the `TypeConverter` with a conversion that converts tensors of polynomial types to multi-dimensional tensors.
* In order to support affine loops, this also declares the `affine` dialect dynamically legal (same rule as for `tensor`).

For example, the pass converts this example as follows:
```llvm
#map = affine_map<(d0) -> (d0)>
func.func @test_bin_ops(%arg0: tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>, %arg1: tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>) -> tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>> {
  %0 = affine.for %i = 0 to 2 iter_args(%a = %arg0) -> tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>> {
    %in = tensor.extract %arg0[%i] :  tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>
    %in_0 = tensor.extract %arg1[%i] :  tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>
    %1 = polynomial.add(%in, %in_0) : !polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>
    %t = tensor.insert %1 into %a[%i] :  tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>
    affine.yield %t :  tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>
  }
  return %0 : tensor<2x!polynomial.polynomial<<cmod=33538049, ideal=#polynomial.polynomial<1 + x**1024>>>>
}
```
```llvm 
func.func @test_bin_ops(%arg0: tensor<2x1024xi25>, %arg1: tensor<2x1024xi25>) -> tensor<2x1024xi25> {
  %0 = affine.for %arg2 = 0 to 2 iter_args(%arg3 = %arg0) -> (tensor<2x1024xi25>) {
    %extracted_slice = tensor.extract_slice %arg0[%arg2, 0] [1, 1024] [1, 1] : tensor<2x1024xi25> to tensor<1024xi25>
    %extracted_slice_0 = tensor.extract_slice %arg1[%arg2, 0] [1, 1024] [1, 1] : tensor<2x1024xi25> to tensor<1024xi25>
    %cst = arith.constant dense<33538049> : tensor<1024xi26>
    %1 = arith.extsi %extracted_slice : tensor<1024xi25> to tensor<1024xi26>
    %2 = arith.extsi %extracted_slice_0 : tensor<1024xi25> to tensor<1024xi26>
    %3 = arith.addi %1, %2 : tensor<1024xi26>
    %4 = arith.remsi %3, %cst : tensor<1024xi26>
    %5 = arith.trunci %4 : tensor<1024xi26> to tensor<1024xi25>
    %inserted_slice = tensor.insert_slice %5 into %arg3[%arg2, 0] [1, 1024] [1, 1] : tensor<1024xi25> into tensor<2x1024xi25>
    affine.yield %inserted_slice : tensor<2x1024xi25>
  }
  return %0 : tensor<2x1024xi25>
}
``` 

This is still WIP, there's a few things left to clean-up/improve:
* This is mostly a generic set of utilities for lowering `ElementwiseMappable` Ops so it should probably move to `Utils.cpp` where `addStructuralConversionPatterns` already is. 
* The `addConversion` to go from tensor-of-poly to multi-dim-poly currently uses the `convertPolynomialType` function, but it could probably be made generic by calling the typeconverter itself to convert the tensor's element in a  black-box fashion.
* The `ConvertAny` pattern needs to be marked as `setHasBoundedRewriteRecursion(true)` because otherwise it will not apply to operations it generates (e.g., the "new" operations created when cloning a region, such as when called on `affine.for`). This seems to be fine (operations are "legalized" in the eyes of the type converter after being visited once) but and the more specific patterns seem to have higher priority, but there could be a case where an operation is marked explicitly illegal but no other pattern matches it, in which case this pattern would probably be applied infinitely. 
* Still need to add tests that actually exercise this logic

Most importantly, I'd still need to extend this concept to `memref`, since that's what actually gets generated by the pipeline in #504, but there's a bit of annoying issue in that we'd somehow have to go from a `memref<2x1024xi25>` to `tensor<1024xi25>` because that's what the `arith` operations that we insert in place of the `polynomial` operations expect.

### DIFF
--- a/lib/Conversion/Utils.cpp
+++ b/lib/Conversion/Utils.cpp
@@ -1,8 +1,12 @@
 #include "lib/Conversion/Utils.h"
 
+#include <cstdint>
+
+#include "mlir/include/mlir/Dialect/Affine/IR/AffineOps.h"  // from @llvm-project
 #include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
 #include "mlir/include/mlir/Dialect/Func/Transforms/FuncConversions.h"  // from @llvm-project
 #include "mlir/include/mlir/Dialect/SCF/Transforms/Patterns.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm - project
 #include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
 
 namespace mlir {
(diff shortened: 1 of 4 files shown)


j2kun:
> in which case this pattern would probably be applied infinitely.

Luckily, MLIR will automatically cut it off after a hard-coded bound, to avoid truly infinite loops.

> annoying issue in that we'd somehow have to go from a memref<2x1024xi25> to tensor<1024xi25>

The main reason for this is that any lowering of linalg requires bufferization. I think it's not all that uncommon to raise "artificial" memrefs back to tensor in a complicated pipeline, but I agree it's not ideal. Something to bring up to the MLIR community I suppose.

j2kun:
Also note, if there is a good point to checkpoint and submit because some new test case is working or some subset of feature is implemented, feel free to add TODOs linked to an issue and we can merge the code in more incrementally

AlexanderViand-Intel:
> Also note, if there is a good point to checkpoint and submit because some new test case is working or some subset of feature is implemented, feel free to add TODOs linked to an issue and we can merge the code in more incrementally

I cleaned this up (moved everything to Utils.cpp and removed the `memref` handling part until I understand what the expected output would even be) and was about to submit it for review, but while writing the test I realized we need to also handle converting  `tensor.from_elements` to `tensor.concat`, so I'll add support for that and (assuming I don't see anything else that needs to be added) mark it ready for review then.

lib/Conversion/Utils.cpp
```diff
@@ -12,6 +16,238 @@ using ::mlir::func::CallOp;
...
+              combinedShape, castConvertedType.getElementType());
+          return combinedType;
+        }
+      }
+    }
+    return type;
+  });
+
+  // Should this be in here? Maybe better to have it be part of
+  // StructuralConversionPatterns?
```
j2kun, line 242: I'd say yes, since it concerns just control flow

lib/Conversion/Utils.h
```diff
@@ -6,6 +6,11 @@
 namespace mlir {
 namespace heir {
 
+// Adds conversion patterns that deal with tensor<..xsource_type>
+// when source_type will be type converted to tensor<...>, too
+void addTensorPatterns(TypeConverter &typeConverter,
```
j2kun, line 11: While the comment is helpful, I think a more explicit name like `populateTensorOfTensorWorkaroundPatterns` would be even better.

lib/Conversion/Utils.cpp
```diff
@@ -12,6 +16,238 @@ using ::mlir::func::CallOp;
...
+
+    // strides are all 1, and we need as many as there are dimensions in
+    // both shape and resultShape together
+    SmallVector<OpFoldResult> strides;
+    for (size_t i = 0; i < shape.size() + resultShape.size(); ++i) {
+      strides.push_back(rewriter.getIndexAttr(1));
+    }
+
+    // now replace op with a new extract_slice with these offsets, sizes and
+    // strides:
```
j2kun, line 108:
```suggestion
```

The other comments are useful but this one is not adding much.

tests/polynomial/lower_add.mlir
```diff
@@ -41,3 +41,51 @@ func.func @test_lower_add_prime_cmod() -> !polynomial.polynomial<#ring_prime> {
   // CHECK: return  [[TRUNC_RESULT]] : [[T]]
   return %poly2 : !polynomial.polynomial<#ring_prime>
 }
+
+func.func @test_lower_add_tensor() -> tensor<2x!polynomial.polynomial<#ring>> {
```
j2kun, line 45:
```suggestion
// CHECK-LABEL: @test_lower_add_tensor
func.func @test_lower_add_tensor() -> tensor<2x!polynomial.polynomial<#ring>> {
```

Though I didn't do it elsewhere in this file due to my past ignorance, each test should have a `CHECK-LABEL` to separate the CHECK directives that are meant for each test function. Otherwise (concurrently with some other bug) a CHECK from before line 44 could match ops in this test, and produce a confusing failure.

Feel free to add the other check-label directives for this file if you have a sec.

j2kun (approved):
Some very minor nits, otherwise extremely solid work!

Before we merge, you'll need to squash your commits into a single commit (it's a weird quirk of Google's internal system, otherwise each of your commits will show up as separate commits in the linear history post-merge).

tests/polynomial/lower_add.mlir
AlexanderViand-Intel: Done (7ee24021b8285b549dd2fd07d8b2fa6beac56cbc)

lib/Conversion/Utils.h
AlexanderViand-Intel: I went with `addTensorOfTensorConversionPatterns` for now, to match the naming with the existing `addStructuralConversionPatterns`

AlexanderViand-Intel:
@j2kun Can you have a look at what's failing here? Thanks!

j2kun:
I'm on it. The google internal systems often require me to add some extra patches to the build files and imports. It's semi-automated, and while I've fixed it and it's ready to merge, it requires another Googler to sign off on it, and everyone else on my team is sick/out today.

lib/Conversion/Utils.cpp
```diff
@@ -12,6 +16,237 @@ using ::mlir::func::CallOp;
...
+
+  target.addDynamicallyLegalDialect<tensor::TensorDialect>([&](Operation *op) {
+    return typeConverter.isLegal(op->getOperandTypes()) &&
+           typeConverter.isLegal(op->getResultTypes());
+  });
+
+  target.addDynamicallyLegalDialect<affine::AffineDialect>([&](Operation *op) {
+    return typeConverter.isLegal(op->getOperandTypes()) &&
+           typeConverter.isLegal(op->getResultTypes());
+  });
```
AlexanderViand-Intel, line 244: In the meantime, I noticed & fixed a minor bug that meant ops like `affine.yield` weren't getting handled properly (popped up while testing the bgv->poly->standard chain) because the legality check previously only considered the OperandTypes but not the ResultTypes.

j2kun: It looks like the original PR before this fix was merged asyncronously to your fix as 348c2252cf44e0378a64545dbeec682a51d83dd1 (which appears to be why this PR was not automatically marked as merged). Could you cherry pick that change to a new PR and close this one?

AlexanderViand-Intel:
Closing & creating new PR for the fix

PR #517, opened 2024-03-11
Add an initial rotate and reduce pattern

j2kun:
This PR adds a `rotate-and-reduce` pattern.

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
+++ b/lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
@@ -1,0 +1,195 @@
+#include "include/Dialect/TensorExt/Transforms/RotateAndReduce.h"
+
+#include "include/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "llvm/include/llvm/ADT/DenseSet.h"              // from @llvm-project
+#include "llvm/include/llvm/ADT/TypeSwitch.h"            // from @llvm-project
+#include "llvm/include/llvm/Support/Debug.h"             // from @llvm-project
+#include "mlir/include/mlir/Analysis/SliceAnalysis.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"   // from @llvm-project
+#include "mlir/include/mlir/IR/Iterators.h"              // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"     // from @llvm-project
+
+#define DEBUG_TYPE "rotate-and-reduce"
+
+namespace mlir {
+namespace heir {
+namespace tensor_ext {
+
+#define GEN_PASS_DEF_ROTATEANDREDUCE
+#include "include/Dialect/TensorExt/Transforms/Passes.h.inc"
+
+/// A pass that searches for a length N sequence of binary operations that
+/// reduces a length N vector to a single scalar, and replaces it with a
+/// logarithmic number of rotations and binary operations.
+struct RotateAndReduce : impl::RotateAndReduceBase<RotateAndReduce> {
+  using RotateAndReduceBase::RotateAndReduceBase;
+
+  template <typename ArithOp>
+  void tryReplace(ArithOp op, DenseSet<Operation *> &visited) {
+    LLVM_DEBUG(llvm::dbgs() << "Trying to replace " << *op << "\n");
+    SetVector<Operation *> backwardSlice;
+    BackwardSliceOptions options;
+    // asserts that the parent op has a single region with a single block.
+    options.omitBlockArguments = false;
+
+    DenseSet<Value> inputTensors;
+    DenseSet<Operation *> visitedReductionOps;
+    DenseSet<unsigned> accessIndices;
+    DenseMap<llvm::StringRef, int> opCounts;
+    opCounts[op->getName().getStringRef()]++;
+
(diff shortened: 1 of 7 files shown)


j2kun:
Update: the failing test `tests/simd/rotate_and_sum.mlir` gets to this point

```
module {
  func.func @simple_sum(%arg0: !secret.secret<tensor<8xi16>>) -> !secret.secret<i16> {
    %c6 = arith.constant 6 : index
    %c5 = arith.constant 5 : index
    %c7 = arith.constant 7 : index
    %0 = secret.generic ins(%arg0 : !secret.secret<tensor<8xi16>>) {
    ^bb0(%arg1: tensor<8xi16>):
      %1 = tensor_ext.rotate %arg1, %c1 : tensor<8xi16>, index
      %2 = arith.addi %1, %arg1 : tensor<8xi16>
      %3 = tensor_ext.rotate %arg1, %c7 : tensor<8xi16>, index
      %4 = tensor_ext.rotate %2, %c5 : tensor<8xi16>, index
      %5 = arith.addi %3, %4 : tensor<8xi16>
      %6 = arith.addi %5, %arg1 : tensor<8xi16>
      %7 = tensor_ext.rotate %6, %c6 : tensor<8xi16>, index
      %8 = arith.addi %3, %7 : tensor<8xi16>
      %9 = arith.addi %8, %arg1 : tensor<8xi16>
      %10 = tensor_ext.rotate %9, %c6 : tensor<8xi16>, index
      %11 = arith.addi %3, %10 : tensor<8xi16>
      %12 = arith.addi %11, %arg1 : tensor<8xi16>
      %extracted = tensor.extract %12[%c7] : tensor<8xi16>
      secret.yield %extracted : i16
    } -> !secret.secret<i16>
  }
}
```

After which rotate-and-sum cannot be applied. It almost seems like a rotate-and-reduce could be applied after the initial unroll as well, to capture simple cases like this one...

Based on my napkin math, the above is doing a nearly linear sum-and-rotate and not achieving the logarithmic optimum.

j2kun:
Sending for review even though there are some FIXMEs left, to solicit opinion on those FIXMEs.

lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
```diff
@@ -0,0 +1,196 @@
...
+    DenseSet<llvm::StringRef> supportedOps = {"arith.addi", "arith.muli"};
+    DenseSet<Operation *> visited;
+    // Traverse the IR in reverse order so that we can eagerly compute backward
+    // slices for each operation.
+    // FIXME: j2kun wonders: is traversing in forward order and computing
+    // backward slices at each step actually bad performance-wise? I need to
+    // test this in isolation. Also wonder: if there's no matched pattern, is
+    // traversing backward and repeatedly computing the same backward slice
+    // going to be too slow? If so, refactor into an analysis pass that caches
+    // the slice at each op.
```
AlexanderViand-Intel, line 181: Is there even a way around worst-case O(n^2) here with the current constraints? If there was an addition over all values except one from a vector of length n, for example.

AlexanderViand-Intel (approved):
I think #520 might require an update to `--insert-rotate` rather than this pass, so this looks good to me! 

HECO's pass (see https://github.com/MarbleHE/HECO/blob/main/src/Passes/hir2hir/InternalOperandBatching.cpp#L103)  is ever so slightly more general, in that it supports any contiguous block of power-of-two length, but I don't think that's even used in any of the examples, so I'm not sure how valuable it is to add that at the moment (especially given that this pass, just like `--insert-rotate` is probably due to be replaced with some kind of analysis)

lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
j2kun: I think what I'd have to do is define a custom dataflow analysis that associates each op in the IR to the multiset of tensor indices accessed by its backward slice. I.e., each op would come with a set

```
(t_1, [i_{1,1}, i_{1,2}, ..., i_{1, k_1}],
(t_2, [i_{2,1}, i_{2,2}, ..., i_{2, k_2}],
...
(t_n, [i_{n,1}, i_{n,2}, ..., i_{n, k_n}],
```

Where `t_i` is a tensor and the list is the set of indices accessed by ops in the backward slice. Then of course the join operation in the dataflow lattice is a particular sort of set union, and we could impose an artificial cap of `TOO_MANY_TENSORS` once a join causes it to combine two tensors, so that the dataflow size stays small (though without an intentional data structure, it will be O(n^2) space).

Then we could just do a linear pass over the ops until we find index sets that are sufficiently large as to benefit from a rotate-and-reduce, and replace the ops.

Actually... I wish I had thought of this first. This idea is not that hard to implement. I kept thinking of having an analysis pass that cached backward slices, but this is much cleaner. We could even have the analysis templated by the op type, and instantiate one for each of `addi`, `muli`.

tests/tensor_ext/rotate_and_reduce.mlir
```diff
@@ -0,0 +1,307 @@
...
+  %10 = arith.addi %9, %3 : i32
+  %11 = arith.muli %10, %4 : i32
+  %12 = arith.addi %11, %5 : i32
+  %13 = arith.addi %12, %6 : i32
+  %14 = arith.addi %13, %7 : i32
+  return %14 : i32
+}
+
+// CHECK-LABEL: @not_supported_missing_indices
+// CHECK-NOT: tensor_ext.rotate
```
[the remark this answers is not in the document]
j2kun: Filed https://github.com/google/heir/issues/521

lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
j2kun: I will check this in tonight as a reference and do try this tomorrow.

PR #524, opened 2024-03-12
Add -convert-elementwise-to-affine

### LINKED ISSUE #143: PolyToStandard: handling tensors of poly?

AlexanderViand-Intel:
Second part to fixing #143 by adding `-convert-elementwise-to-affine` which can be used prior to `-polynomial-to-standard` to rewrite `polynomial` operations over tensors of polynomials to loops over those tensors instead.

It's probably a good idea to throw in some loop unrolling/fusion after using this pass, as it currently creates one `affine.for` per (tensor-operand) `polynomial` op.

My idea was that it's probably relatively easy to adapt LowerElementwiseToLinalg to generate `affine.for` instead of `linalg.generic`. However, it turns out the final pass implemntation is in fact rather different from the linalg version, since `linalg.generic` doesn't need to make the operands explicit, the biggest change is that we now need to generate explicit `tensor.extract` and `tensor.insert` operations


AlexanderViand-Intel:
As far as I can tell, this now works as expected and is ready to review - I rebased it current main to remove the dependency on PR 508 as they're in fact completely orthogonal (just both relevant to #143).

AlexanderViand-Intel:
As part of integrating this into the polynomial pipeline, I changed how the `--heir-polynomial-to-llvm` pipelines handles the actual "to llvm" part from a bunch of manually assembled passes to using the `ConvertToLLVM` framework (that automagically lowers everything that has the appropriate interface). This should (hopefully) make that part of the pipeline more robust to future changes in higher-level lowerings 😃 

With this, we now also have a bgv->polynomial->llvm path (`-bgv-to-polynomial -heir-polynomial-to-llvm`), though given the performance overhead of the generic polynomial-to-standard lowering, I wouldn't recommend trying to actually run the generated code for anything except toy programs 😉

include/Dialect/Polynomial/IR/PolynomialTypes.td
```diff
@@ -14,7 +14,7 @@ class Polynomial_Type<string name, string typeMnemonic, list<Trait> traits = []>
   let mnemonic = typeMnemonic;
 }
 
-def Polynomial : Polynomial_Type<"Polynomial", "polynomial", [MemRefElementTypeInterface]> {
+def Polynomial : Polynomial_Type<"Polynomial", "polynomial", []> {
```
j2kun, line 17:
```suggestion
def Polynomial : Polynomial_Type<"Polynomial", "polynomial"> {
```

nit: remove because the default is an empty list

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
```diff
@@ -0,0 +1,136 @@
...
+#define GEN_PASS_DEF_ELEMENTWISETOAFFINE
+#include "include/Transforms/ElementwiseToAffine/ElementwiseToAffine.h.inc"
+
+// All of this is based on the ElementwiseToLinalg Pass in
+// mlir/lib/Dialect/Linalg/Transforms/ElementwiseToLinalg.cpp
+
+static bool isElementwiseMappableOpOnRankedTensors(Operation *op) {
+  if (!OpTrait::hasElementwiseMappableTraits(op)) return false;
+
+  // FIXME: Test this with `any_of`,
```
j2kun, line 20: If you prefer not to fix this in this PR, file an issue and link it to this line with

```
// TODO(#<issue_number>): message that will show up in the GH issue
```

I wrote https://github.com/j2kun/todo-backlinks to automate managing these over the lifecycle of the issue.

j2kun (approved):
Great work! Do you think there'd be an appetite for this upstream?

AlexanderViand-Intel:
Fixes are in & PR is squashed :) 

> Great work! Do you think there'd be an appetite for this upstream?

Yes, I think this (and your structural pattern conversions stuff + the tensor of tensor logic) really seem like common patterns that MLIR should support out of the box (but any upstream PR is a todo for after Toronto 😉).

j2kun (approved):

PR #525, opened 2024-03-12
WIP: Adding tfhe-rs-bool fpga tests

WoutLegiest:
Updating the tests for e2e tfhe-rs-bool and starting with the tfhe-rs-bool fpga tests

### DIFF
--- a/tests/tfhe_rust_bool/end_to_end_fpga/Cargo.toml
+++ b/tests/tfhe_rust_bool/end_to_end_fpga/Cargo.toml
@@ -2,17 +2,20 @@
 name = "heir-tfhe-rust-integration-test"
 version = "0.1.0"
 edition = "2021"
+default-run = "main"
 
 [dependencies]
 clap = { version = "4.1.8", features = ["derive"] }
 rayon = "1.6.1"
 serde = { version = "1.0.152", features = ["derive"] }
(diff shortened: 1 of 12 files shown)


tests/tfhe_rust_bool/end_to_end_fpga/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+name = "heir-tfhe-rust-integration-test"
+version = "0.1.0"
+edition = "2021"
+default-run = "main"
+
+[dependencies]
+clap = { version = "4.1.8", features = ["derive"] }
+rayon = "1.6.1"
+serde = { version = "1.0.152", features = ["derive"] }
+tfhe = { path = "tfhe-rs/tfhe", features = [
```
asraa, line 11: nit: Could you add a quick comment in the README.md that says that the tfhe library is expected to be a local path dependency? I assume even if I'm using version "X" of the upstream tfhe without the fpga feature, that I should still be able to run locally on my CPU so long as there are no packed operations, right? Otherwise, just specify that this expects a tfhe-rs fork that has a batched API.

asraa (approved):
Thank you!

j2kun (approved):

asraa (approved):

PR #526, opened 2024-03-12
Add a more thorough target slot analysis

j2kun:


lib/Dialect/TensorExt/Transforms/InsertRotate.cpp
```diff
@@ -33,6 +40,31 @@ struct InsertRotate : impl::InsertRotateBase<InsertRotate> {
...
+    }
+
+    LLVM_DEBUG({
+      getOperation()->walk([&](Operation *op) {
+        if (op->getNumResults() == 0) return;
+        auto *targetSlotLattice =
+            solver.lookupState<target_slot_analysis::TargetSlotLattice>(
+                op->getResult(0));
+        llvm::dbgs() << "Target slot for op " << *op << ": "
+                     << targetSlotLattice->getValue() << "\n";
```
asraa, line 64: This may be a silly question, but is this later used with the insert-rotate pattern?

j2kun: I'm working on it in a followup PR.

AlexanderViand-Intel: It think it might be easier to review the analysis and the pass (#530) together, so I'll have a look at that tonight :)

asraa: Ah, thats a follow-up https://github.com/google/heir/pull/531/files

asraa (approved):

AlexanderViand-Intel (approved):
> I believe this is sufficient to thread through the pass and achieve the optimal rotations

Yes, this looks really good! Not only because the taget slots match for each of the ops in an unrolled iteration of the `box_blur` loop nest, but also because it doesn't seem to flag any "excessive" target slots in `hamming_dist` 👍

include/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.h
```diff
@@ -0,0 +1,97 @@
...
+                              SymbolTableCollection &symbolTable)
+      : SparseBackwardDataFlowAnalysis(solver, symbolTable) {}
+  ~TargetSlotAnalysis() override = default;
+  using SparseBackwardDataFlowAnalysis::SparseBackwardDataFlowAnalysis;
+
+  // Given the computed results of the operation, update its operand lattice
+  // values.
+  void visitOperation(Operation *op, ArrayRef<TargetSlotLattice *> operands,
+                      ArrayRef<const TargetSlotLattice *> results) override;
+
```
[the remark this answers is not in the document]
j2kun: I believe these hooks are for adding special behavior at control flow entry and exit points, where by default the framework would have no propagation through them. For something like a liveness analysis 

https://github.com/llvm/llvm-project/blob/2a547f0f6c6b456342b9bfad38787f54f265fc96/mlir/lib/Analysis/DataFlow/LivenessAnalysis.cpp#L104-L106

I don't specifically know _which_ arguments of the branch ops, for example, are "not forwarded" automatically. But for our purposes for now we're assuming there is no control flow.

include/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.h
```diff
@@ -0,0 +1,97 @@
...
+  /// Get a known slot target.
+  const int64_t &getValue() const {
+    assert(isInitialized());
+    return *value;
+  }
+
+  bool operator==(const TargetSlot &rhs) const { return value == rhs.value; }
+
+  /// Join two target slots.
```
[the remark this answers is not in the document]
j2kun: Done! I added it as the docstring of the TargetSlot class.

Issue #520, opened 2024-03-12
Achieve the optimum rotations for insert-rotate on a simple_sum example


j2kun:
E.g., 

```mlir
$  bazel run //tools:heir-opt -- --secretize=entry-function=simple_sum --wrap-generic --canonicalize --cse --full-loop-unroll --insert-rotate --cse --canonicalize --rotate-and-reduce --canonicalize <<EOF
func.func @simple_sum(%arg0: tensor<32xi16> {secret.secret}) -> i16 {
  %c0 = arith.constant 0 : index
  %c0_si16 = arith.constant 0 : i16
  %0 = affine.for %i = 0 to 32 iter_args(%sum_iter = %c0_si16) -> i16 {
    %1 = tensor.extract %arg0[%i] : tensor<32xi16>
    %2 = arith.addi %1, %sum_iter : i16
    affine.yield %2 : i16
  }
  return %0 : i16
}
EOF
```

Should produce an IR with log2(32) = 5 rotations. Perhaps this means `rotate-and-reduce` takes priority before `insert-rotate`, perhaps it means `rotate-and-reduce` should be smarter, perhaps it means `--insert-rotate` should be smarter, or perhaps it means that a pass should happen in between the two to make `rotate-and-reduce` easier to detect the reduction.

AlexanderViand-Intel:
In HECO, we first do the `n-ary` grouping, then the rotation batching, then lower any suitable `n-ary` ops that are left to rotate-and-reduce. I think you could achieve the same effect by having `--insert-rotate` look ahead until it either finds (a) an insert operation, the index of which becomes the target slot or  (b) a non-matching arithmetic operation (in which case the target index is undefined). With a bit of luck, that should emulate the effect of having the arithmetic ops grouped without actually having to group them

j2kun:
@AlexanderViand-Intel I think I figured out why our numbers were different. The version of box_blur I was using from HECO's test files was actually a 3x3 kernel, closer to the Gaussian blur from your paper's Listing 3 (note the `affine.for %j = -1 to 2` instead of `affine.for %j = -1 to 1`)

```
 func.func @box_blur(%arg0: tensor<16xi16> {secret.secret}) -> tensor<16xi16> {
    %c16 = arith.constant 16 : index
    %c4 = arith.constant 4 : index
    %0 = affine.for %x = 0 to 4 iter_args(%arg0_x = %arg0) -> (tensor<16xi16>) {
[...]

PR #530, opened 2024-03-13
Use target slot analysis to simplify `insert-rotate`

j2kun:
Rebased off https://github.com/google/heir/pull/526

This gives us a cleaner implementation of the target slot selection, and has a few benefits:

- Now programs like `hamming_distance` and the `simple_sum` from https://github.com/google/heir/issues/520 materialize their reductions as a sequence of shift-by-1 rotates, which means we can augment `rotate-and-reduce` to support replacing that with a logarithmic number of rotations. In particular, this increased the number of rotations of the hamming_distance test case by one, which is temporary.
- I found a way to incorporate the result of a dataflow analysis into DRR rewrite patterns, which is demonstrated in this PR by first running the dataflow solver, then walking the IR and attaching the output of the analysis as attributes to the relevant ops, and adding a helper NativeCodeCall in DRR to extract the attribute. I found out you can also use this to apply constraints (i.e., apply this pattern only if the op has the relevant attr). This method seems like it would be limited to simple data output by the analysis, not something complex like a struct or piece of a larger data structure.

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
+++ b/lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
@@ -2,6 +2,7 @@
 
 #include <cstdint>
 
+#include "include/Dialect/Secret/IR/SecretOps.h"
 #include "include/Dialect/TensorExt/IR/TensorExtOps.h"
 #include "llvm/include/llvm/ADT/DenseSet.h"              // from @llvm-project
 #include "llvm/include/llvm/ADT/StringRef.h"             // from @llvm-project
@@ -53,6 +54,9 @@ struct RotateAndReduce : impl::RotateAndReduceBase<RotateAndReduce> {
       auto result =
           llvm::TypeSwitch<Operation *, LogicalResult>(upstreamOpPtr)
               .Case<arith::ConstantOp>(
+                  [&](auto upstreamOp) { return success(); })
+              // Ignore generic ops
+              .template Case<secret::GenericOp>(
                   [&](auto upstreamOp) { return success(); })
               .template Case<arith::AddIOp, arith::MulIOp>(
                   [&](auto upstreamOp) {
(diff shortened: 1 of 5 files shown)


AlexanderViand-Intel:
> * I found a way to incorporate the result of a dataflow analysis into DRR rewrite patterns, which is demonstrated in this PR by first running the dataflow solver, then walking the IR and attaching the output of the analysis as attributes to the relevant ops, and adding a helper NativeCodeCall in DRR to extract the attribute. I found out you can also use this to apply constraints (i.e., apply this pattern only if the op has the relevant attr). This method seems like it would be limited to simple data output by the analysis, not something complex like a struct or piece of a larger data structure.

Ohhh, nice! That's a cool little insight, independent of the main point of the PR/pass. We might want to start collecting these things and other "best practices"/"tips and tricks" somewhere so new contributors can discover them?

j2kun:
Rebased and ready for any final review comments.

lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
```diff
@@ -54,6 +55,9 @@ struct RotateAndReduce : impl::RotateAndReduceBase<RotateAndReduce> {
           llvm::TypeSwitch<Operation *, LogicalResult>(upstreamOpPtr)
               .Case<arith::ConstantOp>(
                   [&](auto upstreamOp) { return success(); })
+              // Ignore generic ops
+              .template Case<secret::GenericOp>(
+                  [&](auto upstreamOp) { return success(); })
```
AlexanderViand-Intel, line 60: What's going on with the explicit handling of  `secret.generic` op here? I'm guessing it's because it can be a parent op?What about other "container" ops that provide block arguments (e.g. `affine.for`) then?

tests/simd/hamming_distance.mlir
```diff
@@ -1,5 +1,5 @@
 // RUN: heir-opt --secretize=entry-function=hamming --wrap-generic --canonicalize --cse \
-// RUN:   --full-loop-unroll --insert-rotate --cse --canonicalize \
+// RUN:   --full-loop-unroll --cse --canonicalize --insert-rotate --cse --canonicalize \
```
AlexanderViand-Intel, line 2: I think it would make sense to define this as a pass pipeline, so that we can more easily ensure all `simd` tests use the same pipeline (it's easy to accidentally build a set of passes where, for each program, there's a working pipeline, but there's no single pipeline that works for every program - don't ask me how I now 😉)

asraa: I have been playing around with `secret-to-bgv` so I will probbaly start adding these passes to a pipeline like that!

j2kun: Leaving it for later.

asraa (approved):

PR #545, opened 2024-03-22
poly-to-standard: ensure modular coefficient arthimetic in polymul

### LINKED ISSUE #201: poly: add an extra modulus for lowering mul with non-machine-word sized coefficients
See TODO linking to this issue in `PolyToStandard.cpp`

inbelic:
- Ensures that the coefficent arithmetic is done mod the coefficient
  modulus during accumulation in polynomial multiplication
  - Allows for the output of polymul to retain its original element
    bitsize, and further, reduce the number of bits used in the divMod
    operation after

Resolves #201


inbelic:
After applying the current changes I unfortunately fail all of the tests in `polynomial/runner/*` except `lower_4_mul.mlir`. This is because there is still a `bufferization.clone` call in the lowered `llvm` code.

From inspection, the difference (shown below) in these test cases is that `lower_4_mul` is the only one that requires a truncation at the end of the respective `divMul`. For all other testcases, this appears to be folded away when invoking `--polynomial-to-standard` because it is truncating from and to the same type.

In the previous implementation, the truncate is never folded away because we are still operating on the naivePolymulTensorType. So I suspect the error is a result of this folding behaviour, but I don't understand why this might be the case. Possibly, we clone the memory to do the truncate operation and then the truncate operation is folded away and so the cloned memory remains as dead code, then we don't run an applicable canonicalization that removes it as dead code? But that would be strange.

Would appreciate some pointers to or reasons as to why this might be the case.

Output logs: https://gist.github.com/inbelic/3a522181be6021409cb65a71958ebaf8

j2kun:
Excellent start! What strikes me as most curious is the remaining `unrealized_conversion_cast` op, which suggests some type did not lower properly.

```
%109 = builtin.unrealized_conversion_cast %108 : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> to memref<12xi32, strided<[?], offset: ?>>
```

This first occurs after the `--lower-to-llvm` pass, and it suggests something that should have remained 32-bit was lowered to 64-bit, and hence it is incompatible with the subsequent lowering.

I suspect this is a bug upstream, but I want to take some more time to read through your code to try to figure out why this is happening. I may not have time to do that for a few days because I'm in the middle of a conference.

A (non-minimal) reproduction mlir: https://gist.github.com/j2kun/8d7082604ba93b8988ee0181e8d76f84

Run with

```
bazel run //tools:heir-opt -- --expand-strided-metadata --convert-to-llvm $PWD/test.mlir
```

Output: 

```
   %239 = builtin.unrealized_conversion_cast %238 : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> to memref<12xi32, strided<[?], offset: ?>>
    %241 = builtin.unrealized_conversion_cast %240 : memref<12xi32, strided<[?], offset: ?>> to !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>
```

inbelic:
Hm interesting. Such a lowering error could definitely be a consequence of the changes. I will take a closer look with that in mind.

Of course, no rush. Thanks for the quick response :)

j2kun:
I still don't have an answer, but I have a more minimal reproduction:

```
func.func @__heir_poly_mod_4294967296_1_x12(%arg0: tensor<23xi32>) -> tensor<12xi32> {
  %0 = polynomial.from_tensor %arg0 : tensor<23xi32> -> !polynomial.polynomial<<cmod=4294967296, ideal=#polynomial.polynomial<-1 + x**23>>>
  %1 = polynomial.add (%0, %0) : !polynomial.polynomial<<cmod=4294967296, ideal=#polynomial.polynomial<-1 + x**23>>>
  %2 = polynomial.to_tensor %1 : !polynomial.polynomial<<cmod=4294967296, ideal=#polynomial.polynomial<-1 + x**23>>> -> tensor<23xi32>
  %extracted_slice = tensor.extract_slice %2[0] [12] [1] : tensor<23xi32> to tensor<12xi32>
  return %extracted_slice : tensor<12xi32>
}
```

Which lowers via `polynomial-to-standard` to

```
module {
  func.func @__heir_poly_mod_4294967296_1_x12(%arg0: tensor<23xi32>) -> tensor<12xi32> {
    %0 = arith.addi %arg0, %arg0 : tensor<23xi32>
    %extracted_slice = tensor.extract_slice %0[0] [12] [1] : tensor<23xi32> to tensor<12xi32>
    return %extracted_slice : tensor<12xi32>
  }
}
```

Which then lowers via the rest of the pipeline to 

```
module {                                                                                                                            
  llvm.func @__heir_poly_mod_4294967296_1_x12(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: i64, %arg3: i64, %arg4: i64) -> !llvm.struc
t<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> {                                                                                
    %0 = llvm.mlir.constant(12 : index) : i64                                                                                       
    %1 = llvm.mlir.constant(0 : index) : i64                                                                                        
    %2 = llvm.mlir.constant(23 : index) : i64
    %3 = llvm.mlir.constant(1 : index) : i64
    llvm.br ^bb1(%1 : i64)
  ^bb1(%4: i64):  // 2 preds: ^bb0, ^bb2
    %5 = llvm.icmp "slt" %4, %2 : i64
    llvm.cond_br %5, ^bb2, ^bb3
  ^bb2:  // pred: ^bb1
    %6 = llvm.getelementptr %arg1[%arg2] : (!llvm.ptr, i64) -> !llvm.ptr, i32
    %7 = llvm.mul %4, %arg4  : i64
    %8 = llvm.getelementptr %6[%7] : (!llvm.ptr, i64) -> !llvm.ptr, i32
    %9 = llvm.load %8 : !llvm.ptr -> i32
    %10 = llvm.add %9, %9  : i32
    llvm.store %10, %8 : i32, !llvm.ptr
    %11 = llvm.add %4, %3  : i64
    llvm.br ^bb1(%11 : i64)
  ^bb3:  // pred: ^bb1
    %12 = llvm.mlir.undef : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>
    %13 = llvm.insertvalue %arg0, %12[0] : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> 
    %14 = llvm.insertvalue %arg1, %13[1] : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> 
    %15 = llvm.insertvalue %arg2, %14[2] : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> 
    %16 = llvm.insertvalue %0, %15[3, 0] : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> 
    %17 = llvm.insertvalue %arg4, %16[4, 0] : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> 
    %18 = builtin.unrealized_conversion_cast %17 : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)> to memref<12xi32, s
trided<[?], offset: ?>>
    %19 = bufferization.clone %18 : memref<12xi32, strided<[?], offset: ?>> to memref<12xi32, strided<[?], offset: ?>>
    %20 = builtin.unrealized_conversion_cast %19 : memref<12xi32, strided<[?], offset: ?>> to !llvm.struct<(ptr, ptr, i64, array<1 x
 i64>, array<1 x i64>)>
    llvm.return %20 : !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>
  }
}
```

j2kun:
Please try applying this patch to your PR (`git apply < file.patch` should work).

I think it will be sufficient for this PR, though I'm not sure if this is exactly the right place to put this pass in the pipeline. lower_mul_0.mlir passes with this applied on top of your changes.

```
From 9026247186110df2c61b1c571cc5798b48daf1bb Mon Sep 17 00:00:00 2001
From: Jeremy Kun <j2kun@users.noreply.github.com>
Date: Sat, 23 Mar 2024 06:56:33 -0700
Subject: [PATCH] Add bufferization to memref pass to poly pipeline

---
 tools/BUILD        | 1 +
 tools/heir-opt.cpp | 2 ++
 2 files changed, 3 insertions(+)

diff --git a/tools/BUILD b/tools/BUILD
index fd5b8e52..df1e9b49 100644
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -72,6 +72,7 @@ cc_binary(
         "@llvm-project//mlir:ArithToLLVM",
         "@llvm-project//mlir:ArithTransforms",
+        "@llvm-project//mlir:BufferizationToMemRef",
         "@llvm-project//mlir:BufferizationTransforms",
         "@llvm-project//mlir:ControlFlowToLLVM",
         "@llvm-project//mlir:ConvertToLLVM",
diff --git a/tools/heir-opt.cpp b/tools/heir-opt.cpp
index a7e0ec0b..5ecb6989 100644
--- a/tools/heir-opt.cpp
+++ b/tools/heir-opt.cpp
@@ -34,6 +34,7 @@
 #include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
 #include "mlir/include/mlir/Conversion/AffineToStandard/AffineToStandard.h"  // from @llvm-project
 #include "mlir/include/mlir/Conversion/ArithToLLVM/ArithToLLVM.h"  // from @llvm-project
+#include "mlir/include/mlir/Conversion/BufferizationToMemRef/BufferizationToMemRef.h"  // from @llvm-project
 #include "mlir/include/mlir/Conversion/ConvertToLLVM/ToLLVMPass.h"  // from @llvm-project
 #include "mlir/include/mlir/Conversion/IndexToLLVM/IndexToLLVM.h"  // from @llvm-project
 #include "mlir/include/mlir/Conversion/MemRefToLLVM/MemRefToLLVM.h"  // from @llvm-project
@@ -152,6 +153,7 @@ void polynomialToLLVMPipelineBuilder(OpPassManager &manager) {
   // But lowering to loops also re-introduces affine.apply, so re-lower that
   manager.addNestedPass<FuncOp>(createConvertLinalgToLoopsPass());
   manager.addPass(createLowerAffinePass());
+  manager.addPass(createBufferizationToMemRefPass());
 
   // Cleanup
   manager.addPass(createCanonicalizerPass());
-- 
2.44.0.396.g6e790dbe36-goog

```

j2kun:
BTW, I think it's fine to leave the bufferization patch in this PR, so that the change is atomic with the change that caused its need.

tests/polynomial/lower_mul.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: heir-opt --polynomial-to-standard %s | FileCheck %s
+// RUN: heir-opt --polynomial-to-standard --split-input-file %s | FileCheck %s
```
j2kun, line 1: A little tip: while using split-input-file is fine, if you don't want to use this, then what you do is use `CHECK-LABEL` instead of `CHECK` when matching function names (and use `CHECK-SAME` if you need to capture regexes of function arguments on the same line), and that creates a logical separation between test cases that FileCheck understands. A decent example is here: https://github.com/google/heir/blob/1125bfd29a65f5c31dddbabf84749688a70615e9/tests/secret/distribute_generic.mlir#L3-L4

inbelic:
Awesome. The patch did the trick and there is the expected behaviour. Okay sounds good I will reword the commit so you are the author and with your subject.

j2kun:
>  I will reword the commit so you are the author and with your subject.

No need to bother with that. It will all get squashed anyway before merging (see https://heir.dev/docs/contributing/#pull-request-review-flow)

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -478,21 +481,15 @@ struct ConvertMul : public OpConversionPattern<MulOp> {
...
-          auto lhs = args[0];
-          auto rhs = args[1];
-          auto accum = args[2];
-          if (!inputOutputHaveSameBitwidth) {
-            lhs = b.create<arith::ExtSIOp>(polymulTensorType.getElementType(),
-                                           lhs);
-            rhs = b.create<arith::ExtSIOp>(polymulTensorType.getElementType(),
-                                           rhs);
-          }
+          auto lhs = b.create<arith::ExtSIOp>(intermediateType, args[0]);
```
inbelic, line 484: Question related to these conversions. What are the consequences of doing the conversions inside the linalg generic operation, as opposed to doing this on the entire tensor before and after the generic op? Is there one operation that would allow for better lowering optimizations?

tests/polynomial/lower_mul.mlir
inbelic: Nice tip! In this case I used `--split-input-file` since I changed the coefficient modulus and to avoid having to create new values with different names. But maybe that would create confusion when reading the testcase

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: I think the effect is rather minimal, since we're mostly using this as a "cheap" way to express the convolution and get a free lowering to loops, rather than expressing any sophisticated transformation of the linalg operation.

But to answer your question, I believe if you put it outside the loop, then it would lower to a sequence of multiple loops, which would then likely be fused together by LLVM optimizations (if we went through llc). But I would have to inspect the lowered code to be sure. (you could try it and use `--mlir-print-ir-after-all` to see the intermediate IRs after each pass.

j2kun (approved):
Beautifully done! Two little nitpicks and we should be ready to squash and merge.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -463,12 +460,42 @@ struct ConvertMul : public OpConversionPattern<MulOp> {
...
+    // 2^6.
+    //
+    // Further, when we truncate from the intermediate type back to the
+    // polynomial type, we can't rely on the truncation behaviour to ensure that
+    // the value congruent with respect to cmod. So we will have to manually
+    // convert by computing acc + cmod mod cmod.
+    //
+    // Eg: Let acc = -12. Then arith.remsi acc 7 = -5 : i6, and so,
+    // arith.trunci -5 : i6 -> i3 = 3. -5 is not congruent with 3 mod 7. So we
+    // will need to compute -5 + 7 mod 7 = 2.
```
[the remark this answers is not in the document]
inbelic: The addition should be necessary, otherwise the truncation can cause an error. In the example `-5 = 111011 : i6`, if we directly truncate it to an i3 then we would get `3 = 011 : i3` which are not congruent modulo 7. So we do the addition to ensure that they are in [0,7) before the truncation to avoid such an error.

tests/polynomial/lower_mul.mlir
```diff
@@ -53,43 +52,85 @@
 // CHECK:    ^[[bb0:.*]](%[[DIVIDEND:.*]]: [[NAIVE_POLYMUL_TENSOR_TY]]):
 // CHECK:      %[[DIVISOR:.*]] = arith.constant dense<"0x[[DIVISOR_COEFFS:010*10*]]"> : [[NAIVE_POLYMUL_TENSOR_TY]]
 
-// CHECK:      %c0_i64 = arith.constant 0 : i64
+// CHECK:      %c0_i32 = arith.constant 0 : i32
```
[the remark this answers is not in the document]
inbelic: I added the issue but didn't have the privileges to assign it to you, #551.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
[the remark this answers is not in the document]
j2kun: Thanks!

j2kun (approved):

PR #549, opened 2024-03-23
Add RNS dialect and type

j2kun:
This PR adds:

- an RNS dialect (boilerplate and plumbing)
- a new type `rns<ty1, ..., tyN>`, representing the domain of an RNS decomposition
- a new type interface `RNSBasisTypeInterface` which allows types to determine if they are compatible with other basis types used in the same instance of an `rns` type.

One interesting thing I learned while doing this is that it _is_ possible to attach an out-of-tree type interface to an upstream type, using the `attachInterface` method of the type at the out-of-tree dialect initialization time. We can use this to retroactively attach `RNSBasisTypeInterface` to integer types if that ends up being the best option.

Another interesting thing I learned is that the TypeInterface requires no registration on its own, it is only implicitly registered once a type ends up using it.

A simple example, pulled from `syntax.mlir` test

```
#ideal = #polynomial.polynomial<1 + x**1024>

// random 32-bit primes
#ring_1 = #polynomial.ring<cmod=3721063133, ideal=#ideal>
#ring_2 = #polynomial.ring<cmod=2737228591, ideal=#ideal>
#ring_3 = #polynomial.ring<cmod=3180146689, ideal=#ideal>

!poly_ty_1 = !polynomial.polynomial<#ring_1>
!poly_ty_2 = !polynomial.polynomial<#ring_2>
!poly_ty_3 = !polynomial.polynomial<#ring_3>

!ty = !rns.rns<!poly_ty_1, !poly_ty_2, !poly_ty_3>

func.func @test_syntax(%arg0: !ty) -> !ty {
  return %arg0 : !ty
}
```

In follow-up PRs, I intend to introduce an op for [applying an RNS decomposition](https://github.com/google/heir/issues/558) (agnostic of how that is done), and update polynomial ops to [operate on RNS types](https://github.com/google/heir/issues/559).

### DIFF
--- a/include/Dialect/RNS/IR/RNSTypeInterfaces.td
+++ b/include/Dialect/RNS/IR/RNSTypeInterfaces.td
@@ -1,0 +1,33 @@
+#ifndef INCLUDE_DIALECT_RNS_IR_RNSTYPEINTERFACES_TD_
+#define INCLUDE_DIALECT_RNS_IR_RNSTYPEINTERFACES_TD_
+
+include "mlir/IR/BuiltinTypeInterfaces.td"
+
+def RNSBasisTypeInterface : TypeInterface<"RNSBasisTypeInterface"> {
+  let cppNamespace = "::mlir::heir::rns";
+  let description = [{
+    This interface is required for a type to be used as a parameter
+    to an `rns` type.
+  }];
+  let methods = [
+    InterfaceMethod<
+    /*description=*/[{
+      Returns true if this type is compatible with another type in the
+      same RNS basis. In particular, the set of types used for a single
+      RNS basis are never equal as types, but instead have some common
+      attribute that must be checked here. For example, an RNS type where
+      the basis types are polynomials would return true if the two types
+      are both polynomial types, even if they have different coefficient
+      moduli.
+
+      `isCompatibleWith` must be both commutative and associative, in the sense
+      that `type1.isCompatibleWith(type2)` if and only if
+      `type2.isCompatibleWith(type1)`, and further
+      `type2.isCompatibleWith(type3)` if and only if
+      `type1.isCompatibleWith(type3)`.
+    }],
+    "bool", "isCompatibleWith", (ins "::mlir::Type":$otherRnsBasisType)>
+  ];
+}
+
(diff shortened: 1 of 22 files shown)


asraa (approved):
LGTM to start. I think later on we'll also want to double-RNS polynomials and update the trait so that the mod can match while the ideal is split. But for now, lgtm!

include/Dialect/RNS/IR/RNSTypeInterfaces.td
```diff
@@ -0,0 +1,33 @@
...
+def RNSBasisTypeInterface : TypeInterface<"RNSBasisTypeInterface"> {
+  let cppNamespace = "::mlir::heir::rns";
+  let description = [{
+    This interface is required for a type to be used as a parameter
+    to an `rns` type.
+  }];
+  let methods = [
+    InterfaceMethod<
+    /*description=*/[{
+      Returns true if this type is compatible with another type in the
```
AlexanderViand-Intel, line 15: I'm not super excited about the vague "compatible" naming here, but I'm sure we'll have a better idea of what exactly this should model after we have more examples.

AlexanderViand-Intel (approved):

AlexanderViand-Intel:
> One interesting thing I learned while doing this is that it is possible to attach an out-of-tree type interface to an upstream type, using the attachInterface method of the type at the out-of-tree dialect initialization time. We can use this to retroactively attach RNSBasisTypeInterface to integer types if that ends up being the best option.

> Another interesting thing I learned is that the TypeInterface requires no registration on its own, it is only implicitly registered once a type ends up using it.

Once again, some very cool "hidden" upstream features! I'm guessing you're already collecting these insights somewhere?

j2kun:
> > One interesting thing I learned while doing this is that it is possible to attach an out-of-tree type interface to an upstream type, using the attachInterface method of the type at the out-of-tree dialect initialization time. We can use this to retroactively attach RNSBasisTypeInterface to integer types if that ends up being the best option.
> 
> > Another interesting thing I learned is that the TypeInterface requires no registration on its own, it is only implicitly registered once a type ends up using it.
> 
> Once again, some very cool "hidden" upstream features! I'm guessing you're already collecting these insights somewhere?

I will have to write a blog post studying this in more detail...

PR #569, opened 2024-03-27
Update tfhe-rs-bool-emitter: cggi to tfhe-rs-bool fpga ready code

WoutLegiest:
- Update the tfhe-rs-bool emitter to now emit working FPGA tfhe-rs-bool code
- Remove the and_packed operations
- Update the tests to work with the new emitted code and the removal of the packed operations

### DIFF
--- a/lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
+++ b/lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
@@ -64,7 +64,7 @@ LogicalResult TfheRustBoolEmitter::translate(Operation &op) {
           // Arith ops
           .Case<arith::ConstantOp>([&](auto op) { return printOperation(op); })
           // TfheRustBool ops
-          .Case<AndOp, NandOp, OrOp, NorOp, XorOp, XnorOp, AndPackedOp>(
+          .Case<AndOp, NandOp, OrOp, NorOp, XorOp, XnorOp>(
               [&](auto op) { return printOperation(op); })
           // Tensor ops
           .Case<tensor::ExtractOp, tensor::FromElementsOp>(
@@ -144,11 +144,14 @@ LogicalResult TfheRustBoolEmitter::printOperation(func::FuncOp funcOp) {
 
 LogicalResult TfheRustBoolEmitter::printOperation(func::ReturnOp op) {
   std::function<std::string(Value)> valueOrClonedValue = [&](Value value) {
-    auto cloneStr = "";
+    auto suffix = "";
     if (isa<BlockArgument>(value)) {
-      cloneStr = ".clone()";
+      suffix = ".clone()";
     }
-    return variableNames->getNameForValue(value) + cloneStr;
+    if (isa<tensor::FromElementsOp>(value.getDefiningOp())) {
+      suffix = ".into_iter().cloned().collect()";
+    }
+    return variableNames->getNameForValue(value) + suffix;
   };
 
   if (op.getNumOperands() == 1) {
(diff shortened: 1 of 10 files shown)


lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -165,29 +168,70 @@ void TfheRustBoolEmitter::emitAssignPrefix(Value result) {
...
+        {nonSksOperands[0], nonSksOperands[1]}, [&](Value value) {
+          auto *prefix = value.getType().hasTrait<PassByReference>() ? "&" : "";
+          auto suffix = "";
+          // First check if a DefiningOp exists
+          // if not: comes from function definition
+          mlir::Operation *opParent = value.getDefiningOp();
+          if (opParent) {
+            prefix = isa<tensor::ExtractOp>(opParent) ? prefix : "";
+            prefix =
+                isa<tensor::FromElementsOp>(value.getDefiningOp()) ? "&" : "";
```
j2kun, line 202: The second line appears to completely override the first line. Maybe you meant to include `prefix` in one of the branches of the second line as well?

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -165,29 +168,70 @@ void TfheRustBoolEmitter::emitAssignPrefix(Value result) {
...
+
+  } else {
+    emitAssignPrefix(result);
+
+    auto operandTypesIt = operandTypes.begin();
+    os << variableNames->getNameForValue(sks) << "." << op << "(";
+    os << commaSeparatedValues(nonSksOperands, [&](Value value) {
+      auto *prefix = value.getType().hasTrait<PassByReference>() ? "&" : "";
+      // First check if a DefiningOp exists
+      // if not: comes from function definition
```
j2kun, line 221: Optional: If you want to be extra careful about this: you would cast it to a BlockArgument and check that the owner is a func op.

tests/tfhe_rust_bool/end_to_end_fpga/src/main.rs
```diff
@@ -72,10 +72,11 @@ fn main() {
     let ct_1 = encrypt(flags.input1.into(), &client_key);
     let ct_2 = encrypt(flags.input2.into(), &client_key);
 
-    let ct_1= ct_1.iter().collect();
-    let ct_2= ct_2.iter().collect();
-
+    let t = Instant::now();
     let result = fn_under_test::fn_under_test(&server_key, &ct_1, &ct_2);
+    let run = t.elapsed().as_millis();
```
j2kun, line 77:
```suggestion
    // let t = Instant::now();
    let result = fn_under_test::fn_under_test(&server_key, &ct_1, &ct_2);
    // let run = t.elapsed().as_millis();
```

j2kun: Or delete them

tests/tfhe_rust_bool/end_to_end_fpga/test_cggi_add_bool.mlir
```diff
@@ -0,0 +1,77 @@
...
+
+// RUN: heir-opt --straight-line-vectorize --cggi-to-tfhe-rust-bool -cse -remove-dead-values %s | heir-translate --emit-tfhe-rust-bool > %S/src/fn_under_test.rs
+// RUN: cargo run --release --manifest-path %S/Cargo.toml -- 1 1 | FileCheck %s
+
+#encoding = #lwe.unspecified_bit_field_encoding<cleartext_bitwidth = 1>
+!ct_ty = !lwe.lwe_ciphertext<encoding = #encoding>
+!pt_ty = !lwe.lwe_plaintext<encoding = #encoding>
+
+// CHECK: 01000000
+func.func @fn_under_test(%arg0: tensor<8x!ct_ty>, %arg1: tensor<8x!ct_ty>) -> tensor<8x!ct_ty> {
```
j2kun, line 11: Maybe add a comment indicating what this function is supposed to do? It would help to understand how to fix it if something causes the test to fail in the future.

j2kun (approved):

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -165,29 +168,70 @@ void TfheRustBoolEmitter::emitAssignPrefix(Value result) {
   os << "let " << variableNames->getNameForValue(result) << " = ";
 }
 
+void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
```
asraa, line 171: nit: could you also add a quick assertion here that the value element type is a EncryptedUint (or maybe add this to the doc-string in the header file that it's expected to run on shaped types of encrypted values?)

tests/tfhe_rust_bool/end_to_end_fpga/src/main.rs
WoutLegiest: I have added an additional comment
I know it is not the right thing to do, but it is for obtaining quick timing results when run on fpga (turn the printing off for the user/sys timings)

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
WoutLegiest: The tfhe-rs-bool emitter only produces EncryptedBools
I added an additional check

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
WoutLegiest: Refactored the code

Issue #571, opened 2024-03-27
Port remaining HECO evaluation artifacts and ensure they produce equivalently optimal results.


j2kun:
@AlexanderViand-Intel I may need your help here. I haven't figured out how to build HECO yet (everything succeeds in your repo instructions up to the linking step, then it fails with a bazillion linker errors to LLVM).

What I think I'd want is just the output MLIR from each of the evaluation artifacts, so that I can compare the rotation counts in them with what HEIR produces. I tried manually porting roberts-cross, and I think I see where HEIR is still lacking, but I want some confirmation.

j2kun:
Last one appears to be https://github.com/MarbleHE/HECO/blob/main/evaluation/comparison/heco_input/robertscross_64x64.mlir

AlexanderViand-Intel:
I'm trying to get up to speed after RWC/Easter Holidays, and I saw that you mentioned compile times in the order of 50 minutes in #589, which sounds an order of magnitude worse than what HECO gets. In HECO, I did a few tricks to speed things up, such as moving one or two patterns out of normal canonicalization into their own pass, maybe we can do something similar here.

> @AlexanderViand-Intel I may need your help here. I haven't figured out how to build HECO yet (everything succeeds in your repo instructions up to the linking step, then it fails with a bazillion linker errors to LLVM).

Mh, that's odd - the repo, and the `artifact` tag specifically, should "just" work, having been through the USENIX artifact evaluation process. What linker did you use? `lld` or something else? 

> What I think I'd want is just the output MLIR from each of the evaluation artifacts, so that I can compare the rotation counts in them with what HEIR produces. I tried manually porting roberts-cross, and I think I see where HEIR is still lacking, but I want some confirmation.

Sure, let me run those for you and figure out how best to post them.

j2kun:
That is bizarre about copybara. Here's the commit: https://github.com/google/heir/commit/4d84f04d7d82fa8507e26ef8968b675772def48a

j2kun:
Yeah I was thinking I could extract the needed patterns (converting the cleartext weight tensor to individual constants would suffice), but haven't had the chance to figure out exactly which pattern does that in canonicalize.

AlexanderViand-Intel:
> > What I think I'd want is just the output MLIR from each of the evaluation artifacts, so that I can compare the rotation counts in them with what HEIR produces. I tried manually porting roberts-cross, and I think I see where HEIR is still lacking, but I want some confirmation.
> 
[...]

PR #573, opened 2024-03-28
Port roberts_cross from HECO, and use sparse constant analysis to identify static indices

j2kun:
This PR has three parts:

- Porting the roberts_cross example from HECO, and realizing that it doesn't simplify properly because the pipeline can't identify its static insertion indices
- Adding an SCCP pass before insert-rotate to fix the above, and then seeing that it breaks box_blur for a reason I couldn't determine.
- Swapping a SCCP pass for a "simulated folding" sparse constant analysis, which is now used in insert-rotate for identifying constant that materialize as slot targets

Part of https://github.com/google/heir/issues/571

### DIFF
--- a/lib/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.cpp
+++ b/lib/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.cpp
@@ -1,8 +1,10 @@
 #include "include/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.h"
 
 #include "lib/Dialect/Utils.h"
-#include "llvm/include/llvm/ADT/TypeSwitch.h"              // from @llvm-project
-#include "llvm/include/llvm/Support/Debug.h"               // from @llvm-project
+#include "llvm/include/llvm/ADT/TypeSwitch.h"  // from @llvm-project
+#include "llvm/include/llvm/Support/Debug.h"   // from @llvm-project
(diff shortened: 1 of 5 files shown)


lib/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.cpp
```diff
@@ -22,15 +24,47 @@ void TargetSlotAnalysis::visitOperation(
...
+        const dataflow::Lattice<dataflow::ConstantValue> *insertIndexLattice =
+            sccpAnalysis
+                ->lookupState<dataflow::Lattice<dataflow::ConstantValue>>(
+                    insertIndexValue);
+
+        if (insertIndexLattice) {
+          LLVM_DEBUG(llvm::dbgs()
+                     << "At " << insertOp << " SCCP analysis gives lattice of "
+                     << *insertIndexLattice << "\n");
+        }
```
asraa, line 44: hey! i'm just curious, i ran into needing constant folded strides / sizes once and I came across this method: https://github.com/regehr/llvm-project/blob/ba0763e4cb08015bd0acef5cf05a372d423017b4/mlir/lib/Dialect/Vector/Transforms/VectorEmulateNarrowType.cpp#L88

can you `getAsOpFoldResult` the insert index values?

j2kun: I don't think I can, because in this case the constant analysis is simulated, not materialized in the actual IR.

PR #590, opened 2024-04-03
poly: add initial lowering of NTT op

inbelic:
poly: add initial lowering NTT op
    
      - add implementation of the fast NTT to lower the NTT operation
      - add testing for lowering and a runner

poly: add additional root parameter to RingAttr
    
      - include an optional parameter that computes the 2n-th primitive
        root of a rings ideal degree and its coefficient modulus
      - allow for this to be retrieved from a precomputation or to be
        user-specified
      - Add a script that will compute the 2n-th roots of unity for the
        specified polynomial degree values and n. The values are only
        outputted if the 2n-th root of unity exists.
      - Note: even if the nth root of unity exists it will not be output if
        the 2n-th doesn't exist


inbelic:
I think this is ready for an initial review. @j2kun 

Currently only the NTT lowering is implemented, I would like to make sure that the approach is good before doing the work for the INTT in a similar manner. There are some points that would be good to get feedback on:

- Which (prime, degree) pairs should be included in the generation script.
- In the current implementation, the recursive function `fastNTT` function is unrolled out already. It might be beneficial to make this into an SCF for loop? Or we could generate custom private functions for each degree of the iteration (similar to the implementation of the polynomial modulus)? Functionally they would be the same, but it could improve readability of the lowered code.
- At the moment we construct an empty tensor for each iteration of the fast NTT and then concatenate them to construct our final result. We could reduce the uses of `tensor.empty` to a single one at the start and then operate directly on the single tensor. Although I haven't checked the generated LLVM code to see if it does this for us.

inbelic:
Sweet, no rush. Another thing to think about would be if we should automatically extend the polynomial coefficients to be a power of 2, or if we should issue an error/warning for the user to do so themselves? Right now we just assume it is a power of 2.

inbelic:
Sweet, no rush.

Note to self: I realise now there should also be a warning/error in the case the degree is not a power of 2.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +802,149 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+          iterType, cmod.sextOrTrunc(iterType.getElementTypeBitWidth())));
+
+  // Implementing Equation 23 from https://ieeexplore.ieee.org/document/10177902
+  //
+  // We have that evenNTT = {A_j}, oddNTT = {B_j} and roots = {root^{2j + 1}}
+  // for all 0 <= j < degree/2
+  auto roots = b.create<arith::ConstantOp>(
+      iterType, constructRoots(iterType, root, degree, cmod));
+
+  // Since roots_j * B_j -> [0, cmod^2) then RemUI will compute the modulus
```
j2kun, line 869: I think what you're saying here is that overflow is expected? I worry that this will not behave exactly as you anticipate, because the overflow will implicitly compute a different remainder.

E.g., if cmod = 7 but the underlying storage type is `i3`, then `(6^2 mod 8) mod 7 = 4 mod 7 = 4`, but `6^2 mod 7 = 1`, and so the overflow would break the equivalence.

Feel free to file a TODO for later improvement, I think a carefully crafted test would identify if this is a problem.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +802,149 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+  auto ctPlus = b.create<arith::AddIOp>(iterType, evenNTT, rootsBModded);
+  auto ctPlusModded = b.create<arith::RemUIOp>(ctPlus, cModVec);
+
+  // Since A - rootsB -> (-cmod, cmod) then we can add cmod such that the
+  // range is shifted to (0, 2 * cmod) and use RemUI to compute the modulus
+  auto ctMinus = b.create<arith::SubIOp>(iterType, evenNTT, rootsBModded);
+  auto ctMinusShifted = b.create<arith::AddIOp>(ctMinus, cModVec);
+  auto ctMinusModded = b.create<arith::RemUIOp>(ctMinusShifted, cModVec);
+
+  // Manually concatenate since tensor::ConcatOp is currently not bufferizable
```
j2kun, line 883: Not a problem to do this, but you should be able to get around it by adding this to the pattern set in this pass: https://github.com/llvm/llvm-project/blob/7d4e8c1f3bbfe976f4871c9cf953f76d771b0eda/mlir/include/mlir/Dialect/Tensor/Transforms/Transforms.h#L75

Then create a ConcatOp here and that pattern should automatically convert it to insert_slice ops for you.

j2kun (changes requested):
Don't take my mass of comments poorly, this is really great work!

For a few of your other questions:

> for loop

I think unrolled will actually admit more opportunities for optimization. The only case we'd stand to do better is if we generate affine loops and apply loop fusion techniques to the result. But that seems beyond the scope of this PR.

> avoiding extra tensors

I'd expect lower level optimizations to handle this, but you could also experiment with a combination of `-canonicalize` and `-eliminate-empty-tensors`, which may be smart enough to identify how the insert_slice/extract_slices are forwarded through the unrolled iterations (and if that works, it'd be another reason to keep it unrolled, because forwarding is hard inside a loop).

j2kun:
One more inconvenience: my partial work upstreaming the polynomial dialect (https://github.com/llvm/llvm-project/pull/72081) causes a name conflict here. The fix to those namespace conflicts will probably land in this repo today. The changes that need to be made to this PR to handle this are:

- Use a fully-qualified namespace `::mlir::heir::polynomial::Polynomial` in the C++ API where relevant (https://github.com/google/heir/pull/616, which already landed)
- Use `_polynomial` as the dialect name instead of `polynomial` in mlir code (https://github.com/google/heir/pull/617, which will land as a separate commit later)

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
inbelic: Not quite what I was going for. It was more trying to convey that since both of the inputs to the multiplication are positive within [0, cmod), then we don't need to do the chain of `arith.RemSI -> arith.AddI -> arith.RemUI` with `cmod` to ensure the modulo is computed correctly.

Currently, there is a conversion before we invoke `fastNTT` that will cast it to an intermediate type that has double the bitwidth. For your example that would be `i3 -> i6`, to ensure that there is no overflow.

lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
```diff
@@ -145,9 +146,25 @@ void RingAttr::print(AsmPrinter &p) const {
...
 }
 
+static bool is2nthRoot(const APInt &root, const unsigned degree,
+                       const APInt &cmod) {
+  // root bitwidth may be 1 less then cmod
+  APInt a = APInt(root).zext(cmod.getBitWidth());
+  auto len = (size_t)std::log2((double)degree);
+  for (size_t i = 0; i <= len; ++i) {
+    if (a == 1) return false;
+    a = (a * a).urem(cmod);
```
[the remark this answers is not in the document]
inbelic: Good catch! I switched to your recommended implementation.

More then happy to add unit tests, although I am having some issues in how I should "export" the function to be used in the test file.

scripts/generate_static_roots.py
```diff
@@ -0,0 +1,134 @@
...
+  for cmod, roots in cmod_mapping.items():
+    for degree, root in zip(degrees, roots):
+      if root:
+        output = ENTRY_TEMPLATE.format(cmod=cmod, degree=degree, root=root)
+        outfile.write(output)
+  outfile.write(CLOSING)
+
+def main(args: argparse.Namespace) -> None:
+  degrees = args.degrees if args.degrees else [256, 512, 1024]
+  cmods = args.cmods if args.cmods else [12289, 8380417]
```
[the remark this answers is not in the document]
inbelic: These values were just used as a filler. Other than being the Dilithium and Falcon values there is no other motivation behind them. I added a pair for the large N and will implement a performance test with it. I will also add some random ones then for now, and denote a TODO.

inbelic:
Awesome. I really appreciate the time for all of the feedback!

Okay, sounds good. I will experiment with loop unrolling and see if it is worth it to change.

Thanks for the heads up, will deal with that once I get to rebasing.

lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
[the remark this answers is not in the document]
j2kun: You should be able to add the declaration to PolynomialAttributes.h, then create a dedicated `cc_test` build rule in this directory that depends on the `PolynomialAttributes` build target, and import `PolynomialAttributes.h` in the test file. If you can't get it to work in 15 minutes, let's file an issue and I can add the test in a followup PR.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: I think it should work the first way (using applyPartialConversion), but you may need to mark the concat op as an illegal op for it to trigger the pattern, which is a quirk of the dialect conversion framework. If that doesn't work, file a TODO and let's go back to the insert slice style.

j2kun:
> The next steps will be to implement the INTT

I would support doing this i followup PRs, just so we can incrementally merge things and avoid any headaches rebasing in the future.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
inbelic: Nice, making the op illegal did the trick

lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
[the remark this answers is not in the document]
inbelic: Unfortunately couldn't get it to work on my end. When I try to add the `PolynomialAttributes` target to the `cc_test` build rule it causes linker errors during compilation, both in the existing one and in an independent one. Created a follow-up issue.

inbelic:
Sounds good. I will hopefully get those follow-up prs in the next couple days :)

j2kun:
Let me know when this is ready for a final review.

inbelic:
Okay, took another jab at looking into the performance. For the new test added, it takes ~40 seconds to compile with `--heir-polynomial-to-llvm` and < 1 second for `--polynomial-to-standard` for `n=256` with random coefficients.

My initial suspicion was that my use of `tensor.extract_slice`s to split the vector at each function call would have been the cause. So instead I have changed it to use `tensor.extract` on each element using the bit-reverse order. The other thing I wanted to see, was if creating the `cMod` and `roots` at each function call was generating too many values for `--cse` to deal with efficiently. So I have moved these out to be pre-computed before calling `fastNTT`.

j2kun:
Starting from c6c5b7cd5e8fd8d72b94f0b829b3e0f0d0d9b85f, I ran your test from be317d2 with `-mlir-timing` and see

```
  Total Execution Time: 25.0003 seconds

  ----User Time----  ----Wall Time----  ----Name----
    0.0060 (  0.0%)    0.0060 (  0.0%)  Parser
    0.0011 (  0.0%)    0.0011 (  0.0%)  ElementwiseToAffine
    0.1182 (  0.4%)    0.1182 (  0.5%)  PolynomialToStandard
    0.1586 (  0.6%)    0.1586 (  0.6%)  Canonicalizer
    0.0492 (  0.2%)    0.0492 (  0.2%)  CSE
    0.0000 (  0.0%)    0.0000 (  0.0%)    (A) DominanceInfo
    0.3331 (  1.3%)    0.3329 (  1.3%)  'func.func' Pipeline
    0.1313 (  0.5%)    0.1312 (  0.5%)    ConvertElementwiseToLinalgPass
    0.1132 (  0.4%)    0.1131 (  0.5%)    AffineExpandIndexOps
    0.0885 (  0.3%)    0.0885 (  0.4%)    SimplifyAffineStructures
    0.0954 (  0.4%)    0.0954 (  0.4%)  ConvertAffineToStandard
    0.2173 (  0.8%)    0.2171 (  0.9%)  'func.func' Pipeline
    0.0941 (  0.4%)    0.0940 (  0.4%)    ExpandOps
    0.1232 (  0.5%)    0.1231 (  0.5%)    ExpandStridedMetadata
   16.4575 ( 62.4%)   16.4575 ( 65.8%)  OneShotBufferize
    0.1061 (  0.4%)    0.1061 (  0.4%)  ExpandRealloc
    0.1572 (  0.6%)    0.1572 (  0.6%)  OwnershipBasedBufferDeallocation
    0.2453 (  0.9%)    0.2453 (  1.0%)  Canonicalizer
    0.8840 (  3.4%)    0.8840 (  3.5%)  BufferDeallocationSimplification
    0.1216 (  0.5%)    0.1216 (  0.5%)  LowerDeallocations
    0.1331 (  0.5%)    0.1331 (  0.5%)  CSE
    0.0000 (  0.0%)    0.0000 (  0.0%)    (A) DominanceInfo
    0.2199 (  0.8%)    0.2199 (  0.9%)  Canonicalizer
    0.4434 (  1.7%)    0.4433 (  1.8%)  'func.func' Pipeline
    0.4434 (  1.7%)    0.4433 (  1.8%)    ConvertLinalgToLoopsPass
    0.0813 (  0.3%)    0.0813 (  0.3%)  ConvertAffineToStandard
    0.0818 (  0.3%)    0.0818 (  0.3%)  ConvertBufferizationToMemRef
    0.1679 (  0.6%)    0.1679 (  0.7%)  Canonicalizer
    0.1819 (  0.7%)    0.1819 (  0.7%)  SCCP
    0.0471 (  0.2%)    0.0471 (  0.2%)  CSE
    0.0000 (  0.0%)    0.0000 (  0.0%)    (A) DominanceInfo
    0.1121 (  0.4%)    0.1121 (  0.4%)  SymbolDCE
    0.1036 (  0.4%)    0.1036 (  0.4%)  ArithExpandOpsPass
    0.6725 (  2.5%)    0.6725 (  2.7%)  SCFToControlFlow
    0.3854 (  1.5%)    0.3853 (  1.5%)  'func.func' Pipeline
    0.3854 (  1.5%)    0.3853 (  1.5%)    ExpandStridedMetadata
    0.7060 (  2.7%)    0.7060 (  2.8%)  ConvertToLLVMPass
    0.7442 (  2.8%)    0.7442 (  3.0%)  Canonicalizer
    0.4704 (  1.8%)    0.4704 (  1.9%)  SCCP
    0.2397 (  0.9%)    0.2397 (  1.0%)  CSE
    0.0000 (  0.0%)    0.0000 (  0.0%)    (A) DominanceInfo
    0.1932 (  0.7%)    0.1932 (  0.8%)  SymbolDCE
    1.0333 (  3.9%)    1.0333 (  4.1%)  Output
    0.0131 (  0.0%)    0.0131 (  0.1%)  Rest
   26.3794 (100.0%)   25.0003 (100.0%)  Total
```

I've never seen `OneShotBufferize` take so long.

j2kun:
Here's the mlir that is slow input to OneShotBufferize. Looks like it's ~5k single-op linalg.generic ops. https://gist.github.com/j2kun/af8ac4fbfd7df989d92fa5414c544db1

j2kun:
I suspect the main problem is that the IR after `polynomial-to-standard` consists largely of single-element tensor ops (see below), which get lowered to linalg

```
%449 = arith.muli %from_elements_413, %cst_15 : tensor<1xi40>
%450 = arith.remui %449, %cst_14 : tensor<1xi40>
%451 = arith.addi %from_elements_411, %450 : tensor<1xi40>
%452 = arith.remui %451, %cst_14 : tensor<1xi40>
%453 = arith.subi %from_elements_411, %450 : tensor<1xi40>
%454 = arith.addi %453, %cst_14 : tensor<1xi40>
%455 = arith.remui %454, %cst_14 : tensor<1xi40>
```

Can we just change these to scalar ops? Either in the original lowering for ntt, or else in a pass that identifies `tensor<1xfoo>` and replaces it with scalar ops? I feel like there should be an upstream pass that can do this, but I'm not able to find one quickly.

inbelic:
That is a very handy flag!

That seems promising, thanks for taking a look. I will test out with the scalar ops and also to see what manual fusion that we could do (also for the larger sized tensors). Doing the modulo arithmetic should be straightforward.

Checking I understand correctly, when you say to use an `affine.for`, that would be instead of the current recursive algorithm right?

j2kun:
> Checking I understand correctly, when you say to use an `affine.for`, that would be instead of the current recursive algorithm right?

In reference to your comment https://github.com/google/heir/pull/590#issuecomment-2054587006, " the recursive function fastNTT function is unrolled out already. It might be beneficial to make this into an SCF for loop"

I'm certain that making it a loop (in any form) will alleviate the slow compile time issues. I'm not sure what sort of loop would be reasonable for this. The best would be affine.for, since that would preserve affine loop nests when lowering through linalg. I haven't thought too hard about how bad it would be to use scf.while, since I'm not aware of an upstream loop unroller for scf (ideally we could just use whatever loop we want, apply bufferization, and then unroll to get what we have now back).

inbelic:
Yep, then we are on the same page. I will give it a go to prioritize using `affine.for`

inbelic:
This pull request is continued in #642 with an implementation using `affine.for` loops.

Issue #586, opened 2024-04-02
Determine how to generally identify and handle plaintext-ciphertext ops in `insert-rotate`


j2kun:
Will add a more detailed description soon...

AlexanderViand-Intel:
👍 Isn't the classification just "is the element type of the tensor `secret.secret`"? If that's the case, it might not be worth adding a whole pass/analysis and instead bneing able to provide an "isOperandRestricted" callback might be sufficient?

PR #601, opened 2024-04-06
Replace canonicalize with post-unroll-simplify

j2kun:
This new pass gives an empty set of patterns to the greedy pattern rewrite engine, which ends up applying each op's folding routine, which is enough to enable us to handle all examples in heir_simd_vectorizer

Avoids the slowdown mentioned in https://github.com/google/heir/issues/586


j2kun:
Looks like the change resulted in two additional rotations being added to the box_blur_64x64 IR, so I will investigate that on Monday

j2kun:
I didn't find the source of the additional inserted rotations, but instead added some extra tensor_ext canonicalization patterns that restored the original behavior. Also included some minor cleanup discovered along the way.

asraa (approved):
Nice fix! Is there any downside in just running `(void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));` with empty patterns in InsertRotate before adding the other patterns and running the actual pass? (This would avoid the pass boilerplate, and other passes that need this as a pre-pass can do the same thing)

j2kun:
> Nice fix! Is there any downside in just running `(void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));` with empty patterns in InsertRotate before adding the other patterns and running the actual pass? (This would avoid the pass boilerplate, and other passes that need this as a pre-pass can do the same thing)

The reason I didn't do this is because I wanted to try putting `canonicalize` in between this new pass and `insert-rotate`. I suppose I could put it at the end of the loop unroll pass? I just think having it more visible will make it less surprising.

asraa:
> I wanted to try putting canonicalize in between this new pass and insert-rotate. I suppose I could put it at the end of the loop unroll pass?

Ohhh, I see. Nah, putting it at the end of loop unroll seems like it'd be a surprise, I'd expect it to be applied as a pre-condition in the pass it's needed.

asraa (approved):

Issue #250, opened 2023-11-09
yosys: extend yosys optimization pass to support higher bit-width ops


j2kun:
E.g., turn the tfhe-rs API into black box ops, and run it through yosys/abc, improving upon our simpler lut-based approach.

PR #607, opened 2024-04-09
Dynamically find LUT size in comb-to-cggi

WoutLegiest:
A starter for dynammically finding the LUT size in the comb-to-cggi conversion
For now, once its detect a TruthTableOp, it goes to the default of 3

(Partial) Solve for ToDo #250


j2kun:
LGTM! Please add a lit test that exercises this behavior for a lut size != 3

j2kun (approved):

PR #624, opened 2024-04-17
Secret to BGV to OpenFHE e2e pipeline

j2kun:
This PR runs an end-to-end example (`simple_sum`) through the `heir-simd-vectorizer` pipeline, with the new `bgv-add-client-interface`, down to OpenFHE codegen, and runs the generated code in a test.

Deficiencies (still need to file TODOs):

- OpenFHE's `MakePackedPlaintext` will zero-pad to the ring degree `N`, and this breaks the rotate-and-reduce trick, so instead I have to manually repeat the coefficients cyclically until it fills the whole ciphertext. I'd like to add this to the encode helper, and it suggests a bunch more work to do if we have an input cleartext tensor that is not divisible by the OpenFHE's ring degree (rotate-and-reduce tricks won't work in that case, or will have to be tweaked to account for the true underlying ring degree).
- Support non-tensor input types to the encode codegen
- move `int64_t` casting to a higher-level lowering (say, from BGV to OpenFHE)

The output code generated for simple_sum

```cpp

#include "src/pke/include/openfhe.h" // from @openfhe

using namespace lbcrypto;
using CiphertextT = ConstCiphertext<DCRTPoly>;
using CryptoContextT = CryptoContext<DCRTPoly>;
using EvalKeyT = EvalKey<DCRTPoly>;
using PlaintextT = Plaintext;
using PrivateKeyT = PrivateKey<DCRTPoly>;
using PublicKeyT = PublicKey<DCRTPoly>;

CiphertextT simple_sum(CryptoContextT v0, CiphertextT v1) {
  size_t v2 = 1;
  size_t v3 = 2;
  size_t v4 = 4;
  size_t v5 = 8;
  size_t v6 = 16;
  size_t v7 = 31;
  int64_t v8 = static_cast<int64_t>(v6);
  const auto& v9 = v0->EvalRotate(v1, v8);
  const auto& v10 = v0->EvalAdd(v1, v9);
  int64_t v11 = static_cast<int64_t>(v5);
  const auto& v12 = v0->EvalRotate(v10, v11);
  const auto& v13 = v0->EvalAdd(v10, v12);
  int64_t v14 = static_cast<int64_t>(v4);
  const auto& v15 = v0->EvalRotate(v13, v14);
  const auto& v16 = v0->EvalAdd(v13, v15);
  int64_t v17 = static_cast<int64_t>(v3);
  const auto& v18 = v0->EvalRotate(v16, v17);
  const auto& v19 = v0->EvalAdd(v16, v18);
  int64_t v20 = static_cast<int64_t>(v2);
  const auto& v21 = v0->EvalRotate(v19, v20);
  const auto& v22 = v0->EvalAdd(v19, v21);
  std::vector<int16_t> v23 = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1};
  std::vector<int64_t> v23_cast(std::begin(v23), std::end(v23));
  const auto& v24 = v0->MakePackedPlaintext(v23_cast);
  const auto& v25 = v0->EvalMult(v22, v24);
  int64_t v26 = static_cast<int64_t>(v7);
  const auto& v27 = v0->EvalRotate(v25, v26);
  const auto& v28 = v27;
  return v28;
}

CiphertextT simple_sum__encrypt(CryptoContextT v29, std::vector<int16_t> v30, PublicKeyT v31) {
  std::vector<int64_t> v30_cast(std::begin(v30), std::end(v30));
  const auto& v32 = v29->MakePackedPlaintext(v30_cast);
  const auto& v33 = v29->Encrypt(v31, v32);
  return v33;
}

int16_t simple_sum__decrypt(CryptoContextT v34, CiphertextT v35, PrivateKeyT v36) {
  PlaintextT v37;
  v34->Decrypt(v36, v35, &v37);
  int16_t v38 = v37->GetPackedValue()[0];
  return v38;
}
```

### DIFF
--- a/lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
+++ b/lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
@@ -1,6 +1,7 @@
 #include "include/Target/OpenFhePke/OpenFhePkeEmitter.h"
 
 #include <functional>
+#include <numeric>
 #include <string_view>
 
 #include "include/Analysis/SelectVariableNames/SelectVariableNames.h"
@@ -40,10 +41,10 @@ void registerToOpenFhePkeTranslation() {
         return translateToOpenFhePke(op, output);
       },
       [](DialectRegistry &registry) {
-        registry
-            .insert<func::FuncDialect, openfhe::OpenfheDialect, lwe::LWEDialect,
-                    ::mlir::heir::polynomial::PolynomialDialect,
-                    arith::ArithDialect, tensor::TensorDialect>();
+        registry.insert<arith::ArithDialect, func::FuncDialect,
+                        openfhe::OpenfheDialect, lwe::LWEDialect,
+                        ::mlir::heir::polynomial::PolynomialDialect,
+                        arith::ArithDialect, tensor::TensorDialect>();
       });
 }
 
@@ -63,11 +64,17 @@ LogicalResult OpenFhePkeEmitter::translate(Operation &op) {
           .Case<func::FuncOp, func::ReturnOp>(
               [&](auto op) { return printOperation(op); })
           // Arith ops
-          .Case<arith::ConstantOp>([&](auto op) { return printOperation(op); })
+          .Case<arith::ConstantOp, arith::IndexCastOp>(
+              [&](auto op) { return printOperation(op); })
+          // LWE ops
+          .Case<lwe::RLWEEncodeOp, lwe::RLWEDecodeOp,
+                lwe::ReinterpretUnderlyingTypeOp>(
+              [&](auto op) { return printOperation(op); })
           // OpenFHE ops
           .Case<AddOp, SubOp, MulOp, MulPlainOp, SquareOp, NegateOp, MulConstOp,
                 RelinOp, ModReduceOp, LevelReduceOp, RotOp, AutomorphOp,
-                KeySwitchOp>([&](auto op) { return printOperation(op); })
+                KeySwitchOp, EncryptOp, DecryptOp>(
+              [&](auto op) { return printOperation(op); })
           .Default([&](Operation &) {
             return op.emitOpError("unable to find printer for op");
           });
@@ -117,8 +124,8 @@ LogicalResult OpenFhePkeEmitter::printOperation(func::FuncOp funcOp) {
   }
 
   os << commaSeparatedValues(funcOp.getArguments(), [&](Value value) {
-    auto res = convertType(value.getType());
-    return res.value() + " " + variableNames->getNameForValue(value);
+    return convertType(value.getType()).value() + " " +
+           variableNames->getNameForValue(value);
   });
   os.unindent();
   os << ") {\n";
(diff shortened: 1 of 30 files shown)


j2kun:
OK this monster PR is working now, EXCEPT for one tiny detail: I can't figure out how to get the new `//tests/openfhe/end_to_end:simple_sum_test` to run with yosys enabled by default. It seems to run `heir-opt` with yosys disabled by default, and to make it pass I have to run `bazel test` with `--//:enable_yosys=0`

tests/openfhe/end_to_end/simple_sum.mlir
```diff
@@ -0,0 +1,11 @@
+// heir-opt --mlir-to-openfhe-bgv='entry-function=simple_sum ciphertext-degree=32' %s | bazel-bin/tools/heir-translate --emit-openfhe-pke
```
asraa, line 1: This seems like a documentation (and not a RUN line for a lit test) since the build target contains the flag info and heir-translate commands. In case the flags change (which I suspect ciphertext-degree may at some point be changed), could you add a comment saying something like "the build target executes the following"? At first glance I assumed this was the RUN line.

tools/heir-opt.bzl
```diff
@@ -0,0 +1,54 @@
+"""A rule for running heir-opt."""
```
asraa, line 1: nice!

tests/openfhe/end_to_end/simple_sum.mlir
j2kun: +1 yeah I left it there as documentation

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
```diff
@@ -147,14 +154,22 @@ LogicalResult OpenFhePkeEmitter::printOperation(func::ReturnOp op) {
   return success();
 }
 
-void OpenFhePkeEmitter::emitAssignPrefix(Value result) {
-  os << "auto " << variableNames->getNameForValue(result) << " = ";
+void OpenFhePkeEmitter::emitAutoAssignPrefix(Value result) {
+  os << "const auto& " << variableNames->getNameForValue(result) << " = ";
```
asraa, line 158: are there any return types that you are expecting to always be const references and not pointers? i'm wondering if there should be an assertion on the value type that it's a ciphertext type etc, or are we safe here with always using references?

tests/openfhe/end_to_end/simple_sum.mlir
j2kun: Fixed in 30cdb032

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
j2kun: Added in 235eca8c

asraa (approved):
Thank you for all the fixes to the config settings and adding the compilation flag!

j2kun:
Squashed and ready to pull

Issue #629, opened 2024-04-18
Add design section to documentation


j2kun:
Add a `Design` section to the documentation website with the following pages.

The purpose of this would be to ramp up someone who already knows a decent amount of MLIR, and a decent amount of FHE, but is not familiar with our particular design choices, or might want to slice and dice our passes for their own purposes.

- [x] secret
- [ ] tfhe/cggi pipeline
- [x] "HECO" pipeline
- [ ] polynomial abstraction/math
- [x] dialect taxonomy/diagram
- [ ] exit/entry dialects

asraa:
I can start a Design section on secret today since I'm working on secret.separator at the moment. It'll just be a start / placeholder and we can add more as we go

j2kun:
I'm going to start on the `secret` docs because it's what I have the energy to do right now.

j2kun:
This issue has become stale, and I think most of our documentation work in the near future will go into paper writing. Closing for now.

PR #630, opened 2024-04-19
Add secret design doc

j2kun:
Part of https://github.com/google/heir/issues/629

### DIFF
--- a/docs/content/en/docs/Design/secret.md
+++ b/docs/content/en/docs/Design/secret.md
@@ -1,0 +1,322 @@
+<!-- mdformat off(yaml frontmatter) -->
+---
+title: Secret
+weight: 9
+---
+<!-- mdformat on -->
+
+The [`secret` dialect](https://heir.dev/docs/dialects/secret/) contains types
+and operations to represent generic computations on secret data. It is intended
+to be a high-level entry point for the HEIR compiler, agnostic of any particular
+FHE scheme.
+
+Most prior FHE compiler projects design their IR around a specific FHE scheme,
+and provide dedicated IR types for the secret analogues of existing data types,
+and/or dedicated operations on secret data types. For example, the Concrete
+compiler has `!FHE.eint<32>` for an encrypted 32-bit integer, and `add_eint` and
+similar ops. HECO has `!fhe.secret<T>` that models a generic secret type, but
+similarly defines `fhe.add` and `fhe.multiply`, and other projects are similar.
+
+The problem with this approach is that it is difficult to incorporate the apply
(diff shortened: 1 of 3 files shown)


docs/content/en/docs/Design/secret.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+The canonicalization patterns for `secret.generic` apply a variety of simplifications, such as:
+
+- Removing any unused or non-secret arguments and return values.
+- Hoisting operations in the body of a `generic` that only depend on cleartext values to the enclosing scope.
+- Removing any `generic` ops that use no secrets at all.
+
+These can be used together with the [`secret-distribute-generic`
```
asraa, line 112: It might be worth mentioning that the reason we do split to a single op is that it ends up performing secret type prorogation to the individual ops contained in larger generics. (Removing the need to perform dataflow to figure out whether an op operand contained within a large generic block has a secret type or not)

asraa: If you wanted, you could add a mul_plain operation as example in the snippet below

docs/content/en/docs/Design/secret.md
```diff
@@ -0,0 +1,249 @@
...
+To accomplish this, we have two additional mechanisms.
+One is the pass option `ops-to-distribute` for `-secret-distribute-generic`,
+which allows the user to specify a list of ops that `generic` should be split across,
+and all others left alone. Specifying `affine.for` here will pass `generic` through
+the `affine.for` loop, but leave its body intact.
+This can also be used with the `-unroll-factor` option to the `-yosys-optimizer` pass
+to partially unroll a loop nest and pass the partially-unrolled body through
+the circuit optimizer.
+
+The other mechanism is the `secret.separator` op,
```
asraa, line 221: Another key reason: this allows us to insert boundaries between higher-level units of the program, before lowering those down and requiring an op separator belonging to a lower level dialect. We can save this for the TOSA example.

asraa: Ah you mention it below!

asraa (approved):
Super clear! (At least to me.) Perhaps we want a review from a less familiar reader :)

docs/content/en/docs/Design/secret.md
j2kun: Done.

j2kun:
@AlexanderViand-Intel do you want to read the `secret` doc and see if matches what you had in mind for the level of detail and justification?

AlexanderViand-Intel:
I've finally had a look and I really like the level of detail/abstraction!

Issue #542, opened 2024-03-20
Write canonicalization patterns for NTT/INTT


j2kun:
In particular, immediate ntt/intt pairs can be elided. Depends on https://github.com/google/heir/issues/182

Maokami:
I've created a simple canonicalization pass: `NTT(INTT(x)) => x` and `INTT(NTT(x)) => x`.

Now, I started thinking that it would be good to have the following optimization pass as well:
`INTT(Op(NTT(x1), NTT(x2))) => Op(x1, x2)` where `Op = polynomial.add/sub/mul_scalar` (and the version with NTT and INTT switched.)

Should this be included in the canonicalization pass, or should I create a new pass for it? 😃

j2kun:
I think this second pattern will be a bit more complicated, since Op will be a tensor op in one domain and a polynomial op in the other, and it is context dependent which is cheaper (since it depends on what other neighboring ops are doing, if they are in the coefficient or NTT domain)

j2kun:
I think the right approach to figure out what to do next is:

1. Add a pass that actually introduces ntt/intt ops by replacing polynomial.mul with ntt + tensor mul + intt.
2. Analyze some natural programs involving polynomial multiplications to look for optimization opportunities. Ideally we can do this by lowering some high level programs from BGV to polynomial, but that will involve a lot of extra work to actually implement bgv operations.

That or, comb through the HEaaN.mlir paper looking for optimizations related to this: https://dl.acm.org/doi/pdf/10.1145/3591228

Maokami:
> I think this second pattern will be a bit more complicated, since Op will be a tensor op in one domain and a polynomial op in the other, and it is context dependent which is cheaper (since it depends on what other neighboring ops are doing, if they are in the coefficient or NTT domain)

Ah, you are right! This requires considering much more complex factors than I thought.

I'll tackle the task 1(making a (i)ntt introducing pass) and then skim through the HEaaN.mlir optimizations. 👍

j2kun:
I filed a separate issue for HEaaN.mlir, and we can limit this issue to the basic canonicalization patterns: https://github.com/google/heir/issues/635

I think loop fusion would be a good place to start. CC @inbelic since he is working on the initial ntt lowering.

inbelic:
Sweet. Feel free to ask me any questions on how it is implemented.

[...]

PR #631, opened 2024-04-19
Canonicalization patterns for NTT/INTT

Maokami:
This is a canonicalization pass that transforms `ntt(intt(x))` and `ntt(intt(x))` into `x`.


Maokami:
#542

PR #636, opened 2024-04-21
unit-test 2nth-root-of-unity generator

j2kun:
@inbelic feel free to just patch this commit into your existing PR. Note I did change some names slightly, and added an assertion because some tests failed if the input was not properly modded.

### DIFF
--- a/include/Dialect/Polynomial/IR/NumberTheory.h
+++ b/include/Dialect/Polynomial/IR/NumberTheory.h
@@ -1,0 +1,15 @@
+#include "llvm/include/llvm/ADT/APInt.h"  // from @llvm-project
+
+namespace mlir::heir::polynomial {
+
+/// Returns true if `root` is a primitive nth root modulo `cmod`.
+bool isPrimitiveNthRootOfUnity(const ::llvm::APInt &root, unsigned n,
+                               const ::llvm::APInt &cmod);
+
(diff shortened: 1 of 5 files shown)


include/Dialect/Polynomial/IR/NumberTheory.h
```diff
@@ -0,0 +1,15 @@
+#include "llvm/include/llvm/ADT/APInt.h"  // from @llvm-project
+
+namespace mlir::heir::polynomial {
+
+/// Returns true if `root` is a primitive nth root modulo `cmod`.
+bool isPrimitiveNthRootOfUnity(const ::llvm::APInt &root, unsigned n,
+                               const ::llvm::APInt &cmod);
+
+/// Returns true if `root` is a primitive 2nth root modulo `cmod`.
+inline bool isPrimitive2nthRootOfUnity(const ::llvm::APInt &root, unsigned n,
```
j2kun, line 10: @inbelic TBH, this method might not even be necessary, just call it Nth root and rely on the user to pass in 2*n when needed.

inbelic:
Awesome, will do!

include/Dialect/Polynomial/IR/NumberTheory.h
inbelic: It should be straightforward enough to do so. I can modify that when I cherry-pick it

PR #642, opened 2024-04-24
poly: add initial lowering for NTT op and add root parameter to rings

### LINKED ISSUE #543: Write an initial lowering for NTT/INTT using a standard algorithm
Depends on https://github.com/google/heir/issues/182

I'm not an expert in NTT algorithms, but here are a few references:

- https://github.com/nict-sfl/FasterNTT
- [Number Theoretic Transform and Its Applications in Lattice-based Cryptosystems: A Survey](https://arxiv.org/abs/2211.13546)
- [Speeding up the Number Theoretic Transform for Faster Ideal Lattice-Based Cryptography](https://eprint.iacr.org/2016/504)

inbelic:
- add a standard iterative implementation of the NTT Op
    - add testing for lowering and a simple case runner
    - add performance test that uses random coefficients to prevent
      optimizations on dense values that artifically improves
      compile/runtime

   - include an optional parameter that computes the 2n-th primitive
      root of a rings ideal degree and its coefficient modulus
   - allow for this to be retrieved from a pre-computed or to be
     user-specified
   - Add a script that will compute the 2n-th roots of unity for the
     specified polynomial degree values and n. The values are only
     outputted if the 2n-th root of unity exists.

  Resolve #543


inbelic:
@j2kun Ready for review.

Sorry for letting it grow so much. I should have created a first pull request with the changes adding the `root` parameter. Will try to reduce the size of commits in the future.

Hmm, I am not able to reproduce the build failure locally. Can take another look when I am more awake.

include/Dialect/Polynomial/IR/NumberTheory.h
```diff
@@ -0,0 +1,11 @@
+#include "llvm/include/llvm/ADT/APInt.h"  // from @llvm-project
+
+namespace mlir::heir::polynomial {
+
+/// Returns true if `root` is a primitive nth root modulo `cmod`.
+///
+/// Check if `root` is a 2n-th primitive root modulo `cmod` with `n` = 2*n.
```
j2kun, line 7:
```suggestion
/// To test for a 2k-th primitive root, set `n` = 2*k.
```

lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
```diff
@@ -172,10 +178,37 @@ mlir::Attribute mlir::heir::polynomial::RingAttr::parse(AsmParser &parser,
...
+
+    auto result = parser.parseInteger(root);
+    if (failed(result) || cmod.ule(root.zext(cmod.getBitWidth())) ||
+        !isPrimitiveNthRootOfUnity(root, 2 * poly.getDegree(), cmod)) {
+      parser.emitError(parser.getCurrentLocation(),
+                       "Invalid 2n-th primitive root of unity.");
+      return {};
+    }
+  } else {
+    auto maybeRoot =
```
j2kun, line 199:
```suggestion
    // TODO(#643): replace with a pass that computes roots as needed
    auto maybeRoot =
```

j2kun: I wrote a GitHub action that will link this line to the GH issue.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +803,263 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+  SmallVector<utils::IteratorType> iteratorTypes(1,
+                                                 utils::IteratorType::parallel);
+  AffineExpr d0;
+  bindDims(b.getContext(), d0);
+  SmallVector<AffineMap> indexingMaps = {AffineMap::get(1, 0, {d0}),
+                                         AffineMap::get(1, 0, {d0})};
+  auto out = b.create<arith::ConstantOp>(
+      type,
+      DenseElementsAttr::get(type, APInt(type.getElementTypeBitWidth(), 0L)));
+  auto shuffleOp = b.create<linalg::GenericOp>(
```
j2kun, line 858: I don't have a ton of experience with this, but I _think_ you should be able to replace this with a `tensor.gather` op, using the indices created above as the index argument.

Feel free to leave as-is, but using a dedicated op could make this code simpler.

j2kun:
> Hmm, I am not able to reproduce the build failure locally.

I also cannot reproduce the failure locally. I'll take a look at the CI.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +803,263 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+    auto polyTy = dyn_cast<PolynomialType>(op.getInput().getType());
+    if (!polyTy) {
+      op.emitError(
+          "Can't directly lower for a tensor of polynomials. "
+          "It is expected that it should be lowered to be done elementwise.");
+      return failure();
+    }
+
+    RingAttr ring = polyTy.getRing();
+    auto inputType = convertPolynomialType(polyTy);
```
j2kun, line 894: The converted type should already be present on the `adaptor.getInput().getType()`; the dialect conversion framework uses whatever type converter is provided by this pass to convert the types of all operands and results. The `op` argument to this function contains the original, unconverted operands.

j2kun: Two lines below, I think this means you would also not need the `to_tensor` op at all.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +803,263 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+    //  end
+
+    // Convert the input into Reverse-Bit Order
+    initialValue = computeReverseBitOrder(b, intermediateType, initialValue);
+
+    // Initialize the variables
+    Value initialBatchSize = b.create<arith::ConstantOp>(
+        IndexType::get(b.getContext()), b.getIndexAttr(2));
+
+    Value zero = b.create<arith::ConstantOp>(IndexType::get(b.getContext()),
```
j2kun, line 957: You can make these declarations slightly cleaner with the index-specialized helper

```
builder.create<arith::ConstantIndexOp>(loc, X)
```

Where `X` is the desired index value.

j2kun: There are similar things elsewhere for ConstantIntOp

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +803,263 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+      Value rootExp = b.create<affine::AffineApplyOp>(x.floorDiv(y),
+                                                      ValueRange{n, batchSize});
+
+      auto stagesIP = b.saveInsertionPoint();
+      auto innerLoop = b.create<affine::AffineForOp>(
+          /*lbOperands=*/zero, /*lbMap=*/AffineMap::get(1, 0, x),
+          /*ubOperands=*/rootExp, /*ubMap=*/AffineMap::get(1, 0, x),
+          /*step=*/1, /*iterArgs=*/target);
+      b.setInsertionPointToStart(innerLoop.getBody());
+      {
```
j2kun, line 988: You can avoid the insertion point management and manual nesting by using the `BodyBuilderFn` builder, see for example

https://github.com/llvm/llvm-project/blob/7b06ec073d0bd8abad1274d12900a849bd3829c5/mlir/lib/Dialect/Affine/Transforms/SuperVectorize.cpp#L1357-L1360

You should just be able to give the last argument of `create` as

```
[&](OpBuilder &, Location, Value, ValueRange) {
  ...
}
```

And replace `...` with all the code below.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -802,6 +803,263 @@ func::FuncOp PolynomialToStandard::buildPolynomialModFunc(FunctionType funcType,
...
+        Value target = innerLoop.getRegionIterArgs()[0];
+        Value indexK = b.create<affine::AffineApplyOp>(
+            x * y, ValueRange{batchSize, innerLoop.getInductionVar()});
+
+        auto innerIP = b.saveInsertionPoint();
+        auto arithLoop = b.create<affine::AffineForOp>(
+            /*lbOperands=*/zero, /*lbMap=*/AffineMap::get(1, 0, x),
+            /*ubOperands=*/batchSize,
+            /*ubMap=*/AffineMap::get(1, 0, x.floorDiv(2)),
```
j2kun, line 997: TIL you can use floorDiv in affine maps

j2kun (approved):
LGTM! Just a few optional improvements. When you're ready please manually squash your commits and I'll get it merged.

j2kun:
My guess is that this will fix the build failure:

```
diff --git a/lib/Dialect/Polynomial/IR/BUILD b/lib/Dialect/Polynomial/IR/BUILD
index 32666254..ac791511 100644
--- a/lib/Dialect/Polynomial/IR/BUILD
+++ b/lib/Dialect/Polynomial/IR/BUILD
@@ -38,12 +38,12 @@ cc_library(
         "PolynomialAttributes.cpp",
     ],
     hdrs = [
-        "@heir//include/Dialect/Polynomial/IR:NumberTheory.h",
         "@heir//include/Dialect/Polynomial/IR:PolynomialAttributes.h",
         "@heir//include/Dialect/Polynomial/IR:PolynomialDialect.h",
         "@heir//include/Dialect/Polynomial/IR:StaticRoots.h",
     ],
     deps = [
+        ":NumberTheory",
         ":Polynomial",
         "@heir//include/Dialect/Polynomial/IR:attributes_inc_gen",
         "@heir//include/Dialect/Polynomial/IR:dialect_inc_gen",
```

It's weird that I can't reproduce it locally, but I think it may be because I'm using clang + lld and the CI is using gcc + ld. I don't have time to do a detailed test/fix, but just based on my intuition this is the problem: you have the header there but no dependency on the implementation directly on the `PolynomialAttributes` target.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
inbelic: Unfortunately when trying to use `gather` it requires the output tensor to be strictly a subset of the source.

inbelic:
That fixed the issue. I had accidentally added the dep to the `PolynomialOp` build rule instead of `PolynomialAttributes`.

j2kun (approved):

PR #657, opened 2024-04-29
poly: add pass to convert poly.mul to use ntt ops

### LINKED ISSUE #541: Add a pass that converts polynomial multiplication to NTT/INTT + arith.muli
Cf. https://github.com/google/heir/issues/232#issuecomment-2009684084 for context. Depends on https://github.com/google/heir/issues/182

inbelic:
- We can reduce complexity of polynomial multiplication by using the number-theoretic transform

  - Here we add a pass that will convert all poly.mul operations into a poly.ntt of both operands -> modulus multiplication of the coefficients -> poly.intt back into the polynomial type

Resolves #541

### DIFF
--- a/lib/Dialect/Polynomial/Transforms/NTTRewrites.td
+++ b/lib/Dialect/Polynomial/Transforms/NTTRewrites.td
@@ -1,0 +1,65 @@
+#ifndef LIB_DIALECT_POLYNOMIAL_TRANSFORMS_NTTREWRITES_TD_
+#define LIB_DIALECT_POLYNOMIAL_TRANSFORMS_NTTREWRITES_TD_
+
+include "lib/DRR/Utils.td"
+include "lib/Dialect/Polynomial/IR/PolynomialOps.td"
+include "mlir/Dialect/Arith/IR/ArithOps.td"
+include "mlir/IR/PatternBase.td"
+
+def GetRingAttr : NativeCodeCall<
+      "(dyn_cast<PolynomialType>($0.getType())).getRing()">;
+
+def InputTensorType : NativeCodeCall<
+      "RankedTensorType::get({$0.ideal().getDegree()},"
+      " IntegerType::get($_builder.getContext(),"
+      "  ($0.coefficientModulus() - 1).getActiveBits()),"
+      " $0)">;
+
+def IntermediateTensorType : NativeCodeCall<
+      "RankedTensorType::get({$0.ideal().getDegree()},"
+      " IntegerType::get($_builder.getContext(),"
+      "  2 * ($0.coefficientModulus() - 1).getActiveBits()),"
+      " $0)">;
+
+def CreateCModConstant : NativeCodeCall<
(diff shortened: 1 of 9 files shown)


lib/Dialect/Polynomial/Transforms/Passes.td
```diff
@@ -0,0 +1,20 @@
+#ifndef LIB_DIALECT_POLYNOMIAL_TRANSFORMS_PASSES_TD_
+#define LIB_DIALECT_POLYNOMIAL_TRANSFORMS_PASSES_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def NTTRewrites : Pass<"ntt-rewrites"> {
+  let summary = "Rewrites polynomial operations to their NTT equivalents";
+  let description = [{
+    Applies a set of rewrites to convert polynomial operations to their
```
j2kun, line 9: Are there other rewrites of polynomial operations in terms of NTT? Maybe you're thinking of how you can do a convolution of two vectors in a coefficient encoding using NTT?

If there are not others, let's just name this pass more directly what it is, something like `convert-polynomial-mul-to-ntt`. You can still keep the patterns in `NTTRewrites.td`, and just add the `NTTRewritePolyMul` pattern by name to the pattern set instead of adding all generated patterns.

My reasoning: `ntt-rewrites` is too vague. I would guess it means "optimizing existing NTT ops." Also the handling of poly mul is somehow a very distinguished part of FHE, so I want it to be easy to find :)

lib/Dialect/Polynomial/Transforms/NTTRewrites.td
```diff
@@ -0,0 +1,68 @@
...
+      " $0)">;
+
+def CreateCModConstant : NativeCodeCall<
+      "$_builder.create<arith::ConstantOp>($0.getLoc(), $2,"
+      " DenseElementsAttr::get($2,"
+      "  $1.coefficientModulus().sextOrTrunc($2.getElementTypeBitWidth())))">;
+
+// Computes (n & (n - 1)) == 0. Can only pass in symbols to the DRR constraint
+// so unfortunately need to extract the ring and degree each time...
+def HasDegreePowerOfTwo : Constraint<
```
j2kun, line 31: Might be slightly simpler to wrap it in an APInt and call `APInt::isPowerOf2`, at least since you know the degree has <= 64 bits of precision.

j2kun (approved):
Great work!

j2kun:
Also please squash the commits when you're ready. Thanks!

lib/Dialect/Polynomial/Transforms/Passes.td
inbelic: I agree with the reasoning to be more specific.

I had left it vague to allow for other NTT-related optimizations to also be applied in this pass.

inbelic:
Great thanks. I will be AFK until the 15th, but can fix it up when I have my laptop again. Otherwise, feel free to apply your suggestions and merge. Sorry for the delay

j2kun (approved):

PR #662, opened 2024-04-29
Add an end-to-end BGV/OpenFHE test for dot product

j2kun:
To add an e2e test for dot product, this PR modifies `bgv-add-client-interface` to create separate encryption/decryption functions for each input/output value of the compiled function. Otherwise the codegen would have to be modified to have multiple return values via outparams, which I think is more fuss than it's worth.

### DIFF
--- a/lib/Dialect/BGV/Transforms/AddClientInterface.cpp
+++ b/lib/Dialect/BGV/Transforms/AddClientInterface.cpp
@@ -26,90 +26,50 @@ namespace bgv {
 #define GEN_PASS_DEF_ADDCLIENTINTERFACE
 #include "lib/Dialect/BGV/Transforms/Passes.h.inc"
 
-/// Adds the client interface for a single func. This should only be used on the
-/// "entry" func for the IR being compiled, but there may be multiple.
-LogicalResult convertFunc(func::FuncOp op, bool usePublicKey) {
-  auto module = op->getParentOfType<ModuleOp>();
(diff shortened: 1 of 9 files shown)


lib/Dialect/BGV/Transforms/AddClientInterface.cpp
```diff
@@ -26,48 +26,30 @@ namespace bgv {
...
       }
       rlweParams = argCtTy.getRlweParams();
-      continue;
     }
-
-    // For plaintext arguments, the function is a no-op
-    encFuncArgTypes.push_back(argTy);
-    encFuncResultTypes.push_back(argTy);
   }
+  return rlweParams;
```
asraa, line 42: In case an RLWE params isn't found, does nullptr actually create a failure logicalresult? i think here an attribute can generally be a nullptr, and if it is, then there might be a segfault later.

asraa: it looks like you catch a nullptr rlweParams on line 53 for the encryption func but not the decryption func, but you also check for the logical result failure.

asraa (approved):
nice! just a very minor comment

lib/Dialect/BGV/Transforms/AddClientInterface.cpp
j2kun: Good call. Fixed!

Issue #256, opened 2023-11-16
poly: upstream `polynomial` dialect


j2kun:
- [x] attributes and types: https://github.com/llvm/llvm-project/pull/72081
- [x] basic ops: https://github.com/llvm/llvm-project/pull/89525
- [x] basic canonicalization patterns
- [ ] generic lowering to standard MLIR dialects
- [x] advanced ops (ntt)

PR #675, opened 2024-05-08
Replace HEIR polynomial dialect with upstream dialect

j2kun:
Fixes https://github.com/google/heir/issues/256

Pending

- https://github.com/llvm/llvm-project/pull/91137
- https://github.com/llvm/llvm-project/pull/91410
- https://github.com/llvm/llvm-project/pull/93227
- https://github.com/llvm/llvm-project/pull/93243
- https://github.com/llvm/llvm-project/pull/93368


j2kun:
Marking pull_ready so I can get an internal patch to test with sanitizers

j2kun:
Main remaining problem is that the openfhe end_to_end tests don't build. heir-translate is apparently not able to load the Polynomial dialect's attributes, despite the fact that it's properly loaded upstream.

```
LLVM ERROR: can't create Attribute 'mlir::polynomial::IntPolynomialAttr' because storage uniquer isn't initialized: the dialect was likely not loaded, or the attribute wasn't added with addAttributes<...>() in the Dialect::initialize() method.
```

j2kun:
All the fixes are in, and now we just need to wait until the upstream changes are integrated into HEIR, probably will happen on Monday.

tests/polynomial/runner/generate_test_cases.py
```diff
@@ -43,15 +43,15 @@ def test_setup(i: int):
 #  - expected
 #  - coefficient_list
 TEST_TEMPLATE = """
-#ideal_{{test_number}} = #polynomial.polynomial<{{ideal}}>
-#ring_{{test_number}} = #polynomial.ring<cmod={{cmod}}, ideal=#ideal_{{test_number}}>
-!poly_ty_{{test_number}} = !polynomial.polynomial<#ring_{{test_number}}>
+#ideal_{{test_number}} = #polynomial.int_polynomial<{{ideal}}>
+#ring_{{test_number}} = #polynomial.ring<coefficientType = i32, coefficientModulus={{cmod}} : {{cmod_type}}, polynomialModulus=#ideal_{{test_number}}>
```
asraa, line 47: Given that the coefficient type of the ring is hard-coded at i32, why do we need the `gen_tensor_op` sign extension/truncation ops? It seems like all the tests use i32, and I noticed that the coefficient types in the tests were updated to i32, so maybe the test configs don't need it?

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -4,17 +4,14 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <iostream>
```
asraa, line 7: nit: might have been left from debugging

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -265,32 +307,66 @@ struct ConvertMulScalar : public OpConversionPattern<MulScalarOp> {
...
+        finalOp = b.create<arith::RemSIOp>(adaptor.getPolynomial(), modValue);
+      }
+    }
+
+    rewriter.replaceOp(op, finalOp);
     return success();
   }
 };
 
+bool hasCyclicModulus(MonicMonomialMulOp op) {
```
asraa, line 336: nit: This is OK naming but could you add a quick docstring about the form of this polynomial since it doesn't quite line up exactly with the name? I think it qualifies as a cyclic polynomial (one that's invariant under symmetric group operations on the coefficients since it only has one degree) but it's also not a cyclotomic polynomial (since x^n - 1 may not be irreducible).

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
```diff
@@ -421,13 +497,30 @@ struct ConvertAdd : public OpConversionPattern<AddOp> {
...
+    if (!intCoeffType) {
+      op.emitError()
+          << "Unsupported coefficient type for lowering polynomial add";
+      return failure();
+    }
+    unsigned width = intCoeffType.getIntOrFloatBitWidth();
 
+    // coeffTypeMod is the value 2**width for an i<width>
+    APInt coeffTypeMod = APInt(width + 1, 0);
+    coeffTypeMod.setBit(width);
```
asraa, line 514: is this `intTypeModValue(coeffType)`?

lib/Dialect/Polynomial/Transforms/NTTRewrites.td
```diff
@@ -2,43 +2,45 @@
...
 
+def Nullptr
+  : NativeCodeCall<"nullptr">;
+
 def NTTRewritePolyMul : Pattern<
   (Polynomial_MulOp:$mulOp $p1, $p2),
   [
     // Transform to NTT point-value representation
-    (Polynomial_NTTOp:$p1NTT $p1,
+    (Polynomial_NTTOp:$p1NTT $p1, (Nullptr),
```
asraa, line 41: Is the `(Nullptr)` referring to a null root attribute? I actually didn't realize that the pattern match goes like (Op:$name, $input, $attr)

asraa (approved):
Thank you!!! Just a handful of small nits. So cool that it's on MLIR upstream now

tests/polynomial/runner/generate_test_cases.py
j2kun: I think it would be future proof for other coefficient types. I should add another test, but baby has arrived so I just want to get this one in :)

lib/Dialect/Polynomial/Transforms/NTTRewrites.td
j2kun: It only does that if the attribute is explicitly listed on the op, which it is upstream. It should just correspond exactly to the order of `ins` in the tablegen definition of the op.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: Good catch. Fixed.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: Removed.

lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp
j2kun: Done.

asraa (approved):
Thank you!!

j2kun:
We're still missing the last upstream PR that will fix the remaining test failure

PR #688, opened 2024-05-22
poly: add pass to materialize roots in rings

inbelic:
- Not all rings need a root attribute as they are not used in operations
    that require them. Currently, we attempt to materialize a root for all
    rings at parse time. The introduced pass will only attempt to
    materialize roots for rings that are used in such an operation. This
    allows for the materialization to be just-in-time.

  - Adds a pass that will first denote all rings that are used in
    operations that require a root. Then it does a replacement pass to
    recursively replace those rings.


inbelic:
Will need to rebase onto main once #657 is merged.

inbelic:
The roots attribute will be moved to the operation rather than the polynomial type. As such, this pass will need to modified to fit the new representation. See discord discussion.

PR #692, opened 2024-05-23
Abstract importer classes need an explicit virtual destructor

raghav198:
With Apple Clang 15.0.0, the implicitly generated destructors for `BooleanGateImporter`, `RTLILImporter`, and `LUTImporter` cause a run-time error when attempting to run any pipeline that uses them. Explicitly marking them virtual fixes this.

### DIFF
--- a/lib/Transforms/YosysOptimizer/BooleanGateImporter.h
+++ b/lib/Transforms/YosysOptimizer/BooleanGateImporter.h
@@ -21,6 +21,8 @@ class BooleanGateImporter : public RTLILImporter {
  public:
   BooleanGateImporter(MLIRContext *context) : RTLILImporter(context) {}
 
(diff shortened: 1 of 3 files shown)


j2kun (changes requested):
As I understand, the `virtual` keyword is only needed on the RTLILImporter base class, correct? Otherwise LGTM!

j2kun (approved):

raghav198:
@j2kun I finally got the CLA stuff taken care of, is there anything else I need to do before it can be merged?

j2kun:
You're all good!

asraa:
Yep! From our end we make sure our internal presubmits and transformations work and then trigger the bot to submit it. (see https://heir.dev/docs/contributing/#internal-review-details) Just did :)

PR #694, opened 2024-05-24
fix #693 (SmallVector<T> vs SmallVector<T,4>)

### LINKED ISSUE #693: Compiler error
Steps:
1. Clone the repo (commit `c070cf34f195a5fd9dbbbf95533f5b2e350fd01d`)
2. `bazel build @heir//tools:heir-opt`

This produces an error:
```
lib/Transforms/YosysOptimizer/LUTImporter.cpp: In member function 'virtual llvm::SmallVector<Yosys::RTLIL::SigSpec> mlir::heir::LUTImporter::getInputs(Yosys::RTLIL::Cell*) const':
lib/Transforms/YosysOptimizer/LUTImporter.cpp:62:10: error: could not convert 'reversed' from 'SmallVector<[...],4>' to 'SmallVector<[...],1>'
   62 |   return reversed;
      |          ^~~~~~~~
      |          |
      |          SmallVector<[...],4>
Target //tools:heir-opt failed to build
```

AlexanderViand-Intel:
Fixes #693:

While some compilers allow returning `SmallVector<T,4>` when the function return type is `SmallVector<T>`, others (including gcc 11.4.0) do not. Not sure which behavior is the correct one, but unless setting the default size is critical for performance here, this should be an easy fix for the issue.

### DIFF
--- a/lib/Transforms/YosysOptimizer/BooleanGateImporter.cpp
+++ b/lib/Transforms/YosysOptimizer/BooleanGateImporter.cpp
@@ -40,7 +40,7 @@ mlir::Operation *BooleanGateImporter::createOp(Yosys::RTLIL::Cell *cell,
 SmallVector<Yosys::RTLIL::SigSpec> BooleanGateImporter::getInputs(
     Yosys::RTLIL::Cell *cell) const {
   // Return all non-Y named attributes.
-  SmallVector<Yosys::RTLIL::SigSpec, 4> inputs;
+  SmallVector<Yosys::RTLIL::SigSpec> inputs;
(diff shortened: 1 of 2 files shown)


j2kun (approved):
`SmallVector<T, 4>` would be correct if we had a reason to believe there would usually only be 4 values stored in the vector. Omitting it makes LLVM guess (trying to keep the size of the SmallVector instance around 64 bytes, as per https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h), which in our case is close enough to 4 that it probably doesn't matter. Let's leave out the 4 for simplicity's sake.

Issue #661, opened 2024-04-29
Generate a helper that configures OpenFHE crypto context based on the IR content


j2kun:
Right now the user must manually call functions like `EvalMultKeyGen` if the IR contains `EvalMult`, and `EvalRotateKeyGen` with known rotation shifts (which requires looking at the generated code to determine), as well as `SetMultiplicativeDepth` and `SetPlaintextModulus`.

We should have a pass in the OpenFHE dialect that generates a function that configures the `cryptoContext` appropriately.

Maokami:
I've looked over this issue, and I would like to give it a try :)
I'm currently preparing for my master's thesis defense, so I will probably be able to start it next Tuesday.

asraa:
Also one more note: we can add the pub/priv key generation to the context generation function (https://github.com/google/heir/pull/696#discussion_r1636688307), if we remove the restriction to return only one value.

Maybe something like:

`(context, pubkey, privkey) = my_func__generate_crypto_context()`

or it could go in configure

asraa:
After removing that restriction, we also can remove the return parameter in configure_crypto_context. See https://github.com/google/heir/pull/696#discussion_r1636695254

Maokami:
https://github.com/google/heir/pull/696#issuecomment-2143834856
> Now, I'm trying to determine the appropriate values for `mulDepth` and `plainMod`. My plan is as follows:
> 
> 1. Finding `mulDepth`:
>    
>    * Determine the maximum multiplication depth through forward dataflow analysis.
>    * `muldepth(op) = max_i (muldepth(operand_i(op))) + 1` if `op` is `mulOp` (or `mulPlainOp`)
>    * `muldepth(op) = max_i (muldepth(operand_i(op)))` otherwise
> 2. Finding `plainMod`:
>    
>    * Considering the following post (https://openfhe.discourse.group/t/automaticly-find-a-good-plaintext-modulus/312), I think the best approach is to determine the maximum number of plaintext bits (`nBits`) that can appear in the program and then call OpenFHE's `FirstPrime(nBits, 2*65536)` to calculate `plainMod`. To do this, it seems reasonable to receive the maximum plaintext bit count of the inputs given as arguments to the program and then use forward dataflow analysis to determine the maximum plaintext bits of the program.

Before I forget, I would like to seek some advice and explain what I have attempted so far.

[...]

PR #696, opened 2024-05-25
Draft: Generate a helper function that configures OpenFHE crypto context

Maokami:
Made a pass in the OpenFHE dialect to generate a helper function that configures OpenFHE crypto context. (#661)

edit:
I think the configurations can be split into two phases:
1. Setting appropriate encryption parameters (`CCParams`) when generating the crypto context.
2. Calling the appropriate each `KeyGen` APIs on the generated crypto context.

Therefore, I have decided to create two helper functions for each phase: generate_crypto_context and configure_crypto_context.

Current Status:

- [x] Make `ConfigureCryptoContext` pass in the OpenFHE dialect
- [x] Generate the helper functions
- [x] Add ops and types corresponding to `CCParams` and `CryptoContext` API in OpenFHE
- [x] Add OpenFhePke printers
- [x] Express each `KeyGen`s (~~`KeyGen`~~, `EvalKeyGen`, `EvalRotateKeyGen`) in OpenFHE dialect
- [ ] Find appropriate encryption parameters(`mulDepth`, `plainMod`) by analyzing the given function
- [x] Configurate each keygen (`EvalMultKeyGen`, vector of indices for `EvalRotateKeyGen`) by analyzing the given function
- [ ] Add more tests

### DIFF
--- a/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
@@ -1,0 +1,160 @@
+#include "lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.h"
+
+#include <cstddef>
+#include <set>
+#include <string>
+
+#include "lib/Dialect/Openfhe/IR/OpenfheOps.h"
+#include "lib/Dialect/Openfhe/IR/OpenfheTypes.h"
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"   // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace openfhe {
+
+#define GEN_PASS_DEF_CONFIGURECRYPTOCONTEXT
+#include "lib/Dialect/Openfhe/Transforms/Passes.h.inc"
+
+// Helper function to check if the function has MulOp or MulNoRelinOp
+bool hasMulOp(func::FuncOp op) {
+  bool result = false;
+  op.walk<WalkOrder::PreOrder>([&](Operation *op) {
+    if (isa<openfhe::MulOp>(op) || isa<openfhe::MulNoRelinOp>(op) ||
+        isa<openfhe::MulPlainOp>(op)) {
+      result = true;
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  return result;
+}
+
+// Helper function to find all the rotation indices in the function
+SmallVector<int64_t> findAllRotIndices(func::FuncOp op) {
+  std::set<int64_t> distinctRotIndices;
+  op.walk([&](openfhe::RotOp rotOp) {
+    auto indexAttr =
+        dyn_cast<arith::ConstantOp>(rotOp.getIndex().getDefiningOp())
+            .getValue();
+    int64_t rotIndex = dyn_cast<IntegerAttr>(indexAttr).getInt();
+    distinctRotIndices.insert(rotIndex);
+    return WalkResult::advance();
+  });
+  SmallVector<int64_t> rotIndicesResult(distinctRotIndices.begin(),
+                                        distinctRotIndices.end());
+  return rotIndicesResult;
+}
+
+// function that generates the crypto context with proper parameters
+LogicalResult generateGenFunc(func::FuncOp op, const std::string &genFuncName,
+                              ImplicitLocOpBuilder &builder) {
+  Type openfheContextType =
+      openfhe::CryptoContextType::get(builder.getContext());
+  SmallVector<Type> funcArgTypes;
+  SmallVector<Type> funcResultTypes;
+  funcResultTypes.push_back(openfheContextType);
+
+  FunctionType genFuncType =
+      FunctionType::get(builder.getContext(), funcArgTypes, funcResultTypes);
+  auto genFuncOp = builder.create<func::FuncOp>(genFuncName, genFuncType);
+  builder.setInsertionPointToEnd(genFuncOp.addEntryBlock());
+
+  // TODO(#661) : Calculate the appropriate values by analyzing the function
+  int64_t mulDepth = 2;
+  int64_t plainMod = 4295294977;
+  Type openfheParamsType = openfhe::CCParamsType::get(builder.getContext());
+  Value ccParams = builder.create<openfhe::GenParamsOp>(openfheParamsType,
+                                                        mulDepth, plainMod);
+  Value cryptoContext =
+      builder.create<openfhe::GenContextOp>(openfheContextType, ccParams);
+
+  builder.create<func::ReturnOp>(cryptoContext);
+  return success();
+}
+
+// function that configures the crypto context with proper keygeneration
+LogicalResult generateConfigFunc(func::FuncOp op,
+                                 const std::string &configFuncName,
+                                 bool hasMulOp, SmallVector<int64_t> rotIndices,
+                                 ImplicitLocOpBuilder &builder) {
+  Type openfheContextType =
+      openfhe::CryptoContextType::get(builder.getContext());
+  Type privateKeyType = openfhe::PrivateKeyType::get(builder.getContext());
+
+  SmallVector<Type> funcArgTypes;
+  funcArgTypes.push_back(openfheContextType);
+  funcArgTypes.push_back(privateKeyType);
+
+  SmallVector<Type> funcResultTypes;
+  funcResultTypes.push_back(openfheContextType);
+
+  FunctionType configFuncType =
+      FunctionType::get(builder.getContext(), funcArgTypes, funcResultTypes);
+  auto configFuncOp =
+      builder.create<func::FuncOp>(configFuncName, configFuncType);
+  builder.setInsertionPointToEnd(configFuncOp.addEntryBlock());
+
+  Value cryptoContext = configFuncOp.getArgument(0);
+  Value privateKey = configFuncOp.getArgument(1);
+
+  if (hasMulOp) {
+    builder.create<openfhe::GenMulKeyOp>(cryptoContext, privateKey);
+  }
+  if (!rotIndices.empty()) {
+    builder.create<openfhe::GenRotKeyOp>(cryptoContext, privateKey, rotIndices);
+  }
+
+  builder.create<func::ReturnOp>(cryptoContext);
+  return success();
+}
+
+LogicalResult convertFunc(func::FuncOp op) {
+  auto module = op->getParentOfType<ModuleOp>();
+  std::string genFuncName("");
+  llvm::raw_string_ostream genNameOs(genFuncName);
+  genNameOs << op.getSymName() << "__generate_crypto_context";
+
+  std::string configFuncName("");
+  llvm::raw_string_ostream configNameOs(configFuncName);
+  configNameOs << op.getSymName() << "__configure_crypto_context";
+
+  ImplicitLocOpBuilder builder =
+      ImplicitLocOpBuilder::atBlockEnd(module.getLoc(), module.getBody());
+
+  if (failed(generateGenFunc(op, genFuncName, builder))) {
+    return failure();
+  }
+
+  builder.setInsertionPointToEnd(module.getBody());
+
+  bool hasMulOpResult = hasMulOp(op);
+  SmallVector<int64_t> rotIndices = findAllRotIndices(op);
+  if (failed(generateConfigFunc(op, configFuncName, hasMulOpResult, rotIndices,
+                                builder))) {
+    return failure();
+  }
+  return success();
+}
+
+struct ConfigureCryptoContext
+    : impl::ConfigureCryptoContextBase<ConfigureCryptoContext> {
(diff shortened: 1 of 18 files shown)


j2kun:
FYI https://github.com/google/heir/pull/700 is coming which updates the OpenFHE version in this repo to v1.1.4. I don't think this will have any effect on this PR, but you may want to rebase over it once it merges to be sure.

Maokami:
Thank you for letting me know. I will rebase over it later.

Maokami:
Now, I'm trying to determine the appropriate values for `mulDepth` and `plainMod`. My plan is as follows:

1. Finding `mulDepth`:
   - Determine the maximum multiplication depth through forward dataflow analysis.
   - `muldepth(op) = max_i (muldepth(operand_i(op)))` if `op` is `mulOp` (or `mulOpPlain`)
   - `muldepth(op) = max_i (muldepth(operand_i(op))) + 1` otherwise

2. Finding `plainMod`:
   - Considering the following post (https://openfhe.discourse.group/t/automaticly-find-a-good-plaintext-modulus/312), I think the best approach is to determine the maximum number of plaintext bits (`nBits`) that can appear in the program and then call OpenFHE's `FirstPrime(nBits, 2*65536)` to calculate `plainMod`. To do this, it seems reasonable to receive the maximum plaintext bit count of the inputs given as arguments to the program and then use forward dataflow analysis to determine the maximum plaintext bits of the program.

@j2kun  @asraa, do you have any opinions on this? If there is a better method, please let me know!

asraa:
@Maokami That seems like a good start!

> Finding mulDepth:

Agreed a dataflow analysis with the two conditions makes sense to me. Should the conditions be reversed, so that the computation adds one if `op` is a multiplication op? I know there can be some complexity in this in case there are thousands of addition operations, but for the most part libraries assume it's just the multiplication cost, and we should start with that assumption too.

> To do this, it seems reasonable to receive the maximum plaintext bit count of the inputs given as arguments to the program and then use forward dataflow analysis to determine the maximum plaintext bits of the program.

I agree this is would be necessary to have a good parameter selection. We probably want to have some kind of function argument attributes (like how we mark `secret.secret` types of the input) to mark the range of the input or the max bit width. We can put an attribute in `secret` like `secret.max_bit_width`, or we can start with the assumption that it uses the full range of the integer and that the program is written with minimal bit width to start if you want

j2kun:
Is this PR in a good state to checkpoint with a TODO for the remaining parts? I'd be happy merging it if so.

Maokami:
I'm currently struggling with writing the analyses, but I agree that this PR can be merged as a checkpoint.
Could you please review it? Once done, I'll squash the commits for merging.

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -0,0 +1,161 @@
...
+  SmallVector<Type> funcArgTypes;
+  SmallVector<Type> funcResultTypes;
+  funcResultTypes.push_back(openfheContextType);
+
+  FunctionType genFuncType =
+      FunctionType::get(builder.getContext(), funcArgTypes, funcResultTypes);
+  auto genFuncOp = builder.create<func::FuncOp>(genFuncName, genFuncType);
+  builder.setInsertionPointToEnd(genFuncOp.addEntryBlock());
+
+  // TODO : Calculate the appropriate values by analyzing the function
```
asraa, line 65:
```suggestion
  // TODO(661) : Calculate the appropriate values by analyzing the function
```

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -0,0 +1,161 @@
...
+// Helper function to find all the rotation indices in the function
+SmallVector<int64_t> findAllRotIndices(func::FuncOp op) {
+  SmallVector<int64_t> rotIndices;
+  op.walk([&](openfhe::RotOp rotOp) {
+    // TODO : I'm not sure if this is the right way to get the index
+    auto typedIndex =
+        dyn_cast<arith::ConstantOp>(rotOp.getIndex().getDefiningOp())
+            .getValue();
+    int64_t rotIndex = dyn_cast<IntegerAttr>(typedIndex).getInt();
+    if (std::find(rotIndices.begin(), rotIndices.end(), rotIndex) ==
```
asraa, line 42: I think you could probably accumulate these in a set and then convert into a smallvector if you need!

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    return WalkResult::advance();
+  });
+  return result;
+}
+
+// Helper function to find all the rotation indices in the function
+SmallVector<int64_t> findAllRotIndices(func::FuncOp op) {
+  SmallVector<int64_t> rotIndices;
+  op.walk([&](openfhe::RotOp rotOp) {
+    // TODO : I'm not sure if this is the right way to get the index
```
asraa, line 37: I think that seems fine with me. I think though the rotation index is always going to be statically known, and I wonder whether we should convert this to an integer attribute instead of an SSA value (not in this PR)... @j2kun wdyt?

tests/openfhe/end_to_end/dot_product_8_test.cpp
```diff
@@ -21,18 +21,25 @@ namespace openfhe {
...
+  // cryptoContext->Enable(LEVELEDSHE);
+  // KeyPair<DCRTPoly> keyPair = cryptoContext->KeyGen();
+  // PublicKey<DCRTPoly> publicKey = keyPair.publicKey;
+  // PrivateKey<DCRTPoly> secretKey = keyPair.secretKey;
+  // cryptoContext->EvalMultKeyGen(secretKey);
+  // cryptoContext->EvalRotateKeyGen(secretKey, {1, 2, 4, 7});
+  auto cryptoContext = dot_product__generate_crypto_context();
+  auto keyPair = cryptoContext->KeyGen();
+  auto publicKey = keyPair.publicKey;
+  auto secretKey = keyPair.secretKey;
```
asraa, line 40: Should there be a option in the pass to add public/private key gen directly into the configuration helper?

lib/Dialect/Openfhe/Transforms/Passes.td
```diff
@@ -0,0 +1,20 @@
+#ifndef LIB_DIALECT_OPENFHE_TRANSFORMS_PASSES_TD_
+#define LIB_DIALECT_OPENFHE_TRANSFORMS_PASSES_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def ConfigureCryptoContext : Pass<"openfhe-configure-crypto-context"> {
+  let summary = "Configure the crypto context in OpenFHE";
+  let description = [{
+  TODO
```
asraa, line 9:
```suggestion
     This pass generates helper functions to generate and configure the OpenFHE crypto context for the given function. Generating the crypto context sets the appropriate encryption parameters, while the configuration generates the necessary evaluation keys (relinearization and rotation keys).
          
     For example, for an MLIR function `@my_func`, the generated helpers have the following signatures
     ```mlir
    func.func  @my_func__generate_crypto_context() -> !openfhe.crypto_context
    
    func.func  @my_func__configure_crypto_context(!openfhe.crypto_context, !openfhe.private_key) -> !openfhe.crypto_context
     ```
```

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -0,0 +1,161 @@
...
+  Type openfheContextType =
+      openfhe::CryptoContextType::get(builder.getContext());
+  Type privateKeyType = openfhe::PrivateKeyType::get(builder.getContext());
+
+  SmallVector<Type> funcArgTypes;
+  funcArgTypes.push_back(openfheContextType);
+  funcArgTypes.push_back(privateKeyType);
+
+  SmallVector<Type> funcResultTypes;
+  funcResultTypes.push_back(openfheContextType);
```
asraa, line 92: just curious, but does this need to be returned from the configuration funct if it modifies the context arg?

asraa (commented):
Pass looks good to me! Ready to fix up the test comments, and just a couple of nits/suggestions.

It would be good to add a simple filecheck test that just checks for the presence of the generate/config funcs.

Thank you!

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
Maokami: I think I might not have understood the question correctly. Could you elaborate, please?

tests/openfhe/end_to_end/dot_product_8_test.cpp
Maokami: I agree that this is a good idea.

There is one minor issue though. When I initially created this PR, I also wanted to model the Keygen functionality in the Openfhe dialect like: (pub_key, sec_key) = Openfhe.keygen(context).

However, when I tried to generate code using this, it failed at [OpenFhePkeEmitter](https://github.com/google/heir/blob/9cad9ed31d826550fdc0ce54068a85ac0167a949/lib/Target/OpenFhePke/OpenFhePkeHeaderEmitter.cpp#L78) because it only supports single return type, causing code generation to fail. As far as I know, it seems like mlir no longer supports tuple types. Is there any better solution?

Anyway, I also think it would be great if we could generate the keygen as a helper function, and it might be good to leave a todo for the same issue :)

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
asraa: oh, it might be me not understanding! I *think* this function takes a crypto context and private key, and then generates the needed keys in the context. so if it modifies the context (i'm not sure if it modifies or copies), then i don't think it needs to return the context. so the user would do something like

```
OpenFHEContext context = __generate_crypto_context();
// create pub/priv key
__configure_crypto_context(context, priv_key);
```

tests/openfhe/end_to_end/dot_product_8_test.cpp
asraa: > However, when I tried to generate code using this, it failed at [OpenFhePkeEmitter](https://github.com/google/heir/blob/9cad9ed31d826550fdc0ce54068a85ac0167a949/lib/Target/OpenFhePke/OpenFhePkeHeaderEmitter.cpp#L78) because it only supports single return type, causing code generation to fail. As far as I know, it seems like mlir no longer supports tuple types.

Oh! I think actually that was a restriction that was legacy from the Verilog emitter. I expect that we don't need that here. See this MLIR [test](https://github.com/google/heir/blob/62d5c959f741e003fdae19b9a558ee2fa474dd50/tests/jaxite/emit_jaxite.mlir#L17) for an example of multiple return values.

Agree, I'll note it in the issue thread as a comment! I think I can help take care of that and fix the code assumptions.

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
Maokami: Oh, I see! You're right, it just modifies the context, not makes a new copy, so there's no need to return the context!

Though, again, currently it doesn't work since OpenFhePkeEmitter allows single value return only 😄 (Zero value return is not allowed)

Maokami:
I'll add simple FileCheck tests by tomorrow!

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
asraa: Ah got it! Adding it to that thread, and then I can help fix it up from there.

tests/openfhe/end_to_end/simple_sum_test.cpp
```diff
@@ -19,17 +19,23 @@ namespace heir {
...
 
 TEST(BinopsTest, TestInput1) {
-  CCParams<CryptoContextBGVRNS> parameters;
-  parameters.SetMultiplicativeDepth(2);
-  parameters.SetPlaintextModulus(65537);
-  CryptoContext<DCRTPoly> cryptoContext = GenCryptoContext(parameters);
-  cryptoContext->Enable(PKE);
-  cryptoContext->Enable(KEYSWITCH);
-  cryptoContext->Enable(LEVELEDSHE);
+  // CCParams<CryptoContextBGVRNS> parameters;
```
asraa, line 22: nit: could you please remove the commented out stubs in the test harnesses? ty!

asraa (approved):
Thank you!

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
asraa: I'll also add this to the issue thread.

tests/openfhe/end_to_end/simple_sum_test.cpp
Maokami: sure!

asraa (approved):
Thanks! LGTM @Maokami - please squash and rebase and I'll merge!

asraa:
Hey! @Maokami we should mark this as ready to review and merge before continuining right?

Maokami:
> Hey! @Maokami we should mark this as ready to review and merge before continuining right?

If you mean 'ready to pull', yes!

PR #702, opened 2024-05-29
Add lwe.add, lwe.mul_scalar, and cggi.lut_lincomb

j2kun:
Also make the `cggi.lut*` syntax more consistent by removing parens from the operands.

This is to prepare for a new optimization pass involving more finely-crafted LUT ops by qu4ternion on Discord

### DIFF
--- a/lib/Dialect/CGGI/IR/CGGIOps.td
+++ b/lib/Dialect/CGGI/IR/CGGIOps.td
@@ -21,7 +21,6 @@ class CGGI_Op<string mnemonic, list<Trait> traits = []> :
 }
 
 // --- Operations for a gate-bootstrapping API of a CGGI library ---
-def LWECiphertextLike : TypeOrContainer<LWECiphertext, "ciphertext-like">;
 
 class CGGI_BinaryGateOp<string mnemonic>
   : CGGI_Op<mnemonic, [
@@ -65,10 +64,9 @@ class CGGI_LutOp<string mnemonic, list<Trait> traits = []>
   ElementwiseMappable,
   Scalarizable,
   DeclareOpInterfaceMethods<LUTOpInterface>
-
 ]> {
   let results = (outs LWECiphertextLike:$output);
-  let assemblyFormat = "`(` operands `)` attr-dict `:` qualified(type($output))" ;
+  let assemblyFormat = "operands attr-dict `:` qualified(type($output))" ;
 
   let description = [{
     An op representing a lookup table applied to some number `n` of ciphertexts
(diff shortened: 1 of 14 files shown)


lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -99,4 +98,16 @@ def CGGI_Lut3Op : CGGI_LutOp<"lut3", [AllTypesMatch<["a", "b", "c", "output"]>]>
   let results = (outs LWECiphertextLike:$output);
 }
 
+def CGGI_LutPreparedOp : CGGI_LutOp<"lut_prepared", [AllTypesMatch<["input", "output"]>]> {
```
asraa, line 101: Should there be a verifier that checks that the LUT range is within the ciphertext's encoding's bitwidth? (Or we can leave as a TODO, since I think it might not even then be clear if we have some extra padding bits. But at least it is upper bound)

asraa (approved):
Thanks! 

Part of me doesn't like the name `lut_prepared` but I can't think of anything better besides something like "packing" the inputs

j2kun:
Based on some more conversations I think I will tweak this one a bit more before submitting.

lib/Dialect/CGGI/IR/CGGIOps.td
j2kun: Done

j2kun:
Tweaks done. Now there's a linear combination LUT op with specifiable coefficients. I think we could canonicalize the other two ops in terms of that op easily, and I'll look into that later.

lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -99,4 +97,23 @@ def CGGI_Lut3Op : CGGI_LutOp<"lut3", [AllTypesMatch<["a", "b", "c", "output"]>]>
   let results = (outs LWECiphertextLike:$output);
 }
 
+def CGGI_LutLinCombOp : CGGI_LutOp<"lut_lincomb", [SameOperandsAndResultType]> {
+  let summary = "A variadic-input lookup table with inputs prepared via linear combination.";
+  let arguments = (ins
+    Variadic<LWECiphertext>:$inputs,
+    DenseI32ArrayAttr:$coefficients,
```
asraa, line 104: The coefficient array size must match the input size, I think. Might also mark that as a TODO in the issue

lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -99,4 +97,23 @@ def CGGI_Lut3Op : CGGI_LutOp<"lut3", [AllTypesMatch<["a", "b", "c", "output"]>]>
...
+    Builtin_IntegerAttr:$lookup_table
+  );
+  let description = [{
+    An op representing a lookup table applied to an arbitrary number of input
+    ciphertexts, which are combined according to a static linear combination
+    attached to the op.
+
```
asraa, line 111:
```suggestion
For example,

%2 = cggi.lut_lincomb %0, %1 {coefficients = array<i32: 1, -1>, lookup_table = 68 : index} : !ciphertext

represents applying the lookup table to the linear combination `%0 - %1`.
```

lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -99,4 +97,23 @@ def CGGI_Lut3Op : CGGI_LutOp<"lut3", [AllTypesMatch<["a", "b", "c", "output"]>]>
...
+    E.g., where the semantics of `cggi.lut3` combine the inputs into a single
+    ciphertext using products with plaintexts and sums, this op assumes that
+    work has been done in advance, using `lwe.add` and `lwe.mul_scalar`.
```
asraa, line 114: Is this still true? It seems like the add and mul_scalars are represented in the `coefficients` attribute.

It may also be worth mentioning that the user is expected to ensure that the linear combination does not overflow the error bits.

asraa (commented):
Thanks! Just a couple more nits

j2kun:
I'll just go ahead and write the verifier now :)

lib/Dialect/CGGI/IR/CGGIOps.td
j2kun: Fixed

lib/Dialect/CGGI/IR/CGGIOps.td
j2kun: Added in a verifier

j2kun:
@asraa sorry to keep bothering with updates, but I added the verifier.

asraa (approved):
No prob! Nice verifier tests. Take it away from here :)

PR #711, opened 2024-05-31
add a new arith_ext dialect

### LINKED ISSUE #709: Create a new dialect `arith_ext`
Cf. #635.

Add the initial operations of `barret_reduce` and `subifge`.

inbelic:
- Add operations barrett_reduce and subifge to mirror the required operations of the HEaan.mlir paper

Resolve #709

### DIFF
--- a/lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
+++ b/lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
@@ -1,0 +1,49 @@
+#include "lib/Dialect/ArithExt/IR/ArithExtDialect.h"
+
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
+#include "mlir/include/mlir/IR/DialectImplementation.h"  // from @llvm-project
+
+// NOLINTNEXTLINE(misc-include-cleaner): Required to define ArithExtOps
+#include "lib/Dialect/ArithExt/IR/ArithExtOps.h"
+
+// Generated definitions
+#include "lib/Dialect/ArithExt/IR/ArithExtDialect.cpp.inc"
+
+#define GET_OP_CLASSES
+#include "lib/Dialect/ArithExt/IR/ArithExtOps.cpp.inc"
+
+namespace mlir {
+namespace heir {
+namespace arith_ext {
+
+void ArithExtDialect::initialize() {
+  addOperations<
+#define GET_OP_LIST
+#include "lib/Dialect/ArithExt/IR/ArithExtOps.cpp.inc"
+      >();
+}
+
+LogicalResult BarrettReduceOp::verify() {
+  auto inputType = getInput().getType();
+  unsigned bitWidth;
+  if (auto tensorType = dyn_cast<RankedTensorType>(inputType)) {
+    bitWidth = tensorType.getElementTypeBitWidth();
+  } else if (auto integerType = dyn_cast<IntegerType>(inputType)) {
+    bitWidth = integerType.getWidth();
+  }
+  auto cmod = APInt(64, getModulus());
+  auto expectedBitWidth = (cmod - 1).getActiveBits();
+  if (bitWidth < expectedBitWidth || 2 * expectedBitWidth < bitWidth) {
+    return emitOpError()
+           << "input bitwidth is required to be in the range [w, 2w], where w "
+              "is the smallest bit-width that contains the range [0, modulus). "
+              "Got "
+           << bitWidth << " but w is " << expectedBitWidth << ".";
+  }
+
+  return success();
+}
+
+}  // namespace arith_ext
+}  // namespace heir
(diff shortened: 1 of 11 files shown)


lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -0,0 +1,46 @@
...
+include "mlir/Interfaces/InferTypeOpInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+
+class ArithExt_Op<string mnemonic, list<Trait> traits = []> :
+        Op<ArithExt_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::arith_ext";
+}
+
+def ArithExt_BarrettReduceOp : ArithExt_Op<"barrett_reduce", [Pure, SameOperandsAndResultType]> {
```
j2kun, line 17: can probably move `Pure` to the base class. I can't imagine any arith_ext ops being stateful.

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -0,0 +1,46 @@
...
+  let summary = "Compute the first step of the Barrett reduction.";
+  let description = [{
+    Let q denote a statically known modulo and let b = 2^bitWidth where bitWidth
```
j2kun, line 20:
```suggestion
    Let q denote a statically known modulus and let b = 2^bitWidth where bitWidth
```

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -0,0 +1,46 @@
...
+    is the smallest bitWidth that contains the range [0, q). The Barrett reduce
+    operation computes barret_reduce x = floor(x * floor(b / q) / b).
+
+    Given 0 <= x < q^2, then this will compute (x mod q) or (x mod q) + p.
```
j2kun, line 24:
```suggestion
    Given $0 <= x < q^2$, then this will compute $(x \mod q)$ or $(x \mod q) + p$.
```

note you can use latex mathmode, here and elsewhere if you prefer

j2kun (approved):

inbelic:
Having worked a bit on the lowerings, I have now also added a verify function to the operation on the bitwidth.

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -0,0 +1,48 @@
...
+    Let $q$ denote a statically known modulus and $b = 4^{w}$, where $w$ is the
+    smallest bit-width that contains the range $[0, q)$. The Barrett reduce
+    operation computes `barret_reduce x = x - floor(x * floor(b / q) / b) * q`.
+
+    Given $0 <= x < q^2$, then this will compute $(x \mod q)$ or $(x \mod q) + p$.
+  }];
+
+  let arguments = (ins
+    SignlessIntegerLike:$input,
+    I64Attr:$modulus
```
asraa, line 29: I wonder if the cmod here should be more generic? In the verifier code above you take just the active bits, which is the right approach, and for now I doubt we would be performing mod q muls with greater than 64 bits, but just wanted to note.

lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
```diff
@@ -0,0 +1,46 @@
...
+    bitWidth = tensorType.getElementTypeBitWidth();
+  } else if (auto integerType = dyn_cast<IntegerType>(inputType)) {
+    bitWidth = integerType.getWidth();
+  }
+  auto cmod = APInt(64, getModulus());
+  if (bitWidth != 2 * (cmod - 1).getActiveBits()) {
+    return emitOpError()
+           << "input bitwidth is required to be 2w, where w "
+              "is the smallest bit-width that contains the range [0, modulus).";
+  }
```
asraa, line 39: Should we bound this instead of computing an exact inequality? `bitWidth <= 2 * w`?

Might also be worth outputting the expected 2w bitwidth exactly in the error (e.g. expected 10 bits, got 11)

asraa (approved):
Thanks! Some small questions, but looks good to start!

lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
inbelic: Yep, I think we could allow for the range `w <= bitWidth <= 2 * w`, as if we have a bit-width less than `w` we may truncate the result.

lib/Dialect/ArithExt/IR/ArithExtOps.td
asraa: I think you're right that there are other assumptions sprinkled through the project in our lowerings. SG!

asraa (approved):
Thanks @inbelic !

PR #712, opened 2024-05-31
arith_ext: add transforms for arith.remui patterns

### LINKED ISSUE #710: Add passes to convert `arith.remui` to use the `arith_ext` ops
Cf #635.

Add the following passes to `arith_ext`:

- [ ] Create a pass that converts arith.muli + arith.remui into arith.muli + artih_ext.barret_reduce + arith_ext.subifge when the operands are in the range [0, cmod).
- [ ]  Create a pass that converts arith.addi + arith.remui into arith.addi + arith_ext.subifge when the operands are in the range [0, cmod).
- [ ]  Create a pass that converts arith.subi + arith.remui into arith.subi + arith.addi cmod + arith.subifge when the operands are in the range [0, cmod).

inbelic:
- We can reduce runtime computation by avoiding computing remui and
   instead use a Barrett reduction or SubIfGE operation when possible

Resolve #710

### DIFF
--- a/lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
+++ b/lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
@@ -92,7 +92,8 @@ void ArithExtToArith::runOnOperation() {
   target.addLegalDialect<arith::ArithDialect>();
 
   RewritePatternSet patterns(context);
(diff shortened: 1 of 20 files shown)


inbelic:
Note to self: currently does not have any check if the operands are actually in the range `[0, cmod)`

Issue #547, opened 2024-03-23
Add an overall architecture diagram for the dialects and lowerings


j2kun:
Ultimately it would be good to automate this, but getting a first version ready for new contributors takes priority.

PR #713, opened 2024-06-02
Add dot diagram for an overview of HEIR dialects

eymay:
A dot diagram is provided for visualising the available dialects and
the conversions between them. The graph is heavily influenced by the
CIRCT dialect diagram:
Source: https://github.com/llvm/circt/blob/main/docs/dialects.dot
Visual: https://circt.llvm.org/includes/img/dialects.svg

All dialects supported are shown and the possible conversions between
are represented with edges. Although there might be missing edges or
frontend/targets, this should provide a reasonable starting ground.

If the diagram is modified, the SVG should be recreated with the
`dot -Tsvg "input-dot" > "output-svg"` command to update the figure
in the website.


eymay:
Addresses  #547

eymay:
Hi,
I am interested in the HEIR project and realized it would be helpful for newcomers like me to have a look at the dialect diagram. I tried to reflect the architecture as close as possible. I have checked a diagram sketch from the [talk](https://www.youtube.com/watch?v=kqDFdKUTNA4&t=30m19s) and then checked the codebase for missing pieces. Please let me know of any errors or different stylistic choices :)

asraa (changes requested):
Thank you for this!

I think this is a great start, and we can add our new dialects (arith_ext!) as we go to it!

I wanted to just correct a small arrow and offer a few suggestions:
* The Yosys -> Comb path only enters into the CGGI scheme (comb-to-cggi pass), and not the RLWE scheme (BGV). Is it possible to make the arrow go directly into the CGGI box?
* The arrow from LWE -> Jaxite should probably be CGGI -> Jaxite (since it lowers from the CGGI scheme, even though LWE is used as the types in the CGGI IR)
* We might also want to color the Standard box purple, since that is also from Upstream MLIR
* Should we box around Polynomial and RNS to group them together like you did for LWE with the FHE schemes? 

Wonderful work!

eymay:
Thank you for the review!

>     * The Yosys -> Comb path only enters into the CGGI scheme (comb-to-cggi pass), and not the RLWE scheme (BGV). Is it possible to make the arrow go directly into the CGGI box?
> 
>     * The arrow from LWE -> Jaxite should probably be CGGI -> Jaxite (since it lowers from the CGGI scheme, even though LWE is used as the types in the CGGI IR)
> 

Sure fixed them!

>     * We might also want to color the Standard box purple, since that is also from Upstream MLIR
> 

Now it is purple, also we have the option to expand it to `Linalg`, `Tensor` and `Affine` dialects similar to the Upstream MLIR box. I refrained from that to not clutter the bottom but it could be more explanatory. 

>     * Should we box around Polynomial and RNS to group them together like you did for LWE with the FHE schemes?
> 

I boxed them now but not sure what to call them or whether they should be labeled. For now they are "Polynomial Dialects".

docs/static/dialects.dot
```diff
@@ -0,0 +1,99 @@
...
+    // Invisible edges for layout adjustments
+    edge [style=invis]
+    //This is a real edge but invisible now for simplicity
+    Arith -> Yosys
+    // Keep RNS below FHE Schemes
+    LWE -> RNS
+    // Keep Poly above Targets
+    Polynomial -> {OpenFHE TfheRust TfheRustBool Jaxite}
+    // Keep Verilog below Targets: LeftmostTarget -> Verilog
+    OpenFHE -> Verilog
```
eymay, line 98: Another question is where would the Verilog emission make the most sense. I placed it below the targets but it can be a simple Target in the rank of OpenFHE, TfheRust etc.

asraa: I think I like it where you placed it! I think it is kind of categorically different than OpenFHE/TfheRust/etc since I'm reading those as the MLIR dialects and not the non-MLIR translated code that we can emit using those libraries.

The Verilog emission arrow is slightly nuanced - the `emit-verilog` command actually works on the standard MLIR set above secret, but internally we run it on the standard MLIR within the `secret.generic` region. So I can see as a new user trying to understand the system that it would work off of secret, but also technically it works off standard MLIR. If it's confusing to someone we will find out :)

asraa (approved):
This looks great now! As we go we can add more of the new dialects being added (e.g. the *_ext) dialects, but for a simple new user reference this is perfect without being too dense.

asraa:
Hey @eymay when you're ready, please squash your commits into one (see https://heir.dev/docs/contributing/#pull-request-review-flow) - that way the copybara bot will merge your PR as a single commit :)

PR #715, opened 2024-06-03
arith_ext: add lowering to arith operations

inbelic:
Following up on #711

### DIFF
--- a/lib/Conversion/ArithExtToArith/ArithExtToArith.td
+++ b/lib/Conversion/ArithExtToArith/ArithExtToArith.td
@@ -1,0 +1,36 @@
+#ifndef LIB_CONVERSION_ARITHEXTTOARITH_ARITHEXTTOARITH_TD_
+#define LIB_CONVERSION_ARITHEXTTOARITH_ARITHEXTTOARITH_TD_
+
+include "lib/DRR/Utils.td"
+include "lib/Dialect/ArithExt/IR/ArithExtOps.td"
+include "mlir/Dialect/Arith/IR/ArithOps.td"
+include "mlir/IR/PatternBase.td"
+include "mlir/Pass/PassBase.td"
+
+def ArithExtToArith : Pass<"arith-ext-to-arith", "ModuleOp"> {
+  let summary = "Lower `arith_ext` to standard `arith`.";
+
+  let description = [{
+    This pass lowers the `arith_ext` dialect to their `arith` equivalents.
+  }];
+
+  let dependentDialects = [
+    "mlir::arith::ArithDialect",
+    "mlir::heir::arith_ext::ArithExtDialect",
+  ];
+}
+
+// Using DRR to generate the lowering patterns for specific operations
+
+defvar DefGE = ConstantEnumCase<Arith_CmpIPredicateAttr, "uge">;
+
+def ConvertSubIfGE : Pattern<
+  (ArithExt_SubIfGEOp $x, $y),
+  [
+    (Arith_SubIOp:$subOp $x, $y, DefOverflow),
+    (Arith_CmpIOp:$cmpOp DefGE, $x, $y),
+    (SelectOp $cmpOp, $subOp, $x)
+  ]
+>;
+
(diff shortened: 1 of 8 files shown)


lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -0,0 +1,48 @@
...
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+
+class ArithExt_Op<string mnemonic, list<Trait> traits = [Pure]> :
+        Op<ArithExt_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::arith_ext";
+}
+
+def ArithExt_BarrettReduceOp : ArithExt_Op<"barrett_reduce", [SameOperandsAndResultType]> {
+  let summary = "Compute the first step of the Barrett reduction.";
```
AlexanderViand-Intel, line 18: I think `arith_ext` ops should be `ElementwiseMappaple` wherever possible, and I don't think it'd be very complicated to enable this here, as the lowerings are to `affine` which is already `ElementwiseMappaple` + the ops already take `SinglessIntegerLike`, i.e., also permit tensors/etc.

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -0,0 +1,48 @@
...
+    I64Attr:$modulus
+  );
+  let results = (outs SignlessIntegerLike:$output);
+  let assemblyFormat = "operands attr-dict `:` qualified(type($input))";
+
+  let hasVerifier = 1;
+}
+
+def ArithExt_SubIfGEOp : ArithExt_Op<"subifge", [SameOperandsAndResultType]> {
+  let summary = "Compute (x >= y) ? x - y : x.";
```
AlexanderViand-Intel, line 38: is this just to make the lowering look nicer (`arith_ext.subifge` vs `arith.subi` + `arith.cmpi` + `arith.select`) or are there optimization/alternative lowering options? (I.e., do some targets have this as a native op?). I'm just curious, I'm not against having it "just" to make the lowering nicer :)

AlexanderViand-Intel (commented):
This looks good to me (but does need a rebase/squash) :)

lib/Dialect/ArithExt/IR/ArithExtOps.td
inbelic: The operation is added for the specific HEaan paper optimizations. They show that you can reduce the number of `arith_ext.subifge` operations by relaxing the output ranges.

j2kun (approved):

lib/Conversion/ArithExtToArith/ArithExtToArith.td
```diff
@@ -0,0 +1,36 @@
...
+
+  let dependentDialects = [
+    "mlir::arith::ArithDialect",
+    "mlir::heir::arith_ext::ArithExtDialect",
+  ];
+}
+
+// Using DRR to generate the lowering patterns for specific operations
+
+defvar DefGE = ConstantEnumCase<Arith_CmpIPredicateAttr, "uge">;
```
j2kun, line 25: I wonder if we will always want this to be unsigned (uge = unsigned greater equal) vs signed. I feel that perhaps we should all agree on the semantics of arith_ext ops on this front, i.e., that we always take the representative in [0, modulus) for modular operations.

Still happy to submit and revisit this later.

inbelic: I agree we should discuss this further.

I think if we let it be signed then we will have similar confusion/issue with using RemSI where it treats both operands as signed and we generally don't want the second operand to be signed.

FWIW, in the context of HEaan, they do make the assumption that it is an unsigned input.

AlexanderViand-Intel: I was planning to spend some time in the next meeting on discussing how we model modular arithmetic, and data-type semantics in general, and I think the signedness question would be a good addition to the list :)

asraa (approved):
Thanks!

PR #716, opened 2024-06-05
Add a design doc for the SIMD (HECO batching) optimizations 

AlexanderViand-Intel:
I still need to add a more detailed description of the main `-insert-rotate` pass, but this should already give a bit of an overview over the whole batching pipeline and I'd appreciate any feedback/questions. I realized I've gotten too used to "academic paper" language, so it doesn't sound as "documentation-y" as I wanted 😅 

Oh, and I noticed the `-apply-folders` pass wasn't registered in `heir-opt` while playing around with the pipeline to verify it works as  described, but I guess that should be a separate bug-fix PR.

EDIT: I guess we can either keep this as a draft until Jeremy is back, or merge it once it's completed/polished a bit and add a todo/issue for him to do a pass on it.

### DIFF
--- a/docs/content/en/docs/Design/simd.md
+++ b/docs/content/en/docs/Design/simd.md
@@ -1,0 +1,175 @@
+<!-- mdformat off(yaml frontmatter) -->
+---
+title: SIMD Optimizations
+weight: 9
+---
+<!-- mdformat on -->
+
+HEIR includes a SIMD (Single Instruction, Multiple Data) optimizer which is designed to exploit the restricted SIMD parallelism most (Ring-LWE-based) FHE schemes support (also commonly known as "packing" or "batching").
+Specifically, HEIR incorporates the "automated batching" optimizations (among many other things) from the [HECO](https://github.com/MarbleHE/HECO) compiler.
+The following will assume basic familiarity with the FHE SIMD paradigm and the high-level goals of the optimization, and we refer to the associated HECO [paper](https://www.usenix.org/system/files/usenixsecurity23-viand.pdf), [slides](https://www.usenix.org/system/files/sec23_slides_viand.pdf), [talk](https://www.youtube.com/watch?v=SP3C6gLWIS4) and additional resources on the [Usenix'23 website](https://www.usenix.org/conference/usenixsecurity23/presentation/viand) for an introduction to the topic.
+This documentation will mostly focus on describing how the optimization is realized in HEIR (which differs somewhat from the original implementation) and how the optimization is intended to be used in an overall end-to-end compilation pipeline.
+
+
+## Representing FHE SIMD Operations
+Following the design principle of maintaining programs in standard MLIR dialects as long as possible  (cf. the design rationale behind the [Secret Dialect](https://heir.dev/docs/design/secret/)),
+HEIR uses the MLIR [`tensor` dialect](https://mlir.llvm.org/docs/Dialects/TensorOps/) and [ElementwiseMappable](https://mlir.llvm.org/docs/Traits/#elementwisemappable) operations from the MLIR [`arith` dialect](https://mlir.llvm.org/docs/Dialects/ArithOps/) to represent HE SIMD operations.
+
+We do introduce the HEIR-specific [`tensor_ext.rotate`](https://heir.dev/docs/dialects/tensorext/#tensor_extrotate-heirtensor_extrotateop) operation, which represents a cyclical left-rotation of a tensor.
+Note that, as the current SIMD vectorizer only supports one-dimensional tensors, the semantics of this operation on multi-dimensional tensors are not (currently) defined.
+
+For example, the common "rotate-and-reduce" pattern which results in each element containing the sum/product/etc of the original vector can be expressed as:
+```llvm
+%tensor = tensor.from_elements %i1, %i2, %i3, %i4, %i5, %i6, %i7, %i8 : tensor<8xi16>
+%0 = tensor_ext.rotate %tensor, %c4 : tensor<8xi16>, index
+%1 = arith.addi %tensor, %0 : tensor<8xi16>
+%2 = tensor_ext.rotate %1, %c2 : tensor<8xi16>, index
+%3 = arith.addi %1, %2 : tensor<8xi16>
+%4 = tensor_ext.rotate %3, %c1 : tensor<8xi16>, index
+%5 = arith.addi %3, %4 : tensor<8xi16>
+```
+The `%cN` and `%iN`, which are defined as `%cN = arith.constant N : index` and `%iN = arith.constant N : i16`, respectively, have been omitted for readability.
+
+## Intended Usage
+The `-heir-simd-vectorizer` pipeline transforms a program consisting of loops and index-based accesses into tensors (e.g., `tensor.extract` and `tensor.insert`) into one consisting of SIMD operations (including rotations) on entire tensors.
+While its implementation does not depend on any FHE-specific details or even the Secret dialect, this transformation is likely only useful when lowering a high-level program to an arithmetic-circuit-based FHE scheme (e.g., B/FV, BGV, or CKKS).
+The `-mlir-to-openfhe-bgv` pipeline demonstrates the intended flow: augmenting a high-level program with `secret` annotations, then applying the SIMD optimization (and any other high-level optimizations) before lowering to BGV operations and then exiting to OpenFHE.
+
+> **Warning**
+> The current SIMD vectorizer pipeline supports only one-dimensional tensors.
+> As a workaround, one could reshape all multi-dimensional tensors into one-dimensional tensors, but MLIR/HEIR currently do not provide a pass to automate this process.
+
+Since the optimization is based on heuristics, the resulting program might not be optimal or could even be worse than a trivial realization that does not use ciphertext packing.
+However, well-structured programs generally lower to reasonable batched solutions, even if they do not achieve optimal batching layouts.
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/Design/simd.md
```diff
@@ -0,0 +1,112 @@
...
+    See [Insert Rotate Pass](#insert-rotate-pass) below for more details. This pass is roughly equivalent to the `-batching` pass in the original HECO implementation.
+
+    Doing this rewrite by itself does not represent an optimization, but if we consider what happens to the corresponding code for other indices (e.g.,  `t2[5] = t0[4] + t1[6]`), we see that the pass transforms expressions with the same relative index offsets into the exact same set of rotations/SIMD operations, so the following [Common Subexpression Eliminiation (CSE)](https://en.wikipedia.org/wiki/Common_subexpression_elimination) will remove redundant computations.
+    We apply CSE twice, once directly (which creates new opportunities for canonicalization and folding) and then again after that canonicalization.
+
+* **Cleanup of Redundant Insert/Extract** (`-collapse-insertion-chains -sccp -canonicalize -cse`):
+    Because the `-insert-rotate` pass maintains the concisstency of the IR, it emits a `tensor.extract` operation after the SIMD operation and uses that to replace the original operation (which is valid, as both produce the desired scalar result).
```
j2kun, line 89: consistency (sp)

docs/content/en/docs/Design/simd.md
```diff
@@ -0,0 +1,112 @@
...
+%2 = tensor_ext.rotate %1, %c2 : tensor<8xi16>, index
+%3 = arith.addi %1, %2 : tensor<8xi16>
+%4 = tensor_ext.rotate %3, %c1 : tensor<8xi16>, index
+%5 = arith.addi %3, %4 : tensor<8xi16>
+```
+The `%cN` and `%iN`, which are defined as `%cN = arith.constant N : index` and `%iN = arith.constant N : i16`, respectively, have been omitted for readability.
+
+## Intended Usage
+The `-heir-simd-vectorizer` pipeline transforms a program consisting of loops and index-based accesses into tensors (e.g., `tensor.extract` and `tensor.insert`) into one consisting of SIMD operations (including rotations) on entire tensors.
+While it's implementation does not depend on any FHE-specific details or even the Secret dialect, this transformation is likely only useful when lowering a high-level program to an arithmetic-circuit based FHE scheme (e.g., B/FV, BGV, or CKKS).
```
j2kun, line 35: it's -> its

docs/content/en/docs/Design/simd.md
```diff
@@ -0,0 +1,112 @@
...
+    %1 = tensor_ext.rotate %t1, %c1 : tensor<32xi16>, index
+    %2 = arith.addi %0, %1 : tensor<32xi16>
+    ```
+    i.e., rotating `t0` down by one (31 = -1 (mod 32)) and  `t1` up by one to bring the elements at index 3 and 5, respectively, to the "target" index 4.
+    The pass uses the `TargetSlotAnalysis` to identify the appropriate target index (or ciphertext "slot" in FHE-speak).
+    See [Insert Rotate Pass](#insert-rotate-pass) below for more details. This pass is roughly equivalent to the `-batching` pass in the original HECO implementation.
+
+    Doing this rewrite by itself does not represent an optimization, but if we consider what happens to the corresponding code for other indices (e.g.,  `t2[5] = t0[4] + t1[6]`), we see that the pass transforms expressions with the same relative index offsets into the exact same set of rotations/SIMD operations, so the following [Common Subexpression Eliminiation (CSE)](https://en.wikipedia.org/wiki/Common_subexpression_elimination) will remove redundant computations.
+    We apply CSE twice, once directly (which creates new opportunities for canonicalization and folding) and then again after that canonicalization.
```
j2kun, line 86: Is it worth explaining what those canoncalization patterns are?

j2kun (approved):
Looks like a good start!

docs/content/en/docs/Design/simd.md
```diff
@@ -0,0 +1,112 @@
...
+    As a result, the `-rotate-and-reduce` pass needs to search for sequences of arithmetic operations that correspond to the full folding of a tensor,
+    i.e., patterns such as `t[0]+(t[1]+(t[2]+t[3]+(...)))`, which currently uses a backwards search through the IR, but could be achieved more efficiently through a data flow analysis (c.f. issue [#532](https://github.com/google/heir/issues/523)).
+    In HECO, rotate-and-reduce is handled differently, by identifying sequences of compatible operations prior to batching and rewriting them to "n-ary" operations.
+    However, this approach requires non-standard arithmetic operations and is therefore not suitable for use in HEIR.
+    However, there is likely still an opportunity to make the patterns in HEIR more robust/general (e.g., support constant scalar operands in the fold, or support non-full-tensor folds).
+    See issue [#522](https://github.com/google/heir/issues/522) for ideas on how to make the HEIR pattern more robust/more general.
+
+
+### Insert Rotate Pass
+TODO: Write a detailed description of the rotation insertion pass and the associated target slot analysis.
```
j2kun, line 112: File an issue and replace this with `TODO(#123):` where 123 is replaced with the issue number

docs/content/en/docs/Design/simd.md
AlexanderViand-Intel: I added a brief "tensor text canonicalizations" section to the document and referenced it from here, but I feel like this information really out to go to https://heir.dev/docs/dialects/tensorext/, I'm just not sure where I'd have to put it for it to appear there.

docs/content/en/docs/Design/simd.md
AlexanderViand-Intel: I was originally planning to finish writing this up, but I think I'll take the easy way out and do just that ;)

AlexanderViand-Intel:
Thanks for the review @j2kun! I added a description of the `tensor_ext` canonicalization patterns, fixed the typos you spotted (and an embarrassingly high number of additional typos) and created #721 for the remaining TODO.

Oh, and I should probably mention the original "Design Doc" issue #629 so this PR gets linked from there, too.

docs/content/en/docs/Design/simd.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Intended Usage
+The `-heir-simd-vectorizer` pipeline transforms a program consisting of loops and index-based accesses into tensors (e.g., `tensor.extract` and `tensor.insert`) into one consisting of SIMD operations (including rotations) on entire tensors.
+While its implementation does not depend on any FHE-specific details or even the Secret dialect, this transformation is likely only useful when lowering a high-level program to an arithmetic-circuit-based FHE scheme (e.g., B/FV, BGV, or CKKS).
+The `-mlir-to-openfhe-bgv` pipeline demonstrates the intended flow: augmenting a high-level program with `secret` annotations, then applying the SIMD optimization (and any other high-level optimizations) before lowering to BGV operations and then exiting to OpenFHE.
+
+> **Warning**
+> The current SIMD vectorizer pipeline supports only one-dimensional tensors.
+> As a workaround, one could reshape all multi-dimensional tensors into one-dimensional tensors, but MLIR/HEIR currently do not provide a pass to automate this process.
```
asraa, line 40: Useful warning! Next HEIR meeting I may pick brains at this. I started writing a packing pass that promotes arbitrary tensors into tensors with a desired power of two number of slots with padding/replication, but I'm a little stuck on multi-dimensional tensors (which is probably necessary for matrix multiplication). It seems to me slightly chicken-and-eggy to write a packing pass and then find optimal rotations which depend on the packing. But like you mention below, we will always be relying on heuristics, and so some kind of heuristic will probably be necessary (and maybe we allow for handwritten attributes or options that can indicate a better approach)

docs/content/en/docs/Design/simd.md
```diff
@@ -0,0 +1,175 @@
...
+
+Since the optimization is based on heuristics, the resulting program might not be optimal or could even be worse than a trivial realization that does not use ciphertext packing.
+However, well-structured programs generally lower to reasonable batched solutions, even if they do not achieve optimal batching layouts.
+For common operations such as matrix-vector or matrix-matrix multiplications, state-of-the-art approaches require advanced packing schemes that might map elements into the ciphertext vector in non-trivial ways (e.g., diagonal-major and/or replicated).
+The current SIMD vectorizer will never change the arrangement of elements inside an input tensor and therefore cannot produce the optimal approaches for these operations.
+
+Note, that the SIMD batching optimization is different from, and significantly more complex than, the Straight Line Vectorizer (`-straight-line-vectorize` pass), which simply groups  [ElementwiseMappable](https://mlir.llvm.org/docs/Traits/#elementwisemappable) operations that agree in operation name and operand/result types into vectorized/tensorized versions.
```
asraa, line 47: One question I have that I am happy to add an answer to when I understand is when you want to use one or the either. (Or is it even possible to use straight-line-vectorize to tensor-ify operations and then heir-simd-vectorize after? probably doesn't do anything useful)

docs/content/en/docs/Design/simd.md
asraa: I agree it feels like it should belong to canonicalization in that dialect. I bet we can just add a README.md to each Dialect subfolder and then append that to the markdown generated https://github.com/google/heir/blob/15e510cf97874bac4f7901bc0fe5556e01bc7294/.github/workflows/copy_tblgen_files.sh#L64

right below the header: https://heir.dev/docs/dialects/tensorext/

Maybe the secret design section should actually go in like that too, and maybe each dialect should optionally have  a Design/MoreInfo subpage. We can discuss on an issue

asraa (approved):
Thanks!

PR #720, opened 2024-06-06
register `-apply-folders` pass in heir-opt

AlexanderViand-Intel:
As mentioned in #716: registers the `-apply-folders` pass in `heir-opt` so it can be used from the command line.


asraa (approved):
thanks!

PR #731, opened 2024-06-07
ArithExt Improvements: Arbitrary Coefficient Bitwidth + lowerings of modular add/sub/mul to arith

AlexanderViand-Intel:
This does three things
* removes the `I64Attr` constraint (replacing it with `APIntAttr`) and adds a verifier to ensure that the modulus can actually fit into the underlying operand type.
* adds an `arith_ext.mac %lhs, %rhs, %acc` operation that multiplies the two first operands and then adds the third. 
 This not only saves one remainder operation compared to doing the operations separately, but is also a native operation in some RLWE accelerators, so it's useful to have a way to model it in HEIR
* adds lowerings for `arith_ext.add`/sub/mul/mac to arith. If the results of a modular operation can fit into the underlying operand type, it's just a straighforward operation + remainder, otherwise it'll create `extui` and `remui` operations to temporarily increase the bitwidth.

### DIFF
--- a/lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
+++ b/lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
@@ -3,6 +3,7 @@
 #include "lib/Dialect/ArithExt/IR/ArithExtOps.h"
 #include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"  // from @llvm-project
 #include "mlir/include/mlir/IR/MLIRContext.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/TypeUtilities.h"         // from @llvm-project
 #include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
 #include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
 
@@ -12,6 +13,22 @@ namespace arith_ext {
 
 #define GEN_PASS_DEF_ARITHEXTTOARITH
 #include "lib/Conversion/ArithExtToArith/ArithExtToArith.h.inc"
+
+/// Returns a possibly extended modulus necessary to compute the given operation
+/// without overflow.
+template <typename ValueOrOpResult>
+TypedAttr modulusHelper(IntegerAttr mod, ValueOrOpResult op, bool mul = false) {
+  auto width = getElementTypeOrSelf(op).getIntOrFloatBitWidth();
+  auto modWidth = (mod.getValue() - 1).getActiveBits();
+  width = std::max(width, mul ? 2 * modWidth : modWidth + 1);
+  auto intType = IntegerType::get(op.getContext(), width);
+  auto truncmod = mod.getValue().zextOrTrunc(width);
+  if (auto st = mlir::dyn_cast<ShapedType>(op.getType())) {
+    auto containerType = st.cloneWith(st.getShape(), intType);
+    return DenseElementsAttr::get(containerType, truncmod);
+  }
+  return IntegerAttr::get(intType, truncmod);
+}
 
 namespace rewrites {
 // In an inner namespace to avoid conflicts with canonicalization patterns
@@ -31,7 +48,7 @@ struct ConvertBarrettReduce : public OpConversionPattern<BarrettReduceOp> {
 
     // Compute B = 4^{bitWidth} and ratio = floordiv(B / modulus)
     auto input = adaptor.getInput();
-    auto mod = APInt(64, op.getModulus());
+    auto mod = op.getModulus();
     auto bitWidth = (mod - 1).getActiveBits();
     mod = mod.trunc(3 * bitWidth);
     auto B = APInt(3 * bitWidth, 1).shl(2 * bitWidth);
@@ -92,7 +109,8 @@ void ArithExtToArith::runOnOperation() {
   target.addLegalDialect<arith::ArithDialect>();
 
   RewritePatternSet patterns(context);
-  patterns.add<rewrites::ConvertSubIfGE, ConvertBarrettReduce>(context);
+  rewrites::populateWithGenerated(patterns);
+  patterns.add<ConvertBarrettReduce>(context);
 
   if (failed(applyPartialConversion(module, target, std::move(patterns)))) {
     signalPassFailure();
--- a/lib/Conversion/ArithExtToArith/ArithExtToArith.td
+++ b/lib/Conversion/ArithExtToArith/ArithExtToArith.td
@@ -33,4 +33,121 @@ def ConvertSubIfGE : Pattern<
   ]
 >;
 
+
+def HasEnoughSpaceAddSub: Constraint<CPred<"llvm::cast<IntegerType>(getElementTypeOrSelf($_self.getType())).getWidth() >= ($0.getValue() - 1).getActiveBits() + 1">,
+"underlying type is sufficient for modular add/sub operation without overflow">;
+
+def HasEnoughSpaceMul: Constraint<CPred<"llvm::cast<IntegerType>(getElementTypeOrSelf($_self.getType())).getWidth() >= 2 * ($0.getValue() - 1).getActiveBits()">,
+ "underlying type is sufficient for modular mul operation without overflow">;
+
+def CastModulusAttributeAddSub : NativeCodeCall<"modulusHelper($0,$1,false)">;
+def CastModulusAttributeMul : NativeCodeCall<"modulusHelper($0,$1,true)">;
+
+def ConvertAddSimple : Pattern<
+  (ArithExt_AddOp:$op $x, $y, $mod),
+  [
+    (Arith_AddIOp:$add $x, $y, DefOverflow),
+    (Arith_RemUIOp $add, (Arith_ConstantOp (CastModulusAttributeAddSub $mod, $x)))
+  ],
+  [(HasEnoughSpaceAddSub:$op $mod)],
+  [],
+  (addBenefit 2)
+>;
+
+def ConvertSubSimple : Pattern<
+  (ArithExt_SubOp:$op $x, $y, $mod),
+  [
+    (Arith_SubIOp:$sub $x, $y, DefOverflow),
+    (Arith_RemUIOp $sub, (Arith_ConstantOp (CastModulusAttributeAddSub $mod, $x)))
+  ],
+  [(HasEnoughSpaceAddSub:$op $mod)],
+  [],
+  (addBenefit 2)
+>;
+
+def ConvertMulSimple : Pattern<
+  (ArithExt_MulOp:$op $x, $y, $mod),
+  [
+    (Arith_MulIOp:$mul $x, $y, DefOverflow),
+    (Arith_RemUIOp $mul, (Arith_ConstantOp (CastModulusAttributeMul $mod, $x)))
+  ],
+  [(HasEnoughSpaceMul:$op $mod)],
+  [],
+  (addBenefit 2)
+>;
+
+def ConvertMacSimple : Pattern<
+  (ArithExt_MacOp:$op $x, $y, $acc, $mod),
+  [
+    (Arith_MulIOp:$mul $x, $y, DefOverflow),
+    (Arith_AddIOp:$add $mul, $acc, DefOverflow),
+    (Arith_RemUIOp $add, (Arith_ConstantOp (CastModulusAttributeMul $mod, $x)))
+  ],
+  [(HasEnoughSpaceMul:$op $mod)],
+  [],
+  (addBenefit 2)
+>;
+
+def ConvertAdd : Pattern<
+  (ArithExt_AddOp $x, $y, $mod),
+  [
+    (Arith_ConstantOp:$newmod (CastModulusAttributeAddSub $mod, $x)),
+    (Arith_AddIOp:$add
+      (Arith_ExtUIOp $x,
+        (returnType $newmod)),
+      (Arith_ExtUIOp $y,
+        (returnType $newmod)),
+      DefOverflow),
+    (Arith_TruncIOp:$res
+      (Arith_RemUIOp $add, $newmod))
+  ]
+>;
+
+def ConvertSub : Pattern<
+  (ArithExt_SubOp $x, $y, $mod),
+  [
+    (Arith_ConstantOp:$newmod (CastModulusAttributeAddSub $mod, $x)),
+    (Arith_SubIOp:$sub
+      (Arith_ExtUIOp $x,
+        (returnType $newmod)),
+      (Arith_ExtUIOp $y,
+        (returnType $newmod)),
+      DefOverflow),
+    (Arith_TruncIOp:$res
+      (Arith_RemUIOp $sub, $newmod))
+  ]
+>;
+
+def ConvertMul : Pattern<
+  (ArithExt_MulOp $x, $y, $mod),
+  [
+    (Arith_ConstantOp:$newmod (CastModulusAttributeMul $mod, $x)),
+    (Arith_MulIOp:$mul
+      (Arith_ExtUIOp $x,
+        (returnType $newmod)),
+      (Arith_ExtUIOp $y,
+        (returnType $newmod)),
+      DefOverflow),
+    (Arith_TruncIOp:$res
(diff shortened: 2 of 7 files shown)


lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
```diff
@@ -13,6 +14,20 @@ namespace arith_ext {
...
 #include "lib/Conversion/ArithExtToArith/ArithExtToArith.h.inc"
 
+template <typename ValueOrOpResult>
+TypedAttr modulusHelper(IntegerAttr mod, ValueOrOpResult op, bool mul = false) {
+  auto width = getElementTypeOrSelf(op).getIntOrFloatBitWidth();
+  auto modWidth = (mod.getValue() - 1).getActiveBits();
+  width = std::max(width, mul ? 2 * modWidth : modWidth + 1);
+  auto intType = IntegerType::get(op.getContext(), width);
+  auto truncmod = mod.getValue().sextOrTrunc(width);
+  if (auto st = mlir::dyn_cast_or_null<ShapedType>(op.getType())) {
```
asraa, line 24:
```suggestion
  if (auto st = mlir::dyn_cast<ShapedType>(op.getType())) {
```

I think here we won't ever expect `op.getType()` to be null!

lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
```diff
@@ -13,6 +14,20 @@ namespace arith_ext {
 #define GEN_PASS_DEF_ARITHEXTTOARITH
 #include "lib/Conversion/ArithExtToArith/ArithExtToArith.h.inc"
 
+template <typename ValueOrOpResult>
+TypedAttr modulusHelper(IntegerAttr mod, ValueOrOpResult op, bool mul = false) {
+  auto width = getElementTypeOrSelf(op).getIntOrFloatBitWidth();
+  auto modWidth = (mod.getValue() - 1).getActiveBits();
+  width = std::max(width, mul ? 2 * modWidth : modWidth + 1);
+  auto intType = IntegerType::get(op.getContext(), width);
+  auto truncmod = mod.getValue().sextOrTrunc(width);
```
asraa, line 23: Hmm I think I might need some help, but should we assume this will be an unsigned modulus and zext it instead? FWIW (and I think I'm not totally up to speed on the issues before), that a signed modulus like 2**32 - 1 will be treated like a -1 when doing a signed remsi operation. But the operations below use remui, so I'm unclear if it will end up treating the sign extended modulus as unsigned..

lib/Conversion/ArithExtToArith/ArithExtToArith.td
```diff
@@ -33,4 +33,121 @@ def ConvertSubIfGE : Pattern<
...
+        (returnType $newmod)),
+      (Arith_ExtUIOp $y,
+        (returnType $newmod)),
+      DefOverflow),
+    (Arith_TruncIOp:$res
+      (Arith_RemUIOp $mul, $newmod))
+  ]
+>;
+
+def ConvertMac : Pattern<
```
asraa, line 136: Just curious, to me it seems like you could express `arith_ext.mac` as an `arith_ext.mul` and then `arith_ext.add`, and then utilize the patterns above for lowering to arith.

But like the pattern below, we can save on a `remui`-`trunci` sequence on the intermediate mul with the pattern below. Was that the intention to create a mac operation?

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -42,6 +43,13 @@ def ArithExt_MulOp : ArithExt_BinaryOp<"mul", [Commutative]> {
   }];
 }
 
+def ArithExt_MacOp : ArithExt_Op<"mac", [SameOperandsAndResultType, Pure, ElementwiseMappable]> {
+  let summary = "modular multiplication-and-accumulation operation";
```
asraa, line 47:
```suggestion
  let summary = "modular multiplication-and-accumulation operation";
  
  let description = [{
    The modular multiplication and accumulation operation computes
    
    `arith_ext.mac x, y, acc {modulus = q} = (x * y) + acc \mod q`
  }];
```

lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
```diff
@@ -13,6 +14,20 @@ namespace arith_ext {
 #define GEN_PASS_DEF_ARITHEXTTOARITH
 #include "lib/Conversion/ArithExtToArith/ArithExtToArith.h.inc"
 
+template <typename ValueOrOpResult>
+TypedAttr modulusHelper(IntegerAttr mod, ValueOrOpResult op, bool mul = false) {
```
asraa, line 18:
```suggestion
// Returns a possibly extended modulus necessary to compute the given operation without overflow.
template <typename ValueOrOpResult>
TypedAttr modulusHelper(IntegerAttr mod, ValueOrOpResult op, bool mul = false) {
```

asraa (approved):
Wonderful! Thank you!

lib/Conversion/ArithExtToArith/ArithExtToArith.td
AlexanderViand-Intel: The main reason is actually that `mac` is a native HW instruction in (some? all?) RLWE accelerators, but I also like having it for the efficiency gain ;)  

Though I guess we could also achieve saving the `remui`/etc just with patterns that apply to `add(mul, _)`, but that might make things a bit messy with all the "checking if bits fit" logic.

lib/Dialect/ArithExt/IR/ArithExtOps.td
j2kun: We have mathmode on the website if you want to use it

`arith_ext.mac x, y, z {modulus = q}` computes $(x * y) + z \mod q$

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -42,6 +43,13 @@ def ArithExt_MulOp : ArithExt_BinaryOp<"mul", [Commutative]> {
...
+  let arguments = (ins SignlessIntegerLike:$lhs, SignlessIntegerLike:$rhs, SignlessIntegerLike:$acc, APIntAttr:$modulus);
+  let results = (outs SignlessIntegerLike:$output);
+  let hasVerifier = 1;
+  let assemblyFormat = "$lhs `,` $rhs `,` $acc attr-dict `:` qualified(type($output))";
```
j2kun, line 51:
```suggestion
  let assemblyFormat = "operands attr-dict `:` type($output)";
```

j2kun: You could use this for all the others as well. I have started defaulting to having the base class op define an assembly format as

```
operands attr-dict `:` functional-type(operands, results)
```

And then override the part after the `:` for type-inferrible ops. Generally this gives a good format without having to specify anything op-specific.

tests/arith_ext/arith-ext-to-arith.mlir
```diff
@@ -1,4 +1,229 @@
...
+
+// CHECK-LABEL: @test_lower_simple_add
+// CHECK-SAME: (%[[LHS:.*]]: [[TYPE:.*]], %[[RHS:.*]]: [[TYPE]]) -> [[TYPE]] {
+func.func @test_lower_simple_add(%lhs : i8, %rhs : i8) -> i8 {
+  // CHECK-NOT: arith_ext.add
+  // CHECK: %[[ADD:.*]] = arith.addi %[[LHS]], %[[RHS]] : [[TYPE]]
+  // CHECK: %[[CMOD:.*]] = arith.constant 17 : [[TYPE]]
+  // CHECK: %[[REM:.*]] = arith.remui %[[ADD]], %[[CMOD]] : [[TYPE]]
+  // CHECK: return %[[REM]] : [[TYPE]]
+  %res = arith_ext.add %lhs, %rhs {modulus = 17 }: i8
```
j2kun, line 11: None of these moduli specify a type. You should test it with `modulus = 17 : i128` or similar to make sure that works.

lib/Dialect/ArithExt/IR/ArithExtOps.td
```diff
@@ -16,8 +16,9 @@ class ArithExt_Op<string mnemonic, list<Trait> traits = [Pure]> :
 
 class ArithExt_BinaryOp<string mnemonic, list<Trait> traits = []> :
     ArithExt_Op<mnemonic, traits # [SameOperandsAndResultType, Pure, ElementwiseMappable]>,
-    Arguments<(ins SignlessIntegerLike:$lhs, SignlessIntegerLike:$rhs, I64Attr:$modulus)>,
+    Arguments<(ins SignlessIntegerLike:$lhs, SignlessIntegerLike:$rhs, APIntAttr:$modulus)>,
```
j2kun, line 19: I never quite understood why APIntAttr exists, when it seems to be equivalent to `Builtin_IntegerAttr`, which requires an explicit specification of the integer container bit width. I suspect the difference is that APIntAttr is not a typed attribute, and I'm not sure which is better for our purposes.

j2kun: I guess it doesn't matter as long as the operands are typed, and the modulus fits within those operands.

lib/Conversion/ArithExtToArith/ArithExtToArith.td
asraa: Yeah, it seems pretty messy to try to "lazy remui".
Thank you!

tests/arith_ext/arith-ext-to-arith.mlir
j2kun: Nevermind, it's an untyped attribute.

lib/Dialect/ArithExt/IR/ArithExtOps.td
AlexanderViand-Intel: I think `Builtin_IntegerAttr ` isthe actual tablegen definition (autogeneration source) of `mlir::IntegerAttr`, whereas `APIntAttr` is actually a `Constraint` (basically `CPred<isa<mlir::IntegerAttr>($_self)>`, iirc.). 

I'm not sure `Builtin_IntegerAttr` would be valid in a place where a constraint is expected? Similarily, `APIntAttr` is, afaik, a purely TableGen concept and has no C++ equivalent (I guess, `isa<mlir::IntegerAttr>` is the equivalant thing). Anytime I interact with MLIR's integer type/attribute system, I get a mild headache ;)

lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
j2kun: Agree with zext.

lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
```diff
@@ -29,6 +30,28 @@ void ArithExtDialect::initialize() {
...
+template <typename OpType>
+LogicalResult verifyArithExtOpMod(OpType op) {
+  auto type = getElementTypeOrSelf(op.getResult().getType());
+  unsigned bitWidth = llvm::cast<IntegerType>(type).getWidth();
+  unsigned modWidth = (op.getModulus() - 1).getActiveBits();
+  if (modWidth > bitWidth)
+    return op.emitOpError()
+           << "underlying type's bitwidth must be at least as "
+           << "large as the modulus bitwidth, but got " << bitWidth
+           << " while modulus requires width " << modWidth << ".";
```
j2kun, line 43: Since it's not typed, we should be careful that a value of 2**32 - 1 is not interpreted as -1, as @asraa mentioned in another comment.

Can we add a verification to assert that, even when interpreting the modulus as signed, the value is always > 0, and then add a test for a modulus of 2**64 - 1 (64 is the default APInt bit width).

lib/Conversion/ArithExtToArith/ArithExtToArith.cpp
AlexanderViand-Intel: Thanks, good catch, this should not be signed!

lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
AlexanderViand-Intel: Mh, very good point. I guess we shouldn't just hardcode to unsigned since we want to allow the underlying "container" type to be both signed (for people who prefer $[\floor{-q/2}, \floor{q/2})$) and unsigned (for $[0, q-1]$). For the modulus, though, e really only need unsignged now that I think about it... but I guess it's really the underlying type we care about so let's not mess with the attribute (which the parser makes signless 64-bit IntegerAttr, so if we were to change it, we'd have to always specify the type in the textual representation)

I'm not sure we can hard error on the > 0 check in all cases, though - it's a bit contrived, but someone might want their modulus to be 2^31-1 and use a 31-bit underlying type, which should pass the check here?
Maybe we should error if the underlying type is signed and the modulus isn't >0, warn if the underlying type is signless and the modulus would be <0 if interpreted as signed and not emit any warning/error if the underlying type is unsigned?

lib/Dialect/ArithExt/IR/ArithExtOps.td
j2kun: We use Builtin_IntegerAttr elsewhere in HEIR, and it requires a type specification that corresponds to the chosen bitwidth of the APInt that stores it.

lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
j2kun: > but someone might want their modulus to be 2^31-1 and use a 31-bit underlying type, which should pass the check here?

This is where I ran into issues with the remsi/remui before, since if the input is negative and the modulus is positive, you get the wrong answer either way (remsi gives 0 because the modulus is interpreted as -1, remui gives the wrong answer because the input is interpreted as a positive value mod 2**31).

I also think that this dialect is low-level enough that "preferences" about which coset representative is used are irrelevant. The lowering decides, and if we all agree to use $[0, q-1]$ then we can force it to be simpler that way.

I think specifying an explicit type in the textual IR will be safer, but I'm willing to give this a shot as-is.

AlexanderViand-Intel: Oh, that's very annoying. :/  I guess supporting the  "q=2^31-1 in 31-bit underlying type" would require some kind of type conversion before we lower to `arith`?  That just seems way overkill, so I'll make it a hard error for both signless and signed then!

lib/Dialect/ArithExt/IR/ArithExtOps.td
AlexanderViand-Intel: Interesting! I guess it does make sense that one can use `TypeOrAttrDef`s directly as a constraint, though I haven't seen `Builtin_IntegerAttr` used that way upstream yet. 

I tried swapping ut `APIntAttr` for `Builtin_IntegerAttr`  to see what happens, but before I even got to see what that does to the textual representation, I noticed that `getModulus()` now returns the Attr directly, whereas with `APIntAttr`, the tablegen backend would have that function name be a helper function that returns an `APInt`, and demote the actual `Attr` getter to `getModulusAttr()`. I'd love to say I understand way, but I'm at "MLIR works in mysterious ways" 🤷‍♂️

j2kun:
The reason is that the attribute is now typed, so it has the type as an
additional field.

lib/Dialect/ArithExt/IR/ArithExtDialect.cpp
AlexanderViand-Intel: I ended up having to make it "just" a warning for now because I realized (a) the ops only take `SignlessIntegerLike` so we couldn't test the "non-simple" add/sub cases with the hard error in place and (b) while I tried relaxing the type constraint to a quickly defined `IntegerLike`, I realized actually using unsigned types completely breaks the lowering to `arith`, which insists on `SignlessIntegerLike`. (Why though?)

Since we were already planning to discuss data represetations/signedness stuff in tomorrow's meeting, I'll just add this to the pile :)

j2kun: Signless is required, I knew that. I just meant we should only be constraining the modulus, and interpreting the input values as part of the semantics.

Issue #599, opened 2024-04-05
[BGV]: add lazy relinearization


j2kun:
Ideally this is done in a way that the optimization pass can be shared between all relinearizing ops across all schemes. In particular, we should implement a `RelinearizeLike` trait/interface, and implement the optimization pass against the interface.

ahmedshakill:
Hi @j2kun I would like to work on this. Could you shade some light on how may I approach this. I have gathered some basic knowledge about bgv, openfhe and their conversion dialects. 
About **lazy**-ness: we want to delay the relinearization computation ( and also to find the scope to do so) ? How to introduce laziness to a computation in a pass?
There's  an idea of **lazy loading** in mlir::BytecodeReader. But I got lost and couldn't extract anything helpful from there.

j2kun:
Sure! The basic idea is that an arith.muli lowers to a sequence of bgv.mul and bgv.relin. The relin is necessary to return the key basis from (1, s, s^2) to (1, s), but sometimes the relin op can be delayed if the output of the mul is combined with other values in that same key basis.

For example, if you multiply two pairs of values and then add their results, you can do it with one relin op after the addition, rather than two relin ops after each mul.

The implementation of this feature would, I believe, require some relatively straightforward tablegen patterns.

AlexanderViand-Intel:
While looking at `bgv.mul`, I noticed it (incorrectly) wouldn't allow having sequential multiplications without relineraization in-between. With the fix in (currently PR #733), I'd suggest the simplest way to realize this is to proceed as follows
* make the secret->bgv pass lowering of multiplication not emit any relinearization.
* have a new "lazy relin" pass with two simple patterns that rewrite `mul(mul1, _)` and `mul(_, mul1)` to `mul(relin(mul1),_` and `mul(_, relin(mul1))`, respectively. As Jeremy suggested, these should be pretty straightforward DDR patterns.

j2kun:
I didn't realize that relinearizing after repeated multiplications is a thing people do. I assume that:

- The key basis for that would be (1, s, s^2, s^3, s^4)
- We'd have to implement more generalized lowerings of relinearize for all these new possible key bases (and, e.g., maybe some backends don't support it? Does OpenFHE allow this?)

AlexanderViand-Intel:
> I didn't realize that relinearizing after repeated multiplications is a thing people do. I assume that:
> 
> * The key basis for that would be (1, s, s^2, s^3, s^4)
[...]

PR #733, opened 2024-06-10
fix dimension check in bgv.mul verifier

AlexanderViand-Intel:
Verifier logic for `bgv.mul` output degree was incorrect (and did not match type inference logic, which was correct).

With this fix, you should now be able to express multiplications of already multiplied values (without a relinerization in between) which might be helpful for the  layz relin optimization (#599)


j2kun:
LGTM! Just needs a rebase I believe.

j2kun (approved):

AlexanderViand-Intel:
> LGTM! Just needs a rebase I believe.

Done :)

j2kun (approved):

PR #737, opened 2024-06-14
Naive If-transformation

MeronZerihun:
An implementation for a pass that hoists instructions out of an if-operation and finally substitutes it with a select-operation.


lib/Transforms/IfTransform/IfTransform.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // 2. Hoist instructions in the 'else' block
+    for (auto &operation : ifOp.getElseRegion().getOps()) {
+      auto distance = std::distance(ifOp.getElseRegion().getOps().begin(),
+                                    ifOp.getElseRegion().getOps().end());
+      if (distance == 1) {
+        break;
+      }
+      rewriter.setInsertionPointToStart(ifOp.thenBlock());
+      rewriter.moveOpBefore(&operation, ifOp);
+    }
```
AlexanderViand-Intel, line 47: I think you should be able to use `llvm::make_early_inc_range` to simplify this loop. I'm actually not entirely sure the current solution is "correct" as the `moveOpBefore` might invalidate the for loop iterator (even if it works in practice), but I don't remember enough C++ stuff to know for sure ;)

lib/Transforms/IfTransform/run.sh
AlexanderViand-Intel, line 1: I'm guessing this shouldn't be checked in ;)

AlexanderViand-Intel (changes requested):
👍  Let's work on getting the tests rewritten to FileCheck today and then we can focus on the required data analyses next :)

tests/if_transform/secret_condition_with_secret_tensor.mlir
```diff
@@ -0,0 +1,27 @@
+module {
```
AlexanderViand-Intel, line 1: Placeholder comment to remind me about FileCheck tests

lib/Transforms/IfTransform/IfTransform.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // 2. Hoist instructions in the 'else' block
+    for (auto &operation : ifOp.getElseRegion().getOps()) {
+      auto distance = std::distance(ifOp.getElseRegion().getOps().begin(),
+                                    ifOp.getElseRegion().getOps().end());
+      if (distance == 1) {
+        break;
+      }
```
AlexanderViand-Intel, line 44: I think you should be able to use `llvm::make_early_inc_range` to simplify this loop. I'm actually not entirely sure the current solution is "correct" as the `moveOpBefore` might invalidate the for loop iterator (even if it works in practice), but I don't remember enough C++ stuff to know for sure ;)

tools/heir-opt.cpp
```diff
@@ -503,6 +504,7 @@ int main(int argc, char **argv) {
   registerElementwiseToAffinePasses();
   registerSecretizePasses();
   registerFullLoopUnrollPasses();
+  registerIfTransformPasses();  // Register IfTransfrom pass
```
AlexanderViand-Intel, line 507: I think this comment is a bit redundant ;)
```suggestion
  registerIfTransformPasses();
```

tests/if_transform/secret_condition_with_secret_tensor.mlir
AlexanderViand-Intel: Also: note to add tests with if statements that do not yield any values, including ones that do not have an else branch

lib/Transforms/IfTransform/IfTransform.td
```diff
@@ -0,0 +1,15 @@
+#ifndef LIB_TRANSFORMS_IFTRANSFORM_IFTRANSFORM_TD_
+#define LIB_TRANSFORMS_IFTRANSFORM_IFTRANSFORM_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def IfTransform : Pass<"if-transform"> {
```
j2kun, line 6: Could we name this pass to better describe what it does? Perhaps `convert-if-to-select` like the pattern name.

AlexanderViand-Intel: Great timing, I was just discussing this with Meron in our meeting while you posted this! We will change it to fit MLIR naming convetions :)

lib/Transforms/IfTransform/IfTransform.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // 1. Hoist instructions in the 'then' block
+    for (auto &operation : ifOp.getThenRegion().getOps()) {
+      auto distance = std::distance(ifOp.getThenRegion().getOps().begin(),
+                                    ifOp.getThenRegion().getOps().end());
+      if (distance == 1) {
+        break;
+      }
+      rewriter.setInsertionPointToStart(ifOp.thenBlock());
+      rewriter.moveOpBefore(&operation, ifOp);
+    }
```
AlexanderViand-Intel, line 37: For example (as discussed in our meeting)
```suggestion
    for (auto &operation :
         llvm::make_early_inc_range(ifOp.getElseRegion().getOps())) {
      if (!llvm::isa<scf::YieldOp>(operation)) {
        rewriter.setInsertionPointToStart(ifOp.thenBlock());
        rewriter.moveOpBefore(&operation, ifOp);
      }
    }
```

tests/if_transform/secret_condition_with_secret_tensor.mlir
AlexanderViand-Intel: PS: you should be able to add all tests into a single file

lib/Transforms/IfTransform/IfTransform.cpp
MeronZerihun: This looks more precise, I will do something similar.

lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.td
```diff
@@ -0,0 +1,14 @@
+#ifndef LIB_TRANSFORMS_CONVERTIFTOSELECT_CONVERTIFTOSELECT_TD_
+#define LIB_TRANSFORMS_CONVERTIFTOSELECT_CONVERTIFTOSELECT_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def ConvertIfToSelect : Pass<"convert-if-to-select"> {
+  let summary = "Convert If-operations on secret conditions to select operations.";
+  let description = [{ Conversion for If-operations that evaluate secret condition(s) to their alternative select operations. }];
+  let dependentDialects = [
+    "mlir::scf::SCFDialect",
```
AlexanderViand-Intel, line 10: dependentDialects needs to include the dialects of (new) operations the pass might introduce (see https://mlir.llvm.org/docs/PassManagement/#dependent-dialects).
Without this, a program that has an `scf.if` but not yet any `arith` ops could crash when the pass tries to create the `arith.select` op, due to Arith not being loaded yet.
Technically, I think that means SCF isn't even required (if there ar no `scf` ops in the program, the pass doesn't do anything) but it probably doesn't hurt to force it being loaded.

```suggestion
    "mlir::scf::SCFDialect",
    "mlir::arith::ArithDialect"
```

lib/Transforms/ConvertIfToSelect/BUILD
```diff
@@ -0,0 +1,43 @@
...
+    srcs = ["ConvertIfToSelect.cpp"],
+    hdrs = ["ConvertIfToSelect.h"],
+    deps = [
+        ":pass_inc_gen",
+        "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:SCFDialect",
+        "@llvm-project//mlir:Support",
+        "@llvm-project//mlir:Transforms",
+    ],
```
AlexanderViand-Intel, line 19: It probably gets the Arith dependency transivitely from SCF or something else already, but I think the pass pass should explicitly depend on ArithDialect

```suggestion
    deps = [
        ":pass_inc_gen",
        "@llvm-project//mlir:IR",
        "@llvm-project//mlir:Pass",
        "@llvm-project//mlir:ArithDialect",
        "@llvm-project//mlir:SCFDialect",
        "@llvm-project//mlir:Support",
        "@llvm-project//mlir:Transforms",
    ],
```

lib/Transforms/ConvertIfToSelect/BUILD
```diff
@@ -0,0 +1,43 @@
...
+
+cc_library(
+    name = "ConvertIfToSelect",
+    srcs = ["ConvertIfToSelect.cpp"],
+    hdrs = ["ConvertIfToSelect.h"],
+    deps = [
+        ":pass_inc_gen",
+        "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:SCFDialect",
```
AlexanderViand-Intel, line 16: It probably gets the Arith dependency transivitely from SCF or something else already, but I think the pass pass should explicitly depend on ArithDialect

```suggestion
        "@llvm-project//mlir:SCFDialect",
        "@llvm-project//mlir:ArithDialect",
```

lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
```diff
@@ -0,0 +1,72 @@
+#include "lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.h"
+
+#include "mlir/include/mlir/Dialect/SCF/IR/SCF.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"        // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"     // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/Passes.h"  // from @llvm-project
+
+using namespace mlir;
+using namespace mlir::scf;
```
AlexanderViand-Intel, line 10: While there's sometimes good reasons to use `using namespace ...`, I don't think it's necessary here.
The first one isn't necessary anyway, as everything is already happening in `namespace mlir {...}`.
For the second one, I think it'd actually be clearer to write `scf::IfOp` instead of `IfOp` to make it clearer what exact op you're targeting.

lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
```diff
@@ -0,0 +1,72 @@
...
+  LogicalResult matchAndRewrite(IfOp ifOp,
+                                PatternRewriter &rewriter) const override {
+    // Hoist instructions in the 'then' and 'else' regions
+    auto thenOps = ifOp.getThenRegion().getOps();
+    auto elseOps = ifOp.getElseRegion().getOps();
+
+    rewriter.setInsertionPointToStart(ifOp.thenBlock());
+    for (auto &operation : llvm::make_early_inc_range(
+             llvm::concat<Operation>(thenOps, elseOps))) {
+      if (!llvm::isa<scf::YieldOp>(operation)) {
```
AlexanderViand-Intel, line 30: While we cannot check the necessity of the conversion without the secretness dataflow analysis, I just realized we can probably already thow a hard error here if the operation to be hoisted does not have the `NoMemoryEffect` and `AlwaysSpeculatable` traits (i.e. is `Pure`). 

It'd probably also be a good idea to add a TODO here to extend support to `ConditionallySpeculatable`, which includes operations such as `arith.divui` which is only safely hoistable if we can statically exlcude a zero division.
To do this "properly", first create an issue for the new task, then add `//TODO (#123): ....` (but replacing 123 with your issue's number). This way, the CI system will automatically pick this up and link it with the issue.

tests/convert_if_to_select/check_error_msgs/invalid_conditionals.mlir
```diff
@@ -0,0 +1,39 @@
+// TODO: Add RUN command and file check for checking diagnostic messages
```
AlexanderViand-Intel, line 1: See last comment: if we add the hard fail on non-pure ops, this test file can be enabled (see any [invalid-ops.mlir](https://github.com/google/heir/blob/main/tests/arith_ext/invalid-ops.mlir) file in the various test folders for examples for how to do this.)

tests/convert_if_to_select/expected_outputs.mlir
```diff
@@ -0,0 +1,160 @@
...
+  %0 = secret.generic ins(%inp, %cond : i16, !secret.secret<i1>) {
+    ^bb0(%copy_inp: i16, %secret_cond: i1):
+      %2 = arith.addi %copy_inp, %copy_inp : i16
+      %1 = arith.select %secret_cond, %2, %copy_inp : i16
+      secret.yield %1 : i16
+    } -> !secret.secret<i16>
+    return %0 : !secret.secret<i16>
+}
+
+// -----
```
AlexanderViand-Intel, line 21: Remove all of these from the (non-failing) tests, as they're only parsed by `mlir-opt`/`heir-opt` if you pass `-split-input-file`, and are only needed to test errors.
The `CHECK-LABEL` already does some of the work here in splitting test cases. If you want to incrase the isolation between test cases even further, you can add [the `--enable-var-scope` option](https://llvm.org/docs/CommandGuide/FileCheck.html#cmdoption-FileCheck-enable-var-scope) to the `FileCheck` arguments.

tests/convert_if_to_select/non_secret_conditionals.mlir
AlexanderViand-Intel, line 1: If we go ahead with merging this PR without the dataflow analysis, this file should be removed from this PR and added to the dataflow analysis PR

AlexanderViand-Intel (commented):
Thanks for putting this together! Despite my many (mostly nit-picky) comments, I think this actually looks great, especially the `IfToSelectConversion` Pattern is significantly more elegant/simple than I would have initially expected!

AlexanderViand-Intel (changes requested):
Thanks for putting this together! Despite my many (mostly nit-picky) comments, I think this actually looks great, especially the `IfToSelectConversion` Pattern is significantly more elegant/simple than I would have initially expected!

lib/Transforms/ConvertIfToSelect/BUILD
j2kun: Google has an internal rule that says something like "all included files require direct dependencies exporting those files," but fixing this is part of our internal process for handling contributed PRs. Long story short, if you don't do this then we'll do it automatically for you :)

lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
j2kun: For reporting an error, use `ifOp.emitError() << "message"; return failure();`

But I think it is also fine to omit the safety checks in the interim. Nothing is going to production any time soon :)

tests/convert_if_to_select/check_error_msgs/invalid_conditionals.mlir
j2kun, line 1: Let's put this test file in the same directory as the others. I think so long as there's no RUN line it should not produce any problems. But if it does fail for some reason, you can just do `// RUN: heir-opt %s` and it will just check that the IR is valid.

tests/convert_if_to_select/non_secret_conditionals.mlir
j2kun: You can leave it in with a TODO, since the RUN line doesn't actually run the pass.

j2kun:
In lieu of a proper greeting during the meeting this morning, Hello @MeronZerihun! It's great to have you in our band of merry cryptographers. This PR looks very well done and thorough so far, and I'll give it a second pass after you address Alex's review comments.

A dataflow analysis is the right approach, but in the interim you can get away with something as simple as this, which checks if the condition to the `if` is a direct operand of the `secret.generic` op containing the if:

```cpp
if (auto genericOp = ifOp.getParentOfType<secret::GenericOp>()) {
  if (OpOperand operand = genericOp.getOpOperandForBlockArgument(condition)) {
    // it's an OpOperand to the containing secret.generic
    if (isa<secret::SecretType>(operand.get().getType())) {
      // it's a secret!
    }
  }
}
```

lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
AlexanderViand-Intel: Turns out, `isPure(...)` in C++ doesn't actually check for the equivalent components of the `Pure` trait (a tablegen-only thing ) but instead does a proper check which also handles conditionally speculatable ops, so we get this for free :)

tests/convert_if_to_select/expected_outputs.mlir
```diff
@@ -0,0 +1,160 @@
...
+  %1 = secret.generic ins(%inp: !secret.secret<i16>) {
+    ^bb0(%secret_inp: i16):
+      %2 = arith.cmpi eq, %secret_inp, %0 : i16
+      %4 = arith.addi %secret_inp, %secret_inp : i16
+      %3 = arith.select %2, %4, %secret_inp : i16
+      secret.yield %3 : i16
+  } -> !secret.secret<i16>
+
+  return %1 : !secret.secret<i16>
+}
```
j2kun, line 160: Is the purpose of this file is just to show the expected outputs of `secret_conditionals` in plaintext MLIR instead of regex-laden `CHECK` statements? I don't particularly see a need for this, but if you want to keep it for readability, then you should remove the check statements from this file and just `// RUN: heir-opt %s` to ensure the IR parses and verifies.

```suggestion
// RUN: heir-opt  %s

func.func @secret_condition_with_non_secret_int(%inp: i16, %cond: !secret.secret<i1>) -> !secret.secret<i16> {
  %0 = secret.generic ins(%inp, %cond : i16, !secret.secret<i1>) {
    ^bb0(%copy_inp: i16, %secret_cond: i1):
      %2 = arith.addi %copy_inp, %copy_inp : i16
      %1 = arith.select %secret_cond, %2, %copy_inp : i16
      secret.yield %1 : i16
    } -> !secret.secret<i16>
    return %0 : !secret.secret<i16>
}


func.func @secret_condition_with_secret_int(%inp: !secret.secret<i16>, %cond: !secret.secret<i1>) -> !secret.secret<i16> {
  %0 = secret.generic ins(%inp, %cond : !secret.secret<i16>, !secret.secret<i1>) {
    ^bb0(%secret_inp: i16, %secret_cond: i1):
      %2 = arith.addi %secret_inp, %secret_inp : i16
      %1 = arith.select %secret_cond, %2, %secret_inp : i16
      secret.yield %1 : i16
    } -> !secret.secret<i16>
    return %0 : !secret.secret<i16>
}

func.func @secret_condition_with_secret_int_and_multiple_yields(%inp: !secret.secret<i16>, %cond: !secret.secret<i1>) -> !secret.secret<i16> {
  %0 = secret.generic ins(%inp, %cond : !secret.secret<i16>, !secret.secret<i1>) {
  ^bb0(%secret_inp: i16, %secret_cond: i1):
    %2 = arith.addi %secret_inp, %secret_inp : i16
    %4 = arith.muli %secret_inp, %2 : i16
    %1 = arith.select %secret_cond, %2, %secret_inp : i16
    %3 = arith.select %secret_cond, %4, %secret_inp : i16
    %5 = arith.addi %1, %3 : i16
    secret.yield %5 : i16
  } -> !secret.secret<i16>
  return %0 : !secret.secret<i16>
}

func.func @secret_condition_with_secret_tensor(%inp: !secret.secret<tensor<16xi16>>, %cond: !secret.secret<i1>) -> !secret.secret<tensor<16xi16>> {
  %0 = secret.generic ins(%inp, %cond : !secret.secret<tensor<16xi16>>, !secret.secret<i1>) {
    ^bb0(%secret_inp: tensor<16xi16>, %secret_cond: i1):
      %2 = arith.addi %secret_inp, %secret_inp : tensor<16xi16>
      %3 = arith.muli %secret_inp, %secret_inp : tensor<16xi16>
      %1 = arith.select %secret_cond, %2, %3 : tensor<16xi16>
      secret.yield %1 : tensor<16xi16>
    } -> !secret.secret<tensor<16xi16>>
    return %0 : !secret.secret<tensor<16xi16>>
}

func.func @non_secret_condition_with_secret_tensor(%inp: !secret.secret<tensor<16xi16>>, %cond: i1) -> !secret.secret<tensor<16xi16>> {
  %0 = secret.generic ins(%inp, %cond : !secret.secret<tensor<16xi16>>, i1) {
    ^bb0(%secret_inp: tensor<16xi16>, %copy_cond: i1):
      %1 = scf.if %copy_cond -> (tensor<16xi16>) {
        %2 = arith.addi %secret_inp, %secret_inp : tensor<16xi16>
        scf.yield %2 : tensor<16xi16>
      } else {
        scf.yield %secret_inp : tensor<16xi16>
      }
      secret.yield %1 : tensor<16xi16>
    } -> !secret.secret<tensor<16xi16>>
    return %0 : !secret.secret<tensor<16xi16>>
}

func.func @secret_condition_with_secret_vector(%inp: !secret.secret<vector<4xf32>>, %cond: !secret.secret<i1>) -> !secret.secret<vector<4xf32>> {
  %0 = secret.generic ins(%inp, %cond : !secret.secret<vector<4xf32>>, !secret.secret<i1>) {
  ^bb0(%secret_inp: vector<4xf32>, %secret_cond: i1):
    %2 = arith.addf %secret_inp, %secret_inp : vector<4xf32>
    %1 = arith.select %secret_cond, %2, %secret_inp : vector<4xf32>
    secret.yield %1 : vector<4xf32>
  } -> !secret.secret<vector<4xf32>>
  return %0 : !secret.secret<vector<4xf32>>
}

func.func @tainted_condition(%inp: !secret.secret<i16>) -> !secret.secret<i16>{
  %0 = arith.constant 0 : i16
  %1 = secret.generic ins(%inp: !secret.secret<i16>) {
    ^bb0(%secret_inp: i16):
      %2 = arith.cmpi eq, %secret_inp, %0 : i16
      %4 = arith.addi %secret_inp, %secret_inp : i16
      %3 = arith.select %2, %4, %secret_inp : i16
      secret.yield %3 : i16
  } -> !secret.secret<i16>

  return %1 : !secret.secret<i16>
}
```

j2kun (approved):
LGTM! One optional change suggested, then let's get it merged.

tests/convert_if_to_select/expected_outputs.mlir
AlexanderViand-Intel, line 1: Good catch @j2kun! This shouldn't be in the final PR, as it was purely for development/didactic purposes :)

tests/convert_if_to_select/secret_conditionals.mlir
AlexanderViand-Intel, line 1: Let's add the test for `arith.divui` with a statically-known, non-zero divisor we discussed in our meeting, to make sure we "record" the expected behavior/that this is supported at all.

AlexanderViand-Intel (approved):
LGTM, too, just some minor cleanup around the tests and then we should be able to get this merged👍

AlexanderViand-Intel:
image
@j2kun / @asraa Any idea why the other actions aren't showing up after the last push?

j2kun:
That is bizarre... maybe GitHub had a brain fart. Try force-pushing again with a trivial change?

asraa:
Hey! @MeronZerihun super big apologies, but could you rebase cd909c851bdb948dc14704a9e43d22b56a80052d on top of main? (Sorry for attempting to resolve the commit with my last merge commit, I had a mental fart too and realized we needed to squash commits)

AlexanderViand-Intel:
Oh, wow, this PR is really hitting all the CI issues xD Now we've got a `Google internal checks FAILED`? 
@asraa Is it something we need to fix on our side, or something you can fix internally?

j2kun:
The internal checks failing is normal. It just means we have to click a few
extra buttons to get it to pass internal CI, which is usually build rule
fixes that only apply internally.

asraa:
> Oh, wow, this PR is really hitting all the CI issues xD Now we've got a Google internal checks FAILED?

Ah yes, all good now! I can now take it from here to fix up the build issues and merge :)

PR #751, opened 2024-06-28
explain how to point HEIR to local llvm

j2kun:


docs/content/en/docs/ide_configuration.md
```diff
@@ -65,3 +67,33 @@ The main things that cause a rebuild are:
...
+        # this BUILD file is intentionally empty, because the LLVM project
+        # internally contains a set of bazel BUILD files overlaying the project.
+        build_file_content = "# empty",
+        path = "/path/to/llvm-project",
+    )
+EOF
+```
+
+This will require a full rebuild if the checked-out LLVM commit differs from
+the pinned commit hash in `bazel/import_llvm.bzl`.
```
AlexanderViand-Intel, line 96: Can you expand a bit on the most efficient build workflow with this setup? If you want to run the MLIR tests/etc, do you have to build LLVM twice (once from the LLVM dir (using LLVM's CMake or bazel setup) and then a rebuild from HEIR as bazel sees the changed source files), or is there a way to avoid that (other than `ccache`)? E.g., can I run the mlir tests via HEIR's bazel setup, or is bazel smart enough to realize LLVM has already been built?

AlexanderViand-Intel (approved):
Very nice, thanks for putting this together!

docs/content/en/docs/ide_configuration.md
j2kun: The short answer is you would have to do everything through bazel. Added a commit with more details.

asraa (approved):
Thank you! 

The only small feedback I have is mentioning some examples of when this might be necessary (particularly because we use an in-progress / developing implementation of the polynomial dialect upstream).

j2kun:
> Thank you!
> 
> The only small feedback I have is mentioning some examples of when this might be necessary (particularly because we use an in-progress / developing implementation of the polynomial dialect upstream).

Done

Issue #686, opened 2024-05-22
Add an importer for AutoHoG json netlists


j2kun:
We're discussing integration with AutoHoG (see https://www.x-mol.net/paper/article/1761273823319789568).

The authors have a json spec for their optimized circuit output, and we want to enable this as an entry point to the rest of the compiler. Later, we will likely add an exit point to go from HEIR to AutoHoG's input specification.

asraa:
Just to CC @sbian3 @Lavendes

Lavendes:
The AutoHog JSON file is formatted :
{
    "circuit_name": <circuit_name>
    "ports": {
        <node_id>: <port_details>,
        ...
    },
    "cells": {
        <node_id>: <cell_details>,
        ...
    },
}

Where <port_details> is:
{
    "direction":  <"input" | "output">,
    "bits": <bit_vector>              // a node id list that defines the nodes (gates) connect from the input or connect to the output
}

and <cell_details> is:
{
    "hide_name": <1 | 0>
    "type": <cell_type>               // the gate type, including the conventional gates "AND", "OR", "XOR"... as well as our "HomGateS", "HomGateM"
    "parameters":{}                   // built-in parameters of yosys, default is empty
    "attributes":{}                   // built-in attributes of yosys, default is empty
    "port_directions":                // the inputs and outputs of the gate 
    {
        <port_name>: <"input" | "output">,
        ...
    },                                 
    "connections":                    // node id of the inputs and outputs
    {
        <port_name>: <bit_vector>,
        ...
    },
    "weights":                        // the weight for each input port
    {
        <port_name>: <weight>,        // weight is a int
        ...
    },
    "tableT":                         // the tableT (LUT) for each output port
    {
        <port_name>: <tableT>,        // Lut is a list of a 0-1 sequence, with a maximum length of 32 (under the parameter n=2048). If not full, 0s are padded for the remaining positions during PBS
        ...
    }                                 // If "type" is not "HomGate", both "weights" and "tableT" are empty
}

asraa:
Hey @Lavendes , Thank you for the spec! Two quick questions about it

1. Where is the constant offset for the linear combo of the HomGate? Is it contained in `weights` but with some no-op input port?
2. Are HomGateS and HomGateM the single and multiple output gates? Or is it something else? (So I should expect one tableT for HomGateS and multiple for HomGateM?)

If you have any example inputs you can paste or commit into a test file, that would be greatly appreciated! I can work on the ingestion pass.

[...]

PR #756, opened 2024-06-28
Add AutoHoG JSON importer

j2kun:
Fixes https://github.com/google/heir/issues/686


lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,516 @@
...
+    const int coeffValue = itr->value.GetInt();
+    assert(inputPortNameToOperandIndex.contains(portName) &&
+           "Coefficient port name not found in inputPortNameToOperandIndex");
+    result[inputPortNameToOperandIndex.at(portName)] = coeffValue;
+  }
+}
+
+OwningOpRef<Operation *> translateFromAutoHog(llvm::StringRef inputString,
+                                              MLIRContext *context) {
+  // FIXME: shouldn't have to manually load these dialects!
```
asraa, line 230: Ah, this might be a bug in `TranslateToMLIRRegistration`?

lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,516 @@
...
+  int numInputs = 0;
+  int numOutputs = 0;
+  for (auto &[portName, port] : ioPorts) {
+    if (port.type == PortType::INPUT) {
+      inputPortToTensorIndex[portName] = numInputs++;
+    } else {
+      outputPortToTensorIndex[portName] = numOutputs++;
+    }
+  }
+  // FIXME: detect proper minBitWidth from the circuit
```
asraa, line 281: could you please link this to the importer issue?

lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,516 @@
...
+  int minBitWidth = 3;
+  Type ciphertextType = lwe::LWECiphertextType::get(
+      context, lwe::UnspecifiedBitFieldEncodingAttr::get(context, minBitWidth),
+      lwe::LWEParamsAttr());
+  Type inputType = RankedTensorType::get({numInputs}, ciphertextType);
+  Type outputType = RankedTensorType::get({numOutputs}, ciphertextType);
```
asraa, line 287: I was just curious why make the choice to use single tensor to hold all the inputs - do we even know how the input bits are distributed in different multi-bit inputs? Or how they should be ordered within the tensor?

I wonder if it might also be worth annotating the function inputs with the port names from the JSON (but that would only be possible if the inputs were unpacked)

lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,516 @@
...
+    } else if (strcmp(cellType, "OR") == 0) {
+      op = builder.create<cggi::OrOp>(funcOp.getLoc(), resultTypes, operands);
+    } else if (strcmp(cellType, "XOR") == 0) {
+      op = builder.create<cggi::XorOp>(funcOp.getLoc(), resultTypes, operands);
+    } else if (strcmp(cellType, "XNOR") == 0) {
+      op = builder.create<cggi::XNorOp>(funcOp.getLoc(), resultTypes, operands);
+    } else if (strcmp(cellType, "NOT") == 0) {
+      op = builder.create<cggi::NotOp>(funcOp.getLoc(), resultTypes, operands);
+    } else {
+      assert(false && "Unknown cell type");
```
asraa, line 452: in release mode we would skip this assertion, should we actually return a failure? or maybe at least a debug message?

lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,516 @@
...
+                 << cellPort.cellName.value_or("") << "\n");
+      Operation *op = cellsByName[cellPort.cellName.value().c_str()];
+      int resultIndex =
+          cellOutputPortToResultIndex.at(cellPort.cellName.value_or(""))
+              .at(cellPort.portName);
+      outputValues[outputPortToTensorIndex[portName]] =
+          op->getResult(resultIndex);
+      continue;
+    }
+  }
```
asraa, line 506: should there also be a JSON failure here detecting if there was no connection to that output port? i guess, to me there's also a possibility that an output port is a constant, which i don't think would appear in either of these two cases.

lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,516 @@
...
+  auto &cells = document["cells"];
+  assert(cells.IsObject() && "Expected 'cells' to be an object");
+  for (rapidjson::Value::ConstMemberIterator itr = cells.MemberBegin();
+       itr != cells.MemberEnd(); ++itr) {
+    const rapidjson::Value &cell = itr->value;
+    assert(cell.IsObject() && "Expected cell to be an object");
+    const char *cellName = cell["cell_name"].GetString();
+    cellGraph.addVertex(std::string(cellName));
+    auto &connections = cell["connections"];
+    assert(connections.IsObject() && "Expected 'connections' to be an object");
```
asraa, line 164: same with the comment here and below - these assertions seem like they're catching JSON inputs that would be invalid, and that these would be skipped in release mode and probably cause unexpected confusion. it *might* be worth putting some kind of input error here instead? or we can i suppose always hope the input will be valid..

asraa (commented):
Nice! Definitely a lot of work, so I really appreciate it!

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: lol this will be the third time I fix dialect loading errors in mlir-translate

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: Maybe I am misunderstanding the autohog work, but I was under the  impression that the inputs to the circuit are still single bit ciphertexts.

asraa: No! You are probably right, I think what I'm asking for on second thought would be work that we would need to handle if we export a circuit to an AutoHoG input - we probably want to keep track of some circuit metadata so we can map inputs to their single bit ciphertexts.

The question was more about packing them all into a single argument rather than having unpacked single bit ciphertext inputs. The former I guess just makes the IR easier to read? It won't matter if we use that metadata to create client input/output handlers in the end.

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: The dialects are loaded lazily, when something from that dialect is parsed. Nothing in the input is MLIR in this case.

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: Yeah, I basically just picked arbitrarily. I was thinking about how we often use `tensor<lwe.ciphertext>`, and I kind of expect these IRs to be quite big (a 64-bit adder, e.g.). We can put whatever metadata we want on the function (a DictionaryAttr mapping int to string, for example). I feel like the right time to do that is when the need arises.

asraa: Agreed - noted for now it's somewhat arbitrary and when we have a full integration it might make sense to use metadata/reconsider

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: I believe most of these type assertions would fail on the very next line when iteration is attempted. I'm OK leaving them as debug assertions for now.

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: Done

j2kun:
Still TODO: convert ports from object to array

j2kun:
This one should be ready for a final review now.

lib/Source/AutoHog/AutoHogImporter.cpp
```diff
@@ -0,0 +1,530 @@
...
+  assert(jsonConnections.IsObject() &&
+         "Expected 'connections' to be an object");
+  for (rapidjson::Value::ConstMemberIterator itr =
+           jsonConnections.MemberBegin();
+       itr != jsonConnections.MemberEnd(); ++itr) {
+    const char *localPortName = itr->name.GetString();
+    const rapidjson::Value &portData = itr->value;
+    assert(portData.IsObject() &&
+           "Expected 'connections[...]' to be an object");
+    const char *connectedPortName = portData["port"].GetString();
```
asraa, line 126: I don't really understand why this value is always needed - in toposort.json the AND cell's port `$0$` connection contains both `cell` and `port` - 

```
        "$0$": {
          "cell": "input_cell",
          "port": "$2$"
        },
```

Shouldn't it either be connected to a port or a cell? Why does this field have both? It's also confusing because that port name is the same name as a local port name...

asraa (approved):
Thanks! Really small question about the spec - after this is merged we can file a follow-up issue to add a README.md doc about the spec.

lib/Source/AutoHog/AutoHogImporter.cpp
j2kun: I feel the names are also confusing, but the way I understand it is that `port` refers to one of:

- Global circuit input/output (when `cell` is not present)
- Local cell input/output (when `cell` is present)

The problem is that `$2$` is used many times in the example JSON, in each cell that has at least three inputs. So the way to disambiguate is either to give everything unique names, or add the name of the cell as a sort of namespace for the ports local to that cell.

Issue #729, opened 2024-06-07
test: re-enable openfhe end to end tests after #727 is fixed upstream


asraa:

j2kun:
Here is the upstream WIP https://github.com/llvm/llvm-project/pull/94838

So far I've reproduced the error but don't know how to fix it.

j2kun:
Can reenable once https://github.com/llvm/llvm-project/pull/96242 is integrated

j2kun:
The upstream fix is merged, but I'm still getting segfaults.

PR #760, opened 2024-07-01
Add an analysis for calculating the multicative depth

Maokami:
This PR contains the multicative depth analysis pass for OpenFHE.(#661)

It seems to be working well, but needs refactoring and additional testing.

Testing is blocked by #729.

### DIFF
--- a/lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
+++ b/lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
@@ -1,0 +1,64 @@
+#include "lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h"  // from @llvm-project
+
+#include "lib/Dialect/Openfhe/IR/OpenfheOps.h"
+#include "llvm/include/llvm/ADT/TypeSwitch.h"  // from @llvm-project
+#include "llvm/include/llvm/Support/Debug.h"   // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"    // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"        // from @llvm-project
+#include "mlir/include/mlir/IR/Visitors.h"     // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"    // from @llvm-project
+
+#define DEBUG_TYPE "openfhe-muldepth-analysis"
+
+namespace mlir {
+namespace heir {
+namespace mul_depth_analysis {
+
+void MulDepthAnalysis::visitOperation(
+    Operation *op, ArrayRef<const MulDepthLattice *> operands,
+    ArrayRef<MulDepthLattice *> results) {
+  llvm::TypeSwitch<Operation &>(*op)
+      .Case<openfhe::MulOp, openfhe::MulPlainOp, openfhe::MulConstOp,
+            openfhe::MulNoRelinOp>([&](auto mulOp) {
+        LLVM_DEBUG(
+            { llvm::dbgs() << "Visiting Mul: " << mulOp->getName() << "\n"; });
+        MulDepthLattice *r = results[0];
+        MulDepth lhsMulDepth = operands[1]->getValue();
+        if (lhsMulDepth.isInitialized()) {
+          lhsMulDepth = MulDepth{lhsMulDepth.getValue() + 1};
+        } else {
+          lhsMulDepth = MulDepth{1};
+        }
+        MulDepth rhsMulDepth = operands[2]->getValue();
+        if (rhsMulDepth.isInitialized()) {
+          rhsMulDepth = MulDepth{rhsMulDepth.getValue() + 1};
+        } else {
+          rhsMulDepth = MulDepth{1};
+        }
+        LLVM_DEBUG({
+          llvm::dbgs() << "lhsMulDepth: " << lhsMulDepth << "\n";
+          llvm::dbgs() << "rhsMulDepth: " << rhsMulDepth << "\n";
+        });
+        ChangeResult result = r->join(MulDepth::join(lhsMulDepth, rhsMulDepth));
+        propagateIfChanged(r, result);
+        LLVM_DEBUG({
+          llvm::dbgs() << "MulDepth: " << results[0]->getValue() << "\n";
+        });
+      })
+      .Default([&](Operation &defaultOp) {
+        LLVM_DEBUG(
+            { llvm::dbgs() << "Visiting: " << defaultOp.getName() << "\n"; });
+        for (const MulDepthLattice *operand : operands) {
+          for (MulDepthLattice *r : results) {
+            ChangeResult result = r->join(*operand);
+            propagateIfChanged(r, result);
+          }
+        }
+      });
+}
+
+void MulDepthAnalysis::setToEntryState(MulDepthLattice *lattice) {}
+
+}  // namespace mul_depth_analysis
+}  // namespace heir
+}  // namespace mlir
--- a/lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
+++ b/lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
@@ -1,0 +1,81 @@
+#ifndef LIB_ANALYSIS_MULDEPTHANALYSIS_MULDEPTHANALYSIS_H_
+#define LIB_ANALYSIS_MULDEPTHANALYSIS_MULDEPTHANALYSIS_H_
+
+#include "mlir/include/mlir/Analysis/DataFlow/SparseAnalysis.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Diagnostics.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"    // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"        // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace mul_depth_analysis {
+
+class MulDepth {
+ public:
+  MulDepth() : value(std::nullopt) {}
+  MulDepth(int64_t value) : value(value) {}
+  ~MulDepth() = default;
+
+  bool isInitialized() const { return value.has_value(); }
+
+  const int64_t &getValue() const {
+    assert(isInitialized());
+    return *value;
+  }
+
+  bool operator==(const MulDepth &rhs) const { return value == rhs.value; }
+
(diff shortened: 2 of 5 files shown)


lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "mlir/include/mlir/IR/Operation.h"    // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"        // from @llvm-project
+#include "mlir/include/mlir/IR/Visitors.h"     // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"    // from @llvm-project
+
+#define DEBUG_TYPE "openfhe-muldepth-analysis"
+
+namespace mlir {
+namespace heir {
+namespace mul_depth_analysis {
```
j2kun, line 15: This inner namespace is probably not necessary.

j2kun, line 15: This inner namespace is probably not necessary.

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+void MulDepthAnalysis::visitOperation(
+    Operation *op, ArrayRef<const MulDepthLattice *> operands,
+    ArrayRef<MulDepthLattice *> results) {
+  llvm::TypeSwitch<Operation &>(*op)
+      .Case<openfhe::MulOp, openfhe::MulPlainOp, openfhe::MulConstOp,
+            openfhe::MulNoRelinOp>([&](auto mulOp) {
+        LLVM_DEBUG(
+            { llvm::dbgs() << "Visiting Mul: " << mulOp->getName() << "\n"; });
```
j2kun, line 24: note, for a single line you can omit the `{ }` and the semicolon in the inner statement. Maybe this is just slightly over the line limit?

```suggestion
        LLVM_DEBUG(llvm::dbgs() << "Visiting Mul: " << mulOp->getName() << "\n");
```

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+namespace mlir {
+namespace heir {
+namespace mul_depth_analysis {
+
+void MulDepthAnalysis::visitOperation(
+    Operation *op, ArrayRef<const MulDepthLattice *> operands,
+    ArrayRef<MulDepthLattice *> results) {
+  llvm::TypeSwitch<Operation &>(*op)
+      .Case<openfhe::MulOp, openfhe::MulPlainOp, openfhe::MulConstOp,
```
j2kun, line 21: MulPlain and MulConst I think should not consume multiplicative depth. Or rather, if we want to track those it depends on the const/plain value on the other side, and that requires a more fine-grained noise analysis. I'd say leave them out. cc @AlexanderViand-Intel

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -0,0 +1,81 @@
...
+  bool operator==(const MulDepth &rhs) const { return value == rhs.value; }
+
+  static MulDepth join(const MulDepth &lhs, const MulDepth &rhs) {
+    if (!lhs.isInitialized()) return rhs;
+    if (!rhs.isInitialized()) return lhs;
+    return MulDepth{lhs.getValue() > rhs.getValue() ? lhs.getValue()
+                                                    : rhs.getValue()};
+  }
+
+  void print(raw_ostream &os) const { os << value; }
```
j2kun, line 35: Nit: when debugging there can be many printed values of various analysis, and it's hard to tell which is which without a name.

```suggestion
  void print(raw_ostream &os) const { os << "MulDepth(" << value << ")"; }
```

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -0,0 +1,81 @@
...
+
+ private:
+  std::optional<int64_t> value;
+
+  friend mlir::Diagnostic &operator<<(mlir::Diagnostic &diagnostic,
+                                      const MulDepth &foo) {
+    if (foo.isInitialized()) {
+      return diagnostic << foo.getValue();
+    }
+    return diagnostic << "uninitialized";
```
j2kun, line 45:
```suggestion
    return diagnostic << "MulDepth(uninitialized)";
```

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
```diff
@@ -0,0 +1,64 @@
...
+            openfhe::MulNoRelinOp>([&](auto mulOp) {
+        LLVM_DEBUG(
+            { llvm::dbgs() << "Visiting Mul: " << mulOp->getName() << "\n"; });
+        MulDepthLattice *r = results[0];
```
j2kun, line 25: Add an assert here that there is only one result. I think that will help when this pass is eventually generalized to support multiple dialects via interface.

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
```diff
@@ -0,0 +1,64 @@
...
+        for (const MulDepthLattice *operand : operands) {
+          for (MulDepthLattice *r : results) {
+            ChangeResult result = r->join(*operand);
+            propagateIfChanged(r, result);
+          }
+        }
+      });
+}
+
+void MulDepthAnalysis::setToEntryState(MulDepthLattice *lattice) {}
```
j2kun, line 60: This may not need to be overridden if it does nothing. But if you do override it, it is probably worth explicitly instantiating the lattice to the uninitialized value.

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -0,0 +1,81 @@
...
+#include "mlir/include/mlir/IR/Value.h"        // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace mul_depth_analysis {
+
+class MulDepth {
+ public:
+  MulDepth() : value(std::nullopt) {}
+  MulDepth(int64_t value) : value(value) {}
```
j2kun, line 16: To avoid unexpected conversion casts, prepend this constructor with `explicit`
```suggestion
  explicit MulDepth(int64_t value) : value(value) {}
```

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -0,0 +1,81 @@
...
+  ~MulDepthAnalysis() override = default;
+  using SparseForwardDataFlowAnalysis::SparseForwardDataFlowAnalysis;
+
+  void visitOperation(Operation *op, ArrayRef<const MulDepthLattice *> operands,
+                      ArrayRef<MulDepthLattice *> results) override;
+
+  void setToEntryState(MulDepthLattice *lattice) override;
+
+ private:
+  const DataFlowSolver *sccpAnalysis;
```
j2kun, line 74: This member variable is unused and I believe it can be removed, along with the constructor argument. All we really need for this analysis is that the DataFlowSolver has the other analyses loaded into it.

lib/Dialect/LWE/IR/LWEDialect.cpp
```diff
@@ -177,6 +177,18 @@ LogicalResult ReinterpretUnderlyingTypeOp::verify() {
   return success();
 }
 
+// Verification for RLWE_EncryptOp
+LogicalResult RLWEEncryptOp::verify() {
+  // Placeholder verification
```
j2kun, line 182: I think you will need to rebase over main to remove this spurious extra code, here and in LWEOps.td

j2kun (changes requested):
This is awesome! I left some minor cleanup comments, but otherwise I think all you'd need is a little bit of extra automated tests to pick up differing multiplicative depths in the generated code. You should be able to do all the testing you need without #729 being fixed, though that fix should come soon. The testing would just test this codegen in isolation, and assert that the generated OpenFHE code has calls to set the right multiplicative depth. The errors from #729 only occur when `heir-opt` is combined with `heir-translate` in a test that runs the whole pipeline end to end.

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
Maokami: I might be wrong about this part because I believe I don't fully understand the characteristics of noise.

I agree that having MulPlain and MulConst consume multiplicative depth is a very conservative approach: https://openfhe.discourse.group/t/multiplication-depth-in-bfv/487/7

However, if we leave these out in this analysis, couldn't it cause a problem in programs that contain too many MulPlain/MulConst ops(though I'm not sure if there are any practical use cases for this) due to a lower multiplicative depth than actual?

How about keeping the current conservative approach for now, and then changing it when we can calculate multiplicative depth considering MulPlain/MulConst through a more fine-grained noise analysis?

AlexanderViand-Intel: Mh, this is tricky. If we want to use the MulDepth Analysis for Relin insertion, we'd have to ignore them, while for "rescale" they do count. 
And of course they do add some smaller but noticeable noise so things that heavily use them (e.g., ML inference) would probably get wrong parameters.

I guess if the goal is to get a conservative parameter estimation, we should probably count them.

j2kun:
I think what matters for the immediate pass is whether OpenFHE counts them
when consuming multiplicative depth. Based on the link above, we should
keep it. Let's follow up in a later issue with what we'd need to do to
generalize the analysis so it can be used in other situations.

j2kun (approved):

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
asraa: I remember when the @Maokami first posted the issue I went through a rabbit hole trying to see what other libraries or compilers do. Apparently OpenFHE also counts plaintext-ciphertext multiplications the same as ciphertext multiplications (https://openfhe.discourse.group/t/multiplication-depth-in-bfv/487/5), and decided that we should be conservative for now too.

asraa (approved):
nice!

PR #763, opened 2024-07-01
add `--convert-tensor-to-scalars` pass

AlexanderViand-Intel:
This is a helper pass to remove `tensor.insert/extract`  (with constant indices) and (statically shaped) `tensor<...>` types from the IR by effectively "unrolling" `tensor<axbx!element_type>` into a `TypeRange` of a*b copies of `!element_type`.

This is necessary for targeting RLWE HW accelerators, the first generation of which "understand" only polynomial operations and datatypes. Specifically, this is necessary as passes such as `-bgv-to-polynomial` produce polynomial ops on tensors (e.g., adding two standard ciphertext lowers to a polynomial addition of  two `tensor<2x!polynomial.polynomial>`).
While `-convert-elementwise-to-affine` (see #524) lowers this to (loops over) polynomial operations on individual polynomial values (and the loops can be unrolled via `-full-loop-unroll`), the resulting IR still contains various tensor operations, primarily `tensor.insert/extract`.

This PR introduces a simple pass that replaces tensor types (with static shape) by a `TypeRange`  of dim1xdim2...  copies of the element type via the `OneToNTypeConversion` framework. Note: this framework apparently exists because the standard `DialectConversion` and associated `TypeConverter` are _very_ broken when it comes to handling 1->N type conversions (as I found out the hard way). 

In addition to doing the type conversion, there are also patterns to translate `tensor` operations to corresponding operations on the `ValueRange`, but the list is fairly incomplete right now, as there are only patterns for `tensor.from_elements` and `tensor.insert`. Surprisingly, together with folding (which, because left-over `ValueRanges` are materialized as `tensor.from_elements`, takes care of `tensor.extract`) this is actually already enough for my primary use case. However, I'd like to make this more robust and actually support any tensor ops that can conceptually be "unrolled" this way.

**Remaining ToDos**
- [ ] #1023 (+ more tests)
- [x] Add a cut-off size as a pass argument with a sane default, to prevent very large tensors from being "unrolled" accidentally (In our use case, tensors will be of single-digit sizes)

**Related ToDos:**
- [x] Address `polynomial.ntt/intt/mul_scalar` which aren't `ElementwiseMappable` and therefore not handled by `-convert-elementwise-to-affine` (#143)
- [ ] #1024

### DIFF
--- a/lib/Transforms/TensorToScalars/TensorToScalars.cpp
+++ b/lib/Transforms/TensorToScalars/TensorToScalars.cpp
@@ -1,0 +1,149 @@
+#include "lib/Transforms/TensorToScalars/TensorToScalars.h"
+
+#include <optional>
+
+#include "llvm/include/llvm/Support/Debug.h"            // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"   // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/Transforms/FuncConversions.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/Transforms/OneToNFuncConversions.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/SCF/Transforms/Patterns.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"           // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/OneToNTypeConversion.h"  // from @llvm-project
+namespace mlir {
+namespace heir {
+
+#define GEN_PASS_DEF_TENSORTOSCALARS
+#include "lib/Transforms/TensorToScalars/TensorToScalars.h.inc"
+
+static std::optional<Value> buildFromElementsOp(OpBuilder &builder,
+                                                RankedTensorType resultType,
+                                                ValueRange inputs,
+                                                Location loc) {
+  return builder.create<tensor::FromElementsOp>(loc, resultType, inputs);
+}
+
+static std::optional<SmallVector<Value>> buildExtractOps(OpBuilder &builder,
+                                                         TypeRange resultTypes,
+                                                         Value input,
+                                                         Location loc) {
+  // This pass only operates on tensors of static shape
+  RankedTensorType inputType = dyn_cast<RankedTensorType>(input.getType());
+  if (!inputType || !inputType.hasStaticShape()) return {};
+
+  // Create extract ops in "natural" order (dimension-by-dimension)
+  SmallVector<Value> values;
+  for (auto dim : inputType.getShape()) {
+    for (int i = 0; i < dim; ++i) {
+      Value index = builder.create<arith::ConstantIndexOp>(loc, i);
+      Value element = builder.create<tensor::ExtractOp>(loc, input, index);
+      values.push_back(element);
+    }
+  }
+  return values;
+}
+
+class ConvertFromElementsOp
+    : public OneToNOpConversionPattern<tensor::FromElementsOp> {
+ public:
+  using OneToNOpConversionPattern<
+      tensor::FromElementsOp>::OneToNOpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      tensor::FromElementsOp op, OpAdaptor adaptor,
+      OneToNPatternRewriter &rewriter) const override {
+    // OneToNConversion has no Conversion-level illegality handling
+    if (typeConverter->isLegal(op)) return failure();
+
+    // This pass only operates on tensors of static shape,
+    // but no check is necessary here as from_elements' shape is always static
+
+    // Replace the current op with the flattened operands.
+    // This should already match the "natural" order expected by this pass.
+    rewriter.replaceOp(op, adaptor.getFlatOperands(),
+                       adaptor.getResultMapping());
+    return success();
+  }
+};
+
+class ConvertInsertOp : public OneToNOpConversionPattern<tensor::InsertOp> {
+ public:
+  using OneToNOpConversionPattern<tensor::InsertOp>::OneToNOpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      tensor::InsertOp op, OpAdaptor adaptor,
+      OneToNPatternRewriter &rewriter) const override {
+    // OneToNConversion has no Conversion-level illegality handling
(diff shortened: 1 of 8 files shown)


lib/Transforms/TensorToScalars/TensorToScalars.td
```diff
@@ -0,0 +1,35 @@
...
+  let description = [{
+    This pass will convert a static-shaped tensor type to a TypeRange
+    containing product(dim) copies of the element type of the tensor.
+    This pass currently includes two patterns:
+    1. It converts tensor.from_elements operations to
+    the corresponding scalar inputs.
+    2. It converts tensor.insert operations by updating the
+    ValueRange corresponding to the converted input and r
+    updating it with the scalar to be inserted.
+    It also applies folders greedily to simplify, e.g., extract(from_elements).
```
j2kun, line 18: Nit (markdown formatting)

```suggestion
    This pass currently includes two patterns:
    
    1. It converts tensor.from_elements operations to
    the corresponding scalar inputs.
    2. It converts tensor.insert operations by updating the
    ValueRange corresponding to the converted input and r
    updating it with the scalar to be inserted.
    
    It also applies folders greedily to simplify, e.g., extract(from_elements).
```

j2kun:
> This is necessary for targeting RLWE HW accelerators

Explicitly, this means it has no control flow?

lib/Transforms/TensorToScalars/TensorToScalars.cpp
```diff
@@ -0,0 +1,144 @@
...
+  using TensorToScalarsBase::TensorToScalarsBase;
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+
+    OneToNTypeConverter typeConverter;
+    typeConverter.addConversion([](Type type) { return type; });
+    typeConverter.addConversion(
+        [](RankedTensorType tensorType,
+           SmallVectorImpl<Type> &types) -> std::optional<LogicalResult> {
```
j2kun, line 113: maybe in your blog post you can explain why one might return `std::nullopt` in this function :)

lib/Transforms/TensorToScalars/TensorToScalars.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return failure();
+    }
+
+    // Compute the insertion offset (in dimension-by-dimension order):
+    int64_t multiplier = 1;
+    int64_t offset = 0;
+    for (auto [dim, idx] :
+         llvm::zip(op.getResult().getType().getShape(), op.getIndices())) {
+      offset +=
+          idx.getDefiningOp<arith::ConstantIndexOp>().value() * multiplier;
```
j2kun, line 89: instead of looping twice you may as well check for the failure condition here. IIUC, `dyn_cast` is relatively expensive in MLIR.

lib/Transforms/TensorToScalars/TensorToScalars.cpp
```diff
@@ -0,0 +1,144 @@
...
+           SmallVectorImpl<Type> &types) -> std::optional<LogicalResult> {
+          if (!tensorType.hasStaticShape()) return failure();
+          int count = 1;
+          for (auto dim : tensorType.getShape()) {
+            if (dim > 0) count *= dim;
+          }
+          types = SmallVector<Type>(count, tensorType.getElementType());
+          return success();
+        });
+    typeConverter.addArgumentMaterialization(buildFromElementsOp);
```
j2kun, line 122: Are you actually seeing unrealized_conversion_cast ops generated in your IR? As far as I understand, this should only happen at the very end of the pass if the conversion can't completely succeed.

This may be part of my not knowing how the one-to-N conversion stuff works, but in normal dialect conversion you only need materializations if the conversion logic as a whole is split over multiple passes.

AlexanderViand-Intel:
> > This is necessary for targeting RLWE HW accelerators
> 
> Explicitly, this means it has no control flow?

Yes, this was something that came up at the HW summit in Leuven last month: at this point, most of the accelerators don't seem to have a native sense of control flow, expecting the host to supply a a simple stream of instructions.

lib/Transforms/TensorToScalars/TensorToScalars.cpp
AlexanderViand-Intel: Good catch, this should definitively "soft" fail

AlexanderViand-Intel:
> **Related ToDos:**
> 
> * [ ]  Address `polynomial.ntt/intt/mul_scalar` which aren't `ElementwiseMappable` and therefore not handled by `-convert-elementwise-to-affine` ([PolyToStandard: handling tensors of poly? #143](https://github.com/google/heir/issues/143))

That turned out to be easier than expected, see PR here: #769

AlexanderViand-Intel:
Since the pass is functional for our purposes as-is, I've made the "extension to tensor.slice/etc" a separate issue (linked from the pass header) and marked this as ready-for-review.

j2kun (approved):

lib/Transforms/TensorToScalars/TensorToScalars.cpp
```diff
@@ -0,0 +1,149 @@
...
+      // We can only support statically known indices
+      // that have been constant-folded to a single arith.constant op
+      auto cidx = idx.getDefiningOp<arith::ConstantIndexOp>();
+      if (!cidx) return failure();
+      offset += cidx.value() * multiplier;
+      multiplier *= dim;
+    }
+
+    // get converted "tensor" operand from op (likely a unrealized_builtin_cast)
+    SmallVector<Value> elements = adaptor.getOperands()[1];
```
asraa, line 98: nit: small readability thing, but you can grab `adaptor.getScalar()` and `adaptor.getDest()` instead of indexing into the operands

asraa (approved):
Nice, thank you for pushing this PR through!

lib/Transforms/TensorToScalars/TensorToScalars.cpp
AlexanderViand-Intel: Nice catch, thanks! I'm not sure why I didn't originally use them here

AlexanderViand-Intel: Ah, because the function takes an `OpAdaptor` and not a `tensor::InsertOp::Adaptor`. Let me see if it works with that, too.

AlexanderViand-Intel: Unfortunately, it does not - looks like this is a limitation of the 1toNConversion framework.

asraa: argh! thank you for trying anyway, sorry about that! TIL

AlexanderViand-Intel:
@asraa I tried addressing your nit, but unfortunately it seems like it's a limitation of how the 1toN conversion framework works. Modulo "feedback/copybara — Google internal checks FAILED" this should be good to go :)

AlexanderViand-Intel:
@asraa  I reabsed this onto main and moved the tests to `tests/transforms/convert_tensor_to_scalars` in keeping with the recent test reorg.

asraa:
Thank you!

PR #764, opened 2024-07-01
split `-mlir-to-openfhe-bgv` pipeline

AlexanderViand-Intel:
Factors out the secretize-vectorize-BGVize part of the `-mlir-to-openfhe-bgv` pipeline into its own `-mlir-to-bgv` pipeline, in order to improve usability/test writing when not targeting OpenFHE (e.g., LLVM or RLWE accelerators).

### DIFF
--- a/tools/heir-opt.cpp
+++ b/tools/heir-opt.cpp
@@ -420,8 +420,8 @@ void tosaToBooleanFpgaTfhePipeline(const std::string &yosysFilesPath,
 }
 #endif
 
-struct MlirToOpenFheBgvPipelineOptions
-    : public PassPipelineOptions<MlirToOpenFheBgvPipelineOptions> {
+struct MlirToBgvPipelineOptions
+    : public PassPipelineOptions<MlirToBgvPipelineOptions> {
   PassOptions::Option<std::string> entryFunction{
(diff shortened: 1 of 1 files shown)


asraa (approved):
Agree, good change to have!

tools/heir-opt.cpp
```diff
@@ -452,6 +452,12 @@ void mlirToOpenFheBgvPipelineBuilder(
   auto secretToBgvOpts = SecretToBGVOptions{};
   secretToBgvOpts.polyModDegree = options.ciphertextDegree;
   pm.addPass(createSecretToBGV(secretToBgvOpts));
+}
+
+void mlirToOpenFheBgvPipelineBuilder(OpPassManager &pm,
+                                     const MlirToBgvPipelineOptions &options) {
+  // lower to BGV
+  mlirToBgvPipelineBuilder(pm, options);
```
asraa, line 460: we can re-factor later once we have a ckks pass too, but maybe we should just have two levels of pass pipelines (1) mlir-to-bgv, mlir-to-ckks, mlir-to-cggi, etc, and then (2) something like lower-to-openfhe that bundles all the *-to-openfhe passes.

PR #769, opened 2024-07-05
extend `--convert-elementwise-to-affine` to support ops with scalar operands

### LINKED ISSUE #143: PolyToStandard: handling tensors of poly?

### LINKED ISSUE #534: Test "ElementwiseToAffine" support for ops with both tensor and scalar operands
The `ElementwiseToLinalg` pass in MLIR only supports operations where all operands are tensors and this requirement has been left in place for the `ElementwiseToAffine` pass. However, the reason that this limitation is necessary for the linalg version doesn't exist in the affine version, which should already support scalar operands as-is (well, after replacing the `all_of` in `isElementwiseMappableOpOnRankedTensors` with `any_of`) . However, `polynomial` operations have a `SameOperandsAndResultType` constraint, so to test whether this feature actually works as expected, we'd need to either allow mixed tensor-scalar operations and/or simply test this pass with a different `ElementwiseMappable` op.

AlexanderViand-Intel:
As originally pointed out in #143 and recently mentioned in #763, the `--convert-elementwise-to-affine` pass did not handle ops with "mixed" tensor/scalar operands, such as  `polynomial.mul_scalar`.

It turns out this was a much easier fix than anticipated, at least for `mul_scalar` as `ElementwiseMappable` does not (as I wrongly assumed) imply `SameOperandsAndResultType`. In fact, there was already a ToDo (#534) that suggested the three-character change necessary to make this work.
However, this still does not work for `polynomial.ntt/intt`, as `ElementwiseMappable` _does_ require all _tensor_ operands to agree, and that wouldn't hold for a "tensorized" ntt/intt (cf. https://github.com/google/heir/issues/143#issuecomment-2108448494)

With this change, any `ElementwiseMappable` operation with at least one tensor operand is converted into an `affine.for`, and tensor operands are replaced with `tensor.extract`s, while scalar operands are left unmodified. This result is probably what one would expect (despite the semantics for this not being formally noted anywhere upstream, afaik), effectively "broadcasting" the scalar to the tensor operand(s).

PS: I noticed the tests for this pass also had some mostly meaningless "expected output" "tests" in it from development so I used the opportunity to remove those.

-------
Resolves #534
(Mostly, see https://github.com/google/heir/pull/763#issuecomment-2211012794) Resolves #143

### DIFF
--- a/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
+++ b/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
@@ -27,9 +27,7 @@ namespace heir {
 static bool isElementwiseMappableOpOnRankedTensors(Operation *op) {
   if (!OpTrait::hasElementwiseMappableTraits(op)) return false;
 
-  // TODO(#534): Test ElementwiseToAffine with `any_of` constraints
-  // as the pass should (in theory) support scalar operands, too
-  return llvm::all_of(op->getOperandTypes(),
+  return llvm::any_of(op->getOperandTypes(),
                       [](Type type) { return isa<RankedTensorType>(type); });
 }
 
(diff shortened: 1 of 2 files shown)


j2kun:
LGTM! How convenient that you did it right the first time :)

j2kun (approved):

asraa (approved):

Issue #727, opened 2024-06-06
segfault in heir-translate


j2kun:
Repro:

```
bazel run //tools:heir-opt -- --mlir-to-openfhe-bgv='entry-function=dot_product ciphertext-degree=8' $PWD/tests/openfhe/end_to_end/dot_product_8.mlir > $PWD/out.mlir
bazel run //tools:heir-translate -- --emit-openfhe-pke-header $PWD/out.mlir > out.h
```

j2kun:
Error:

`LLVM ERROR: can't create Attribute 'mlir::polynomial::IntPolynomialAttr' because storage uniquer isn't initialized: the dialect was likely not loaded, or the attribute wasn't added with addAttributes<...>() in the Dialect::initialize() method.`

j2kun:
From my investigation in discord: 

ok I figured out the problem. The parser requires me to have a namespaced attribute or else it will fail to load the dialect
[3:17 PM]j2kun: this is weird. If I have an attribute like this: ring = #polynomial.ring<coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = #polynomial.int_polynomial<1 + x**8>> it works, but if I have ring = #polynomial.ring<coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**8>> then the parser doesn't load the attribute
[3:17 PM]j2kun: This is due to this check here: https://github.com/llvm/llvm-project/blob/19bbbcbedcbddcfc39202d7d801508a20baf83b6/mlir/lib/AsmParser/AttributeParser.cpp#L321-L323
[3:18 PM]j2kun: and this happens because the int_polynomial attribute omits the namespaced prefix in the default printer https://github.com/llvm/llvm-project/blob/19bbbcbedcbddcfc39202d7d801508a20baf83b6/mlir/lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp#L22
[3:18 PM]j2kun: Any advice here?

j2kun:
Alex Zinenko on Discord:

> This does look like a bug in the parser to me. If a dialect is
registered, the parser should be able to load it on-demand. One shouldn't
need to preload the dialect for parsing only.

j2kun:
I can work on a fix over the next few days, but in the mean time we can disable the tests

asraa:
This fixes it, but I don't know if it's "right" https://discord.com/channels/636084430946959380/642426447167881246/1248730040145809581

[...]

PR #743, opened 2024-06-21
Reenable openfhe end to end tests

j2kun:
Fixes https://github.com/google/heir/issues/729

Waiting on https://github.com/llvm/llvm-project/pull/96242 to be integrated, then rebase

PR #771, opened 2024-07-06
re-enable end-to-end openFHE tests (tmp workaround)

AlexanderViand-Intel:
This is a temporary workaround for the upstream dialect loading bug (See #727). It simply prepends a dummy polynomial attribute to the input file before calling heir_translate. To-Do's still reference #729 (re-enable tests) as this doesn't really "resolve" the issue.

Even for a temporary hack, I guess it'd be nicer to do this only in `openfhe_end_to_end_test` and not the general `heir_translate` rule, but this was quicker and I just wanted the ability to run end-to-end tests back 😉 (This might also be nice to have back for the mult-depth analysis tests in #760).

I'm not sure if we actually want to merge this, but I'm guessing this might be useful to others for debugging/testing until the upstream issue is fixed ([llvm PR #96667](https://github.com/llvm/llvm-project/pull/96667)), even just as something to cherry-pick.

### DIFF
--- a/tests/openfhe/end_to_end/BUILD
+++ b/tests/openfhe/end_to_end/BUILD
@@ -1,50 +1,49 @@
 # See README.md for setup required to run these tests
 
-# load("@heir//tests/openfhe/end_to_end:test.bzl", "openfhe_end_to_end_test")
+load("@heir//tests/openfhe/end_to_end:test.bzl", "openfhe_end_to_end_test")
 
 package(default_applicable_licenses = ["@heir//:license"])
 
-# TODO(#729): Re-enable tests when upstream dialect loader is fixed
-# openfhe_end_to_end_test(
-#     name = "binops_test",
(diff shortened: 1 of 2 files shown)


j2kun:
I'm fine with this. I don't see why the upstream review is taking so long, maybe folks are on vacation or traveling to conferences.

j2kun (approved):

j2kun:
BTW, the upstream fix was merged and we should see it in HEIR in a day or two.

AlexanderViand-Intel:
> BTW, the upstream fix was merged and we should see it in HEIR in a day or two.

Any chance it's already in the most recent LLVM integration that just happened?

asraa:
I checked, it was!

AlexanderViand-Intel:
> I checked, it was!

Great, closing this in favor of #743 then :)

PR #775, opened 2024-07-08
add nightly release workflow

j2kun:


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,35 @@
...
+        shell: bash
+        run: |
+            bazel-bin/tools/heir-opt --help
+            bazel-bin/tools/heir-translate --help
+            bazel-bin/tools/heir-lsp --help
+
+      - name: GH Release
+        uses: softprops/action-gh-release@a74c6b72af54cfa997e81df42d94703d6313a2d0 #pin@v2.0.6
+        with:
+          fail_on_unmatched_files: true
```
asraa, line 30: add a prerelase flag:

```suggestion
          prerelease: true
```

asraa, line 30: as a follow-up we can pass the sha's from the previous step into a step that makes a file with their checksums / date / commit of the nightly and add a `body_path` flag to that file here

asraa (approved):
Looks good to start! Could you also add some info in https://heir.dev/docs/getting_started/ about using the nightly build?

.github/workflows/nightly.yml
asraa: My bad! I didn't mean to remove the previous line in the suggestion!

j2kun: I think actually that line is unnecessary because the previous step would fail.

asraa: True!

Issue #782, opened 2024-07-10
Update source code in nightly release 


asraa:
It looks like the release assets are updated 

https://github.com/google/heir/releases/tag/nightly

but the source code isn't.

I think we might want to delete the release as a whole in the action instead of overwriting just the assets (not sure why it doesn't overwrite the source... i'll check the action options)

asraa:
I guess the nightly tag isn't updated

PR #783, opened 2024-07-10
fix: fix updating nightly tag and release

asraa:
Fixes https://github.com/google/heir/issues/782

The release tag wasn't updated each night, even though the release assets are overrided.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -24,6 +24,13 @@ jobs:
             bazel-bin/tools/heir-translate --help
             bazel-bin/tools/heir-lsp --help
 
+      - name: Delete previous existing nightly tag and release
+        env:
+          GH_TOKEN: $${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+            gh release view "nightly" && gh release delete "nightly" -y --cleanup-tag
+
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -24,6 +24,13 @@ jobs:
             bazel-bin/tools/heir-translate --help
             bazel-bin/tools/heir-lsp --help
 
+      - name: Delete previous existing nightly tag and release
+        env:
+          GH_TOKEN: $${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+            gh release view "nightly" && gh release delete "nightly" -y --cleanup-tag
```
AlexanderViand-Intel, line 32: Why the `gh release view` here?

AlexanderViand-Intel (approved):
Thanks for fixing this!

j2kun (approved):

.github/workflows/nightly.yml
j2kun: I believe this is to make sure the release exists before trying to delete it.

asraa: Yep! Won't happen often (maybe if we manually delete it) but it will ensure that we don't fail on delete

PR #788, opened 2024-07-12
Cornami, Inc : contribute CMake build files

ai-mannamalai:
Provide CMakeLists.txt files to build all of the core components of HEIR Project (except Yosys and GTest bits)
Contributed by Cornami, Inc, www.cornami.com 
Two ways to build the code:

## Using LLVM build folder
```bash
$ cmake -B builddir -S ./ -DMLIR_DIR=/path/to/llvm-project/build/lib/cmake/mlir -DLLVM_DIR=/path/to/llvm-project/build/lib/cmake/llvm && cmake --build builddir
```

## Using Installed LLVM / MLIR
```bash
$ cmake -S ./ -B builddir -DENABLE_INSTALLED_MLIR=True -DCMAKE_PREFIX_PATH=/usr/local/lib/cmake -DMLIR_CMAKE_DIR=/usr/local/lib/cmake/mlir/ -DLLVM_CMAKE_DIR=/usr/local/lib/cmake/llvm/ &&  cmake --build builddir
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,88 @@
+cmake_minimum_required(VERSION 3.13.4)
+
+if(POLICY CMP0116)
+  cmake_policy(SET CMP0116 OLD)
+endif()
+
+project(HEIR LANGUAGES CXX C)
+
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "C++ standard to conform to")
+
+find_package(MLIR REQUIRED CONFIG)
+
+message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+
+option(ENABLE_YOSYS "Enable YOSYS for yosys-optimizer pass (optional)" FALSE)
+option(ENABLE_INSTALLED_MLIR "Use installed MLIR" FALSE)
+
+if(ENABLE_YOSYS)
+   message("Yosys enabled! Make sure ABC and Yosys development files and libraries are on the path")
+else()
+   add_compile_options(-DHEIR_NO_YOSYS=1)
+endif()
+
+
+set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/bin)
+set(LLVM_LIBRARY_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/lib)
+set(MLIR_BINARY_DIR ${CMAKE_BINARY_DIR})
+
+
+
+list(APPEND CMAKE_MODULE_PATH "${MLIR_CMAKE_DIR}")
+list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
+include(TableGen)
+include(AddLLVM)
+include(AddMLIR)
+include(HandleLLVMOptions)
+
+set(INCLUDE_MLIR_LLVM_HEADERS "${CMAKE_BINARY_DIR}/inc/")
+add_custom_command(OUTPUT ${INCLUDE_MLIR_LLVM_HEADERS}
+  COMMAND mkdir -p inc/llvm/include
+  COMMAND mkdir -p inc/mlir/include
+  WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+  COMMENT "Creating link to LLVM and MLIR headers"
+  VERBATIM
+)
+
+if(ENABLE_INSTALLED_MLIR)
+add_custom_target(HEIRIncludeLLVM
+                  BYPRODUCTS ${CMAKE_BINARY_DIR}/inc/llvm/include/llvm
+                  COMMAND [ -e llvm ] && echo "" || ln -s ${LLVM_INCLUDE_DIRS}/llvm/ ./
+                  DEPENDS ${INCLUDE_MLIR_LLVM_HEADERS}
+                  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/inc/llvm/include/"
+                  )
+
+add_custom_target(HEIRIncludeMLIR
+                  BYPRODUCTS ${CMAKE_BINARY_DIR}/inc/mlir/include/mlir
+                  COMMAND [ -e mlir ] && echo "" || ln -s ${MLIR_INCLUDE_DIRS}/mlir/ ./
+                  DEPENDS ${INCLUDE_MLIR_LLVM_HEADERS}
+                  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/inc/mlir/include/")
+message(STATUS "Adding MLIR and LLVM includes.")
+else()
+add_custom_target(HEIRIncludeLLVM
+          COMMAND echo "")
+add_custom_target(HEIRIncludeMLIR
+          COMMAND echo "")
+message(STATUS "Using MLIR and LLVM from build folders of ${MLIR_DIR} and ${LLVM_DIR}")
+endif()
+
+include_directories(${CMAKE_BINARY_DIR}/inc)
+if(ENABLE_INSTALLED_MLIR)
+include_directories(${LLVM_INCLUDE_DIRS})
+include_directories(${MLIR_INCLUDE_DIRS})
+else()
+include_directories(${LLVM_INCLUDE_DIRS})
+include_directories(${MLIR_INCLUDE_DIRS})
+include_directories(${LLVM_INCLUDE_DIRS}/../..)
+include_directories(${MLIR_INCLUDE_DIRS}/../..)
+endif()
+include_directories(${PROJECT_SOURCE_DIR})
+include_directories(${PROJECT_BINARY_DIR})
+link_directories(${LLVM_BUILD_LIBRARY_DIR})
+add_definitions(${LLVM_DEFINITIONS})
+
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
+set(HEIRLIBDIR ${CMAKE_CURRENT_SOURCE_DIR})
+add_subdirectory(lib)
+add_subdirectory(tools)
--- a/lib/Analysis/CMakeLists.txt
+++ b/lib/Analysis/CMakeLists.txt
@@ -1,0 +1,5 @@
+add_subdirectory(RotationAnalysis)
+add_subdirectory(TargetSlotAnalysis)
+add_subdirectory(SelectVariableNames)
+add_subdirectory(MulDepthAnalysis)
+add_subdirectory(SecretnessAnalysis)
--- a/lib/Analysis/RotationAnalysis/CMakeLists.txt
+++ b/lib/Analysis/RotationAnalysis/CMakeLists.txt
@@ -1,0 +1,17 @@
+add_mlir_conversion_library(MLIRHEIRAnalysisROT
+        RotationAnalysis.cpp
+
+        ADDITIONAL_HEADER_DIRS
+        ${CMAKE_CURRENT_SOURCE_DIR}
+
+        DEPENDS
+        MLIRTensorExtOpsIncGen
+
+        LINK_LIBS PUBLIC
+        MLIRTensorExt
+        LLVMSupport
+        MLIRAnalysis
+        MLIRArithDialect
+        MLIRIR
+        MLIRSupport
+)
--- a/lib/Analysis/SelectVariableNames/CMakeLists.txt
+++ b/lib/Analysis/SelectVariableNames/CMakeLists.txt
@@ -1,0 +1,10 @@
+add_mlir_conversion_library(MLIRHEIRAnalysisSVN
+        SelectVariableNames.cpp
+
+        ADDITIONAL_HEADER_DIRS
+        ${CMAKE_CURRENT_SOURCE_DIR}
+
+        LINK_LIBS PUBLIC
+        MLIRIR
+        LLVMSupport
+)
--- a/lib/Conversion/CMakeLists.txt
+++ b/lib/Conversion/CMakeLists.txt
@@ -1,0 +1,31 @@
+add_mlir_dialect_library(MLIRHEIRUtils
+    PARTIAL_SOURCES_INTENDED
+    Utils.cpp
+
+    ADDITIONAL_HEADER_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+    DEPENDS
+    MLIRTensorExtOpsIncGen
+    MLIRTensorExtPassesIncGen
+    MLIRTensorExtInsertRotateIncGen
+    MLIRTensorExtCanonicalizationIncGen
+
+    LINK_LIBS PUBLIC
+    MLIRIR
+    MLIRInferTypeOpInterface
+    MLIRArithDialect
+    MLIRSupport
+    MLIRDialect
+    MLIRIR
+)
+
+add_subdirectory(ArithExtToArith)
+add_subdirectory(BGVToOpenfhe)
+add_subdirectory(BGVToPolynomial)
+add_subdirectory(CGGIToTfheRust)
+add_subdirectory(CGGIToTfheRustBool)
+add_subdirectory(CombToCGGI)
+add_subdirectory(MemrefToArith)
+add_subdirectory(PolynomialToStandard)
+add_subdirectory(SecretToBGV)
--- a/lib/Dialect/ArithExt/CMakeLists.txt
+++ b/lib/Dialect/ArithExt/CMakeLists.txt
@@ -1,0 +1,18 @@
+add_subdirectory(IR)
+add_mlir_dialect_library(MLIRArithExtDialect
+        IR/ArithExtDialect.cpp
+
+        ADDITIONAL_HEADER_DIRS
+        ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+        DEPENDS
+        MLIRArithExtIncGen
+        MLIRArithExtOpsIncGen
+
+        LINK_LIBS PUBLIC
+        LLVMSupport
+        MLIRArithDialect
+        MLIRIR
+        MLIRInferTypeOpInterface
+        MLIRSupport
+)
--- a/lib/Dialect/ArithExt/IR/CMakeLists.txt
+++ b/lib/Dialect/ArithExt/IR/CMakeLists.txt
@@ -1,0 +1,9 @@
+set(LLVM_TARGET_DEFINITIONS ArithExtDialect.td)
+mlir_tablegen(ArithExtDialect.h.inc -gen-dialect-decls)
+mlir_tablegen(ArithExtDialect.cpp.inc -gen-dialect-defs)
+add_public_tablegen_target(MLIRArithExtIncGen)
+
+set(LLVM_TARGET_DEFINITIONS ArithExtOps.td)
+mlir_tablegen(ArithExtOps.h.inc -gen-op-decls)
+mlir_tablegen(ArithExtOps.cpp.inc -gen-op-defs)
+add_public_tablegen_target(MLIRArithExtOpsIncGen)
--- a/lib/Dialect/BGV/CMakeLists.txt
+++ b/lib/Dialect/BGV/CMakeLists.txt
@@ -1,0 +1,22 @@
+add_subdirectory(IR)
+add_subdirectory(Transforms)
+add_mlir_dialect_library(MLIRBGV
+        IR/BGVDialect.cpp
+        Transforms/AddClientInterface.cpp
+
+        ADDITIONAL_HEADER_DIRS
+        ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+        DEPENDS
+        MLIRBGVIncGen
+        MLIRBGVOpsIncGen
+        MLIRBGVPassesIncGen
+
+        LINK_LIBS PUBLIC
+        MLIRIR
+        MLIRInferTypeOpInterface
+        MLIRArithDialect
+        MLIRSupport
+        MLIRDialect
+        MLIRIR
+)
--- a/lib/Dialect/CGGI/CMakeLists.txt
+++ b/lib/Dialect/CGGI/CMakeLists.txt
@@ -1,0 +1,27 @@
+add_subdirectory(IR)
+add_subdirectory(Transforms)
+add_mlir_dialect_library(MLIRCGGI
+        IR/CGGIDialect.cpp
+        IR/CGGIOps.cpp
+        Transforms/SetDefaultParameters.cpp
+
+        ADDITIONAL_HEADER_DIRS
+        ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+        DEPENDS
+        MLIRCGGIIncGen
+        MLIRCGGIOpsIncGen
+        MLIRCGGIPassesIncGen
+        MLIRCGGIAttributesIncGen
+
+        LINK_LIBS PUBLIC
+        MLIRHEIRInterfaces
+        MLIRLWE
+        MLIRIR
+        MLIRInferTypeOpInterface
+        MLIRArithDialect
+        MLIRSupport
+        MLIRDialect
+        MLIRIR
+)
+add_dependencies(MLIRCGGI MLIRCGGIAttributesIncGen)
--- a/lib/Dialect/CMakeLists.txt
+++ b/lib/Dialect/CMakeLists.txt
@@ -1,0 +1,46 @@
+add_subdirectory(ArithExt)
+# IR + Transforms
+add_subdirectory(LWE)
+
+
+add_subdirectory(Jaxite)
+add_subdirectory(PolyExt)
+add_subdirectory(Openfhe)
+add_subdirectory(RNS)
+add_subdirectory(TfheRust)
+add_subdirectory(TfheRustBool)
+add_subdirectory(Polynomial)
+
+add_subdirectory(BGV)
+add_subdirectory(CGGI)
+add_subdirectory(Secret)
+
+set(LLVM_TARGET_DEFINITIONS HEIRInterfaces.td)
+mlir_tablegen(HEIRInterfaces.h.inc  --gen-op-interface-decls  -name heir)
+mlir_tablegen(HEIRInterfaces.cpp.inc  --gen-op-interface-defs -name heir)
+add_public_tablegen_target(MLIRHEIRInterfacesOpsIncGen)
+
+add_mlir_dialect_library(MLIRHEIRInterfaces
+    HEIRInterfaces.cpp
+
+    ADDITIONAL_HEADER_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+    DEPENDS
+    MLIRHEIRInterfacesOpsIncGen
+    HEIRIncludeMLIR
+    HEIRIncludeLLVM
+
+    LINK_LIBS PUBLIC
+    LLVMSupport
+    MLIRIR
+    MLIRInferTypeOpInterface
+    MLIRSupport
+    MLIRDialect
+)
+
+
+add_subdirectory(Comb) #- requires HEIR interface
+add_dependencies(MLIRCombOpsIncGen MLIRHEIRInterfacesOpsIncGen)
+
+add_subdirectory(TensorExt)
--- a/lib/Source/AutoHog/CMakeLists.txt
+++ b/lib/Source/AutoHog/CMakeLists.txt
@@ -1,0 +1,47 @@
+include(ExternalProject)
+
+# Download RapidJSON
+# TODO (#818): automatically extract the commit from the bazel WORKSPACE file
+ExternalProject_Add(
+    rapidjson
+    PREFIX "vendor/rapidjson"
+    GIT_REPOSITORY "https://github.com/Tencent/rapidjson.git"
+    GIT_TAG f54b0e47a08782a6131cc3d60f94d038fa6e0a51
+    TIMEOUT 10
+    CMAKE_ARGS
+        -DRAPIDJSON_BUILD_TESTS=OFF
+        -DRAPIDJSON_BUILD_DOC=OFF
+        -DRAPIDJSON_BUILD_EXAMPLES=OFF
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+    INSTALL_COMMAND ""
+    UPDATE_COMMAND ""
+)
+
+# Prepare RapidJSON (RapidJSON is a header-only library)
+ExternalProject_Get_Property(rapidjson source_dir)
+set(RAPIDJSON_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/vendor/rapidjson/src/rapidjson/include/)
+
+add_mlir_dialect_library(MLIRAutoHogImporter
+    AutoHogImporter.cpp
+
+    ADDITIONAL_HEADER_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/
+
+    DEPENDS
+    rapidjson
+
+    DEPENDS
+    rapidjson
+
+    LINK_LIBS PUBLIC
+    MLIRCGGI
+    MLIRLWE
+    LLVMSupport
+    MLIRIR
+    MLIRSupport
+    MLIRArithDialect
+    MLIRTensorDialect
+    MLIRTranslateLib
+)
+target_include_directories(MLIRAutoHogImporter PUBLIC ${RAPIDJSON_INCLUDE_DIR})
--- a/lib/Transforms/CMakeLists.txt
+++ b/lib/Transforms/CMakeLists.txt
@@ -1,0 +1,14 @@
+add_subdirectory(ApplyFolders)
+add_subdirectory(ConvertIfToSelect)
+add_subdirectory(ElementwiseToAffine)
+add_subdirectory(ForwardStoreToLoad)
+add_subdirectory(FullLoopUnroll)
+add_subdirectory(Secretize)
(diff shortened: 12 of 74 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+project(HEIR LANGUAGES CXX C)
+
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "C++ standard to conform to")
+
+find_package(MLIR REQUIRED CONFIG)
+
+message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+
+option(ENABLE_YOSYS "Enable YOSYS for BFV and Comb dialects (optional)" FALSE)
```
j2kun, line 16:
```suggestion
option(ENABLE_YOSYS "Enable YOSYS for BFV and Comb dialects (optional)" FALSE)
```
```suggestion
option(ENABLE_YOSYS "Enable YOSYS for yosys-optimizer pass (optional)" FALSE)
```

lib/Analysis/SelectVariableNames/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_mlir_dialect_library(MLIRHEIRAnalysisSVN
+        SelectVariableNames.cpp
+
+        ADDITIONAL_HEADER_DIRS
+        ${CMAKE_CURRENT_SOURCE_DIR}
+
+	LINK_LIBS PUBLIC
```
j2kun, line 7: Looks like a mix of tabs, spaces, and the space-indented parts don't all have the same space indentation. Can you `expandtab`? Or maybe something like

```bash
sed -i 's/^\s\{2,\}/    /g' **/CMakeLists.txt
```

j2kun:
Great work! I think we should get a quick agreement about the expectations for maintaining this build going forward. Some questions:

- Does the CMake build have to pass CI to submit a PR? If so, we should add the CI workflow in this or the next PR. If not, do we expect CMake users to periodically fix a broken CMake build (this is what upstream LLVM does for the bazel build).
- Are we expecting the user to bring their own LLVM repo and set up the path? If so, we should document how to do that somewhere, ideally so it shows up on `heir.dev`, but for now having the CI workflow document it would suffice.
- How will we keep the LLVM commit in sync between the bazel and CMake builds? At least in the CI, can we parse `bazel/import_llvm.bzl` for the commit hash for this?

asraa:
> Does the CMake build have to pass CI to submit a PR? If so, we should add the CI workflow in this or the next PR. If not, do we expect CMake users to periodically fix a broken CMake build (this is what upstream LLVM does for the bazel build).

I was about to broach an issue to talk about this - I would like to see a CI workflow added (probably in a follow-up, because I expect some non-trivial work to handle pulling in and caching LLVM). I would hope right now that it's non-blocking, and that users who are interested in the CMake proactively fix it (we can get the CI to auto-file an issue saying CMake is broken and tag "CMake build maintainers").

> How will we keep the LLVM commit in sync between the bazel and CMake builds? At least in the CI, can we parse bazel/import_llvm.bzl for the commit hash for this?

+1

lib/Analysis/SelectVariableNames/CMakeLists.txt
ai-mannamalai: I something similar - fixed up now.

ai-mannamalai:
> > Does the CMake build have to pass CI to submit a PR? If so, we should add the CI workflow in this or the next PR. If not, do we expect CMake users to periodically fix a broken CMake build (this is what upstream LLVM does for the bazel build).
> 
> I was about to broach an issue to talk about this - I would like to see a CI workflow added (probably in a follow-up, because I expect some non-trivial work to handle pulling in and caching LLVM). I would hope right now that it's non-blocking, and that users who are interested in the CMake proactively fix it (we can get the CI to auto-file an issue saying CMake is broken and tag "CMake build maintainers").
> 
> > How will we keep the LLVM commit in sync between the bazel and CMake builds? At least in the CI, can we parse bazel/import_llvm.bzl for the commit hash for this?
> 
> +1

- We should likely add CMake build in the minimum for PRs to pass otherwise things will diverge quickly
- I'm not expert in building LLVM quickly but I assume other projects downstream of MLIR will be testing in similar way with CMake and we can reuse their methods for CI
- Currently this CMakeLists.txt assumes the mlir/include/ and llvm/include/ folders to exist - this needs some TLC
- Currently the heir-opt and heir-lsp do not built and we have to fix the link errors stopping this build

ai-mannamalai:
Updated pre-commit checks as per docs https://heir.dev/docs/getting_started/

lib/Transforms/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+add_subdirectory(ConvertIfToSelect)
+add_subdirectory(ElementwiseToAffine)
+add_subdirectory(ForwardStoreToLoad)
+add_subdirectory(FullLoopUnroll)
+add_subdirectory(Secretize)
+add_subdirectory(StraightLineVectorizer)
+add_subdirectory(UnusedMemref)
+
+if(ENABLE_YOSYS)
+    message(FATAL_ERROR "Yosys transforms not supported in CMake build yet; use Bazel instead.")
```
AlexanderViand-Intel, line 11: Can you create an issue for this and link to it here + anywhere else changes would have to be made? 
(there's a CI thing that keeps track of mentions for each issue)
```suggestion
    # TODO (#999): Support Yosys in cmake build.
    message(FATAL_ERROR "Yosys transforms not supported in CMake build yet; use Bazel instead.")
```

lib/Analysis/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_subdirectory(RotationAnalysis)
```
AlexanderViand-Intel, line 1: It seems like `MulDepthAnalysis` and `SecretnessAnalysis` don't yet have a CMakeLists.txt (they were recently added)
```suggestion
add_subdirectory(MulDepthAnalysis)
add_subdirectory(RotationAnalysis)
add_subdirectory(SecretnessAnalysis)
```

lib/Dialect/ArithExt/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+add_subdirectory(IR)
+add_mlir_dialect_library(MLIRArithExt
```
AlexanderViand-Intel, line 2: It would be nice if the prefix was consistent across all CMake targets, currently it seems like some are prefixed with `MLIR` and some with `MLIRHEIR`. Maybe we just want to go for `HEIR`?

```suggestion
add_mlir_dialect_library(HEIRArithExt
```

lib/Dialect/BGV/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+add_subdirectory(IR)
+add_subdirectory(Transforms)
+add_mlir_dialect_library(MLIRBGV
+        IR/BGVDialect.cpp
+        Transforms/AddClientInterface.cpp
```
AlexanderViand-Intel, line 5: I'm not sure if this is an issue, but in Bazel, the transforms are in their own target (and therefore library), whereas in the current cmake, they're being combined into a single target.
One downside to this that I could forsee is keeping dependency lists in sync between Bazel and CMake, as CMake would have some of the dependencies "hidden".

lib/Analysis/RotationAnalysis/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_mlir_dialect_library(MLIRHEIRAnalysisROT
```
AlexanderViand-Intel, line 1: I'm not sure if `add_mlir_dialect_library` is the right thing to use here, given that this is an analysis pass? 
According to https://mlir.llvm.org/docs/Tutorials/CreatingADialect/ it should be `add_mlir_conversion_library`?

Also, I think it'd be nice to keep the names of targets consistent across Bazel and CMake (modulo some prefix for CMake), e.g., `HEIRRotationAnalysis` for this target.

tools/CMakeLists.txt
AlexanderViand-Intel, line 1: There's a lot going on here and I'm not quite sure what's happening - if I compare this to `mlir-opt` upstream (https://github.com/llvm/llvm-project/blob/main/mlir/tools/mlir-opt/CMakeLists.txt) or to my own old `heco`-opt (https://github.com/MarbleHE/HECO/blob/main/CMakeLists.txt) they seem to be a lot simpler, but maybe you tried that approach and ran into some issues?

AlexanderViand-Intel (changes requested):
Thanks so much for getting the ball rolling on CMake support for heir! While I've grown to appreciate bazel (although that might just be the stockholm syndrome talking 😉 ) I still remember how badly I missed cmake at the beginning. 

I expect this will make it quite a bit easier for people familiar with MLIR/C++ but not bazel to get started working with and on HEIR :) It might also make upstreaming features/reproducing bugs upstream/etc a bit easier, since there's already a working cmakelists to work from.

I spotted a few issues in my preliminary review and was a bit confused by the /tools CMakeLists.txt, so instead of trying to debug the heir-lsp issues directly, I figured I'd post my comments as-is - with a bit of luck, the lsp thing will sort itself out as part of the general cleanup :)

lib/Dialect/BGV/CMakeLists.txt
AlexanderViand-Intel: Oh, and I'm also wondering if there's anyway we can use `add_mlir_dialect` for the dialects themseves, as thathandles a LOT of the tablegen/etc stuff internally. See https://mlir.llvm.org/docs/Tutorials/CreatingADialect/#tablegen-targets and https://github.com/llvm/llvm-project/blob/main/mlir/cmake/modules/AddMLIR.cmake#L162 for what it does. I _think_ it continues gracefully if any of the files are missing (at least, I used it in HECO for dialects that only had a single *.td).

Alternatively, maybe we can define our own add_heir_dialect that works with our, slightly different, directory structure. My hope is that the simpler we make each CMakeLists.txt file, the more likely we'll be able to keep everything nicely in sync between Bazel and CMake

AlexanderViand-Intel:
> * Does the CMake build have to pass CI to submit a PR? If so, we should add the CI workflow in this or the next PR. If not, do we expect CMake users to periodically fix a broken CMake build (this is what upstream LLVM does for the bazel build).

I think it'd be nice to add this to the CI (and PR requirements). If we extend the /scripts/template.py generator to emit the basic CMakeLists.txt things, too, the overhead should be pretty low for most people.  If something doesn't work "out of the box", I'd be happy to help fix up any PRs with failing CMake.

> * Are we expecting the user to bring their own LLVM repo and set up the path? If so, we should document how to do that somewhere, ideally so it shows up on `heir.dev`, but for now having the CI workflow document it would suffice.
> * How will we keep the LLVM commit in sync between the bazel and CMake builds? At least in the CI, can we parse `bazel/import_llvm.bzl` for the commit hash for this?

I think we could address both of these at the same time by using `ExternalProject_Add` to download and build LLVM via CMake, similar to what bazel does. It should be possible to write a bit of cmake magic to grab the tag from the bazel import_llvm.bzl file, though I probably won't get around to playing around with that until after the WAHC submission deadline (07/26)

lib/Dialect/ArithExt/CMakeLists.txt
ai-mannamalai: I'd just go with what BUILD / Bazel is using so we are consistent with the products of the build

lib/Analysis/CMakeLists.txt
ai-mannamalai: ok added it

lib/Dialect/ArithExt/CMakeLists.txt
ai-mannamalai: resolved in PR updated on Jul 17th, 11PM (PST)

lib/Dialect/BGV/CMakeLists.txt
ai-mannamalai: we had this conversation on Discord; you mentioned you didn't want to touch source files nor the BUILD Bazel files. Indeed my original implementation used the add_mlir_dialect macros in CMake but I had to remove them in place of piecemeal sub-macros to enable the builds to match the unique uses of the HEIR project.

As for the  sub targets in MLIRBGV I should go back and do what is done in BUILD folder; this horrible hack is a side-effect of rushing to get this to work

ai-mannamalai:
Updated to build all the **heir-opt, heir-lsp** and **heir-translate** using CMakeLists.txt

ai-mannamalai:
remaining item seems to be how to hook up all this into the CI system; honestly I'd like to tackle this in a second PR.
cc: @asraa @AlexanderViand-Intel

lib/Analysis/RotationAnalysis/CMakeLists.txt
ai-mannamalai: TBH there is over 60+ files linked against the heir-opt binary; otherwise its plain and simple.

tools/CMakeLists.txt
ai-mannamalai: there are 60+ libraries linked to build the heir-opt which makes this CMake file to be quite long; its comparable with the tools/BUILD file and not in upstream etc. apples :: apples comparison IMHO

lib/Analysis/RotationAnalysis/CMakeLists.txt
ai-mannamalai: done

ai-mannamalai:
> > * Does the CMake build have to pass CI to submit a PR? If so, we should add the CI workflow in this or the next PR. If not, do we expect CMake users to periodically fix a broken CMake build (this is what upstream LLVM does for the bazel build).
> 
> I think it'd be nice to add this to the CI (and PR requirements). If we extend the /scripts/template.py generator to emit the basic CMakeLists.txt things, too, the overhead should be pretty low for most people. If something doesn't work "out of the box", I'd be happy to help fix up any PRs with failing CMake.
> 
> > * Are we expecting the user to bring their own LLVM repo and set up the path? If so, we should document how to do that somewhere, ideally so it shows up on `heir.dev`, but for now having the CI workflow document it would suffice.
> > * How will we keep the LLVM commit in sync between the bazel and CMake builds? At least in the CI, can we parse `bazel/import_llvm.bzl` for the commit hash for this?
> 
> I think we could address both of these at the same time by using `ExternalProject_Add` to download and build LLVM via CMake, similar to what bazel does. It should be possible to write a bit of cmake magic to grab the tag from the bazel import_llvm.bzl file, though I probably won't get around to playing around with that until after the WAHC submission deadline (07/26)

this can be done in a followup PR IMO; issue #818

lib/Dialect/BGV/CMakeLists.txt
AlexanderViand-Intel: Let's merge this PR as-is, and I'll create an issue for me to wirte a custom set of `add_heir_{dialect/conversion/transform/analysis}` CMake macros. With those in place (and updating the right variables), we can then also simplify the tools build to be as simple as the upstream ones.

AlexanderViand-Intel:
I'm getting the following errors during configuration (`cmake -B build -DMLIR_DIR=../llvm-project/build/lib/cmake/mlir`):

``` 
  Found erroneous configuration for source file Utils.cpp

  LLVM's build system enforces that all source files are added to a build
  target, that exactly one build target exists in each directory, and that
  this target lists all files in that directory.  If you want multiple
  targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
  specification, though it is discouraged.

  Please update /home/aviand/code/heir/lib/Conversion/CMakeLists.txt
  ```
```
Found erroneous configuration for source file UnusedMemRef.cpp

  LLVM's build system enforces that all source files are added to a build
  target, that exactly one build target exists in each directory, and that
  this target lists all files in that directory.  If you want multiple
  targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
  specification, though it is discouraged.

  Please update
  /home/aviand/code/heir/lib/Transforms/UnusedMemRef/CMakeLists.txt
```

```
  Found erroneous configuration for source file UnusedMemRef.cpp

  LLVM's build system enforces that all source files are added to a build
  target, that exactly one build target exists in each directory, and that
  this target lists all files in that directory.  If you want multiple
  targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
  specification, though it is discouraged.

  Please update
  /home/aviand/code/heir/lib/Transforms/UnusedMemRef/CMakeLists.txt
```

File: lib/Dialect/BGV/CMakeLists.txt (hunk 6 above)
Comment (ai-mannamalai) [reply lib/Dialect/BGV/CMakeLists.txt:5]: sounds good to me - thanks @AlexanderViand-Intel

File: lib/Transforms/UnusedMemRef/CMakeLists.txt [hunk 8]
```diff
@@ -0,0 +1,20 @@
+set(LLVM_TARGET_DEFINITIONS UnusedMemref.td)
+mlir_tablegen(UnusedMemref.h.inc -gen-pass-decls -name UnusedMemRef)
+add_public_tablegen_target(MLIRHeirUnusedMemRefIncGen)
+
+add_mlir_dialect_library(MLIRHeirUnusedMemRef
+    PARTIAL_SOURCES_INTENDED
+    UnusedMemref.cpp
```
AlexanderViand-Intel, line 7: mismatched case causes build issues on linux
```suggestion
    UnusedMemRef.cpp
```

lib/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+add_mlir_dialect_library(MLIRHEIRUtils
+    PARTIAL_SOURCES_INTENDED
+    utils.cpp
```
AlexanderViand-Intel, line 3:
```suggestion
    Utils.cpp
```

lib/Source/AutoHog/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# AutoHoG importer library
+find_package(rapidjson)
```
AlexanderViand-Intel, line 2: Since the RapidJSON dependency probably isn't going to be updated frequently, I think it's save to already include it in this PR with a hardcoded git commit.

```suggestion
include(ExternalProject)

# Download RapidJSON
# TODO (#818): automatically extract the commit from the bazel WORKSPACE file 
ExternalProject_Add(
    rapidjson
    PREFIX "vendor/rapidjson"
    GIT_REPOSITORY "https://github.com/Tencent/rapidjson.git"
    GIT_TAG f54b0e47a08782a6131cc3d60f94d038fa6e0a51
    TIMEOUT 10
    CMAKE_ARGS
        -DRAPIDJSON_BUILD_TESTS=OFF
        -DRAPIDJSON_BUILD_DOC=OFF
        -DRAPIDJSON_BUILD_EXAMPLES=OFF
    CONFIGURE_COMMAND ""
    BUILD_COMMAND ""
    INSTALL_COMMAND ""
    UPDATE_COMMAND ""
)

# Prepare RapidJSON (RapidJSON is a header-only library)
ExternalProject_Get_Property(rapidjson source_dir)
set(RAPIDJSON_INCLUDE_DIR ${source_dir}/include)
```

AlexanderViand-Intel (changes requested):
With your last update + these additional changes below, the project now configures succesfully for me :) 
 
However when I try to build (`cmake --build build --target heir-opt`), I get  an error that it cannot find `/mlir/include/mlir/IR/Builders.h`

```
fatal error: mlir/include/mlir/IR/Builders.h: No such file or directory
    4 | #include "mlir/include/mlir/IR/Builders.h"               // from @llvm-project
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

ai-mannamalai:
> With your last update + these additional changes below, the project now configures succesfully for me :)
> 
> However when I try to build (`cmake --build build --target heir-opt`), I get an error that it cannot find `/mlir/include/mlir/IR/Builders.h`
> 
> ```
> fatal error: mlir/include/mlir/IR/Builders.h: No such file or directory
>     4 | #include "mlir/include/mlir/IR/Builders.h"               // from @llvm-project
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> ```

I fixed this bug in the most recent push at 2:40PM PST.

ai-mannamalai:
> I'm getting the following errors during configuration (`cmake -B build -DMLIR_DIR=../llvm-project/build/lib/cmake/mlir`):
> 
> ```
>   Found erroneous configuration for source file Utils.cpp
> 
>   LLVM's build system enforces that all source files are added to a build
>   target, that exactly one build target exists in each directory, and that
>   this target lists all files in that directory.  If you want multiple
>   targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
>   specification, though it is discouraged.
> 
>   Please update /home/aviand/code/heir/lib/Conversion/CMakeLists.txt
> ```
> 
>  ```
> Found erroneous configuration for source file UnusedMemRef.cpp
> 
>   LLVM's build system enforces that all source files are added to a build
>   target, that exactly one build target exists in each directory, and that
>   this target lists all files in that directory.  If you want multiple
>   targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
>   specification, though it is discouraged.
> 
>   Please update
>   /home/aviand/code/heir/lib/Transforms/UnusedMemRef/CMakeLists.txt
> ```
> 
>  ```
>   Found erroneous configuration for source file UnusedMemRef.cpp
> 
>   LLVM's build system enforces that all source files are added to a build
>   target, that exactly one build target exists in each directory, and that
>   this target lists all files in that directory.  If you want multiple
>   targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
>   specification, though it is discouraged.
> 
>   Please update
>   /home/aviand/code/heir/lib/Transforms/UnusedMemRef/CMakeLists.txt
> ```
> 
> LLVM commit is acc159aea1e641e3694ab8fe5faa231788077011 (current integrate for HEIR), CMake version is 3.28.3

resolved

lib/Dialect/CGGI/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+        ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+        DEPENDS
+        MLIRCGGIIncGen
+        MLIRCGGIOpsIncGen
+        MLIRCGGIPassesIncGen
+        MLIRCGGIAttributesIncGen
+
+        LINK_LIBS PUBLIC
+        MLIRIR
```
AlexanderViand-Intel, line 18:
```suggestion
        LINK_LIBS PUBLIC
        MLIRHEIRInterfaces
        MLIRLWE
        MLIRIR
```

lib/Dialect/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+add_mlir_dialect_library(MLIRHEIRInterfaces
+    HEIRInterfaces.cpp
+
+    ADDITIONAL_HEADER_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/IR
+
+    DEPENDS
+    MLIRHEIRInterfacesOpsIncGen
+    HEIRIncludeMLIR
+    HEIRIncludeLLVM
```
AlexanderViand-Intel, line 32:
```suggestion
    DEPENDS
    MLIRHEIRInterfacesOpsIncGen
```

lib/Source/AutoHog/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# Prepare RapidJSON (RapidJSON is a header-only library)
+ExternalProject_Get_Property(rapidjson source_dir)
+set(RAPIDJSON_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/vendor/rapidjson/src/rapidjson/include/)
+
+add_mlir_dialect_library(MLIRAutoHogImporter
+    AutoHogImporter.cpp
+
+    ADDITIONAL_HEADER_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/
+
```
AlexanderViand-Intel, line 30:
```suggestion
    ADDITIONAL_HEADER_DIRS
    ${CMAKE_CURRENT_SOURCE_DIR}/
    
    DEPENDS
    rapidjson

```

CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+                  DEPENDS ${INCLUDE_MLIR_LLVM_HEADERS}
+                  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/inc/llvm/include/"
+                  )
+
+add_custom_target(HEIRIncludeMLIR
+                  BYPRODUCTS ${CMAKE_BINARY_DIR}/inc/mlir/include/mlir
+                  COMMAND [ -e mlir ] && echo "" || ln -s ${MLIR_INCLUDE_DIRS}/mlir/ ./
+                  DEPENDS ${INCLUDE_MLIR_LLVM_HEADERS}
+                  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/inc/mlir/include/")
+message(STATUS "Adding MLIR and LLVM includes.")
```
AlexanderViand-Intel, line 58:
```suggestion

```

CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+
+include_directories(${CMAKE_BINARY_DIR}/inc)
+include_directories(${LLVM_INCLUDE_DIRS})
+include_directories(${MLIR_INCLUDE_DIRS})
```
AlexanderViand-Intel, line 62: This little trick removes the need for the custom command (which unfortunately didn't work for me):
* first set of includes is for heir code (which uses a very long include path for some reason)
* second set is so that includes in included mlir/llvm code (shorter path) work, too
```suggestion
include_directories(${LLVM_INCLUDE_DIRS}/../..)
include_directories(${MLIR_INCLUDE_DIRS}/../..)
include_directories(${LLVM_INCLUDE_DIRS})
include_directories(${MLIR_INCLUDE_DIRS})
```

README.md
```diff
@@ -33,6 +33,12 @@ There are many ways to contribute to HEIR:
         to alert us of papers that should be investigated and incorporated into
         HEIR.
 
+## Building with CMake
+HEIR supports building with CMake with exception of building Yosys pass, tests, and examples at the time of writing.
+```bash
+$ cmake -S ./ -B builddir -DCMAKE_PREFIX_PATH=/usr/local/lib/cmake -DMLIR_CMAKE_DIR=/usr/local/lib/cmake/mlir/ -DLLVM_CMAKE_DIR=/usr/local/lib/cmake/llvm/
+```
+
```
AlexanderViand-Intel, line 41: The other build documentation is in docs/content/docs/getting_started.md, and that's where I'd put the CMake documentation, too. However, I'd actually hold off on that a bit, until the CMake stuff is more mature.
```suggestion
```

tools/CMakeLists.txt
```diff
@@ -0,0 +1,163 @@
...
+        MLIRConvertToLLVMPass
+        MLIRFuncDialect
+        MLIRFuncToLLVM
+        MLIRFuncTransforms
+        MLIRIndexToLLVM
+        MLIRLLVMDialect
+        MLIRLinalgTransforms
+        MLIRMemRefDialect
+        MLIRMemRefToLLVM
+        MLIRMemRefTransforms
```
AlexanderViand-Intel, line 118:
```suggestion
        MLIRMemrefToArithMemrefGlobalReplace
        MLIRMemRefToLLVM
        MLIRMemRefTransforms
```

AlexanderViand-Intel (changes requested):
With the below changes, I can get it to configure (`cmake -B build -DMLIR_DIR=../llvm-project/build/lib/cmake/mlir`), and compile (`cmake --build build --target heir-opt -j`), and the same for `heir-lsp` and `heir-translate`  :)

CMakeLists.txt
ai-mannamalai: Okay - let me see

ai-mannamalai:
> With the below changes, I can get it to configure (`cmake -B build -DMLIR_DIR=../llvm-project/build/lib/cmake/mlir`), and compile (`cmake --build build --target heir-opt -j`), and the same for `heir-lsp` and `heir-translate` :)

this seems to work on my Linux machine against a build LLVM folder; however it does not work against the installed LLVM / MLIR which is a kicker.

AlexanderViand-Intel:
> > With the below changes, I can get it to configure (`cmake -B build -DMLIR_DIR=../llvm-project/build/lib/cmake/mlir`), and compile (`cmake --build build --target heir-opt -j`), and the same for `heir-lsp` and `heir-translate` :)
> 
> this seems to work on my Linux machine against a build LLVM folder; however it does not work against the installed LLVM / MLIR which is a kicker. The former code changes would do that. Let me see if I can resolve the two worlds

That's odd! Though I guess installed LLVM isn't really supportable long-term anyway, because of the LLVM integrate synchronization.

ai-mannamalai:
> > > With the below changes, I can get it to configure (`cmake -B build -DMLIR_DIR=../llvm-project/build/lib/cmake/mlir`), and compile (`cmake --build build --target heir-opt -j`), and the same for `heir-lsp` and `heir-translate` :)
> > 
> > 
> > this seems to work on my Linux machine against a build LLVM folder; however it does not work against the installed LLVM / MLIR which is a kicker. The former code changes would do that. Let me see if I can resolve the two worlds
> 
> That's odd! Though I guess installed LLVM isn't really supportable long-term anyway, because of the LLVM integrate synchronization.

I introduced a optional compile pathway like this,
```bash
$ cmake -S ./ -B builddir -DENABLE_INSTALLED_MLIR=TRUE -DCMAKE_PREFIX_PATH=/usr/local/lib/cmake -DMLIR_CMAKE_DIR=/usr/local/lib/cmake/mlir/ -DLLVM_CMAKE_DIR=/usr/local/lib/cmake/llvm/  && cmake —build builddir -j 
```

the validation of both pathways is checked out on my Apple Mac OSX build
https://gist.github.com/ai-mannamalai/7d097bb610422889e8ec465b8ccea039

ai-mannamalai:
If you're OK with the changes we can squash them and enable the tests etc.

ai-mannamalai:
x-checked after final push works as expected with the optional flag.

ai-mannamalai:
> x-checked after final push works as expected with the optional flag.

rebase as 1 commit

CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+endif()
+
+include_directories(${CMAKE_BINARY_DIR}/inc)
+if(ENABLE_INSTALLED_MLIR)
+include_directories(${LLVM_INCLUDE_DIRS})
+include_directories(${MLIR_INCLUDE_DIRS})
+else()
+include_directories(${LLVM_INCLUDE_DIRS}/../..)
+include_directories(${MLIR_INCLUDE_DIRS}/../..)
+endif()
```
AlexanderViand-Intel, line 77: For the non-installed CMAKE option, both the "long" and "short" form of the include paths are needed, but otherwise everything seems to work 👍 
```suggestion
else()
include_directories(${LLVM_INCLUDE_DIRS})
include_directories(${MLIR_INCLUDE_DIRS})
include_directories(${LLVM_INCLUDE_DIRS}/../..)
include_directories(${MLIR_INCLUDE_DIRS}/../..)
endif()
```

AlexanderViand-Intel (approved):
Awesome!

ai-mannamalai:
thanks @AlexanderViand-Intel - I x-checked as well and rebase to single commit;

AlexanderViand-Intel (approved):

AlexanderViand-Intel:
> thanks @AlexanderViand-Intel - I x-checked as well and rebase to single commit;

@ai-mannamalai Did you push the latest rebase? I still see 2 commits in the PR :) 

Oddly enough, the "PR review status / require-single-commit-on-approval (pull_request_review)" check has nevertheless passed - @asraa: any idea what's going on there?

ai-mannamalai:
@AlexanderViand-Intel - I did not rebase to single commit - you were right - forgot to push earlier. Now should be fixed up.

lib/Dialect/ArithExt/IR/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+mlir_tablegen(ArithExtDialect.h.inc -gen-dialect-decls)
+mlir_tablegen(ArithExtDialect.cpp.inc -gen-dialect-defs)
+add_public_tablegen_target(MLIRArithExtIncGen)
+
+set(LLVM_TARGET_DEFINITIONS ArithExtOps.td)
+mlir_tablegen(ArithExtOps.h.inc -gen-op-decls)
+mlir_tablegen(ArithExtOps.cpp.inc -gen-op-defs)
+add_public_tablegen_target(MLIRArithExtOpsIncGen)
+
+add_mlir_doc(ArithExtDialect ArithExtDialect arithext/ -gen-dialect-doc --dialect=arithext)
```
j2kun, line 11: We don't need any mlir docs, since CMake is not load-bearing for generating them. Let's remove all these calls.

j2kun (changes requested):
I won't be able to provide much of a meaningful review, so it seems LGTM. CC @asraa to let here know this is a light review before being pull into Google

lib/Dialect/ArithExt/IR/CMakeLists.txt
ai-mannamalai: ok

ai-mannamalai: - remove add_mlir_doc as per review from @Jeremy Khun in latest push.

j2kun (approved):

AlexanderViand-Intel:
Pinging @asraa :)

asraa:
> Pinging @asraa :)

Sorry about that! @code-perspective was prepping the internal merge, I think it should be ready to merge soon

PR #789, opened 2024-07-12
Restricts ops and dialects

### LINKED ISSUE #768: Add an (optional) pass parameter to `--convert-elementwise-to-affine` to restrict it to certain dialects and/or operations
Currently, `--convert-elementwise-to-affine` applies to all `ElementwiseMappable` operations with `RankedTensorType` operands. 
This is fine for the current intended usage in HEIR, but the pass would be more widely useable with this added feature.

ahmedshakill:
Adds options to restrict dialects and ops to run `--convert-elementwise-to-affine` pass on.

Resolves #768

### DIFF
--- a/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
+++ b/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
@@ -130,10 +130,16 @@ struct ElementwiseToAffine
     RewritePatternSet patterns(context);
 
     patterns.add<ConvertAnyElementwiseMappableOpOnRankedTensors>(context);
-    target.markUnknownOpDynamicallyLegal([](Operation *op) {
-      // TODO (#768): to make this pass more widely applicable, it should take
-      // a dialect and/or list of operations to restrict the conversion to
-      return !isElementwiseMappableOpOnRankedTensors(op);
+    target.markUnknownOpDynamicallyLegal([&](Operation *op) {
+      bool convertAll = convertDialects.empty() && convertOps.empty();
+      bool convertDialect = llvm::is_contained(
+          convertDialects, op->getDialect()->getNamespace().str());
+      bool convertOp =
+          llvm::is_contained(convertOps, op->getName().getStringRef().str());
+
+      if (convertAll || convertDialect || convertOp)
+        return !isElementwiseMappableOpOnRankedTensors(op);
+      return true;
     });
 
     if (failed(applyPartialConversion(getOperation(), target,
--- a/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
+++ b/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
@@ -8,10 +8,30 @@ def ElementwiseToAffine : Pass<"convert-elementwise-to-affine"> {
   let description = [{
     This pass lowers ElementwiseMappable operations over tensors
     to affine loop nests that instead apply the operation to the underlying scalar values.
+
+    Usage:
+    '--convert-elementwise-to-affine=convert-ops=arith.mulf '
+      restrict conversion to mulf op from arith dialect.
+
+    '--convert-elementwise-to-affine=convert-ops=arith.addf,arith.divf  convert-dialects=bgv'
+      restrict conversion to addf and divf ops from arith dialect and all of the ops in bgv dialect.
+
+    --convert-elementwise-to-affine=convert-dialects=arith
+      restrict conversion to arith dialect so ops only from arith dialect is processed.
+
+    --convert-elementwise-to-affine=convert-ops=arith.addf,arith.mulf
+      restrict conversion only to these two ops - addf and mulf - from arith dialect.
+
   }];
   let dependentDialects = [
     "mlir::affine::AffineDialect",
     "mlir::tensor::TensorDialect"
+  ];
+  let options = [
+    ListOption<"convertOps","convert-ops", "std::string",
+                "comma-separated list of ops to run this pass on ">,
+    ListOption<"convertDialects","convert-dialects", "std::string",
+                "comma-separated list of dialects to run this pass on ">,
   ];
 }
 
--- a/tests/polynomial/elementwise_to_affine.mlir
+++ b/tests/polynomial/elementwise_to_affine.mlir
@@ -1,52 +1,112 @@
-// RUN: heir-opt --convert-elementwise-to-affine %s | FileCheck --enable-var-scope %s
+// THE FOLLOWING SHOULD CONVERT BOTH ADD AND MUL
+
+// RUN: heir-opt --convert-elementwise-to-affine %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_MUL %s
+
+// RUN: heir-opt '--convert-elementwise-to-affine=convert-ops=polynomial.add,polynomial.mul_scalar' %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_MUL %s
+
+// RUN: heir-opt '--convert-elementwise-to-affine=convert-ops=polynomial.add,polynomial.mul_scalar  convert-dialects=arith' %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_MUL %s
+
+// RUN: heir-opt --convert-elementwise-to-affine=convert-dialects=polynomial %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_MUL %s
+
+// RUN: heir-opt '--convert-elementwise-to-affine=convert-dialects=polynomial convert-ops=arith.addi' %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_MUL %s
+
+// THE FOLLOWING SHOULD ONLY CONVERT ADD (AND "ARITH" OPS FOR SOME, BUT THERE ARE NONE IN THE TESTS)
+
+// RUN: heir-opt --convert-elementwise-to-affine=convert-ops=polynomial.add %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_NOTMUL %s
+
+// RUN: heir-opt '--convert-elementwise-to-affine=convert-dialects= convert-ops=polynomial.add' %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_NOTMUL %s
+
+// RUN: heir-opt '--convert-elementwise-to-affine=convert-dialects=arith convert-ops=polynomial.add' %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_ADD --check-prefix=CHECK_NOTMUL %s
+
+// THE FOLLOWING SHOULD ONLY CONVERT MUL (AND "ARITH" OPS FOR SOME, BUT THERE ARE NONE IN THE TESTS)
+
+// RUN: heir-opt --convert-elementwise-to-affine=convert-ops=polynomial.mul_scalar %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_NOTADD --check-prefix=CHECK_MUL %s
+
+// RUN: heir-opt '--convert-elementwise-to-affine=convert-dialects= convert-ops=polynomial.mul_scalar' %s \
+// RUN: | FileCheck --enable-var-scope --check-prefix=CHECK --check-prefix=CHECK_NOTADD --check-prefix=CHECK_MUL %s
+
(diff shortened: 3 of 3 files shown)


lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
```diff
@@ -13,6 +13,12 @@ def ElementwiseToAffine : Pass<"convert-elementwise-to-affine"> {
     "mlir::affine::AffineDialect",
     "mlir::tensor::TensorDialect"
   ];
+  let options = [
```
asraa, line 16: could you please add a quick comment in the pass description to describe how to use these options? e.g. a user can *either* set dialects or ops (it's not necessary to to dialects=polynomial if using ops=polynomial.add)

tests/polynomial/elementwise_to_affine.mlir
```diff
@@ -1,5 +1,6 @@
 // RUN: heir-opt --convert-elementwise-to-affine %s | FileCheck --enable-var-scope %s
-
+// RUN: heir-opt --convert-elementwise-to-affine=ops-to-consider="polynomial.add,polynomial.mul_scalar" %s | FileCheck --enable-var-scope %s
```
asraa, line 2: for simplicity, should we rename to `restrict-ops`, and `restrict-dialects`? welcome for any feedback!

ahmedshakill: these names sounds good.

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
ahmedshakill: From the semantic side, I am thinking
`'restric-ops=polynomial.add,polynomial.mul_scalar restrict-dialects=polynomial'`   would allow all instructions in polynomial dialect. 
And 
` 'restrict-ops=polynomial.add,polynomial.mul_scalar restrict-dialects=bgv'`
in the above case it will allow two ops from polynomial case and all ops from bgv.

asraa: I agree that makes sense to me! Especially the second example you had would be a good example to add to the description

ahmedshakill:
Hi, I am getting an  error like this
`failed to legalize operation 'builtin.module' that was explicitly marked illegal` 
while running this 
`RUN: heir-opt --convert-elementwise-to-affine=restrict-dialects=polynomial %s | FileCheck --enable-var-scope %s`
where I am restricting dialect to only polynomial. Should I also include some other dialect besides polynomial here?

j2kun:
IMO the builtin dialect should always be considered legal (in the code), and you may also want to add in purely structural dialects like func. Not because the lowering will do anything to them, but to ensure they're considered legal.

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
ahmedshakill: Sure

ahmedshakill:
It seems it's almost ready. Would you take a look.

asraa (commented):
Logic seems well-structured to me!

If you want to add tests with different results in that test file, you can also use FileCheck's `--check-prefix` option (https://llvm.org/docs/CommandGuide/FileCheck.html#cmdoption-FileCheck-check-prefix), we have a few examples in the repo.

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
```diff
@@ -130,10 +129,29 @@ struct ElementwiseToAffine
...
       // TODO (#768): to make this pass more widely applicable, it should take
       // a dialect and/or list of operations to restrict the conversion to
-      return !isElementwiseMappableOpOnRankedTensors(op);
+
+      auto isDialectInList =
+          llvm::any_of(restrictDialects, [&](std::string dialectName) {
+            return dialectName == op->getDialect()->getNamespace().str();
+          });
+      auto isOpInList = llvm::any_of(restrictOps, [&](std::string opName) {
+        return opName == op->getName().getStringRef().str();
```
AlexanderViand-Intel, line 141: You might be able to simplify these a bit by using `llvm::is_contained` (See [this upstream example of a similar pass option](https://github.com/AlexanderViand-Intel/llvm-project/blob/ebeaebf7f961f8d8c1043e0a5d3dc33758acfdb4/mlir/lib/Dialect/Bufferization/Transforms/Bufferize.cpp#L256-L259))

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
```diff
@@ -130,10 +129,29 @@ struct ElementwiseToAffine
...
+
+      if (!restrictDialects.empty() && !restrictOps.empty()) {
+        return (isOpInList || isDialectInList) ? !is_EWMtoRT : true;
+      } else if (!restrictDialects.empty()) {
+        return isDialectInList ? !is_EWMtoRT : true;
+      } else if (!restrictOps.empty()) {
+        return isOpInList ? !is_EWMtoRT : true;
+      } else {
+        return !is_EWMtoRT;
+      }
```
AlexanderViand-Intel, line 154: Rather than having this complex if/else here, I'd just set `isDialectInList` to true if there's no filter list specified, same for `isOpInList` (maybe rename them to `isDialectElegible` and `isOpElegibile` to make it less confusing), then you can simplify this to a single line:
```suggestion
        return (isDialectElegible || isOpElegibile) ? !is_EWMtoRT : true;
```

AlexanderViand-Intel (changes requested):
Thanks for putting this together and sorry for the delay on reviewing this! 

I've suggested a few changes, and there are also a few issues with formatting/etc, so make sure that you have pre-commit setup correctly (follow the steps from [Getting Started | Developing in HEIR](https://heir.dev/docs/getting_started/#developing-in-heir)).

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
ahmedshakill: Yeah, that much if-else there is complex and confusing there. 
But the test with no restriction 
`// RUN: heir-opt --convert-elementwise-to-affine %s | FileCheck --enable-var-scope %s`
fails with the simplification.

I find at least this much to be necessary
` if (restrictDialects.empty() && restrictOps.empty()) {
        return !is_EWMtoRT;
      }`
 `else {
        return (isDialectElegible || isOpElegible) ? !is_EWMtoRT : true;
      }`

or is there any other way to know that no restriction was provided and we can return the  `!is_EWMtoRT`

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
```diff
@@ -8,11 +8,29 @@ def ElementwiseToAffine : Pass<"convert-elementwise-to-affine"> {
   let description = [{
     This pass lowers ElementwiseMappable operations over tensors
     to affine loop nests that instead apply the operation to the underlying scalar values.
+
```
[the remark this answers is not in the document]
ahmedshakill: switched to arith ops

tests/polynomial/elementwise_to_affine.mlir
```diff
@@ -1,4 +1,7 @@
 // RUN: heir-opt --convert-elementwise-to-affine %s | FileCheck --enable-var-scope %s
+// RUN: heir-opt '--convert-elementwise-to-affine=restrict-ops=polynomial.add,polynomial.mul_scalar  restrict-dialects=polynomial' %s | FileCheck --enable-var-scope %s
+// RUN: heir-opt --convert-elementwise-to-affine=restrict-dialects=polynomial %s | FileCheck --enable-var-scope %s
+// RUN: heir-opt --convert-elementwise-to-affine=restrict-ops=polynomial.add,polynomial.mul_scalar %s | FileCheck --enable-var-scope %s
```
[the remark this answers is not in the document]
ahmedshakill: I've added some tests in the file. please let me know what you think.

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
asraa: > I find at least this much to be necessary

I think I agree, I'm struggling to work out the boolean logic in a simpler way (so I'd argue that if it's taking me this long to figure it out then what you have is more readable right now).

tests/polynomial/elementwise_to_affine.mlir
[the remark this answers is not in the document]
asraa: The test RUNs themselves look good to me now! If I can just suggest testing the positive conditions with the CHECK_POLY_MUL/CHECK_POLY_ADD - I think right now you are checking the cases of poly mul/add *not* being converted, but it might be better to check that the ops being restricted to are also being selectively converted.

In that way though you wouldn't be able to combine both CHECK-* conditions on the both case, which is maybe why you had done it that way.

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
AlexanderViand-Intel: My suggestion was to move the `empty` checks up into the same variable as the `contained` check:
```cpp
auto isDialectElegible =
    restrictDialects.empty() ||
    llvm::is_contained(restrictDialects,
                        op->getDialect()->getNamespace().str());
auto isOpElegible =
    restrictOps.empty() ||
    llvm::is_contained(restrictOps, op->getName().getStringRef().str());

auto is_EWMtoRT = isElementwiseMappableOpOnRankedTensors(op);

return (isDialectElegible && isOpElegible) ? !is_EWMtoRT : true;
 ```
I did have a logical error in my original suggestion, though, as it should be `&&` for the final line

File: tests/polynomial/elementwise_to_affine.mlir (hunk 6 above)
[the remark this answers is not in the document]
Comment (AlexanderViand-Intel) [reply tests/polynomial/elementwise_to_affine.mlir]: Here's a version that doesn't require the separate test functions at the bottom:

https://gist.github.com/AlexanderViand-Intel/2cee3b1b816028a83c4516935974d63a

**Review by AlexanderViand-Intel — CHANGES_REQUESTED** [review]
See comments in discussions above for suggestions on the logic thing & tests

EDIT: never mind, playing around with it a bit more made me realize I'm not sure I understand the intended semantcs/what they should be properly. Will come back to this later today with fresh eyes!

File: tests/polynomial/elementwise_to_affine.mlir
Comment (AlexanderViand-Intel) [inline tests/polynomial/elementwise_to_affine.mlir:1]: Here's a version (already with the renamed command) that doesn't need separate functions + tests _a lot_ more variations: https://gist.github.com/AlexanderViand-Intel/2cee3b1b816028a83c4516935974d63a

File: lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp [hunk 7]
```diff
@@ -130,10 +130,19 @@ struct ElementwiseToAffine
...
+          restrictDialects, op->getDialect()->getNamespace().str());
+      auto isOpElegible =
+          llvm::is_contained(restrictOps, op->getName().getStringRef().str());
+
+      auto is_EWMtoRT = isElementwiseMappableOpOnRankedTensors(op);
+
+      if (restrictDialects.empty() && restrictOps.empty()) {
+        return !is_EWMtoRT;
+      } else {
+        return (isDialectElegible || isOpElegible) ? !is_EWMtoRT : true;
```
AlexanderViand-Intel, line 144: You were right, there really is no way to avoid the empty %% empty check, sorry about sending you on a wild goose chase there! 
```suggestion
      bool convertAll = convertDialects.empty() && convertOps.empty();
      bool convertDialect = llvm::is_contained(
          convertDialects, op->getDialect()->getNamespace().str());
      bool convertOp =
          llvm::is_contained(convertOps, op->getName().getStringRef().str());

      if (convertAll || convertDialect || convertOp)
        return !isElementwiseMappableOpOnRankedTensors(op);
      return true;
```

AlexanderViand-Intel (approved):
Sorry about the back and forth with this, I think I was a bit confused about what the semantics of the options should be, but your interpretation makes a lot more sense!

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.td
```diff
@@ -8,11 +8,31 @@ def ElementwiseToAffine : Pass<"convert-elementwise-to-affine"> {
...
   }];
   let dependentDialects = [
     "mlir::affine::AffineDialect",
     "mlir::tensor::TensorDialect"
   ];
+  let options = [
+    ListOption<"restrictOps","restrict-ops", "std::string",
+                "comma-separated list of ops to run this pass on ">,
+    ListOption<"restrictDialects","restrict-dialects", "std::string",
```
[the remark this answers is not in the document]
ahmedshakill: renamed accordingly

tests/polynomial/elementwise_to_affine.mlir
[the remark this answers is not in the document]
ahmedshakill: Thanks for investing so much time. There's a lot to learn about FileCheck !

AlexanderViand-Intel (approved):
LGTM! Just squash/rebase and this should be pull-ready :)

AlexanderViand-Intel:
@ahmedshakill   I noticed you keep running into CI formatting issues, so make sure you've followed the steps at https://heir.dev/docs/development/#pre-commit

ahmedshakill:
I should've done that earlier. Setting up that now.

AlexanderViand-Intel (approved):

asraa (approved):
Thank you so much for all the patience @ahmedshakill !

ahmedshakill:
Thanks to all of you for your support.

PR #778, opened 2024-07-09
Add Secretness Analysis to convert-if-to-select 

MeronZerihun:
Added a static dataflow analysis for secret types and used it in the convert-if-to-select transformation pass to selectively apply the transformation on scf.if operations that evaluate secret conditions.

PR #793, opened 2024-07-12
Bug fix for visitOperation() not propagating taints for constants and operands with unknown secretness state

MeronZerihun:
My secretness analysis in PR #778 didn't propagate the Secretness state when operations don't have operands, e.g. a ConstantOp, since my original code only sets the Secretness state if operands exist. 

Theoretically, given an operation, all operands should have a Secretness state set before state propagates to the operation's result. Seeing that it's possible to miss setting the Secretness state due to buggy code, I have also added code to set the Secretness state of an operation's result to 'None' if we in case find an operand whose Secretness state has not been initialized.


j2kun (changes requested):
LGTM. Is there a simple test case you can add to exercise the behavior?

MeronZerihun (commented):
I added a new test case called set_secretness_for_constant(...) under 'secret_condtionals.mlir' that has this behavior.

j2kun (approved):

Issue #780, opened 2024-07-10
Add Documentation for `MulDepthAnalysis`


AlexanderViand-Intel:
I didn't notice this when the PR was up, but `MulDepthAnalysis.h` does not have any documentation (no comments on `class MulDepth` for IDE's/doc generators to pick up).  For comparison, see `TargetSlotAnalysis.h` or `RotationAnalysis.h`. 

While the MulDepth analysis is a straight forward concept, we should still to document a few things, e.g., what counts as a multiplication (recall the discussion we had about plaintext/scalar muls).

Original PR: 
* #760

Maokami:
Good point! I will do that as soon as possible.

PR #799, opened 2024-07-19
Add Documentation for MulDepthAnalysis

Maokami:
This PR contains adding more comments in MulDepthAnalysis. (#780)

### DIFF
--- a/lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
+++ b/lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
@@ -9,14 +9,45 @@
 namespace mlir {
 namespace heir {
 
+/// This class represents the multiplicative depth in an FHE program.
+///
+/// The multiplicative depth of a circuit refers to the maximum number of
+/// multiplication gates on any path from a circuit input to an output. In a
+/// more general FHE program, the multiplicative depth is the maximum number of
+/// ciphertext-ciphertext or plaintext-ciphertext multiplication operations on
+/// any use-def chain.
+///
+/// In FHE, multiplicative depth bounds the number of operations that can be
+/// sequentially performed on a ciphertext before the noise introduced during
+/// those operations renders the ciphertext undecryptable.
+///
+/// For example, the multiplicative depth of the following pseudo-code is 2:
+///
+///     %0 = arith.constant 0 : multiplicative depth = 0
+///     %1 = FHE.mul %0 %0 : multiplicative depth = 1
+///     %2 = FHE.add %0 %1 : multiplicative depth = 1
+///     %3 = FHE.mul %2 %0 : multiplicative depth = 2
+///
+/// In more detail, the multiplicative depth for an arbitrary operation `op`
+/// can be defined inductively as follows.
+///
+///     MulDepth(z) = case
+///       mul_op(x, y):
+///         max(MulDepth(x), MulDepth(y)) + 1
+///       any_op(operands):
+///         max(map(MulDepth, operands))
+
 class MulDepth {
  public:
   MulDepth() : value(std::nullopt) {}
   explicit MulDepth(int64_t value) : value(value) {}
   ~MulDepth() = default;
 
+  /// Whether the multiplicative depth is initialized. It can be uninitialized
+  /// when the state hasn't been set during the analysis.
   bool isInitialized() const { return value.has_value(); }
 
+  /// Get a known multiplicative depth.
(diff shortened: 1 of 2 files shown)


lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -9,31 +9,75 @@
...
+///     \[
+///     \text{MulDepth}(\text{op}) =
+///     \begin{cases}
+///     \max\limits_{i}
+///     \left(\text{MulDepth}(\text{operand}_i(\text{op}))\right) + 1 & \text{if
+///     op is multiplication} \\ \max\limits_{i}
+///     \left(\text{MulDepth}(\text{operand}_i(\text{op}))\right) &
+///     \text{otherwise} \end{cases}
+///     \]
+
```
Maokami, line 46: I'm not sure this latex comment can be supported?

j2kun: It is not supported in the comments here, but if you put it in the tablegen, it will be rendered as latex on heir.dev

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -9,31 +9,75 @@
 namespace mlir {
 namespace heir {
 
+/// This class represents the multiplicative depth in an FHE program.
+///
+/// The multiplicative depth in an FHE program refers to the maximum number of
```
j2kun, line 14: I would rephrase this more precisely:

> The multiplicative depth of a circuit refers to the maximum number
> of multiplication gates on any path from a circuit input to
> an output. In a more general FHE program, the multiplicative depth
> is the maximum number of ciphertext-ciphertext or plaintext-ciphertext
> multiplication operations on any use-def chain.
> 
> In FHE, multiplicative depth bounds the number of operations
> that can be sequentially performed on a ciphertext before the
> noise introduced during those operations renders the ciphertext 
> undecryptable.

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
j2kun: What about this pseudocode instead?

```
MulDepth(z) = case
  mul_op(x, y):
    max(MulDepth(x), MulDepth(y)) + 1
  any_op(operands):
    max(map(MulDepth, operands))
```

j2kun (approved):
Minor suggestions only!

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
Maokami: nice!

j2kun (approved):

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
```diff
@@ -9,31 +9,68 @@
...
   bool operator==(const MulDepth &rhs) const { return value == rhs.value; }
 
+  /// Join two multiplicative depth.
   static MulDepth join(const MulDepth &lhs, const MulDepth &rhs) {
     if (!lhs.isInitialized()) return rhs;
     if (!rhs.isInitialized()) return lhs;
+    // If both are initialized, the larger value should be chosen for a sound
+    // analysis of the multiplicative depth. If both are initialized, the larger
+    // value between the two should be selected to soundly analyze the
+    // multiplicative depth.
```
asraa, line 65: Maybe this was a copy-paste thing?

```suggestion
    // If both are initialized, the larger value should be chosen for a sound
    // analysis of the multiplicative depth.
```

asraa (commented):
Looks great!

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
Maokami: Oops!

Maokami: @asraa Could you possibly squash merge this commit? Otherwise, I can do it after 12 hours :)

AlexanderViand-Intel:
Thanks for putting this together!

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.h
asraa: done!

Issue #809, opened 2024-07-23
--wrap-generic rewrapping code already in secret.generic


MeronZerihun:
I am not sure if this is a behavior of any benefit but found it worth mentioning. I had a code region already secretized and wrapped by `secret.generic`. I ran secretize + wrap-generic again and found that the resulting MLIR code does contain another `secret.generic` wrapping around the existing secret.generic code. Here's the MLIR code and the command I ran:
```llvm
// input program
func.func @simple_if(%arg0: !secret.secret<i1>, %arg1: !secret.secret<i16>) -> !secret.secret<i16> {
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<i1>, !secret.secret<i16>) {
    ^bb0(%arg2: i1, %arg3: i16):
      %1 = scf.if %arg2 -> (i16) {
        %2 = arith.addi %arg3, %arg3 : i16
        scf.yield %2 : i16
      } else {
        %2 = arith.muli %arg3, %arg3 : i16
        scf.yield %2 : i16
      }
      secret.yield %1 : i16
    } -> !secret.secret<i16>
    return %0 : !secret.secret<i16>
  }

// output I got
func.func @simple_if(%arg0: !secret.secret<!secret.secret<i1>>, %arg1: !secret.secret<!secret.secret<i16>>) -> !secret.secret<!secret.secret<i16>> {
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<!secret.secret<i1>>, !secret.secret<!secret.secret<i16>>) {
    ^bb0(%arg2: !secret.secret<i1>, %arg3: !secret.secret<i16>):
      %1 = secret.generic ins(%arg2, %arg3 : !secret.secret<i1>, !secret.secret<i16>) {
      ^bb0(%arg4: i1, %arg5: i16):
        %2 = scf.if %arg4 -> (i16) {
          %3 = arith.addi %arg5, %arg5 : i16
          scf.yield %3 : i16
        } else {
          %3 = arith.muli %arg5, %arg5 : i16
          scf.yield %3 : i16
        }
        secret.yield %2 : i16
      } -> !secret.secret<i16>
      secret.yield %1 : !secret.secret<i16>
    } -> !secret.secret<!secret.secret<i16>>
    return %0 : !secret.secret<!secret.secret<i16>>
  }
```

Command I ran:
```bash
[...]

PR #816, opened 2024-07-23
secretize: don't secretize already secret func args

asraa:
Fixes https://github.com/google/heir/issues/809

### DIFF
--- a/tests/secretize/idempotent.mlir
+++ b/tests/secretize/idempotent.mlir
@@ -1,0 +1,16 @@
+// RUN: heir-opt -secretize -wrap-generic -secretize -wrap-generic %s | FileCheck %s
+
+// This is a regression test that ensures that this sequence of patterns doesn't
+// double-wrap types a secret.generic block. One fix was to ensure that
+// secretize does not double wrap secret types.
+
+module {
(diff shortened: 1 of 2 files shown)


tests/secretize/idempotent.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: heir-opt -secretize -wrap-generic -secretize %s | FileCheck %s
```
AlexanderViand-Intel, line 1: Maybe make the test a bit more explicit, as a change/bug in wrap-generic could cause it to regress to double-wrapping even for non-doubled-up `secret<secret<types>>`
```suggestion
// RUN: heir-opt -secretize -wrap-generic -secretize -wrap-generic %s | FileCheck %s
```

AlexanderViand-Intel (approved):
Thanks for fixing this so quickly! Just a tiny (optional) suggestion:

tests/secretize/idempotent.mlir
asraa: nice! done. also added a check to ensure that there's only one secret.generic in that case

MeronZerihun (commented):
LGTM :)

Issue #785, opened 2024-07-10
Encoding "Taxonomy"


AlexanderViand-Intel:
In order to avoid cluttering up the discussion of #762 (Enc/Dec and Randomness handling), I figured I'd create a new issue to (a) help myself get my thoughts straight on the various mappings/encodings and (b) get some feedback, especially with regards to the TFHE perspective on RLWE.

## 1. Application Data (e.g. `i32`) vs. "Cleartext" (e.g. $\mathbb{F}_p$) Semantics
Not sure if "Cleartext" is the right name here, but we usually have a disconnect between the semantics of the actual data type in the application (e.g., `i32` or `f32`) and what we actually end up achieving under HE (usually finite field arithmetic with prime modulus). I know this isn't the case when we do bit-wise encryption (be it in TFHE or in BFV/BGV) but I'm not sure how 4-bit TFHE works here. Clearly, it's something to consider for "word-wise" FHE like BGV/BGV and especially CKKS, with its approximate nature.

Of course, for many applications, the difference between computation mod $2^{32}$ and a 32-bit prime is not important, but a compiler translation from `i32` to `mod p` isn't technically "correct" and we should probably think about how to expose this to the developer/frontend. I guess this is conceptually similar to truncation and/or deciding approximation precision (e.g. LUT size) which are also relevant to TFHE, but in the mod p/polynomial approximation case we get with BFV/BGV and CKKS, "bits of precision" might not be sufficient as a metric/language to communicate the behavior.

## 2. Cleartext (e.g. $\mathbb{F}_p$) -> Plaintext Encoding (e.g., $R_p = \mathbb{Z}_p[X] / X^N + 1$)
For RLWE, the plaintext space is a polynomial ring, and we're generally much more interested in encoding _integers_ (at this point, usually implicitly already considered mod p?). 
There are lots of ways to do this, starting with the trivial solution of just setting the constant coefficient to your integer and all other coefficients to zero, to decomposing your integer and putting a digit into each coefficient, to various forms of CRT-based "packing" or "batching" that allows one to achieve SIMD-style semantics. By far the most common is "full" packing, but you can also trade off a lower number of "slots" for larger (i.e., $\mathbb{F}_{p^d}$ for $d > 1$) slots.
CKKS has some extra complexity here as the message space is technically a vector of complex numbers but of course it's approximate and in fixed-(ish)-point representation, so basically just more SIMD integer vector stuff. 

[...]

PR #825, opened 2024-07-25
Add `arith.select` Pattern to `--secret-to-bgv`

AlexanderViand-Intel:
This adds the following rewrite pattern:
```llvm
%res = arith.select %c, %t, %f
// turns into
%ct = bgv.mul %c, %t
%nc = bgv.negate %c  // EDIT: actually a bug, this should be logical negation, i.e., bgv.sub %one, %c 
%ncf = bgv.mul %nc, %f
%sum = bgv.add %nc, %ncf
%res = bgv.relinearize %sum 
```

However, as-is, this pattern actually produces invalid IR: `%c` has (element/self) type `i1` while `%t` and `%f` can have any IntegerLike type, e.g., `i32`. In the current TypeConverter, this means that the condition is mapped to a ciphertext with encoding `#lwe.polynomial_evaluation_encoding<cleartext_start = 1, cleartext_bitwidth = 1>` while the other values are mapped to ciphertexts with encoding `#lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>`.  Since `bgv.mul` requires the types of its operands to match, this is invalid IR.

In the current draft, I've simply hardcoded the typeconverter to map everything to 32 bits (that's why this is a Draft PR only for now), but we'll need to think about how to handle such cases. I think this is a special case of "what is the plaintext-FHE semantics mapping" (Step 1 in #785).

### DIFF
--- a/lib/Conversion/SecretToBGV/SecretToBGV.cpp
+++ b/lib/Conversion/SecretToBGV/SecretToBGV.cpp
@@ -156,6 +156,25 @@ class SecretGenericOpMulConversion
   }
 };
 
+class SecretGenericOpSelectConversion
+    : public SecretGenericOpConversion<arith::SelectOp, bgv::MulOp> {
+ public:
+  using SecretGenericOpConversion<arith::SelectOp,
+                                  bgv::MulOp>::SecretGenericOpConversion;
+
+  void replaceOp(secret::GenericOp op, TypeRange outputTypes, ValueRange inputs,
+                 ConversionPatternRewriter &rewriter) const override {
+    // inputs = [condition, true_value, false_value]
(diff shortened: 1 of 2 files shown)


AlexanderViand-Intel:
@asraa In today's meeting, I said this only needed a small bugfix, because I only remembered that I still need to swap `bgv.negate %c` to `bgv.sub_plain %one, %c`, but I forgot about the *real* blocker here, which is the typing issue, so this does indeed hang on the attribute/type rework (though probably not even the LWE one, but a future one for cleartext<>plaintext semantics cleanup)

j2kun:
In the interest of pruning the PR page, I'm closing old PRs. Feel free to create a new PR if/when the code has been rebased and updated.

PR #838, opened 2024-07-30
Working BooleanLineVectorizer

WoutLegiest:
- Implementation of the BoolLineVectorizer, a pass in the CGGI dialect
- Adding a test for the pass above
- Expansion of the CGGI Dialect to support the representation of gates through attributes
- Small refactoring of the TfheRustBool Dialect

### DIFF
--- a/lib/Dialect/CGGI/IR/CGGIAttributes.td
+++ b/lib/Dialect/CGGI/IR/CGGIAttributes.td
@@ -3,6 +3,14 @@
 
 include "lib/Dialect/CGGI/IR/CGGIDialect.td"
 include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/BuiltinDialect.td"
+include "mlir/IR/CommonAttrConstraints.td"
+
+class CGGI_Attr<string name, string attrMnemonic, list<Trait> traits = []>
+    : AttrDef<CGGI_Dialect, name, traits> {
+  let mnemonic = attrMnemonic;
+  let assemblyFormat = "`<` struct(params) `>`";
+}
 
 def CGGI_CGGIParams : AttrDef<CGGI_Dialect, "CGGIParams"> {
   let mnemonic = "cggi_params";
@@ -21,5 +29,27 @@ def CGGI_CGGIParams : AttrDef<CGGI_Dialect, "CGGIParams"> {
 
   let assemblyFormat = "`<` struct(params) `>`";
 }
+
+def CGGIGate_Attr : CGGI_Attr<"CGGIGate", "cggi_gate"> {
+  let summary = "An Attribute containing an array of strings to store bool gates";
+
+  let description = [{
+    This attributes stores a list of string identifiers for Boolean gates.
+
+     This used in the `cggi.packed` operation to indicate the boolean gate that applies pairwise between elements of two ciphertext arrays. For example,
+
+     %0 = cggi.packed %a, %b {gates = #cggi.cggi_gate<"and", "xor">} : tensor<2x!lwe.lwe_ciphertext>
+
+     applies an "and" gate between the first elements of %a and %b and an xor gate between the second elements.
+  }];
+
+  let parameters = (ins
+      ArrayRefParameter<"mlir::StringAttr">: $gate);
+
+  let assemblyFormat = "`<` $gate `>`";
+}
+
+
+
 
 #endif  // LIB_DIALECT_CGGI_IR_CGGIATTRIBUTES_TD_
--- a/lib/Dialect/CGGI/IR/CGGIOps.td
+++ b/lib/Dialect/CGGI/IR/CGGIOps.td
@@ -3,12 +3,14 @@
 
 include "lib/Dialect/HEIRInterfaces.td"
 include "lib/Dialect/CGGI/IR/CGGIDialect.td"
+include "lib/Dialect/CGGI/IR/CGGIAttributes.td"
 
 include "mlir/Dialect/Polynomial/IR/PolynomialAttributes.td"
 include "lib/Dialect/LWE/IR/LWETypes.td"
 
 include "mlir/IR/OpBase.td"
 include "mlir/IR/BuiltinAttributes.td"
+include "mlir/IR/CommonAttrConstraints.td"
 include "mlir/Interfaces/InferTypeOpInterface.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
 
@@ -55,6 +57,19 @@ def CGGI_NotOp : CGGI_Op<"not", [
   let results = (outs LWECiphertextLike:$output);
   let assemblyFormat = "operands attr-dict `:` qualified(type($output))";
   let summary = "Logical NOT of two ciphertexts";
+}
+
+def CGGI_PackedOp : CGGI_Op<"packed_gates", [
+    Pure,
+    SameOperandsAndResultType
+]> {
+  let arguments = (ins
+    CGGIGate_Attr:$gates,
+    LWECiphertextLike:$lhs,
+    LWECiphertextLike:$rhs
+  );
+
+  let results = (outs LWECiphertextLike:$output);
 }
 
 class CGGI_LutOp<string mnemonic, list<Trait> traits = []>
--- a/lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
+++ b/lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
@@ -1,5 +1,10 @@
-#include "lib/Transforms/StraightLineVectorizer/StraightLineVectorizer.h"
+#include "lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.h"
 
+#include <string>
+
+#include "lib/Dialect/CGGI/IR/CGGIAttributes.h"
+#include "lib/Dialect/CGGI/IR/CGGIOps.h"
+#include "lib/Dialect/LWE/IR/LWEAttributes.h"
 #include "lib/Graph/Graph.h"
 #include "llvm/include/llvm/ADT/TypeSwitch.h"          // from @llvm-project
 #include "llvm/include/llvm/Support/Debug.h"           // from @llvm-project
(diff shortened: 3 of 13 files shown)


lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -57,6 +59,17 @@ def CGGI_NotOp : CGGI_Op<"not", [
   let summary = "Logical NOT of two ciphertexts";
 }
 
+def CGGI_PackedOp : CGGI_Op<"packed", [
+    Pure]> {
```
asraa, line 63: This is also a Commutative operation, right? I also think you could use `SameOperandsAndResultType` - even if you have an attribute, i think that the type matching only occurs on values (I found an example, at least: https://github.com/google/jax/blob/b1066ee41373991bb66c5a56b2874dc5455b7f70/jaxlib/mosaic/dialect/tpu/tpu.td#L237)

lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -57,6 +59,17 @@ def CGGI_NotOp : CGGI_Op<"not", [
   let summary = "Logical NOT of two ciphertexts";
 }
 
+def CGGI_PackedOp : CGGI_Op<"packed", [
```
asraa, line 62: nit: maybe `packed_gates`?

lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -57,6 +59,17 @@ def CGGI_NotOp : CGGI_Op<"not", [
...
+    Pure]> {
+  let arguments = (ins
+    CGGIGate_Attr:$gates,
```
asraa, line 65: could this op use a verifier to validate that the gate attr is the same length as the lhs/rhs/output? (I think lhs/rhs/output sizes will be verified if you add SameOperandsAndResultType)

lib/Dialect/CGGI/IR/CGGIAttributes.td
```diff
@@ -22,4 +30,20 @@ def CGGI_CGGIParams : AttrDef<CGGI_Dialect, "CGGIParams"> {
   let assemblyFormat = "`<` struct(params) `>`";
 }
 
+def CGGIGate_Attr : CGGI_Attr<"CGGIGate", "cggi_gate"> {
+  let summary = "An Attribute containing an array of strings to store bool gates";
+
+  let description = [{
+    Attribute to store what kind of Boolean gate we are working on
```
asraa, line 37:
```suggestion
    This attributes stores a list of string identifiers for Boolean gates.
    
     This used in the `cggi.packed` operation to indicate the boolean gate that applies pairwise between elements of two ciphertext arrays. For example,
     
     %0 = cggi.packed %a, %b {gates = #cggi.cggi_gate<"and", "xor">} : tensor<2x!lwe.lwe_ciphertext>

     applies an "and" gate between the first elements of %a and %b and an xor gate between the second elements.
```

lib/Dialect/CGGI/Transforms/Passes.td
```diff
@@ -19,4 +19,20 @@ def SetDefaultParameters : Pass<"cggi-set-default-parameters"> {
   let dependentDialects = ["mlir::heir::cggi::CGGIDialect"];
 }
 
+def BooleanLineVectorizer : Pass<"cggi-boolean-line-vectorize"> {
+  let summary = "Group different logic gates with the packed API";
+  let description = [{
+    This pass groups indepentend logic gates into a single call of the packed
+    operations.
```
asraa, line 26:
```suggestion
    This pass groups independent logic gates into a single call of the packed
    operations.
```

lib/Dialect/CGGI/Transforms/Passes.td
```diff
@@ -19,4 +19,20 @@ def SetDefaultParameters : Pass<"cggi-set-default-parameters"> {
...
+
+    Pass is intended for the `FPT` tfhe-rs API.
```
asraa, line 28: nit: might benefit from a short example, or some description that this is an improvement over straight line vectorizer.

lib/Dialect/CGGI/Transforms/Passes.td
```diff
@@ -19,4 +19,20 @@ def SetDefaultParameters : Pass<"cggi-set-default-parameters"> {
...
+  }];
+
+  let dependentDialects = [
+    "tensor::TensorDialect",
+  ];
+
+  let dependentDialects = ["mlir::heir::cggi::CGGIDialect"];
```
asraa, line 35: I think you can get away without cggi as a dependent dialect since it must have already been loaded as a dialect to parse the input. (https://mlir.llvm.org/getting_started/Faq/#registered-loaded-dependent-whats-up-with-dialects-management)

lib/Dialect/TfheRustBool/IR/TfheRustBoolTypes.td
```diff
@@ -28,4 +28,27 @@ def TfheRustBool_ServerKey : TfheRustBool_Type<"ServerKey", "server_key", [PassB
   let summary = "The boolean server key required to perform homomorphic operations.";
 }
 
+
+class TfheRustBool_BooleanGate<string gate>
+  : TfheRustBool_Type<"BooleanGate" # gate, "bin_gate" # gate> {
+  let summary = "Enum placeholder to define the binary gate" # gate # " to perform as homomoprhic operation.";
+}
+
+foreach i = ["And", "Or", "Xor", "Nand", "Nor", "Xnor"] in {
```
asraa, line 37: this is awesome! (any reason not do something similar for the CGGI attributes? is it more difficult to do with a string attr enum?)

lib/Dialect/TfheRustBool/IR/TfheRustBoolOps.td
```diff
@@ -71,4 +73,19 @@ def MuxOp : TfheRustBool_Op<"mux", [
   let results = (outs TfheRustBool_Encrypted:$output);
 }
 
+
+// ToDo: Change to an only tensor type?
```
asraa, line 77: just to make sure, did you want to merge the TfheRustBool files with this PR or the follow-up #842?

tests/cggi/boolean_line_vectorizer.mlir
```diff
@@ -0,0 +1,75 @@
+// RUN: heir-opt --cggi-boolean-line-vectorize %s
+
+#encoding = #lwe.unspecified_bit_field_encoding<cleartext_bitwidth = 1>
+!ct_ty = !lwe.lwe_ciphertext<encoding = #encoding>
+!pt_ty = !lwe.lwe_plaintext<encoding = #encoding>
+
+// CHECK-LABEL: add_one
+// CHECK-COUNT-1: cggi.packed %[[*]], %[[*]] {gates = #cggi.cggi_gate<"and", "xor">} : (tensor<2x!lwe.lwe_ciphertext
```
asraa, line 8: oh no! only one cggi.packed haha.

could you also add a syntax test in tests/cggi/ that tests the syntax of this operation so we have it as documentation? (if you add the verifier would also be good to test that error condition on the gate length)

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
```diff
@@ -0,0 +1,224 @@
...
+namespace mlir {
+namespace heir {
+namespace cggi {
+
+#define GEN_PASS_DEF_BOOLEANLINEVECTORIZER
+#include "lib/Dialect/CGGI/Transforms/Passes.h.inc"
+
+bool areCompatibleBool(Operation *lhs, Operation *rhs) {
+  if (
+      // lhs->getName() != rhs->getName() ||
```
asraa, line 32: nit: remove this! (yay!)

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
```diff
@@ -0,0 +1,224 @@
...
+      lhs->getDialect() != rhs->getDialect() ||
+      lhs->getResultTypes() != rhs->getResultTypes() ||
+      lhs->getAttrs() != rhs->getAttrs()) {
+    return false;
+  }
+  return OpTrait::hasElementwiseMappableTraits(lhs);
```
asraa, line 38: this might be somewhat fragile right? since there might be other elementwise operations that aren't boolean gates..? maybe add a comment or a TODO if this can be made better with a BooleanPackableGate trait on the CGGI_BinaryGateOp's?

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
```diff
@@ -0,0 +1,224 @@
...
+          str = "nand";
+        } else if (isa<cggi::XorOp>(op)) {
+          str = "xor";
+        } else if (isa<cggi::XNorOp>(op)) {
+          str = "xnor";
+        } else if (isa<cggi::OrOp>(op)) {
+          str = "or";
+        } else if (isa<cggi::NorOp>(op)) {
+          str = "nor";
+        }
```
asraa, line 138: might be worth a final if statement that fails / prints LLVM_DEBUG statement in case there is an incompatible operation that we haven't added yet?

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
```diff
@@ -0,0 +1,224 @@
...
+        llvm::dbgs() << "\n";
+      });
+
+      auto oplist = CGGIGateAttr::get(&context, vectorizedGateOperands);
+
+      auto vectorizedOp = builder.create<cggi::PackedOp>(
+          key->getLoc(), tensorType, oplist, vectorizedOperands[0],
+          vectorizedOperands[1]);
+
+      vectorizedOp->getResult(0).setType(tensorType);
```
asraa, line 183: interesting! just curious, why was this needed? I would assume the first operand in the builder above was for the result type.

asraa (commented):
Nice job!!!

lib/Dialect/TfheRustBool/IR/TfheRustBoolOps.td
WoutLegiest: All of the TfheRustBool files are done in 842

Issue #813, opened 2024-07-23
Fixing the nightly build


WoutLegiest:
https://github.com/google/heir/actions/runs/10050313017/job/27777982734

AlexanderViand-Intel:
Unfortunately, it seems like even with the fix from @asraa, something is still going wrong, as the current "nightly" release says it's from 3 weeks ago?

asraa:
weird! packages write was supposed to fix the perm error. Let me actually debug this (and probably add a step to post an issue automatically if there's a failure so we can tell when it's broken)

j2kun:
https://github.com/google/heir/releases/tag/nightly hooray!

PR #840, opened 2024-07-30
ci: fix typo in gh token env var

asraa:
I found a typo! Hopefully this fixes (maybe) https://github.com/google/heir/issues/813, i'll manually run and close the issue manually if it works. Follow-up to add an auto issue filing on failure

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Delete previous existing nightly tag and release
         env:
-          GH_TOKEN: $${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


j2kun (approved):

AlexanderViand-Intel (approved):
good catch! Even with the GH highlighting, I had to look twice to see the typo! 

>  close the issue manually if it works.

Your current PR description includes the GH magic keyword "fixes #issue" so it'll auto-close when this gets merged.

asraa:
> GH magic keyword "fixes #issue" so it'll auto-close when this gets merged.

actually didn't know if that would work without being on a separate new line! reading the docs i think it would :) i'll update to break the pattern

PR #842, opened 2024-07-30
Adapting the TfheRustBool dialect to pass the packed api + update cggi to tfhe rust bool conversion pass

WoutLegiest:
Follow up work of https://github.com/google/heir/pull/838

### DIFF
--- a/lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
+++ b/lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
@@ -1,8 +1,9 @@
-#include "lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.h"
+#include "lib/Dialect/CGGI/Transforms/BooleanVectorizer.h"
 
-#include <string>
+#include <cstdint>
 
 #include "lib/Dialect/CGGI/IR/CGGIAttributes.h"
+#include "lib/Dialect/CGGI/IR/CGGIEnums.h"
 #include "lib/Dialect/CGGI/IR/CGGIOps.h"
 #include "lib/Dialect/LWE/IR/LWEAttributes.h"
 #include "lib/Utils/Graph/Graph.h"
@@ -18,13 +19,13 @@
 #include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
 #include "mlir/include/mlir/Transforms/Passes.h"  // from @llvm-project
 
-#define DEBUG_TYPE "bool-line-vectorizer"
+#define DEBUG_TYPE "bool-vectorizer"
 
 namespace mlir {
 namespace heir {
 namespace cggi {
 
-#define GEN_PASS_DEF_BOOLEANLINEVECTORIZER
+#define GEN_PASS_DEF_BOOLEANVECTORIZER
 #include "lib/Dialect/CGGI/Transforms/Passes.h.inc"
 
 bool areCompatibleBool(Operation *lhs, Operation *rhs) {
@@ -119,27 +120,32 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
           {static_cast<int64_t>(bucket.size())}, elementType);
 
       SmallVector<Value, 4> vectorizedOperands;
-      SmallVector<StringAttr, 4> vectorizedGateOperands;
+      SmallVector<CGGIBoolGateEnumAttr, 4> vectorizedGateOperands;
 
       for (auto *op : bucket) {
-        std::string str;
-        if (isa<cggi::AndOp>(op)) {
-          str = "and";
-        } else if (isa<cggi::NandOp>(op)) {
-          str = "nand";
-        } else if (isa<cggi::XorOp>(op)) {
-          str = "xor";
-        } else if (isa<cggi::XNorOp>(op)) {
-          str = "xnor";
-        } else if (isa<cggi::OrOp>(op)) {
-          str = "or";
-        } else if (isa<cggi::NorOp>(op)) {
-          str = "nor";
-        } else {
-          LLVM_DEBUG(llvm::dbgs()
-                     << "Try to parse boolean operation that does not exist.");
-        }
-        vectorizedGateOperands.push_back(StringAttr::get(&context, str));
+        auto gateStr =
+            llvm::TypeSwitch<Operation *, FailureOr<CGGIBoolGateEnum>>(op)
+                .Case<cggi::AndOp>(
+                    [](AndOp op) { return CGGIBoolGateEnum::bg_and; })
+                .Case<cggi::NandOp>(
+                    [](NandOp op) { return CGGIBoolGateEnum::bg_nand; })
+                .Case<cggi::XorOp>(
+                    [](XorOp op) { return CGGIBoolGateEnum::bg_xor; })
+                .Case<cggi::XNorOp>(
+                    [](XNorOp op) { return CGGIBoolGateEnum::bg_xnor; })
+                .Case<cggi::OrOp>(
+                    [](OrOp op) { return CGGIBoolGateEnum::bg_or; })
+                .Case<cggi::NorOp>(
+                    [](NorOp op) { return CGGIBoolGateEnum::bg_nor; })
+                .Default([op](Operation *) {
+                  LLVM_DEBUG(llvm::dbgs()
+                             << "parsing unsupported boolean operation" << op);
+                  return FailureOr<CGGIBoolGateEnum>();
+                });
+        assert(succeeded(gateStr));
+
+        vectorizedGateOperands.push_back(
+            CGGIBoolGateEnumAttr::get(&context, gateStr.value()));
       }
 
       // Group the independent operands over the operations
@@ -177,7 +183,7 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
         llvm::dbgs() << "\n";
       });
 
-      auto oplist = CGGIGateAttr::get(&context, vectorizedGateOperands);
+      auto oplist = CGGIBoolGatesAttr::get(&context, vectorizedGateOperands);
 
       auto vectorizedOp = builder.create<cggi::PackedOp>(
           key->getLoc(), tensorType, oplist, vectorizedOperands[0],
@@ -205,9 +211,8 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
   return madeReplacement;
 }
 
-struct BooleanLineVectorizer
-    : impl::BooleanLineVectorizerBase<BooleanLineVectorizer> {
-  using BooleanLineVectorizerBase::BooleanLineVectorizerBase;
+struct BooleanVectorizer : impl::BooleanVectorizerBase<BooleanVectorizer> {
+  using BooleanVectorizerBase::BooleanVectorizerBase;
 
   void runOnOperation() override {
     MLIRContext &context = getContext();
--- a/lib/Dialect/CGGI/Transforms/Passes.td
+++ b/lib/Dialect/CGGI/Transforms/Passes.td
@@ -19,11 +19,12 @@ def SetDefaultParameters : Pass<"cggi-set-default-parameters"> {
   let dependentDialects = ["mlir::heir::cggi::CGGIDialect"];
 }
 
-def BooleanLineVectorizer : Pass<"cggi-boolean-line-vectorize"> {
+def BooleanVectorizer : Pass<"cggi-boolean-vectorize"> {
   let summary = "Group different logic gates with the packed API";
   let description = [{
     This pass groups independent logic gates into a single call of the packed
-    operations.
+    operations. Pass is based on the straight-line-vectorizer, but is fundamentally different.
+    This pass combines any type of boolean gates and is not restricted to combining the same type of gate operand.
 
     Pass is intended for the `FPT` tfhe-rs API, where `packed_gates` function get a
     the boolean gates are passed as a string vector and a left and right vector of ciphertexts.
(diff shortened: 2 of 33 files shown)


ai-mannamalai:
@WoutLegiest - can you consider adding the CMake files for the new dialect and the transform pass ? It should be straight forward

lib/Dialect/CGGI/Transforms/Passes.td
```diff
@@ -19,4 +19,20 @@ def SetDefaultParameters : Pass<"cggi-set-default-parameters"> {
   let dependentDialects = ["mlir::heir::cggi::CGGIDialect"];
 }
 
+def BooleanLineVectorizer : Pass<"cggi-boolean-line-vectorize"> {
```
j2kun, line 22: nit: instead of "line vectorizer" it should be "straight line vectorizer". You could also just call it `BooleanVectorizer`

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
```diff
@@ -0,0 +1,224 @@
...
+namespace mlir {
+namespace heir {
+namespace cggi {
+
+#define GEN_PASS_DEF_BOOLEANLINEVECTORIZER
+#include "lib/Dialect/CGGI/Transforms/Passes.h.inc"
+
+bool areCompatibleBool(Operation *lhs, Operation *rhs) {
+  if (
+      // lhs->getName() != rhs->getName() ||
```
j2kun, line 32: nit: remove comment if not needed.

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
```diff
@@ -0,0 +1,224 @@
...
+      Type elementType = key->getResultTypes()[0];
+      RankedTensorType tensorType = RankedTensorType::get(
+          {static_cast<int64_t>(bucket.size())}, elementType);
+
+      SmallVector<Value, 4> vectorizedOperands;
+      SmallVector<StringAttr, 4> vectorizedGateOperands;
+
+      for (auto op : bucket) {
+        std::string str;
+        if (isa<cggi::AndOp>(op)) {
```
j2kun, line 126: nit: better to use `TypeSwitch`

```
  std::string str = llvm::TypeSwitch<Operation &, std::string>(op)
      .Case<cggi::AndOp>([](auto op) { return std::string("and"); })
      .Case<cggi::NandOp>([](auto op) { return std::string("nand"); })
      ...
      .Default([&](Operation &) { return "unknown"; });
```

Alternatively, could probably do string munging on the operation name.

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
j2kun, line 1: This looks like largely a repeat of the straight line vectorizer. May want to add a comment somewhere explaining what makes the two passes incompatible to be shared code.

tests/cggi/boolean_line_vectorizer.mlir
```diff
@@ -0,0 +1,75 @@
+// RUN: heir-opt --cggi-boolean-line-vectorize %s
+
+#encoding = #lwe.unspecified_bit_field_encoding<cleartext_bitwidth = 1>
+!ct_ty = !lwe.lwe_ciphertext<encoding = #encoding>
+!pt_ty = !lwe.lwe_plaintext<encoding = #encoding>
+
+// CHECK-LABEL: add_one
+// CHECK-COUNT-1: cggi.packed %[[*]], %[[*]] {gates = #cggi.cggi_gate<"and", "xor">} : (tensor<2x!lwe.lwe_ciphertext
```
j2kun, line 8: I feel it would be worth adding a few more assertions for the rest of the program.

j2kun (changes requested):

lib/Dialect/CGGI/Transforms/BooleanLineVectorizer.cpp
asraa: Hmm I wonder if the `tryBoolVectorizeBlock` can just take a function callback to the compatible op logic..

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
```diff
@@ -122,24 +122,19 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
...
+            llvm::TypeSwitch<Operation *, std::string>(op)
+                .Case<cggi::AndOp>([](AndOp op) { return std::string("and"); })
+                .Case<cggi::NandOp>(
+                    [](NandOp op) { return std::string("nand"); })
+                .Case<cggi::XorOp>([](XorOp op) { return std::string("xor"); })
+                .Case<cggi::XNorOp>(
+                    [](XNorOp op) { return std::string("xnor"); })
+                .Case<cggi::OrOp>([](OrOp op) { return std::string("or"); })
+                .Case<cggi::NorOp>([](NorOp op) { return std::string("nor"); })
+                .Default([](Operation *) { return std::string("nor"); });
```
asraa, line 135: I'm assuming that TypeSwitch required you to return a string here so for the default you returned some string - you could also return `FailureOr<std::string>` from the typeswitch and then return a `failure()` in the default case. The LLVM_DEBUG/ `llvm_unreachable` could be returned if the gate_str was a failure.

lib/Dialect/TfheRustBool/IR/TfheRustBoolAttributes.td
```diff
@@ -0,0 +1,31 @@
...
+    : AttrDef<TfheRustBool_Dialect, name, traits> {
+  let mnemonic = attrMnemonic;
+  let assemblyFormat = "`<` struct(params) `>`";
+}
+
+def TfheRustBoolGate_Attr : TfheRustBool_Attr<"TfheRustBoolGate", "tfhe_rust_bool_gate"> {
+  let summary = "An Attribute containing an array of strings to store bool gates";
+
+  let description = [{
+    Attribute to store what kind of Boolean gate we are working on
```
asraa, line 19: Might want to add a similar doc string to the gate attr from CGGI that has an example

lib/Dialect/TfheRustBool/IR/TfheRustBoolAttributes.td
```diff
@@ -0,0 +1,31 @@
...
+  }];
+
+  let parameters = (ins
+      ArrayRefParameter<"mlir::StringAttr">: $gate);
```
asraa, line 23: Just realized you could add a Verification method to check whether the string is one of the ones you expect 
https://mlir.llvm.org/docs/DefiningDialects/AttributesAndTypes/#verification

If you share the same code with CGGI (actually, can this gate attribute just be a common attribute between the two?), then the Default typeswitched case above would actually be an unreachable since verification would catch the bad gate string

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
```diff
@@ -205,9 +200,8 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
   return madeReplacement;
 }
 
-struct BooleanLineVectorizer
-    : impl::BooleanLineVectorizerBase<BooleanLineVectorizer> {
-  using BooleanLineVectorizerBase::BooleanLineVectorizerBase;
+struct BooleanVectorizer : impl::BooleanVectorizerBase<BooleanVectorizer> {
```
asraa, line 203: like Jeremy's comment (I think on the other PR?) you might want to comment why the vectorizer code isn't shared with the straight-line-vectorizer (e.g. it has a different compatible op, or that you could improve it beyond the straight-line-vectorizer)

asraa (commented):
Thank you!

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
WoutLegiest: Thanks, was looking for something like this!

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
WoutLegiest: Wrote a small explanation in the description of the pass

lib/Dialect/TfheRustBool/IR/TfheRustBoolAttributes.td
j2kun: I'm not sure what Asra means by sharing code here (CGGI does not have a packed gate op?), but for verification, you could consider changing this to an ArrayRefParameter of an enum type. You can see how this is done here: https://github.com/google/heir/blob/114e8c3d564273d60e36d398f7a795eca60d4bcd/lib/Dialect/Random/IR/RandomEnums.td#L11-L16

The tabelgen definition linked above creates a vanilla C++ enum class (not actually an attribute), and you can just `ArrayRefParameter<"mlir::heir::foo::YourEnumName">` to put it on an attribute or op.

asraa: Oh, I meant sharing the verifier for an array of gates with the CGGI gate attribute.
https://github.com/google/heir/blob/114e8c3d564273d60e36d398f7a795eca60d4bcd/lib/Dialect/CGGI/IR/CGGIAttributes.td#L33

(As in, this gate attribute should just be used in both the cggi packed op and the tfhe one. Here's the CGGI packed op: https://github.com/google/heir/blob/114e8c3d564273d60e36d398f7a795eca60d4bcd/lib/Dialect/CGGI/IR/CGGIOps.td#L62

j2kun: I stand corrected! I feel having an enum (which does not need to be attached to any dialect) would avoid the need for a verifier entirely. Then both ops can use ArrayRefParameter<enum> and we can cut down on tablegen code.

lib/Dialect/TfheRustBool/IR/TfheRustBoolOps.td
```diff
@@ -49,6 +51,18 @@ def TfheRustBool_NorOp  : TfheRustBool_BinaryGateOp<"nor">  { let summary = "Log
 def TfheRustBool_XorOp : TfheRustBool_BinaryGateOp<"xor"> { let summary = "Logical XOR of two TFHE-rs Bool ciphertexts."; }
 def TfheRustBool_XnorOp : TfheRustBool_BinaryGateOp<"xnor"> { let summary = "Logical XNOR of two TFHE-rs Bool ciphertexts."; }
 
+def TfheRustBool_PackedOp : TfheRustBool_Op<"gates_packed", [
+    Pure]> {
+  let arguments = (ins
+    TfheRustBool_ServerKey:$serverKey,
+    TfheRustBoolGate_Attr:$gates,
```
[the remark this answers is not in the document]
WoutLegiest: Tried to implement it that way and got an error: 
`lib/Dialect/TfheRustBool/IR/TfheRustBoolOps.td:54:5: error: unexpected def type; only defs deriving from TypeConstraint or Attr or Property are allowed
def TfheRustBool_PackedOp : TfheRustBool_Op<"gates_packed", [
    ^`

ai-mannamalai:
adding a PR https://github.com/WoutLegiest/heir/pull/1 into your branch to enable CMake build in order

lib/Dialect/TfheRustBool/IR/TfheRustBoolOps.td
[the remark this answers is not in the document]
WoutLegiest: I changed the attributes to use an Enum definition. Those enums still needed to be wrapped in an Attribute before it can be used as an input to the Operations

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
```diff
@@ -119,27 +120,28 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
...
+                    [](XorOp op) { return CGGIBoolGateEnum::bg_xor; })
+                .Case<cggi::XNorOp>(
+                    [](XNorOp op) { return CGGIBoolGateEnum::bg_xnor; })
+                .Case<cggi::OrOp>(
+                    [](OrOp op) { return CGGIBoolGateEnum::bg_or; })
+                .Case<cggi::NorOp>(
+                    [](NorOp op) { return CGGIBoolGateEnum::bg_nor; })
+                .Default(
+                    [](Operation *) { return FailureOr<CGGIBoolGateEnum>(); });
+
```
asraa, line 142:
```suggestion
assert(succeeded(op))
```

or return false if failed because the pass would fail otherwise

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
```diff
@@ -119,27 +120,28 @@ bool tryBoolVectorizeBlock(Block *block, MLIRContext &context) {
...
+                .Case<cggi::XorOp>(
+                    [](XorOp op) { return CGGIBoolGateEnum::bg_xor; })
+                .Case<cggi::XNorOp>(
+                    [](XNorOp op) { return CGGIBoolGateEnum::bg_xnor; })
+                .Case<cggi::OrOp>(
+                    [](OrOp op) { return CGGIBoolGateEnum::bg_or; })
+                .Case<cggi::NorOp>(
+                    [](NorOp op) { return CGGIBoolGateEnum::bg_nor; })
+                .Default(
+                    [](Operation *) { return FailureOr<CGGIBoolGateEnum>(); });
```
asraa, line 141:
```suggestion
                    [](Operation *) { 
                      LLVM_DEBUG(llvm::dbgs()
                           << "parsing unsupported boolean operation" << op);
                      return FailureOr<CGGIBoolGateEnum>(); 
                    });
```

or op.emitError

asraa (commented):
Thank you! Just a couple minor nits now, but otherwise LGTM!

asraa (approved):
Thank you!

asraa:
Sorry about this! I think this just needed a rebase before merging

asraa:
Going to rebase this one myself!

asraa (approved):

Issue #843, opened 2024-07-30
[style] Consistent header guards HEIR_LIB_* vs LIB_*


asraa:
This is just a style / formatting issue. It seems like half our headers and TD files use `HEIR_LIB_*` and half use `LIB_*`. I suspect that we had two periods of externalizing code / copybara header guard transforms and also moving the include/ dir into the lib/ dir that probably caused some confusion.

j2kun:
Here's the macro I have in my neovim config, which creates/updates the include guard automatically. It does not include the `HEIR_` prefix.

https://github.com/j2kun/nvim-config/blob/b8ef6c4bdc7425fa2db537caa4202f094cf5d1aa/lua/mappings.lua#L61-L72

```lua
local function build_include_guard()
  -- project relative filepath
  local abs_path = vim.fn.expand("%")
  local rel_path = vim.fn.fnamemodify(abs_path, ":~:.")

  -- screaming case
  local upper = string.upper(rel_path)
  -- underscore separated
  local underscored = string.gsub(upper, "[./]", "_")
  -- trailing underscore
  return underscored .. "_"
end
```

asraa:
Looks like the tests/ folder is also still including THIRD_PARTY_HEIR_TESTS_ -
https://github.com/search?q=repo%3Agoogle%2Fheir%20THIRD_PARTY_HEIR_&type=code

PR #844, opened 2024-07-30
make include guards consistent

j2kun:
Fixes https://github.com/google/heir/issues/843


asraa (approved):
THANK YOU! 

You totally don't have to, but can you add a really simple regex for our pre commit search and replace? if you just specify a search it will fail. It could match "#ifndef HEIR_".

Just to make sure it doesn't creep in via copy-pasting from an old version / bad manual add.

j2kun:
Done

PR #850, opened 2024-07-31
Split `--bgv-to-poly` into `--bgv-to-lwe --lwe-to-polynomial`

### LINKED ISSUE #828: Split <RLWE_SCHEME>-to-polynomial into <RLWE_SCHEME>-to-lwe and lwe-to-polynomial
+1 we probably want to have this `--lwe-to-polynomial` be a common pass that is applied within BGV -> standard, CKKS -> standard, and even TFHE -> standard.

Adding!

Minimally for this PR we can move the type converter here into a separate file (LweToPolynomial/TypeConverter.cpp), have both passes use it for now, and then in a future PR split and refactor the `--bgv-to-polynomial` into `--bgv-to-lwe` so that it doesn't use the type converter anymore. But let's chat tomorrow if that's the direction we want to go

_Originally posted by @asraa in https://github.com/google/heir/pull/779#discussion_r1672269568_

AlexanderViand-Intel:
Resolves #828

This is mostly just a "mechanical" split, moving patterns from one pass to the other. There's probably lots of opportunities for improvement/cleanup in the future. For example, to keep the bgv->lwe mapping simple, I introduced `lwe.radd`, `lwe.rsub`, `lwe.rnegate` and `lwe.rmul`. 

The `r` prefix is there because there was already an `lwe.add` defined only over LWE ciphertexts, but we can probably drop that and just change the type constraints for these ops to be something along the lines of `Or<LWECiphertextLike,RLWECiphertxtLike>`


j2kun (approved):

AlexanderViand-Intel:
squashed and rebased

asraa (approved):

lawrencekhlim (approved):

PR #802, opened 2024-07-19
Loop Transformation: Secret-dependent scf.for to affine.for with constant bounds

MeronZerihun:
Fixes #801. Rewrites scf.for operations with secret loop bound(s) to use data independent affine.for operations with (static/constant) loop bound(s). Uses scf.if operation to check the current index with the secret bounds and conditionally set the for-loop's result. 

This is built as part of the data-oblivious transformation pipeline (#777) where an If-Transformation pass is then applied to get a data-oblivious program (assuming we don't have secret-dependent memory reads/writes).

PR #853, opened 2024-07-31
guard llvm::errs with LLVM_DEBUG

j2kun:
I saw these messages while running non-debug mode. Quick cleanup


AlexanderViand-Intel:
I think #802 actually removes those entirely, but it might be nicer to keep them around but guarded.

MeronZerihun:
> I think #802 actually removes those entirely, but it might be nicer to keep them around but guarded.

I messed up because I made those changes on `loop-dev` instead of `if-dev`. Please let me know if I need to migrate the changes.

Issue #866, opened 2024-08-03
Access Transformation: Add support for multi-dimensional tensors


MeronZerihun:
Our current implementation only supports secret-index reads/writes for single dimensional tensor. We need to add support for multi-dimensional tensor read/write operations at:
1) Secret indices where accesses at each dimension should be transformed, and
2) Mixed secret-and-public indices where only secret-index accesses at respective dimensions should be transformed.

Issue #777, opened 2024-07-09
Data-oblivious Programming and its Transformations


MeronZerihun:
# Data-oblivious Programming

A data-oblivious program is one that decouples data input from program execution. Such programs exhibit control-flow and memory access patterns that are independent of their input(s). This programming model is necessary for expressing FHE programs. There are 3 major transformations applied to convert a conventional program into a data-oblivious program:

### (1) If-Transformation

If-operations conditioned on inputs create data-dependent control-flow in programs. `scf.if` operations should atleast define a 'then' region (true path) and are always terminated `scf.yield` even when `scf.if` doesn't produce a result. To convert data-dependent `scf.if` operation to an equivalent set of data-oblivious operations in MLIR, we hoist all safely speculatable operations in the `scf.if` operation and convert the `scf.yield` operation to an `arith.select` operation. The following code snippet demonstrates an application of this transformation.

```llvm
// Before applying If-transformation
func.func @my_function(%input : i1) -> () {
  ...
  // Violation: %input is used as a condition causing a data-dependent branch
  %result =`%input -> (i16) {
        %a = arith.muli %b, %c : i16
        scf.yield %a : i16
      } else {
        scf.yield %b : i16
      }
  ...
}

// After applying If-transformation
func.func @my_function(%input : i16) -> (){
  ...
  %a = arith.muli %b, %c : i16
  %result = arith.select %input, %a, %b : i16
  ...
}
  ```

We have implemented a `ConvertIfToSelect` pass that transforms operations with secret-input conditions and with only Pure operations (i.e., operations that have no memory side effect and are speculatable) in their body. <strong>This transformation cannot be applied to operations when side effects are present in only one of the two regions. </strong>Although possible, we currently do not support transformations for operations where both regions have operations with matching side effects.

### (2) Loop-Transformation

Loop statements with input-dependent conditions (bounds) and number of iterations introduce data-dependent branches that violate data-obliviousness. We can convert such loops to become data-oblivious by replacing input-dependent conditionals (bounds) with static input-independent parameters (e.g. defining a constant upper bound), replacing early loop exits with update operations where the value returned from the loop is selectively updated using conditional predication. In MLIR, loops are expressed using either `affine.for`, `scf.for` or `scf.while` operations. 

[...]

PR #871, opened 2024-08-04
Access Transformation: Convert secret tensor.extract and tensor.insert to non-simd conditional alternative

MeronZerihun:
Converts `tensor.extract`/`tensor.insert` operations at secret index to non-simd conditional `tensor.extract`/`tensor.insert` operations respectively. The transformation is built as part of the data-oblivious transformation pipeline (https://github.com/google/heir/issues/777).

TODO(#866 ): Current transformation doesn't support multi-dimensional tensors


lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,146 @@
...
+  using OpRewritePattern<tensor::InsertOp>::OpRewritePattern;
+
+ public:
+  SecretInsertToStaticInsertConversion(DataFlowSolver *solver,
+                                       MLIRContext *context)
+      : OpRewritePattern(context), solver(solver) {}
+
+  LogicalResult matchAndRewrite(tensor::InsertOp insertOp,
+                                PatternRewriter &rewriter) const override {
+    // TODO(#866): Add support for writes on multi-dimensional tensors
```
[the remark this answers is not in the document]
AlexanderViand-Intel: (and to add a test)

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,150 @@
...
+          b.create<scf::YieldOp>(loc, forOp.getRegionIterArgs().front());
+        });
+
+    // Create YieldOp for affine.for
+    SmallVector<Value> results(ifOp->getOpResults());
+    builder.create<affine::AffineYieldOp>(results);
+
+    // Replace the old tensor.insert op with forOp's result
+    rewriter.replaceOp(extractOp, forOp);
+
```
AlexanderViand-Intel, line 112: Make sure to update the secretness of all new operations that you create.
For an example of how to do that, see https://github.com/google/heir/blob/f33691abaf3ad6dafb47a8f655ee630d8a45490a/lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp#L91-L96

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,150 @@
...
+          // Yield dummyValue
+          b.create<scf::YieldOp>(loc, forOp.getRegionIterArgs().front());
+        });
+
+    // Create YieldOp for affine.for
+    SmallVector<Value> results(ifOp->getOpResults());
+    builder.create<affine::AffineYieldOp>(results);
+
+    // Replace the old tensor.insert op with forOp's result
+    rewriter.replaceOp(extractOp, forOp);
```
AlexanderViand-Intel, line 111: reminder to also update secretness of things in this pass

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    auto *indexSecretnessLattice =
+        solver->lookupState<SecretnessLattice>(index);
+
+    // Implies that the tensor.extract op is newly created by this pass
+    if (!indexSecretnessLattice) {
+      extractOp->emitWarning()
+          << "Secretness for tensor.extract index has not been set";
+      return failure();
+    }
```
AlexanderViand-Intel, line 54: I think the handling here should be similar to the newly fixed ConvertIfToSelect, where a null value (which shouldn't happen with the new approach to updating secretness for all newly created values) or an "unknown" (aka unitialized) value still trigger the conversion, just to be safe.

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    auto *indexSecretnessLattice =
+        solver->lookupState<SecretnessLattice>(index);
+
+    // Implies that the tensor.insert op is newly created by this pass
+    if (!indexSecretnessLattice) {
+      insertOp->emitWarning()
+          << "Secretness for tensor.insert index has not been set";
+      return failure();
+    }
```
AlexanderViand-Intel, line 56: reminder to update this logic in this pass, too

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    bool isIndexSecret = indexSecretnessLattice->getValue().getSecretness();
```
AlexanderViand-Intel, line 58: I think this might still cause crashes, because you're checking only for the lattice element being null (so the `->` is safe) but not for unitialzied, so the `.getSecretness` might crash with an assert

AlexanderViand-Intel (changes requested):
I think the actual pass/transformations look great! All my comments are about tangential stuff like handling secretness (driven by my recenty deep dive into that), testing and gracefully failing. 

Sorry if it's a bit verbose/disorganized, I wasn't super consistent with whether I commented on stuff in Insert or Extract, so most things have a "real" comment in one and a "mirror" comment in the other.

My only real requested change on the actual transformation logic is to switch from using the `size` attribute to just grabbing the size from the tensor type directly.

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+ public:
+  SecretExtractToStaticExtractConversion(DataFlowSolver *solver,
+                                         MLIRContext *context)
+      : OpRewritePattern(context), solver(solver) {}
+
+  LogicalResult matchAndRewrite(tensor::ExtractOp extractOp,
+                                PatternRewriter &rewriter) const override {
+    // TODO(#866): Add support for reads on multi-dimensional tensors
+    auto index = extractOp->getOperand(1);
```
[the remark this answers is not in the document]
asraa: > Alternatively, the tensor type might also define a way to get the number of dimensions directly, but I don't remember of the top of my head.

That would be `getRank()`! so it could be `extractOp.getTensor().getType().getRank() != 1`

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+  using OpRewritePattern<tensor::ExtractOp>::OpRewritePattern;
+
+ public:
+  SecretExtractToStaticExtractConversion(DataFlowSolver *solver,
+                                         MLIRContext *context)
+      : OpRewritePattern(context), solver(solver) {}
+
+  LogicalResult matchAndRewrite(tensor::ExtractOp extractOp,
+                                PatternRewriter &rewriter) const override {
+    auto index = extractOp->getOperand(1);
```
asraa, line 43: Is the reason why you aren't using `op.getIndices()` so that you could extract out just the first index? If so, maybe move this below the TODO for multi-dimensional tensors since it relies on that assumption, or use `op.getIndices()[0]` after that TODO.

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.td
```diff
@@ -0,0 +1,55 @@
+#ifndef LIB_TRANSFORMS_CONVERTSECRETINSERTTOSTATICINSERT_CONVERTSECRETINSERTTOSTATICINSERT_TD_
+#define LIB_TRANSFORMS_CONVERTSECRETINSERTTOSTATICINSERT_CONVERTSECRETINSERTTOSTATICINSERT_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def ConvertSecretInsertToStaticInsert : Pass<"convert-secret-insert-to-static-insert"> {
+  let summary = "Convert tensor.insert operations on secret index to static insert operations.";
+  let description = [{Converts tensor.insert operations that write to secret index to alternative static tensor.insert operations that inserts the inserted value at each index and conditionally selects the newly produced tensor that contains the value at the secret index.
+  Note: Running this pass alone does not result in a data-oblivious program; we have to run the `--convert-if-to-select` pass to the resulting program to convert the secret-dependent If-operation to a Select-operation.
```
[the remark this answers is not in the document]
asraa: For these line breaks, if you use VSCode I install the https://marketplace.visualstudio.com/items?itemName=Google.reflowlist extension and then run Reflow on the section and it'll add the line breaks (ctrl+shift+p to get to command prompt, select reflow when your cursor is on that paragraph)

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    auto indexSecretness = indexSecretnessLattice->getValue();
+
+    bool isIndexSecret =
+        !indexSecretness.isInitialized() ||
+        (indexSecretness.isInitialized() && indexSecretness.getSecretness());
+
+    // If index is not secret, no transformation is needed
+    if (!isIndexSecret) return failure();
```
AlexanderViand-Intel, line 71: This has a few issues, most importantly it'll crash on a missing lattice due to line 64 dereferencing a null pointer ;) 
I'd suggest following the same approach as I did in fixing ConvertIfToSelect (cf. #889):


```suggestion
    // use secretness from lattice or, if no lattice found, set to unknown
    auto indexSecretness = indexSecretnessLattice ? indexSecretnessLattice->getValue() : Secretness();

    // If (1) lattice is not found or (2) lattice is set to unknown, 
    // apply transformation anyway but emit a warning
    if (!indexSecretness.isInitialized()) {
      extractOp->emitWarning()
          << "Secretness for tensor.extract index is unknown. "
             "Conservatively, the transformation will be applied:";
    }

    // If index is known to be public, no transformation is needed
    if (indexSecretness.isInitialized() && !indexSecretness.getSecretness())
        return failure();
```

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    auto indexSecretness = indexSecretnessLattice->getValue();
+
+    bool isIndexSecret =
+        !indexSecretness.isInitialized() ||
+        (indexSecretness.isInitialized() && indexSecretness.getSecretness());
+
+    // If index is not secret, no transformation is needed
+    if (!isIndexSecret) return failure();
```
AlexanderViand-Intel, line 72: See equivalent comment on extract.

tests/convert_secret_extract_to_static_extract/secret_extracts.mlir
```diff
@@ -0,0 +1,26 @@
...
+      // CHECK-NEXT: %[[INNER_FOR:.*]] = affine.for %[[J:.*]] = 0 to 32
+      // CHECK-NEXT: %[[COND:.*]] = arith.cmpi eq, %[[J]], %[[SECRET_INDEX:.*]] : index
+      // CHECK-NEXT: %[[EXTRACTED:.*]] = tensor.extract %[[TENSOR:.*]][%[[J]]]
+      // CHECK: %[[IF:.*]] = scf.if %[[COND:.*]] -> (i16) {
+      // CHECK-NEXT:   scf.yield %[[EXTRACTED]] : i16
+      // CHECK-NEXT: } else {
+      // CHECK-NEXT:   scf.yield %[[DUMMY_VAL:.*]] : i16
+      %1 = affine.for %i = 0 to 32 iter_args(%sum = %c0_i16) -> (i16) {
+        %extracted = tensor.extract %arg2[%arg3] {size = 32} : tensor<32xi16>
+        %extracted_0 = tensor.extract %arg2[%i] {size = 32} : tensor<32xi16>
```
AlexanderViand-Intel, line 18: I noticed one of your tests has a loop in the input (extract) and one doesn't (insert). I guess seeing that it handles the nested loop case is nice, so I'd suggest having one non-loop and one loop test case for each test file.

tests/convert_secret_insert_to_static_insert/secret_inserts.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: heir-opt --convert-secret-insert-to-static-insert %s | FileCheck %s
+
+// CHECK-LABEL: @insert_to_secret_index
```
AlexanderViand-Intel, line 3: see equivalent comment in extract

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+        resultSecretness = Secretness();
+      }
+    }
+
+    for (auto result : results) {
+      auto *lattice = solver->getOrCreateState<SecretnessLattice>(result);
+      solver->propagateIfChanged(lattice, lattice->join(resultSecretness));
+    }
+  }
+};
```
AlexanderViand-Intel, line 180: I'm not a fan of this, as it duplicates logic that should only be in the secrectness analysis (I guess one could say it violates [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and [SoC](https://en.wikipedia.org/wiki/Separation_of_concerns))
I think you should be able to avoid this, as you only ever create operations for which you KNOW what the secretness should be.

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+    auto forOp = builder.create<affine::AffineForOp>(0, size, 1, iterArgs);
+    // Set secretness for induction variable
+    setValueToSecretness(solver, forOp.getInductionVar(), Secretness(false));
+
+    builder.setInsertionPointToStart(forOp.getBody());
+
+    // Check if the current index is equal to the secret index
+    auto cond = builder.create<arith::CmpIOp>(arith::CmpIPredicate::eq,
+                                              forOp.getInductionVar(), index);
+    setSecretnessOfResults(solver, cond->getOperands(), cond->getResults());
```
AlexanderViand-Intel, line 101: We know that the new `arith.cmpi` compares a public variable (the induction variable) with the secret-or-unknown index, so we can just directly set the secretness to `indexSecretness` via `setValueToSecretness`

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    // Extract value at current index
+    auto newExtractOp = builder.create<tensor::ExtractOp>(
+        extractOp.getTensor(), forOp.getInductionVar());
+    setSecretnessOfResults(solver, extractOp->getOperands(),
+                           extractOp->getResults());
```
AlexanderViand-Intel, line 107: Here, it should suffice to set the `tensor.extract` op's secretness to the secretness of the `tensor` we're extracting from (the new index is now public, and there are no other operands). Let's call this `tensorSecretness`?

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+          b.create<scf::YieldOp>(loc, newExtractOp.getResult());
+        },
+        [&](OpBuilder &b, Location loc) {
+          // Yield dummyValue
+          b.create<scf::YieldOp>(loc, forOp.getRegionIterArgs().front());
+        });
+    // TODO (#888): Secretness is valid for this region-bearing op only when
+    // index is strictly secret; secretness won't be valid if index has unknown
+    // secretness (i.e., when transformation occurs in the conservative case)
+    setSecretnessOfResults(solver, ifOp->getOperands(), ifOp->getResults());
```
AlexanderViand-Intel, line 122: I don't think #888 is the right issue here, as `scf.if` DOES implement the `RegionBranchOpInterface`. Issue #888 is about stuff like `tensor.generate` which has regions, but no control flow interfaces.

In fact, I don't think there's even an issue here - you should be able to set the secreness directly, because it's just the join of `indexSecretness` with `tensorSecretness`. The extract op has secretness `indexSecretnes`, the `iter_arg` should also have `indexSecretness` and finally the condition also has `indexSecretness`. So, once more, we should be able to use the direct setting method rather than re-computing them.

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    // Replace the old tensor.insert op with forOp's result
+    rewriter.replaceOp(insertOp, forOp);
+
+    return success();
+  }
+
+ private:
+  DataFlowSolver *solver;
+  static inline void setValueToSecretness(DataFlowSolver *solver, Value value,
```
AlexanderViand-Intel, line 121: See equivalent comment in extract

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,196 @@
...
+    SmallVector<Value> iterArgs = {tensor};
+    auto forOp = builder.create<affine::AffineForOp>(0, size, 1, iterArgs);
+    setValueToSecretness(solver, forOp.getInductionVar(), Secretness(false));
+
+    builder.setInsertionPointToStart(forOp.getBody());
+
+    // Check if the current index is equal to the secret index
+    auto cond = builder.create<arith::CmpIOp>(arith::CmpIPredicate::eq,
+                                              forOp.getInductionVar(), index);
+    setSecretnessOfResults(solver, cond->getOperands(), cond->getResults());
```
AlexanderViand-Intel, line 87: See equivalent comment in extract.

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,196 @@
...
+        },
+        [&](OpBuilder &b, Location loc) {
+          // Yield old tensor
+          b.create<scf::YieldOp>(loc, forOp.getRegionIterArgs().front());
+        });
+
+    // TODO (#888): Secretness is valid for this region-bearing op only when
+    // index is strictly secret; secretness won't be valid if index has unknown
+    // secretness (i.e., when transformation occurs in the conservative case)
+    setSecretnessOfResults(solver, ifOp->getOperands(), ifOp->getResults());
```
AlexanderViand-Intel, line 107: See equivalent comment in extract.

AlexanderViand-Intel (changes requested):
There's still a few issues with how you're handling the secretness analysis edge cases and updating, once those are worked out, this should be good to merge.

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
MeronZerihun: Does this work?

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
MeronZerihun: Not sure if I understand why the iter_arg takes indexSecretness, can you please clarify?

AlexanderViand-Intel: You're right `iter_arg` should be `indexSecretness` combined with `tensorSecretness`.

AlexanderViand-Intel: Since "join" does something different than what we need, I'd say we extract the "combining secretness" logic into a function in the Secretness class, so we can use it in situations like this.

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    auto *extractOpLattice =
+        solver->lookupState<SecretnessLattice>(extractOp.getResult());
+
+    auto extractOpSecretness = extractOpLattice ? extractOpLattice->getValue() :
+                                                /*unknown*/ Secretness();
+
+    // Placeholder for extracted value
+    auto dummyValue = builder.create<arith::ConstantOp>(
+        argType, builder.getIntegerAttr(argType, 0));
```
AlexanderViand-Intel, line 85: ### While this is perfectly fine/correct, I think it might be nicer for things like vectorization to have the dummy value be `tensor.extract %t[c0]` (just makes everything more homogenous, especially if you consider what the unrolled loop would look like, and no need to create a dummy ptxt/ctxt value when lowering to actual schemes).

AlexanderViand-Intel: That also means the secretness of `dummyValue` becomes `tensorSecretness`

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.td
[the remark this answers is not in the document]
MeronZerihun: Oh okay, I will do that!

lib/Transforms/ConvertSecretExtractToStaticExtract/ConvertSecretExtractToStaticExtract.cpp
```diff
@@ -0,0 +1,211 @@
...
+          b.create<scf::YieldOp>(loc, newExtractOp.getResult());
+        },
+        [&](OpBuilder &b, Location loc) {
+          // Yield previous value
+          b.create<scf::YieldOp>(loc, forOp.getRegionIterArgs().front());
+        });
+    // Set secretness for ifOp results: Combine indexSecretness with
+    // tensorSecretness
+    auto combinedSecretness =
+        Secretness::join(indexSecretness, tensorSecretness);
```
AlexanderViand-Intel, line 130: I don't think `join` is correct here. If one of the two is unknown/uninitialized and the other is "false", we want the combinedSecretness to also be unknown/uninitialized.
Effectively, you want the same thing as https://github.com/google/heir/blob/028e48f451a19d46a2dd18180b8c5fb51ffbe92c/lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp#L42-L68 
So my recommendation would be to factor that logic out into a separate function `combine` in the `Secretness` class that takes a range or SmallVector of `Secretness`. You can then use it both in `visitOperation` and in the Extract/Insert Conversion.


To go from the `ArrayRef<*SecretnessLattice> operands` you have in `visitOperation` to the right input for that method , you can use `llvm::to_vector<2>(llvm::map_range(operands, [](SecretnessLattice *v) -> Secretness(v->getValue())))` (a quick search in the repo will show plenty of examples, mine probably won't compile as-is).

lib/Transforms/ConvertSecretInsertToStaticInsert/ConvertSecretInsertToStaticInsert.cpp
```diff
@@ -0,0 +1,196 @@
...
+          b.create<scf::YieldOp>(loc, newInsertOp.getResult());
+        },
+        [&](OpBuilder &b, Location loc) {
+          // Yield old tensor
+          b.create<scf::YieldOp>(loc, forOp.getRegionIterArgs().front());
+        });
+    // Set secretness for ifOp results: Combine indexSecretness with
+    // tensorSecretness
+    auto combinedSecretness =
+        Secretness::join(indexSecretness, tensorSecretness);
```
AlexanderViand-Intel, line 117: see equivalent comment on extract

AlexanderViand-Intel (changes requested):
One final requested change for Secretness handling, sorry!

AlexanderViand-Intel (approved):
LGTM!

j2kun (approved):

PR #883, opened 2024-08-08
RNS Support

AlexanderViand-Intel:
This a DRAFT to elicit design feedback on how we should handle RNS support in HEIR. Towards that goal, it simply *switches* most things to RNS (rather than supporting RNS and non-RNS), and it also removes various verifiers/etc for ease of implementation.

### Types
This uses an explicit `RNS` type, which is essentially a "Tuple" of Types. Note that the builtin Tuple type is not suitable, as it is not a valid tensor element type, and we want tensors or RNS-polynomials as they are the natural lowering for an (RNS) RLWECiphertext). 

Rather than adding `!rns.rns<...>` and our RNS dialect to the llvm repo, this adds an rns type to the polynomial dialect (See the changes to LLVM in my fork [here](https://github.com/llvm/llvm-project/compare/main...AlexanderViand-Intel:llvm-project:polyfix), but see also below for ideas how to avoid that. 

In order to support polynomial ops on (a) normal polynomials (b) RNS polynomials and (c) tensors of both, it also adds a new TypeConstraint and uses it for Polynomial operations.

In order to track that BGV/LWE operations are happening in RNS (which changes things such as how a "modswitch" is lowered rather dramatically), it currently introduces an RNSRLWECiphertext. This is not very elegant, but I'm not quite sure how to make it nicer. Maybe the "RNS-ness" can instead become an Attribute on RLWECiphertext? See also #876 (PR on LWE Attrs).

### Passes
This PR adds an `--expand-rns` pass that is similar to `--convert-elementwise-to-affine` (which can be used to convert polynomial ops on tensors of (rns-of) polys to loops with operations on "scalar" (rns-of) polynomials). It also updates LWE-to-polynomial to support RNS polynomials.

The intended order of passes is `-<scheme>-to-LWE --LWE-to-polynomial` (which will create poly operations on `tensor<rns<poly1,poly2,poly3>>`) followed by `--convert-elementwise-to-affine` , possibly `--convert-tensor-to-scalars` (from PR #763) and finally `--expand-rns`.

Finally, the PR switches the `--secret-to-bgv` pass to RNS, and changes the pass options from asking for the number of bits in the ctxt modulus to asking for a list of moduli.

### DIFF
--- a/lib/Dialect/LWE/IR/LWETypes.td
+++ b/lib/Dialect/LWE/IR/LWETypes.td
@@ -46,7 +46,17 @@ def RLWECiphertext : LWE_Type<"RLWECiphertext", "rlwe_ciphertext"> {
   );
 }
 
-def RLWECiphertextLike : TypeOrContainer<RLWECiphertext, "ciphertext-like">;
+def RNSRLWECiphertext : LWE_Type<"RNSRLWECiphertext", "rns_rlwe_ciphertext"> {
+  let summary = "A type for RLWE ciphertexts";
+  let parameters = (ins
+    "::mlir::Attribute":$encoding,
+    "RNSRLWEParamsAttr":$rlwe_params,
+    "Type":$underlying_type
+  );
+}
+
+def RNSorNonRNSRLWECiphertext : AnyTypeOf<[RLWECiphertext,RNSRLWECiphertext], "rns or non-rns rlwe ctxt">;
+def RLWECiphertextLike : TypesOrContainer<[RLWECiphertext,RNSRLWECiphertext], "ciphertext-like">;
 
 def RLWESecretKey : LWE_Type<"RLWESecretKey", "rlwe_secret_key"> {
   let summary = "A secret key for RLWE";
--- a/bazel/import_llvm.bzl
+++ b/bazel/import_llvm.bzl
@@ -7,7 +7,7 @@ load(
 
 def import_llvm(name):
     """Imports LLVM."""
-    LLVM_COMMIT = "0c25f85e5b88102363c0cd55e1946053d5827e99"
+    LLVM_COMMIT = "28727812fbf7ac5223aa9418a2ba8dd0e86fb2e9"
 
     new_git_repository(
         name = name,
@@ -16,5 +16,5 @@ def import_llvm(name):
         build_file_content = "# empty",
         commit = LLVM_COMMIT,
         init_submodules = False,
-        remote = "https://github.com/llvm/llvm-project.git",
+        remote = "https://github.com/AlexanderViand-Intel/llvm-project.git",
     )
--- a/lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
+++ b/lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
@@ -322,17 +322,18 @@ struct ConvertExtractOp : public OpConversionPattern<ExtractOp> {
     }
     int64_t offset = offsetAttr.getInt();
 
-    auto ctTy = op.getInput().getType();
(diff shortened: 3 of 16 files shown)


j2kun:
🤩

Just FYI, I have commit access in LLVM, so I can review and submit any upstream changes you have to polynomial dialect

lib/Dialect/LWE/IR/LWETypes.td
```diff
@@ -46,7 +46,17 @@ def RLWECiphertext : LWE_Type<"RLWECiphertext", "rlwe_ciphertext"> {
   );
 }
 
-def RLWECiphertextLike : TypeOrContainer<RLWECiphertext, "ciphertext-like">;
+def RNSRLWECiphertext : LWE_Type<"RNSRLWECiphertext", "rns_rlwe_ciphertext"> {
+  let summary = "A type for RLWE ciphertexts";
+  let parameters = (ins
+    "::mlir::Attribute":$encoding,
+    "RNSRLWEParamsAttr":$rlwe_params,
```
j2kun, line 53: I think a good approach to avoiding duplication here would be to have `RLWEParams` take a more generic attribute for the ring. Instead of a ring attr, it could be a generic Attribute with a verifier, or [`AnyAttributeOf` constraint](https://github.com/llvm/llvm-project/blob/458fa12202868281f11680a18878d57defcccc9f/mlir/include/mlir/IR/CommonAttrConstraints.td#L171) for the legal RNS/non-RNS ring attirbute options.

AlexanderViand-Intel: Referencing #876 as I guess this should be taken into consideration in the LWE Attr&Types overhaul :)

ZenithalHourlyRate:
I was trying to reuse this PR and found either the type way `!polynomial.rns` or the attr way `RNSRingAttr` for `!polynomial.polynomial` not satisfying.

> having a `!polynomial.rns` type is also bad, as we want to support RNS on integers/ctxts/etc, too, we realized we might be able to switch polynomial operations to require a certain interface on its operand's types rather than specific types

I like the idea of supporting RNS also for integers/ctxts. I think the current NewLWECiphertextType still can not handle RNS quite well. And the current ModArith dialect should also handle rns integer arithmetic.

I think this can be technically done in the following way:

* Introduce `ModularIntegerType` for `ModArith`, and RNSIntegerType is naturally `tuple<ModularIntegerType>`
* Allow `polynomial.ring<coefficientType=(i32|f32|Zp|tuple<Zp1, Zp2>)>` so `coefficientModulus` is no longer needed for `polynomial` type.
* NewLWECiphertextType uses such ring.

j2kun:
FYI, I'm planning to start working on this actively next week when I'm back from leave.

AlexanderViand-Intel:
> I think this can be technically done in the following way:
> 
> * Introduce `ModularIntegerType` for `ModArith`, and RNSIntegerType is naturally `tuple<ModularIntegerType>`
> * Allow `polynomial.ring<coefficientType=(i32|f32|Zp|tuple<Zp1, Zp2>)>` so `coefficientModulus` is no longer needed for `polynomial` type.
> * NewLWECiphertextType uses such ring.

This (minus the modarith, which didn't exist back then) is actually how I implemented RNS for my first hacky proof-of-concept. It has the advantage of being really easy to implement, and enabling very easy lowerings. However, I don't think it's a good solution overall, as it (a) makes the polynomial type messy, breaking the abstraction in a really weird FHE-specific way and (b) doesn't really allow us to re-use any of the RNS handling logic for things such as plaintext-level RNS (e.g., for bignum support in CGGI).

AlexanderViand-Intel:
Closing in favor of #1126 and related changes.

Issue #888, opened 2024-08-08
Add support for non-control-flow region-bearing ops in SecretnessAnalysis


AlexanderViand-Intel:
The fix for #839 conservatively marks region-bearing ops not handled by the analysis framework directly (i.e. no `RegionBranchOpInterface` trait)  as "unknown" secretness (unless there's a secret operand, in which case it's secret irrespective of regions). This is correct, but overly conservative, as we could set the secretness to false if none of the operations in the op's regions are "secret" (That's still somewhat conservative, but probably the best we can do without assuming something about the operation's semantics). Supporting this requires implementing `visitNonControlFlowArguments`.

AlexanderViand-Intel:
Note, that there is also still a bug for control-flow ops, as the SparseDataFlowAnalysis handles RegionBranchOpInterface ops itself (i.e., not calling the visitOperation(...) we define in our Analysis) upstream does not correctly consider, e.g., the condition in an If Statement, only resolving the analysis as join of the if/else branch results. It seems like the analyiss framework only considers statements, not expressions?

It _might_ be possible to work around this via "visitNonControlFlowArguments()", but it sounds like this would not be called on the if condition, either.

This bug does not actually cause issues in our main pipelines, as the convert-if-to-select pass correctly recognizes such if's, rewrites them, and causes the analysis to be re-run.

j2kun:
Note to self (h/t Maks): this may be fixable by using a dense analysis rather than a sparse analysis. It defines `visitRegionBranchControlFlowTransfer`

makslevental:
> Note to self (h/t Maks): this may be fixable by using a dense analysis rather than a sparse analysis. It defines `visitRegionBranchControlFlowTransfer`

That's not the right one - `scf.if` doesn't transfer controlflow between the regions does it? unlike e.g. `scf.while`. I believe what you/we want is this one

```c++
  /// Visit a program point within a region branch operation with predecessors
  /// in it. This can either be an entry block of one of the regions of the
  /// parent operation itself.
  void visitRegionBranchOperation(ProgramPoint *point,
                                  RegionBranchOpInterface branch,
                                  AbstractDenseLattice *after);
```

j2kun:
This should be fixed by https://github.com/llvm/llvm-project/pull/140797, which adds default propagation of lattices through ReturnLike region terminators.

> Note, that there is also still a bug for control-flow ops,

This will still need its own fix.

PR #889, opened 2024-08-08
SecretnessAnalysis and ConvertIfToSelect fixes

### LINKED ISSUE #839: Secretness Analysis is potentially incorrect for ops that are not `IsolatedFromAbove`
I think that, as written, the analysis might only be correct for operations that are `IsolatedFromAbove` (i.e., do not use SSA values from above/outside their own region), which the `scf` operations aren't:

```llvm
%p = ... : i32 // something public
%s = ... : i32 // something that's (potentially only "transitively") secret
%r = scf.if(%cond : i1) -> i32 { // assume a public condition
    %sum = arith.addi %s, %s : i32 // secret
    %scf.yield %sum : i32 // also secret
} else {
    %scf.yield %p : i32 // doesn't really matter for this example
}
%ss = arith.addi %r, %r : i32 // should ALSO be secret, but won't be marked as such?
```

AlexanderViand-Intel:
Fixes #839 by conservatively treating ops with regions (except control-flow ops, which the analysis framework handles directly) as "unknown" secretness (except if they have secret operands, in which case they're already "secret" anyway). See #888 for how this could be made a bit less conservative (not a high priority though, imho).

Investigating that issue also brought up some crashes in `ConvertIfToSelect`, because (a) it did not update the secretness of the operations it created and (b) it did not consider that a secretness could be absent or uninitialized. This PR fixes both issues. I guess this might also be an issue in the other data-oblivious transformations (@MeronZerihun can you please check that?)

Finally, this PR also adds a convenient (DEBUG-only) "printer" for the secretness analysis to `--convert-if-to-select`, which should make further fixes/improvements a bit easier. This works in the same way as the TargetSlot analysis, i.e., walking the IR and adding attributes. I wonder if there's a better place to put this debug logic, like maybe a debug-only pass that does nothing but run and the secretness analysis and augment the IR (@j2kun?)

### DIFF
--- a/lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
+++ b/lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
@@ -39,6 +39,11 @@ void SecretnessAnalysis::visitOperation(
   auto resultSecretness = Secretness();
   bool isUninitializedOpFound = false;
 
+  // Handle operations without operands (e.g. arith.constant)
+  if (operands.empty()) {
+    resultSecretness.setSecretness(false);
+  }
+
   for (const SecretnessLattice *operand : operands) {
     const Secretness operandSecretness = operand->getValue();
     if (!operandSecretness.isInitialized()) {
@@ -50,15 +55,16 @@ void SecretnessAnalysis::visitOperation(
       break;
   }
 
-  if (resultSecretness.isInitialized()) {
-    // Uninitialized operand: theoretically this should not happen, but it's
-    // easy to catch it this way if it does
-    if (isUninitializedOpFound && !resultSecretness.getSecretness()) {
+  // Uninitialized operand: "false" needs to be reverted to "unknown"
+  // "secret" can remain, as "unknown + secret = secret"
+  // As region-bearing ops are not yet supported in the secretness analysis
+  // (except for control-flow, which the analysis framework handles directly),
+  // we apply the same conservative logic if any regions are detected
+  // TODO (#888): Handle region-bearing ops via visitNonControlFlowArguments
+  if (isUninitializedOpFound || operation->getNumRegions()) {
+    if (resultSecretness.isInitialized() && !resultSecretness.getSecretness()) {
       resultSecretness = Secretness();
     }
-  } else {
-    // A constant operation
-    resultSecretness.setSecretness(false);
   }
 
   for (SecretnessLattice *result : results) {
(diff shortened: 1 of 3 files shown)


j2kun:
> I wonder if there's a better place to put this debug logic, like maybe a debug-only pass that does nothing but run and the secretness analysis and augment the IR.

That sounds fair. I think there's probably a further step we could take, adding a util function like

```cpp
template <typename LatticeTy>
void attachLatticeAsAttribute(
    MLIRContext *context, 
    Operation *rootOp, 
    const std::string& attrName, 
    SomeCallbackType setAttr) {
  OpBuilder builder(context);
  rootOp->walk([&](Operation *op) {
    if (op->getNumResults() == 0) return;
    auto *lattice =
        solver.lookupState<LatticeTy>(op->getResult(0));
    if (!lattice) {
      op->setAttr(attrName, builder.getStringAttr("null"));
      return;
    }
    if (!lattice->getValue().isInitialized()) {
      op->setAttr(attrName, builder.getStringAttr("unknown"));
      return;
    }
    setAttr(op, lattice);
  });

  llvm::dbg() << "Op after attaching lattice attributes: " << *rootOp;
}
```

j2kun (approved):

lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
```diff
@@ -50,15 +54,16 @@ void SecretnessAnalysis::visitOperation(
...
 
-  if (resultSecretness.isInitialized()) {
-    // Uninitialized operand: theoretically this should not happen, but it's
-    // easy to catch it this way if it does
-    if (isUninitializedOpFound && !resultSecretness.getSecretness()) {
+  // Uninitialized operand: "false" needs to be reverted to "unknown"
+  // "secret" can remain, as "unknown + secret = secret"
+  // As region-bearing ops are not yet supported in the secretness analysis
+  // (except for control-flow, which the analysis framework handles directly),
+  // we appply the same conservative logic if any regions are detected
```
MeronZerihun, line 61: Not sure how this wasn't caught by the codespell check:
```suggestion
  // we apply the same conservative logic if any regions are detected
```

MeronZerihun (changes requested):
LGTM :)

lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
```diff
@@ -39,6 +39,10 @@ void SecretnessAnalysis::visitOperation(
   auto resultSecretness = Secretness();
   bool isUninitializedOpFound = false;
```
MeronZerihun, line 41:
```suggestion
// Propagate secretness for constants
```

MeronZerihun (approved):

MeronZerihun:
> This PR fixes both issues. I guess this might also be an issue in the other data-oblivious transformations (@MeronZerihun can you please check that?)

I think the other transformations will also have similar issues; I will make sure to replicate the fix and create a PR.

asraa (approved):
Thanks! I'll merge with the comment fixes

AlexanderViand-Intel:
> Thanks! I'll merge with the comment fixes

Fixes pushed & rebased

Issue #891, opened 2024-08-08
Extend `--lwe-add-client-interface` to schemes other than BGV


AlexanderViand-Intel:
The current lowering only supports RLWECiphertexts and also implicitly assumes BGV-style encode/encrypt.
Somewhat related to #882 (adding encrypt/encode lowerings to poly/etc for things other than BGV)

Both probably require LWE Attr/Type fixes (see #758 and #876) before they can be realized.

AlexanderViand-Intel:
> This also reminds me that we should extend the pass that adds client helper encrypt/decrypt functions to support the rust codegen path.

_Originally posted by @j2kun in https://github.com/google/heir/issues/897#issuecomment-2282367968_

PR #892, opened 2024-08-08
rename `--bgv-add-client-interface` to `--lwe-add-client-interface`

### LINKED ISSUE #859: Rename/Move `--bgv-add-client-interface` to `--add-client-interface`
We already use `!lwe.rlwe_ciphertext` which (in theory, see #785) carries all the information needed to understand how to encode and encrypt a value into that ciphertext type and this pass uses nothing BGV specific as far as I can tell. 
With the addition of CKKS, it'd be nice to have one generic pass that can generate the client helper functions for all schemes. 
This should also be useful in eventually supporting scheme-switching.

When doing the switch, it'd probably ge a good idea to create an issue and add a //TODO to the code for LWE encode/encrypt support

AlexanderViand-Intel:
Resolves #859 :
> We already use !lwe.rlwe_ciphertext which (in theory, see https://github.com/google/heir/issues/785) carries all the information needed to understand how to encode and encrypt a value into that ciphertext type and this pass uses nothing BGV specific as far as I can tell.
With the addition of CKKS, it'd be nice to have one generic pass that can generate the client helper functions for all schemes. This should also be useful in eventually supporting scheme-switching.

Introduces TODO #891


j2kun (approved):

asraa (approved):
LGTM - I think this probably works for all RLWE schemes that have an encode and encrypt step (which I think all of them will)? I assume the lowering from LWE dialect manifests the scheme differences according to their types/attributes.

Issue #847, opened 2024-07-31
Pare down the build time and binary size by opting in to just the passes we need


j2kun:
The MLIR build time has a lot of stuff for code gen for various architectures that we don't target (e.g., SPIRV) and passes and dialects that we don't use (e.g., gpu, arm*, etc.). We could likely save a lot of build time by opting-in to including just the dialects and passes we need in heir-opt, and that would also likely cut down the size of the final built binary.

j2kun:
My system:

12th Gen Intel(R) Core(TM) i9-12900K 3.20 GHz, 64.0 GB RAM (24 cores)
Ubuntu clang version 17.0.6 (++20231209124227+6009708b4367-1~exp1~20231209124336.77)

From a clean build (`bazel clean --expunge`), `bazel-build ...:all` takes **(fill in here) minutes**

j2kun:
Note taking as I go:

There are some dependencies that seem unnecessary, which I might be able to clean up upstream. For example, 

```bash
bazel query 'somepath(//tools:heir-opt, @llvm-project//llvm:include/llvm/BinaryFormat/ELFRelocs/Lanai.def)'
//tools:heir-opt
@llvm-project//mlir:LLVMDialect
@llvm-project//llvm:BinaryFormat
@llvm-project//llvm:include/llvm/BinaryFormat/ELFRelocs/Lanai.def
```

LLVM dialect pulls in basically all of LLVM, even though it doesn't seem reasonable to run LLVM analyses themselves on the LLVM _MLIR dialect_.

```bash
bazel query 'somepath(//tools:heir-opt, @llvm-project//mlir:IRDLDialect)'
//tools:heir-opt
@llvm-project//mlir:MlirOptLib
@llvm-project//mlir:IRDLDialect
```

It also seems that the MlirOptLib target depends on all upstream dialects, though it's not clear why.

j2kun:
After https://github.com/google/heir/pull/895, the build time does not meaningfully change.

j2kun:
One thing I'm trying to do, but haven't quite figured out how to do, is to run a profile of the build to  get some sort of idea of what takes up most of the build time.

j2kun:
Trying again, but more specific (requires bazel version 7.0+):

```bash
bazel clean --expunge
bazel build --profile build.profile //tools:heir-opt
bazel analyze-profile build.profile
```

## Before



## After

```
=== PHASE SUMMARY INFORMATION ===

Total launch phase time                              0.785 s    0.19%
[...]

PR #895, opened 2024-08-09
Try to reduce bazel build time and binary size

j2kun:
Addresses https://github.com/google/heir/issues/847

This didn't reduce the build time all that much (30 minutes -> 25 minutes on my machine), but it does reduce the build time of `heir-opt` alone, and makes `heir-opt` 61 MiB when built with optimizations, vs the current 161 MiB


j2kun:
Copybara migration failed, patching internally.

PR #898, opened 2024-08-11
fix gcc build of rapidjson

inbelic:
- update rapidjson repo to include commit that removes assignment of
      read-only member error

see https://github.com/Tencent/rapidjson/issues/2277#issuecomment-2238470733

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -340,7 +340,7 @@ git_repository(
 git_repository(
     name = "rapidjson",
     build_file = "//bazel/openfhe:rapidjson.BUILD",
-    commit = "f54b0e47a08782a6131cc3d60f94d038fa6e0a51",
+    commit = "9bd618f545ab647e2c3bcbf2f1d87423d6edf800",
     remote = "https://github.com/Tencent/rapidjson.git",
 )
 


j2kun:
!!! 

> The fix was merged in 2016 😱

j2kun (approved):

AlexanderViand-Intel:
> !!!
> 
> > The fix was merged in 2016 😱

It looks like rapidjson is basically abandonware :( There are dozens of issues on their repo asking for new releases, especially since one lone maintainer does seem to occasionally merge a few fixes. 

@j2kun? Do we we need the speed of rapidjson, or could we swap to something more well-maintend, like nlohmann/json?

j2kun:
The obstacle is mainly what libraries are already inside Googles monorepo. I can check, and if not it's a bit of extra work to integrate a new library (and I'd be the long term maintainer of that new library internally)

asraa:
> The obstacle is mainly what libraries are already inside Googles monorepo. I can check, and if not it's a bit of extra work to integrate a new library (and I'd be the long term maintainer of that new library internally)

Actually rapidjson is so abandonware that we get internal presubmit warning asking us to not use it and to use nlohmann/json. I used to be on the team that added that presubmit warning because our json fuzzers for rapidjson had so many open bugs. I can migrate it over! I'll open a bug.

asraa (approved):

Issue #897, opened 2024-08-10
heir-translate won't ignore non-entry-point functions and errors since they have not been booleanized


kevaundray:
## Aim

I am trying to go from a c program to tfhe-bool-rust.

## Issue

An unexpected error occurs in heir-translate, which I believe is due to the non-entrypoint function not being booleanized or removed. This is a guess from the error that gets outputted. Below I give instructions on how to reproduce this. 

## The C Program

```c
// main.cc
int foo(int x) {
  return x * x;
}

int isqrt(int num) {
  int res = 0;
  int bit = 1 << 14;  // ((unsigned) INT16_MAX + 1) / 2.

  for (int i = 0; i < 8; ++i) {
    if (num >= res + bit) {
      num -= res + bit;
      res = (res >> 1) + bit;
    } else {
      num >>= 1;
    }
    bit >>= 2;
  }
  return foo(res);
}
```

This was copied from Jeremy's polygeist PR and modified to have the `foo`  function call. 

Note: The one in Jeremy's PR required me to unroll the loops for it eventually be translated to tfhe-rust-bool.

## Polygeist

I convert it from C to standard MLIR using polygeist with the following call:

```sh
cgeist -function=* -raise-scf-to-affine -S -O3 main.cc > main.mlir
```

**Polygeist output**

```mli
[...]

PR #905, opened 2024-08-12
heir-translate: skip unsupported functions with a warning

j2kun:
Fixes https://github.com/google/heir/issues/897


j2kun:
Marking as a draft because, while it works, I want to do some additional refactoring to support this in both the tfhe-rust and tfhe-rust-bool emitters.

asraa (approved):
Thank you!

Issue #786, opened 2024-07-11
Fix "Integrate LLVM" breaking main


AlexanderViand-Intel:
The (at the time of writing) most recent LLVM Integrate commit (497630d) fails to build due to an error in LLVM ([action output](https://github.com/google/heir/actions/runs/9881300266/job/27291789139)):
```
external/llvm-project/mlir/lib/Conversion/ArithToEmitC/ArithToEmitC.cpp:18:10: fatal error: mlir/Dialect/EmitC/Transforms/TypeConversions.h: No such file or directory
   18 | #include "mlir/Dialect/EmitC/Transforms/TypeConversions.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

I'm actually not entirely sure what's going on here, as the file still seems to exist upstream? Maybe an issue with the bazel overlay? Either way, it'd be nice if there was a way to set it up so that the system does _not_ Integrate LLVM if it fails our CI and instead merely create a PR/issue.

asraa:
Was this somehow fixed in https://github.com/google/heir/commit/72f06750e7170477e7ffcce603710775244e4018?

I think the issue might be that internally since we are not a production issue our build failures don't *block* submission of their integrations, but I *can* block copybara from updating the LLVM hash. I'll get that in today!

asraa:
It looks like someone had forced submitted that integrate PR - what happened was the external Bazel LLVM BUILD file wasn't updated properly and missed that dependency. 

Solution is still to block copybara for that! WIP

j2kun:
It should be blocking LLVM integrate, but we can't stop a force submit in
general. I think it will just be an occasional hazard to deal with.

asraa:
I talked to the MLIR rotation folks, who said that this could happen again since the branch they integrate upstream changes to can locally patch the BUILD files before upstream LLVM reconciles them.

Then the question is if copybara can delay posting the LLVM SHA updates until CI is green - like @j2kun said I don't think that's really possible with the way copybara works right now. Probably it would only be possible if we had some kind of staged branch that included a check on the open source build (and somehow had an easy way to keep it in sync). 

I think right now we have to do patch maintenance when that does happen, and I can add some docs for how to do that. Hopefully it's rare enough that the upstream fix isn't integrated at the same time as the breakage :/

AlexanderViand-Intel:
[...]

PR #912, opened 2024-08-16
build: patch llvm bazel build

asraa:
See https://github.com/google/heir/issues/786


AlexanderViand-Intel (approved):
👍 builds locally and on CI, thanks for putting this together so quickly :)

Issue #922, opened 2024-08-20
Adding early-exits to MLIR


MeronZerihun:
Currently, we don't have constructs to express early-exits in MLIR. As early-exits are common in many applications, we find it important to add support in loops. 

## State-of-the-art
In MLIR, a terminator is allowed to transfer the control flow either to another block in the same region or to the parent operation of the region. The parent operation may transfer the control to some other block in the same region as it is located itself, but it must be a terminator for this. Our current loop constructions are strictly defined to contain limited number of regions and only one terminator operation. In the `scf` and `affine` dialects, the for loop is defined to contain one region that contains its body. Defining multiple basic block in the for-loop body seems unachieveable since the loop only allows a single terminator within the region. The while and do-while loop defined by the `scf` dialect actually allow two regions: before and after regions but this regions are defined to fulfill the semantic meaning of how these loops operate. We also have a limit on the number and the type of terminators we can use in these regions.

## Proposed design

Ideally, we want to build new loops that either has (1) multiple regions or (2) multiple basic blocks as follows. Please note that the multiple region proposal is similar to this [RFC proposal](https://discourse.llvm.org/t/rfc-region-based-control-flow-with-early-exits-in-mlir/76998). 

#### Example of a C-like program with `break`:
```C
int foo(int input, int max, int threshold){
    int sum = 0
    for(int i = 0; i < max; i += 1){
        sum = sum + input
        if( i >= threshold)
            break
    }
    return sum
}
```

#### Using multiple basic blocks within the same region:
```mlir
func.func @foo(%input : i32, %max : index, %threshold : index) -> i32{
    %zero = arith.constant 0 : index
    %one = arith.constant 1 : index
    %result = scf.breakable_for %i = %zero to %max step %one iter_args(%arg = %input) -> i32 {
        %sum = arith.addi %input, %input : i32
        scf.cond_br %cond, ^bb1(%sum : i32), ^bb2(%sum : i32)

        ^bb1(%sum : i32):
            // scf.break terminator returns %sum and passes control back to parent op, i.e. the scf.breakable_for op
            scf.break %sum : i32
        ^bb2(%sum : i32):
            // scf.yield terminator yields %sum for the next iteration
            scf.yield %sum : i32
    }
    return %result : i32
}


```

#### Using multiple regions: 
```mlir
func.func @foo(%input : i32, %max : index, %threshold : index) -> i32{
    %zero = arith.constant 0 : index
    %one = arith.constant 1 : index
[...]

PR #923, opened 2024-08-20
Design documentation for data-oblivious transformations

MeronZerihun:
Added design doc for data-oblivious transformation

### DIFF
--- a/docs/content/en/docs/Design/do_transformation.md
+++ b/docs/content/en/docs/Design/do_transformation.md
@@ -1,0 +1,235 @@
+---
+title: Data-oblivious Transformations
+weight: 9
+---
+
+A data-oblivious program is one that decouples data input from program
+execution. Such programs exhibit control-flow and memory access patterns that
+are independent of their input(s). This programming model, when applied to
+encrypted data, is necessary for expressing FHE programs. There are 3 major
+transformations applied to convert a conventional program into a data-oblivious
+program:
+
+### (1) If-Transformation
+
+If-operations conditioned on inputs create data-dependent control-flow in
+programs. `scf.if` operations should at least define a 'then' region (true path)
+and always terminate with `scf.yield` even when `scf.if` doesn't produce a
+result. To convert a data-dependent `scf.if` operation to an equivalent set of
+data-oblivious operations in MLIR, we hoist all safely speculatable operations
+in the `scf.if` operation and convert the `scf.yield` operation to an
+`arith.select` operation. The following code snippet demonstrates an application
+of this transformation.
+
+```mlir
+// Before applying If-transformation
+func.func @my_function(%input : i1 {secret.secret}) -> () {
+  ...
+  // Violation: %input is used as a condition causing a data-dependent branch
+  %result =`%input -> (i16) {
+        %a = arith.muli %b, %c : i16
+        scf.yield %a : i16
+      } else {
+        scf.yield %b : i16
+      }
+  ...
+}
+
+// After applying If-transformation
+func.func @my_function(%input : i16 {secret.secret}) -> (){
+  ...
+  %a = arith.muli %b, %c : i16
+  %result = arith.select %input, %a, %b : i16
+  ...
+}
+```
+
+We implement a `ConvertIfToSelect` pass that transforms operations with
+secret-input conditions and with only Pure operations (i.e., operations that
+have no memory side effect and are speculatable) in their body. **This
+transformation cannot be applied to operations when side effects are present in
+only one of the two regions.** Although possible, we currently do not support
+transformations for operations where both regions have operations with matching
+side effects. When side effects are present, the pass fails.
+
+### (2) Loop-Transformation
+
+Loop statements with input-dependent conditions (bounds) and number of
+iterations introduce data-dependent branches that violate data-obliviousness. To
+convert such loops into a data-oblivious version, we replace input-dependent
+conditionals (bounds) with static input-independent parameters (e.g. defining a
+constant upper bound), and early-exits with update operations where the value
+returned from the loop is selectively updated using conditional predication. In
+MLIR, loops are expressed using either `affine.for`, `scf.for` or `scf.while`
+operations.
+
+> \[!NOTE\] Early exiting from loops is not supported in `scf` and `affine`, so
+> early exits are not supported in this pipeline. TODO(#922): support early
+> exits
+
+- `affine.for`: This operation lends itself well to expressing data oblivious
+  programs because it requires constant loop bounds, eliminating input-dependent
+  limits.
+
+```mlir
+ %sum_0 = arith.constant 0.0 : f32
+ // The for-loop's bound is a fixed constant
+ %sum = affine.for %i = 0 to 10 step 2
+ iter_args(%sum_iter = %sum_0) -> (f32) {
+   %t = affine.load %buffer[%i] : memref<1024xf32>
+   %sum_next = arith.addf %sum_iter, %input : f32
+   affine.yield %sum_next : f32
+ }
+ ...
+```
+
+- `scf.for`: In contrast to affine.for, scf.for does allow input-dependent
+  conditionals which does not adhere to data-obliviousness constraints. A
+  solution to this could be to either have the programmer or the compiler
+  specify an input-independent upper bound so we can transform the loop to use
+  this upper bound and also carefully update values returned from the for-loop
+  using conditional predication. Our current solution to this is for the
+  programmer to add the lower bound and worst case upper bound in the static
+  affine loop's `attributes` list.
+
+```mlir
+func.func @my_function(%value: i32 {secret.secret}, %inputIndex: index {secret.secret}) -> i32 {
+  ...
+  // Violation: for-loop uses %inputIndex as upper bound which causes a secret-dependent control-flow
+  %result = scf.for %iv = %begin to %inputIndex step %step_value iter_args(%arg1 = %value) -> i32 {
+    %output = arith.muli %arg1, %arg1 : i32
+    scf.yield %output : i32
+  }{lower = 0, upper = 32}
+  ...
+}
+
+// After applying Loop-Transformation
+func.func @my_function(%value: i32 {secret.secret}, %inputIndex: index {secret.secret}) -> i32 {
+  ...
+  // Build for-loop using lower and upper values from the `attributes` list
+  %result = affine.for %iv = 0 to  step 32 iter_args(%arg1 = %value) -> i32 {
+    %output = arith.muli %arg1, %agr1 : i32
+    %cond = arith.cmpi eq, %iv, %inputIndex : index
+    %newOutput = arith.select %cond, %output, %arg1
+    scf.yield %newOutput : i32
+    }
+  ...
+}
+```
+
+- `scf.while`: This operation represents a generic while/do-while loop that
+  keeps iterating as long as a condition is met. An input-dependent while
+  condition introduces a data-dependent control flow that violates
+  data-oblivious constraints. For this transformation, the programmer needs to
+  add the `max_iter` attribute that describes the maximum number of iterations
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+
+A data-oblivious program is one that decouples data input from program execution. Such programs exhibit control-flow and memory access patterns that are independent of their input(s). This programming model is necessary for expressing FHE programs. There are 3 major transformations applied to convert a conventional program into a data-oblivious program:
+
+### (1) If-Transformation
+
+If-operations conditioned on inputs create data-dependent control-flow in programs. `scf.if` operations should at least define a 'then' region (true path) and always terminate with `scf.yield` even when `scf.if` doesn't produce a result. To convert a data-dependent `scf.if` operation to an equivalent set of data-oblivious operations in MLIR, we hoist all safely speculatable operations in the `scf.if` operation and convert the `scf.yield` operation to an `arith.select` operation. The following code snippet demonstrates an application of this transformation.
+
+```mlir
+// Before applying If-transformation
+func.func @my_function(%input : i1) -> () {
```
AlexanderViand-Intel, line 16: Given that the transformations now only apply to secret values, I think it's a good idea for the documentation to reflect that:
```suggestion
func.func @my_function(%input : i1 {secret.secret}) -> () {
```

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+        %a = arith.muli %b, %c : i16
+        scf.yield %a : i16
+      } else {
+        scf.yield %b : i16
+      }
+  ...
+}
+
+// After applying If-transformation
+func.func @my_function(%input : i16) -> (){
```
AlexanderViand-Intel, line 29:
```suggestion
func.func @my_function(%input : i16 {secret.secret}) -> (){
```

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+ // The for-loop's bound is a fixed constant
+ %sum = affine.for %i = 0 to 10 step 2
+ iter_args(%sum_iter = %sum_0) -> (f32) {
+   %t = affine.load %buffer[%i] : memref<1024xf32>
+   %sum_next = arith.addf %sum_iter, %input : f32
+   affine.yield %sum_next : f32
+ }
+ ...
+```
+- `scf.for`: Similar to affine.for operation, scf.for operation requires a body region that exactly contains one block that [terminates with `scf.yield`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scffor-scfforop), which makes implementing early loop exits within the same level of abstraction a challenge. However, scf.for does allow input-dependent conditionals which does not adhere to data-obliviousness constraints. A solution to this could be to either have the programmer or the compiler specify an input-independent upper bound so we can transform the loop to use this upper bound and also carefully update values returned from the for-loop using conditional predication. Our current solution to this is simply hardcoding the upper bounds in our programs. We are unsure how to express this bounds in the IR and we would love to get suggestions that promote reusing existing MLIR frontends.
```
AlexanderViand-Intel, line 64: This is no longer accurate, as we now have a way to model this.

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+    %output = arith.muli %arg1, %arg1 : i32
+    scf.yield %output : i32
+  }
+  ...
+}
+
+// After applying Loop-Transformation
+func.func @my_function(%value: i32, %inputIndex: index) -> i32 {
+  ...
+  // %MAX_INDEX is a constant that defines the maximum possible index value
```
AlexanderViand-Intel, line 79: code snippet should be updated to reflect the new attribute-based approach

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+  %result = scf.for %iv = %begin to %MAX_INDEX step %step_value iter_args(%arg1 = %value) -> i32 {
+    %output = arith.muli %arg1, %agr1 : i32
+    %cond = arith.cmpi eq, %iv, %inputIndex : index
+    %newOutput = arith.select %cond, %output, %arg1
+    scf.yield %newOutput : i32
+    }
+  ...
+}
+```
+- `scf.while`: This operation represents a generic while/do-while loop that keeps iterating as long as a condition is met. An input-dependent while condition introduces a data-dependent control flow that violates data-oblivious constraints. It's challenging to implement early-exits since the scf.while is also expected to [terminate with `scf.yield`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop).
```
AlexanderViand-Intel, line 89: Probably a good point to reference https://github.com/google/heir/issues/922

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+    %output = arith.select %cond, %mul, %iter_arg
+    scf.yield %output
+  }
+  ...
+  return
+}
+
+```
+### (3) Access-Transformation
+Input-dependent memory access cause data-dependent memory footprints. A naive data-oblivious solution to this maybe doing read-write operations over the entire data structure while only performing the desired save/update operation for the index of interest. For simplicity, we only look at load/store operations for tensors as they are well supported structures in high-level MLIR likely emitted by most frontends. We drafted the following non-SIMD and SIMD-friendly approaches for this transformation:
```
AlexanderViand-Intel, line 126: remove the SIMD-friendly version, as we decided against implementing that (instead hoping that vectorization will be able to handle it)

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,227 @@
...
+### Related Issues and PRs
+* [#777](https://github.com/google/heir/issues/777)
+* #737
+* #778
+
+### Open Questions
+- If Transformation: we need to handle cases where both the 'then' and 'else' regions contain operations with matching side effects
+- Loop transformation: we currently don't support for do-while loops
+- Loop transformation: Besides hard-coding the bound in the attributes list, it will be useful to express upper bound constants in the IR when the upper bound in code is data-dependent (One suggested solution is to do an integer-range analysis)
+- Access transformation: we currently don't support multi-dimensional tensors
```
AlexanderViand-Intel, line 227:
```suggestion
```

AlexanderViand-Intel (commented):
I know it's still a daft, but here's some suggestions on minimal changes to make to update it to the current state

AlexanderViand-Intel (approved):

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,179 @@
+<!-- mdformat off(yaml frontmatter) -->
+---
+title: Data-oblivious Transformations
+weight: 9
+---
+<!-- mdformat on -->
```
j2kun, line 6:
```suggestion
---
title: Data-oblivious Transformations
weight: 9
---
```

Prompted by this PR, I have been tinkering with the auto-formatter in https://github.com/google/heir/pull/937. If you remove these html-style comments (as in that PR) then you should be able to run `mdformat --wrap=80` to get this PR to a consistent formatting style.

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,179 @@
...
+
+A data-oblivious program is one that decouples data input from program execution. Such programs exhibit control-flow and memory access patterns that are independent of their input(s). This programming model is necessary for expressing FHE programs. There are 3 major transformations applied to convert a conventional program into a data-oblivious program:
```
j2kun, line 8:
```suggestion
A data-oblivious program is one that decouples data input from program execution. Such programs exhibit control-flow and memory access patterns that are independent of their input(s). This programming model, when applied to encrypted data, is necessary for expressing FHE programs. There are 3 major transformations applied to convert a conventional program into a data-oblivious program:
```

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,179 @@
...
+// After applying If-transformation
+func.func @my_function(%input : i16 {secret.secret}) -> (){
+  ...
+  %a = arith.muli %b, %c : i16
+  %result = arith.select %input, %a, %b : i16
+  ...
+}
+  ```
+
+We implement a `ConvertIfToSelect` pass that transforms operations with secret-input conditions and with only Pure operations (i.e., operations that have no memory side effect and are speculatable) in their body. <strong>This transformation cannot be applied to operations when side effects are present in only one of the two regions. </strong>Although possible, we currently do not support transformations for operations where both regions have operations with matching side effects.
```
j2kun, line 37:
```suggestion
We implement a `ConvertIfToSelect` pass that transforms operations with secret-input conditions and with only Pure operations (i.e., operations that have no memory side effect and are speculatable) in their body. **This transformation cannot be applied to operations when side effects are present in only one of the two regions.** Although possible, we currently do not support transformations for operations where both regions have operations with matching side effects. When side effects are present, the pass fails.
```

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,179 @@
...
+### (2) Loop-Transformation
+<!--
+1) a. Input-dependent conditional (while-loop)
+        while(input < 0)
+   b. Input-dependent bound (for-loop)
+        for (i = 1; i < input; i++)
+2) Input-dependent number of iterations:
+      for/while(...)
+        if (input == val){ break; }
+-->
```
j2kun, line 48:
```suggestion
```

Delete the comment

docs/content/en/docs/Design/do_transformation.md
```diff
@@ -0,0 +1,179 @@
...
+  %result = affine.for %iv = 0 to  step 32 iter_args(%arg1 = %value) -> i32 {
+    %output = arith.muli %arg1, %agr1 : i32
+    %cond = arith.cmpi eq, %iv, %inputIndex : index
+    %newOutput = arith.select %cond, %output, %arg1
+    scf.yield %newOutput : i32
+    }
+  ...
+}
+```
+- `scf.while`: This operation represents a generic while/do-while loop that keeps iterating as long as a condition is met. An input-dependent while condition introduces a data-dependent control flow that violates data-oblivious constraints. It's challenging to implement early-exits since the scf.while is also expected to [terminate with `scf.yield`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop). We are working on adding early-exit support to high-level MLIR as described in this [issue](https://github.com/google/heir/issues/922). For this transformation, the programmer needs to add the `max_iter` attribute that describes the maximum number of iterations the loop runs which we then use the value to build our static `affine.for` loop.
```
j2kun, line 89: > We are working on adding early-exit support

Generally speaking it's best to leave "we will do" statements out of documentation. They often do not get completed (for better or worse) and it just becomes documentation rot. Or they get completed and the doc becomes stale as well. There's no shame in it, it's just better to state what the code does now as well as possible. Updating the docs to reflect the future will be a job for the people who implement those features.

Best would be to add a callout block like this near the start of the section

```
> [!NOTE]  
> Early exiting from loops is not supported in `scf` and `affine`, 
> so early exits are not supported in this pipeline.
> TODO(#922): support early exits
```

Then the TODO would be linked to this docs so people know to update it.

j2kun (changes requested):
Looks good! A few nits.

j2kun:
Now you should be able to rebase over main and run `pre-commit run --all-files` to apply the formatter automatically.

AlexanderViand-Intel:
@j2kun I've integrated your suggestions and rebased onto current main, should hopefully good to go now :)

j2kun (approved):

PR #927, opened 2024-08-21
[mod_arith] add reduce op

inbelic:
- add a reduce operation which will take in an arbitrary integer, x,
      and compute y in [0, q) such that y is congruent to x modulo q

    - includes a naive lowering of the op (does not take into account
      operand ranges)

### DIFF
--- a/lib/Conversion/ModArithToArith/ModArithToArith.td
+++ b/lib/Conversion/ModArithToArith/ModArithToArith.td
@@ -42,6 +42,7 @@ def HasEnoughSpaceMul: Constraint<CPred<"llvm::cast<IntegerType>(getElementTypeO
 
 def CastModulusAttributeAddSub : NativeCodeCall<"modulusHelper($0,$1,false)">;
 def CastModulusAttributeMul : NativeCodeCall<"modulusHelper($0,$1,true)">;
+def CastModulusAttributeReduce : NativeCodeCall<"modulusHelper($0,$1,false,false)">;
 
 def ConvertAddSimple : Pattern<
   (ModArith_AddOp:$op $x, $y, $mod),
@@ -150,6 +151,14 @@ def ConvertMac : Pattern<
       (Arith_ExtUIOp:$extacc $acc, (returnType $newmod)), DefOverflow),
     (Arith_TruncIOp:$res
       (Arith_RemUIOp $add, $newmod))
+  ]
+>;
+
+def ConvertReduce : Pattern<
+  (ModArith_ReduceOp $x, $mod),
+  [
+    (Arith_ConstantOp:$newmod (CastModulusAttributeAddSub $mod, $x)),
+    (Arith_RemUIOp (Arith_AddIOp (Arith_RemSIOp $x, $newmod), $newmod, DefOverflow), $newmod)
   ]
 >;
 
--- a/tests/mod_arith/mod-arith-to-arith.mlir
+++ b/tests/mod_arith/mod-arith-to-arith.mlir
@@ -229,6 +229,34 @@ func.func @test_lower_mac_vec(%lhs : tensor<4xi8>, %rhs : tensor<4xi8>, %acc : t
 
 // -----
 
+// CHECK-LABEL: @test_lower_reduce
+// CHECK-SAME: (%[[ARG:.*]]: [[TENSOR_TYPE:.*]]) -> [[TENSOR_TYPE]] {
+func.func @test_lower_reduce(%arg : tensor<4xi8>) -> tensor<4xi8> {
+  // CHECK: %[[CMOD:.*]] = arith.constant dense<17> : [[TENSOR_TYPE]]
+
+  // CHECK: %[[MOD:.*]] = arith.remsi %[[ARG]], %[[CMOD]] : [[TENSOR_TYPE]]
+  // CHECK: %[[SHIFT:.*]] = arith.addi %[[MOD]], %[[CMOD]] : [[TENSOR_TYPE]]
+  // CHECK: %[[RES:.*]] = arith.remui %[[SHIFT]], %[[CMOD]] : [[TENSOR_TYPE]]
+  %res = mod_arith.reduce %arg { modulus = 17 } : tensor<4xi8>
+  return %res : tensor<4xi8>
+}
+
+// -----
+
+// CHECK-LABEL: @test_lower_reduce_int
+// CHECK-SAME: (%[[ARG:.*]]: [[INT_TYPE:.*]]) -> [[INT_TYPE]] {
+func.func @test_lower_reduce_int(%arg : i8) -> i8 {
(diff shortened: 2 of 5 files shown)


inbelic:
Note: in our current implementation we assume that all `mod_arith` ops take in unsigned integers as parameters, with the exception of this new `reduce` op. In this manner, the `reduce` op would act as an "entry point" into the other `mod_arith` operations. Because of this assumption `mod_arith.[add|sub|mul]` can all lower directly to an `arith.[add|sub|mul] -> arith.remui mod`. Having these different input restrictions might not be the most intuitive for a user though.

The alternative would be that instead of lowering directly to `arith.remui`, we could lower to `arith.[add|sub|mul] -> mod_arith.reduce` to remove the unsigned restriction on the operands of `mod_arith`. And then use an `int-range-analysis` pass to deduce how to lower the `reduce` optimally. I think this would lead to some weird/complex uses of `extui/trunci` to ensure the operand of `reduce` is always signed, or, we would need to know if the previous `arith.[add|sub|mul]` had signed/unsigned operands.

This pull request assumes the first option.

WDYT? @asraa @j2kun @AlexanderViand-Intel

tests/mod_arith/mod-arith-to-arith.mlir
```diff
@@ -223,6 +223,34 @@ func.func @test_lower_mac_vec(%lhs : tensor<4xi8>, %rhs : tensor<4xi8>, %acc : t
...
+
+// CHECK-LABEL: @test_lower_reduce
+// CHECK-SAME: (%[[ARG:.*]]: [[TENSOR_TYPE:.*]]) -> [[TENSOR_TYPE]] {
+func.func @test_lower_reduce(%arg : tensor<4xi8>) -> tensor<4xi8> {
+  // CHECK: %[[CMOD:.*]] = arith.constant dense<17> : [[TENSOR_TYPE]]
+
+  // CHECK: %[[MOD:.*]] = arith.remsi %[[ARG]], %[[CMOD]] : [[TENSOR_TYPE]]
+  // CHECK: %[[SHIFT:.*]] = arith.addi %[[MOD]], %[[CMOD]] : [[TENSOR_TYPE]]
+  // CHECK: %[[RES:.*]] = arith.remui %[[SHIFT]], %[[CMOD]] : [[TENSOR_TYPE]]
+  %res = mod_arith.reduce %arg { modulus = 17 } : tensor<4xi8>
```
j2kun, line 236: I ran into a lot of confusion when I did stuff like this where the modulus was close to some (implicitly signed) container type's upper bound, causing the modulus to be negative. Could we make this attribute typed? Then the user would need to specify `{ modulus = 17 : i32 }`. I think APIntAttr supports this, so it would just be a change to the test code. (If I'm forgetting, maybe `IntegerAttr` or `"::mlir::IntegerAttr"` will do it.

j2kun (changes requested):
I like this approach. Unless @AlexanderViand-Intel has any objections, I only have nitpicks.

lib/Conversion/ModArithToArith/ModArithToArith.td
```diff
@@ -150,4 +151,12 @@ def ConvertMac : Pattern<
   ]
 >;
 
+def ConvertReduce : Pattern<
+  (ModArith_ReduceOp $x, $mod),
+  [
+    (Arith_ConstantOp:$newmod (CastModulusAttributeAddSub $mod, $x)),
```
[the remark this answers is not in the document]
j2kun: And if it's a meaningful difference, why didn't tests catch this?

AlexanderViand-Intel:
> WDYT? @asraa @j2kun @AlexanderViand-Intel

Mh, seems like both options are somewhat annoying. I really like the idea of the first approach, having a `mod_arith.reduce` that is then lowered to either `remui` (+ potential "adapter" code to make sure signedness is aligned, as in your recent bug-fix PR) or to a more "advanced" reduction (e.g., barret) where possible. Though I agree converting inputs to match the signedness is annoying.

Let's discuss this at today's meeting (assuming you can join @inbelic?). I wonder if this isn't more of a symptom of us not having a more principled approach to how we deal with signdness (or lack thereof) throughout the compiler.

j2kun:
IMO having an initial step to ensure an invariant is enforced throughout the rest of the program is a small price to pay for the simpler transformations it enables.

inbelic:
Apologies for not making the meeting, hectic schedule atm.

IIRC from the design meeting. Trying to add signedness to the int type would only really be possible upstream (very unlikely), and creating a custom (unsigned)-int type would be unwieldy. So we decided to encode this information to the operations (same as the Montgomery form and constant), and propagate that information down.

Thinking out loud: not sure if an operation verification can depend on a data-flow analysis, but if so, then we can use the `int-range-analysis` and fail verification if unsignedness can't be assumed.

j2kun:
I don't believe verification can depend on dataflow analysis (also it is run quite frequently so it shouldn't)

lib/Conversion/ModArithToArith/ModArithToArith.td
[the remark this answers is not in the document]
inbelic: Great catch! The new function should prevent the modulus type ever being extended further than the underlying.

The used modulus was not big enough to trigger then clause for `add` or `sub` to require an extra bit and didn't catch the incorrect implementation.

Updated code/test to trigger and handle that case correctly.

tests/mod_arith/mod-arith-to-arith.mlir
inbelic: I ran into the same confusion now when updating to a bigger modulus (217 which becomes -39 in `i8`). Having added the verify function, the `APInt` is only considered negative when specified as `{ modulus = 217 : i8 }`. For any larger user-specified bit-width it will be considered positive (and default in this case). However, since it is used as an `i8` it will be printed using the `i8` signed interpretation in the IR.

Are you suggesting that we could enforce the unsigned interpretation to be printed in the IR?

j2kun: Not only do I think we can't force the unsigned version to be printed, but when it is interpreted as signed and wraps around to negative, that will actually execute incorrectly (and potentially lower incorrectly), particularly in cases where we use the mod operator (e.g., 217 mod 255 != -39 mod 255, even though 217 = -39 mod 256).

I think we should perhaps have a verifier that ensures the modulus's type is either explicitly unsigned, or else large enough to ensure no 2s complement overflow.

j2kun (approved):
LGTM with the understanding that there may be some extra caveats to handle around the modulus type and signed overflow.

Please squash and we can merge it in.

asraa:
This is just missing a rebase!

inbelic:
Whoops, didn't see/get notified of the build fail. I will have access to my laptop again on Sunday and will rebase.

j2kun (approved):

PR #931, opened 2024-08-23
DO-Transformation: Added native-to-do pass pipeline to heir-opt

MeronZerihun:
Create a pass pipeline for taking a native program and transforming it to a data-oblivious program.


tools/heir-opt.cpp
```diff
@@ -674,6 +730,10 @@ int main(int argc, char **argv) {
       "to OpenFHE C++ code.",
       mlirToOpenFheBgvPipelineBuilder);
 
+  PassPipelineRegistration<>(
+      "native-to-do", "Transforms a native program to data-oblivious program",
```
j2kun, line 734: Can we rename it to convert-to-data-oblivious?

j2kun (changes requested):

AlexanderViand-Intel (approved):
LGTM! 

I think a technically unrelated pipeline (`--heir-basic-mlir-to-llvm`) also made it into the PR, but since we're planning to use that for some of the data-oblivious overhead benchmarking, I feel like that's a feature and not a bug ;)

j2kun (approved):

tools/heir-opt.cpp
```diff
@@ -461,6 +499,7 @@ void mlirToBgvPipelineBuilder(OpPassManager &pm,
   // Secretize inputs
   pm.addPass(createSecretize(SecretizeOptions{options.entryFunction}));
   pm.addPass(createWrapGeneric());
+  pm.addPass(createConvertIfToSelect());
```
asraa, line 502: if the data oblivious transforms are being added to the bgv pipeline, maybe all the passes should be run via a call to `convertToDataObliviousPipelineBuilder(pm);`?

asraa (commented):
thanks! just one small nit

AlexanderViand-Intel:
@asraa I addressed your last nit + rebased onto current main, should hopefully be good to go now (assuming all checks pass) :)

PR #935, opened 2024-08-25
Move int64_t casting from translate to opt bgv_to_openfhe

### LINKED ISSUE #647: move int64_t casting to `bgv-to-openfhe`
I originally wrote the `bgv-to-openfhe` lowering without a complete understanding of the limitations of the OpenFHE API, which results in a decent amount of cruft related to casting types during OpenFHE codegen. For example, the input to an encode op in OpenFHE must be `int64_t`, but the codegen supports inputs of type `int16_t` and casts them in C++ generated code.

This is relatively small, but since this is part of the OpenFHE API, it would be more coherent to represent those constraints as `openfhe` op type constraints, and insert casting ops in the lowering from BGV to OpenFHE, and then have simpler codegen for the cast ops directly.

eymay:
OpenFHE `MakePackedPlaintext` function used for encoding
accepts `std::vector<int64_t>` as input requiring casting
for smaller sized ints. The casting is moved to the `bgv_to_openfhe`
pass by emitting an `arith.extsi` cast op in the higher level.
This simplifies the emission and better represents openfhe dialect
constraints.

Resolves #647

### DIFF
--- a/lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
+++ b/lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
@@ -369,6 +369,41 @@ struct ConvertExtractOp : public OpConversionPattern<ExtractOp> {
   }
 };
 
+struct ConvertEncodeOp : public OpConversionPattern<lwe::RLWEEncodeOp> {
+  using OpConversionPattern<lwe::RLWEEncodeOp>::OpConversionPattern;
+
+  // OpenFHE has a convention that all inputs to MakePackedPlaintext are
+  // std::vector<int64_t>, so we need to cast the input to that type.
+  LogicalResult matchAndRewrite(
+      lwe::RLWEEncodeOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    auto tensorTy =
+        mlir::dyn_cast<RankedTensorType>(adaptor.getInput().getType());
+    // TODO(#646): support scalar inputs to encode.
+    if (!tensorTy) return op.emitOpError() << "Unsupported input type";
+
+    auto elementTy = tensorTy.getElementType();
+    auto intTy = mlir::dyn_cast<IntegerType>(elementTy);
+    if (!intTy)
+      return op.emitOpError() << "Input element type must be an integer type";
+
+    if (intTy.getWidth() >= 64)
+      return op.emitError() << "No supported packing technique for integers "
+                               "bigger than 64 bits.";
+
+    auto int64Ty = rewriter.getIntegerType(64);
+    auto newTensorTy = RankedTensorType::get(tensorTy.getShape(), int64Ty);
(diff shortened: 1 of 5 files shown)


lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
```diff
@@ -328,6 +328,24 @@ LogicalResult OpenFhePkeEmitter::printOperation(arith::ConstantOp op) {
   return success();
 }
 
+LogicalResult OpenFhePkeEmitter::printOperation(arith::ExtSIOp op) {
+  // OpenFHE has a convention that all inputs to MakePackedPlaintext are
+  // std::vector<int64_t>, so we need to emit casting to that type.
+
```
j2kun, line 334:
```suggestion
  // OpenFHE has a convention that all inputs to MakePackedPlaintext are
  // std::vector<int64_t>, so earlier stages in the pipeline emit typecasts.

```

lib/Conversion/BGVToOpenfhe/BGVToOpenfhe.cpp
```diff
@@ -369,6 +369,39 @@ struct ConvertExtractOp : public OpConversionPattern<ExtractOp> {
...
+        mlir::dyn_cast<RankedTensorType>(adaptor.getInput().getType());
+    // TODO(#646): support scalar inputs to encode.
+    if (!tensorTy) return op.emitOpError() << "Unsupported input type";
+
+    auto elementTy = tensorTy.getElementType();
+    auto intTy = mlir::dyn_cast<IntegerType>(elementTy);
+    if (!intTy)
+      return op.emitOpError() << "Input element type must be an integer type";
+
+    if (intTy.getWidth() >= 64) return failure();
```
j2kun, line 390:
```suggestion
    if (intTy.getWidth() >= 64) {
      return op.emitError() 
        << "No supported packing technique for integers bigger than 64 bits.";
    }
```

j2kun (changes requested):
Well done! Two minor nits, then let's get it merged.

eymay:
Thanks for the review! I have applied the suggestions and amended the commit for easier merging if you do not mind.

asraa (approved):
Thanks for such a quick turnaround! The PR looks great!

tests/bgv_to_openfhe/cast.mlir
```diff
@@ -0,0 +1,42 @@
+// RUN: heir-opt --bgv-to-openfhe %s | FileCheck %s
```
asraa, line 1: very small suggestion if you have time - to add a quick test on the output of one of these tests (extsi + encode) to the openfhe emitter test: https://github.com/google/heir/blob/6ba86a86beee53c25f12a70b73a7511100567263/tests/openfhe/ops.mlir

eymay: Sure, I pushed a commit for it. I can squash them if it looks good.

Issue #880, opened 2024-08-08
Add support for Secret Key Encryption in lowering LWE to Polynomial


lawrencekhlim:
Currently, our lowering for LWE to Polynomial only supports public key encryption (see LWEToPolynomial.cpp and #874). We need to add support for Secret Key Encryption as well.

Issue #881, opened 2024-08-08
Configure the Generated Randomness in LWE to Polynomial lowering


lawrencekhlim:
In our encryption lowering inside LWEToPolynomial.cpp, we need to configure some of the parameters for the randomness. Parameters that need to be configured include:
* Seed
* Standard Deviation of the Discrete Gaussian distribution

PR #936, opened 2024-08-26
Add support for lowering RLWEEncryptOps with Secret Key to Poly Dialect

markoristicc:
Attempted to tackle the issue outlined by #880

achieved this by lowering an RLWEEncryptOp to generate a ciphertext of the form (u,<u,s> + m + e) composed of Polynomial Dialect operations.

### DIFF
--- a/lib/Conversion/LWEToPolynomial/LWEToPolynomial.cpp
+++ b/lib/Conversion/LWEToPolynomial/LWEToPolynomial.cpp
@@ -206,23 +206,60 @@ struct ConvertRLWEEncrypt : public OpConversionPattern<RLWEEncryptOp> {
             });
 
     ImplicitLocOpBuilder builder(loc, rewriter);
+
+    auto index0 = builder.create<arith::ConstantIndexOp>(0);
+    auto dimension =
+        inputT.getRing().getPolynomialModulus().getPolynomial().getDegree();
+
+    auto elementType = rewriter.getIntegerType(inputT.getRing()
+                                                   .getCoefficientModulus()
+                                                   .getType()
+                                                   .getIntOrFloatBitWidth());
+
+    auto tensorParams = RankedTensorType::get({dimension}, elementType);
+
+    // TODO (#881): Add pass options to change the seed (which is currently
+    // hardcoded to 0 with index).
+    // TODO (#873) : Clean up usage of Random Dialect below using num_bits
+    // (currently hardcoded to 32).
+
+    // Initialize random number generator with seed.
+    auto generateRandom =
+        builder.create<random::InitOp>(index0, builder.getI32IntegerAttr(32));
+
+    // Create a uniform discrete random distribution with generated values of
+    // -1, 0, 1.
+    auto uniformDistributionType = random::DistributionType::get(
+        getContext(), random::Distribution::uniform);
+
+    auto uniformDistribution =
+        builder.create<random::DiscreteUniformDistributionOp>(
+            uniformDistributionType, generateRandom,
+            builder.getI32IntegerAttr(-1), builder.getI32IntegerAttr(2));
+
+    // Generate random u polynomial from uniform random ternary distribution
+    auto uTensor =
+        builder.create<random::SampleOp>(tensorParams, uniformDistribution);
+    auto u =
+        builder.create<polynomial::FromTensorOp>(uTensor, inputT.getRing());
+
+    // Create a discrete Gaussian distribution
+    auto discreteGaussianDistributionType = random::DistributionType::get(
+        getContext(), random::Distribution::gaussian);
+
+    auto discreteGaussianDistribution =
+        builder.create<random::DiscreteGaussianDistributionOp>(
+            discreteGaussianDistributionType, generateRandom,
+            builder.getI32IntegerAttr(0), builder.getI32IntegerAttr(5));
+    // TODO (#881): Add pass options to configure stdev
+    // (which is currently hardcoded to 5)
+
     if (isPublicKey) {
-      auto index0 = builder.create<arith::ConstantIndexOp>(0);
       tensor::ExtractOp publicKey0 =
           builder.create<tensor::ExtractOp>(key, ValueRange{index0});
       auto index1 = builder.create<arith::ConstantIndexOp>(1);
       tensor::ExtractOp publicKey1 =
           builder.create<tensor::ExtractOp>(key, ValueRange{index1});
-
-      auto dimension =
-          inputT.getRing().getPolynomialModulus().getPolynomial().getDegree();
-
-      auto elementType = rewriter.getIntegerType(inputT.getRing()
-                                                     .getCoefficientModulus()
-                                                     .getType()
-                                                     .getIntOrFloatBitWidth());
-
-      auto tensorParams = RankedTensorType::get({dimension}, elementType);
 
       // constantT is 2**(cleartextBitwidth), and is used for scalar
       // multiplication.
@@ -230,43 +267,6 @@ struct ConvertRLWEEncrypt : public OpConversionPattern<RLWEEncryptOp> {
       // attributes.
       auto constantT = builder.create<arith::ConstantOp>(
           builder.getI32IntegerAttr(1 << cleartextBitwidth));
-
-      // TODO (#881): Add pass options to change the seed (which is currently
-      // hardcoded to 0 with index0).
-      // TODO (#873) : Clean up usage of Random Dialect below using num_bits
-      // (currently hardcoded to 32).
-
-      // Initialize random number generator with seed.
-      auto generateRandom =
-          builder.create<random::InitOp>(index0, builder.getI32IntegerAttr(32));
-
-      // Create a uniform discrete random distribution with generated values of
-      // -1, 0, 1.
-      auto uniformDistributionType = random::DistributionType::get(
-          getContext(), random::Distribution::uniform);
-
-      auto uniformDistribution =
-          builder.create<random::DiscreteUniformDistributionOp>(
-              uniformDistributionType, generateRandom,
-              builder.getI32IntegerAttr(-1), builder.getI32IntegerAttr(2));
-
-      // generate random u polynomial from uniform random ternary distribution
-      auto uTensor =
-          builder.create<random::SampleOp>(tensorParams, uniformDistribution);
-      auto u =
-          builder.create<polynomial::FromTensorOp>(uTensor, inputT.getRing());
-
-      // Create a discrete Gaussian distribution
-      auto discreteGaussianDistributionType = random::DistributionType::get(
-          getContext(), random::Distribution::gaussian);
-
-      auto discreteGaussianDistribution =
-          builder.create<random::DiscreteGaussianDistributionOp>(
-              discreteGaussianDistributionType, generateRandom,
-              builder.getI32IntegerAttr(0),
-              builder.getI32IntegerAttr(
-                  5));  // TODO (#881): Add pass options to configure stdev
-                        // (which is currently hardcoded to 5)
 
       // generate random e0 polynomial from discrete gaussian distribution
       auto e0Tensor = builder.create<random::SampleOp>(
@@ -299,11 +299,35 @@ struct ConvertRLWEEncrypt : public OpConversionPattern<RLWEEncryptOp> {
           llvm::ArrayRef<Value>({ciphertext0, ciphertext1}));
       rewriter.replaceOp(op, ciphertext);
     } else {  // secret key
-      // TODO (#880): Add support for secret key encryption. Left as future
-      // work.
-      op.emitError()
-          << "`lwe.rlwe_encrypt` doesn't yet support secret key encryption.";
-      return failure();
+      // We only support secret key encryption with a single polynomial (typical
+      // RLWE parameters, whereas CGGI may use a larger number of polynomials
+      // for the secret key).
+      if (cast<RankedTensorType>(key.getType()).getNumElements() != 1) {
+        return op.emitError()
+               << "`lwe.rlwe_encrypt` only supports secret keys with a single "
+                  "polynomial, got secret key type "
+               << key.getType();
+      }
+
+      // Generate random e polynomial from discrete gaussian distribution
+      auto eTensor = builder.create<random::SampleOp>(
+          tensorParams, discreteGaussianDistribution);
+      auto e =
+          builder.create<polynomial::FromTensorOp>(eTensor, inputT.getRing());
+
+      // TODO (#882): Other encryption schemes (e.g. CKKS) may multiply the
+      // noise or key differently. Add support for those cases.
+      // ciphertext0 = u
+      // Compute ciphertext1 = <u,s> + m + e
+      auto keyPoly = builder.create<tensor::ExtractOp>(key, ValueRange{index0});
+      auto us = builder.create<polynomial::MulOp>(u, keyPoly);
+      auto usM = builder.create<polynomial::AddOp>(us, input);
+      auto ciphertext1 = builder.create<polynomial::AddOp>(usM, e);
+
+      // ciphertext = (u, ciphertext0)
+      auto ciphertext = builder.create<tensor::FromElementsOp>(
+          llvm::ArrayRef<Value>({u, ciphertext1}));
+      rewriter.replaceOp(op, ciphertext);
     }
 
     return success();
--- a/tests/lwe_to_polynomial/test_rlwe_sk_encrypt.mlir
+++ b/tests/lwe_to_polynomial/test_rlwe_sk_encrypt.mlir
@@ -1,0 +1,36 @@
+// RUN: heir-opt %s --lwe-to-polynomial | FileCheck %s
+
+#encoding = #lwe.polynomial_coefficient_encoding<cleartext_start=15, cleartext_bitwidth=15>
+#my_poly = #polynomial.int_polynomial<1 + x**1024>
+#ring = #polynomial.ring<coefficientType = i32, coefficientModulus = 7917 : i32, polynomialModulus=#my_poly>
+#rlwe_params = #lwe.rlwe_params<dimension=2, ring=#ring>
+!plaintext_rlwe = !lwe.rlwe_plaintext<encoding = #encoding, ring = #ring, underlying_type=i3>
+!ciphertext_rlwe = !lwe.rlwe_ciphertext<encoding = #encoding, rlwe_params = #rlwe_params, underlying_type=i3>
+!rlwe_key = !lwe.rlwe_secret_key<rlwe_params=#rlwe_params>
+
+func.func @test_rlwe_sk_encrypt(%arg0: !plaintext_rlwe, %arg1: !rlwe_key) -> !ciphertext_rlwe {
+  // CHECK-NOT: lwe.rlwe_encrypt
+
+  // CHECK: %[[ZERO:.*]] = arith.constant 0 : index
+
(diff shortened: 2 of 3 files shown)


markoristicc:
A lot of the lowering code (generating uniform distribution Op, Gaussian distribution, e) is common between both the public key and private key encryption (unless I did this wrong which is very much possible). I chose to keep them separated in their if-else bodies, but let me know if I should move them to be outside of the if-else.

lawrencekhlim:
Go ahead and add a test for your code; the lowering otherwise looks good! Feel free to see [encrypt_ops.mlir](https://github.com/markoristicc/heir/blob/RLWESKLowering/tests/lwe_to_polynomial/encrypt_ops.mlir) for an example.

lawrencekhlim (changes requested):

lib/Conversion/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -299,11 +299,79 @@ struct ConvertRLWEEncrypt : public OpConversionPattern<RLWEEncryptOp> {
...
+
+      // Create a discrete Gaussian distribution
+      auto discreteGaussianDistributionType = random::DistributionType::get(
+          getContext(), random::Distribution::gaussian);
+
+      auto discreteGaussianDistribution =
+          builder.create<random::DiscreteGaussianDistributionOp>(
+              discreteGaussianDistributionType, generateRandom,
+              builder.getI32IntegerAttr(0), builder.getI32IntegerAttr(5));
+      // TODO: Add pass options to configure stdev
```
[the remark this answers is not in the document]
j2kun: Can we attach it to an open GH issue? (it will be autodetected if the comment says `TODO(#xxx): msg` where `xxx` is the github issue id.

lib/Conversion/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -299,11 +299,79 @@ struct ConvertRLWEEncrypt : public OpConversionPattern<RLWEEncryptOp> {
           llvm::ArrayRef<Value>({ciphertext0, ciphertext1}));
       rewriter.replaceOp(op, ciphertext);
     } else {  // secret key
-      // TODO (#880): Add support for secret key encryption. Left as future
-      // work.
-      op.emitError()
-          << "`lwe.rlwe_encrypt` doesn't yet support secret key encryption.";
-      return failure();
+      auto index = builder.create<arith::ConstantIndexOp>(0);
```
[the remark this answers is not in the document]
markoristicc: Will do

markoristicc:
> Go ahead and add a test for your code; the lowering otherwise looks good! Feel free to see [encrypt_ops.mlir](https://github.com/markoristicc/heir/blob/RLWESKLowering/tests/lwe_to_polynomial/encrypt_ops.mlir) for an example.

Sorry, but should this be another test case? Or should I simply integrate it with the encrypt_ops.mlir test, adding a RLWEEncryptOp that uses a private key to the IR?

lawrencekhlim:
I think it makes sense to simply integrate it with the existing test case, but even if you separate it out and rename the files to pk_encrypt_ops and sk_encrypt_ops, then that's fine too.

lib/Conversion/LWEToPolynomial/LWEToPolynomial.cpp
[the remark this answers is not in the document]
markoristicc: Not sure if there's an open GH issue associated with this error (unless I missed it). If you guys want to create one I could add it in before I merge?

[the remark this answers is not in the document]
lawrencekhlim: Issue #881 is what you're looking for.

markoristicc:
Currently am experiencing a strange error: When modifying encrypt_ops.mlir to use a secret key rather than a public one, my code experiences an error on line 310, 

`auto elementType = rewriter.getIntegerType(inputT.getRing()
                                                     .getCoefficientModulus()
                                                     .getType()
                                                     .getIntOrFloatBitWidth());`
                                                     
with the error being

'Running command line: bazel-bin/tools/heir-opt --lwe-to-polynomial /Users/markoristic/Projects/openSource/heir/tests/lwe_to_polynomial/encrypt_ops.mlir

!lwe.rlwe_plaintext<encoding = #lwe.polynomial_coefficient_encoding<cleartext_start = 15, cleartext_bitwidth = 15>, ring = <coefficientType = i32, coefficientModulus = 7917 : i32, polynomialModulus = <1 + x**1024>>, underlying_type = i3>
7917 : i32
i32
Assertion failed: (isIntOrFloat() && "only integers and floats have a bitwidth"), function getIntOrFloatBitWidth, file Types.cpp, line 128.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.'

I added in some .dump() calls in the debug trace, for getRing(), getCoefficientModulus(), and getType(). As you can see, the result of getType is i32, so I'm unsure as to why getIntOrFloatBitwidth is failing. Is this truly, like the error message suggests, an MLIR bug, or is there something I am doing incorrectly here?

Interestingly enough, this error only occurs when trying to compile an RLWEEncryptOp with a secret key, not the public key, even though inputT is the PlaintextType. Wouldn't this mean we would expect to see the same error when using a secret key and public key, as the plaintext is unchanged?

lawrencekhlim:
Could you share with us your test file?

markoristicc:
Updated the MR to show how I modified encrypt_ops.mlir. Like mentioned, I naively replaced the public key with the secret key, using the same RLWE_params as input. I didn't end up modifying the Plaintext handling though, so I'm a little confused as towards the cause of the error w.r.t. inputT

lawrencekhlim:
I'm still investigating the code, but when I change the test to use a secret key without incorporating your changes to LWEToPolynomial.cpp, the test runs properly, so it tells me that the problem is with your code somewhere and not the test yet.

lawrencekhlim:
1. Could someone else take a look at the bug here? It doesn't seem to me that `lwe.rlwe_secret_key` is special in any way (based on checking LWEOps.td), yet it seems to break something else that looks unrelated.

asraa:
I'll take a look! I'll get back to you by the afternoon

asraa:
Hey @markoristicc here's a gist with a diff that fixes the issue (it doesn't fix the test because it's incorrect, but it fixes the assert failure).

https://gist.github.com/asraa/5bb50d68a8b61d089f24d2555b443f08

I didn't see the error happening at 310, to me it was happening at the line where `auto keyTensor` was being created. This was my stack trace from the PR:
```
    ** Insert  : 'polynomial.from_tensor'(0x55f11633a8e0)
heir-opt: external/llvm-project/mlir/lib/IR/Types.cpp:128: unsigned int mlir::Type::getIntOrFloatBitWidth() const: Assertion `isIntOrFloat() && "only integers and floats have a bitwidth"' failed.
Stack dump:
0.      Program arguments: /usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/execroot/heir/bazel-out/k8-dbg/bin/tools/heir-opt --lwe-to-polynomial /usr/local/google/home/asraa/git/heir/tests/lwe_to_polynomial/encrypt_ops.mlir --debug
 #0 0x000055f0d85a6016 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:723:22
 #1 0x000055f0d85a63ea PrintStackTraceSignalHandler(void*) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:798:1
 #2 0x000055f0d85a3892 llvm::sys::RunSignalHandlers() /proc/self/cwd/external/llvm-project/llvm/lib/Support/Signals.cpp:105:20
 #3 0x000055f0d85a5927 SignalHandler(int) /proc/self/cwd/external/llvm-project/llvm/lib/Support/Unix/Signals.inc:413:1
 #4 0x00007f1ac2e591a0 (/lib/x86_64-linux-gnu/libc.so.6+0x3d1a0)
 #5 0x00007f1ac2ea70ec __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #6 0x00007f1ac2e59102 gsignal ./signal/../sysdeps/posix/raise.c:27:6
 #7 0x00007f1ac2e424f2 abort ./stdlib/abort.c:81:7
 #8 0x00007f1ac2e42415 _nl_load_domain ./intl/loadmsgcat.c:1177:9
 #9 0x00007f1ac2e51d32 (/lib/x86_64-linux-gnu/libc.so.6+0x35d32)
#10 0x000055f0d8266f28 mlir::Type::getIntOrFloatBitWidth() const /proc/self/cwd/external/llvm-project/mlir/lib/IR/Types.cpp:129:49
#11 0x000055f0d4a9556e mlir::detail::ShapedTypeTrait<mlir::TensorType>::getElementTypeBitWidth() const /proc/self/cwd/bazel-out/k8-dbg/bin/external/llvm-project/mlir/include/mlir/IR/BuiltinTypeInterfaces.h.inc:225:5
#12 0x000055f0d4a85cc5 mlir::polynomial::FromTensorOp::build(mlir::OpBuilder&, mlir::OperationState&, mlir::Value, mlir::polynomial::RingAttr) /proc/self/cwd/external/llvm-project/mlir/lib/Dialect/Polynomial/IR/PolynomialOps.cpp:26:52
#13 0x000055f0d442a847 mlir::polynomial::FromTensorOp mlir::OpBuilder::create<mlir::polynomial::FromTensorOp, mlir::Value&, mlir::polynomial::RingAttr>(mlir::Location, mlir::Value&, mlir::polynomial::RingAttr&&) /proc/self/cwd/external/llvm-project/mlir/include/mlir/IR/Builders.h:514:22
#14 0x000055f0d4426faf mlir::polynomial::FromTensorOp mlir::ImplicitLocOpBuilder::create<mlir::polynomial::FromTensorOp, mlir::Value&, mlir::polynomial::RingAttr>(mlir::Value&, mlir::polynomial::RingAttr&&) /proc/self/cwd/external/llvm-project/mlir/include/mlir/IR/ImplicitLocOpBuilder.h:68:3
#15 0x000055f0d4424a45 mlir::heir::lwe::ConvertRLWEEncrypt::matchAndRewrite(mlir::heir::lwe::RLWEEncryptOp, mlir::heir::lwe::RLWEEncryptOpAdaptor, mlir::ConversionPatternRewriter&) const 
#16 0x000055f0d43a8847 mlir::OpConversionPattern<mlir::heir::lwe::RLWEEncryptOp>::matchAndRewrite(mlir::Operation*, llvm::ArrayRef<mlir::Value>, mlir::ConversionPatternRewriter&) const /proc/self/cwd/external/llvm-project/mlir/include/mlir/Transforms/DialectConversion.h:544:27
#17 0x000055f0d7c45f23 mlir::ConversionPattern::matchAndRewrite(mlir::Operation*, mlir::PatternRewriter&) const /proc/self/cwd/external/llvm-project/mlir/lib/Transforms/Utils/DialectConversion.cpp:1650:25
#18 0x000055f0d7cb9e8f mlir::PatternApplicator::matchAndRewrite(mlir::Operation*, mlir::PatternRewriter&, llvm::function_ref<bool (mlir::Pattern const&)>, llvm::function_ref<void (mlir::Pattern const&)>, llvm::function_ref<llvm::LogicalResult (mlir::Pattern const&)>)::'lambda'()::operator()() const /proc/self/cwd/external/llvm-project/mlir/lib/Rewrite/PatternApplicator.cpp:212:46
#19 0x000055f0d7cba8ec void llvm::function_ref<void ()>::callback_fn<mlir::PatternApplicator::matchAndRewrite(mlir::Operation*, mlir::PatternRewriter&, llvm::function_ref<bool (mlir::Pattern const&)>, llvm::function_ref<void (mlir::Pattern const&)>, llvm::function_ref<llvm::LogicalResult (mlir::Pattern const&)>)::'lambda'()>(long) /proc/self/cwd/external/llvm-project/llvm/include/llvm/ADT/STLFunctionalExtras.h:46:40
#20 0x000055f0d49a111e llvm::function_ref<void ()>::operator()() const /proc/self/cwd/external/llvm-project/llvm/include/llvm/ADT/STLFunctionalExtras.h:68:62
```

There's a comment in the gist explaining what's going on - but basically `key` is already a tensor, and so you were trying to create a tensor out of `tensor<1xpolynomial>`, and it assert failed trying to get the int or float bit width of a tensor type.

Hopefully that works, given that I'm not getting the same error line!

markoristicc:
Ahhh that makes sense. Initially I was using key directly, however was faced with an error stating that ''polynomial.mul' op requires the same type for all operands and results' 

As one argument was a tensor and one was a polynomial, I thought the correct course of action was to convert key to a tensor, although in hindsight choosing to make both arguments tensors instead of polynomials makes literally no sense. Thanks for helping to catch the error, much appreciated :)

tests/lwe_to_polynomial/sk_encrypt_op.mlir
```diff
@@ -0,0 +1,32 @@
...
+  // CHECK-DAG: %[[PRNG:.*]] = random.init_prng %[[SEED:.*]]
+  // CHECK-DAG: %[[UNIFORM:.*]] = random.discrete_uniform_distribution %[[PRNG]]
+  // CHECK-DAG: %[[GAUSSIAN:.*]] = random.discrete_gaussian_distribution %[[PRNG]]
+
+  // CHECK-DAG: %[[SK:.*]] = tensor.extract %arg1[%[[ZERO]]]
+  // CHECK-DAG: %[[SAMPLE_U:.*]] = random.sample %[[UNIFORM]]
+  // CHECK-DAG: %[[U:.*]] = polynomial.from_tensor %[[SAMPLE_U]]
+  // CHECK: %[[SAMPLE_E:.*]] = random.sample %[[GAUSSIAN]]
+  // CHECK: %[[E:.*]] = polynomial.from_tensor %[[SAMPLE_E]]
+  // CHECK-DAG: %[[SK_TIMES_U:.*]] = polynomial.mul %[[SK]], %[[U]]
```
lawrencekhlim, line 25: When I run the test, it doesn't pass for me. Here are some changes I made for it to pass on my machine:
```
  // CHECK-DAG: %[[SAMPLE_E:.*]] = random.sample %[[GAUSSIAN]]
  // CHECK-DAG: %[[E:.*]] = polynomial.from_tensor %[[SAMPLE_E]]
  // CHECK-DAG: %[[SK_TIMES_U:.*]] = polynomial.mul %[[U]], %[[SK]]
```
Unfortunately, SAMPLE_E and E come in a different order than what the test currently requires. Hence, I modified them to be CHECK-DAG's instead of CHECK. 

Also, I noticed that the polynomial multiplication multiplies U by SK instead of SK by U, so I've reordered them. Now it passes for me.

lawrencekhlim (changes requested):

tests/lwe_to_polynomial/sk_encrypt_op.mlir
lawrencekhlim: Should probably rename `SK_TIMES_U` to `U_TIMES_SK` for naming consistency

markoristicc: Yeah the tests were still failing at the point of the last commit - I got them fixed up tho and committed them, but unfortunately I messed up a bit on the rebasing and included some extraneous commits, which I will now remove

markoristicc: Ok ended up fixing the commit history ... separated the tests into 2 files to keep the FileCheck Naming a bit more legible. lmk if you want me to change anything with the sk_encrypt_op.mlir file or the pk_encrypt_op.mlir file.

lawrencekhlim (commented):
Besides that, everything else looks good! Thanks!

tests/lwe_to_polynomial/pk_encrypt_op.mlir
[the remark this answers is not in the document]
markoristicc: totally valid, didn't realize I'd made some small changes.

tests/lwe_to_polynomial/sk_encrypt_op.mlir
```diff
@@ -0,0 +1,36 @@
...
+#encoding = #lwe.polynomial_coefficient_encoding<cleartext_start=15, cleartext_bitwidth=15>
+#my_poly = #polynomial.int_polynomial<1 + x**1024>
+#ring = #polynomial.ring<coefficientType = i32, coefficientModulus = 7917 : i32, polynomialModulus=#my_poly>
+#rlwe_params = #lwe.rlwe_params<dimension=2, ring=#ring>
+!plaintext_rlwe = !lwe.rlwe_plaintext<encoding = #encoding, ring = #ring, underlying_type=i3>
+!ciphertext_rlwe = !lwe.rlwe_ciphertext<encoding = #encoding, rlwe_params = #rlwe_params, underlying_type=i3>
+!rlwe_key = !lwe.rlwe_secret_key<rlwe_params=#rlwe_params>
+
+func.func @test_rlwe_encrypt(%arg0: !plaintext_rlwe, %arg1: !rlwe_key) -> !ciphertext_rlwe {
+  // CHECK-NOT: lwe.rlwe_encrypt
```
lawrencekhlim, line 12: Oh one more thing: could this test be renamed to @test_rlwe_sk_encrypt? Also do the same for the public key encrypt?

tests/lwe_to_polynomial/test_rlwe_sk_encrypt.mlir
```diff
@@ -0,0 +1,36 @@
...
+
+#encoding = #lwe.polynomial_coefficient_encoding<cleartext_start=15, cleartext_bitwidth=15>
+#my_poly = #polynomial.int_polynomial<1 + x**1024>
+#ring = #polynomial.ring<coefficientType = i32, coefficientModulus = 7917 : i32, polynomialModulus=#my_poly>
+#rlwe_params = #lwe.rlwe_params<dimension=2, ring=#ring>
+!plaintext_rlwe = !lwe.rlwe_plaintext<encoding = #encoding, ring = #ring, underlying_type=i3>
+!ciphertext_rlwe = !lwe.rlwe_ciphertext<encoding = #encoding, rlwe_params = #rlwe_params, underlying_type=i3>
+!rlwe_key = !lwe.rlwe_secret_key<rlwe_params=#rlwe_params>
+
+func.func @test_rlwe_encrypt(%arg0: !plaintext_rlwe, %arg1: !rlwe_key) -> !ciphertext_rlwe {
```
lawrencekhlim, line 11: Apologies, I wrote something confusing. I meant to say that this func.func here should be renamed to test_rlwe_sk_encrypt, not the file name itself (and the same for the pk encrypt test). This is just so that it's easier to disambiguate the two if there are UIs that run through all of these tests and one of them fails.

markoristicc: ahhhh got it my bad

markoristicc: ok should be supported now

lawrencekhlim (approved):

asraa (approved):
That looks great to me, great work @markoristicc! And thank you for the review @lawrencekhlim

markoristicc:
Is there anything I need to do w.r.t the Copybara process?

asraa:
> Is there anything I need to do w.r.t the Copybara process?

No, thank you! Let me ping someone to approve it, that was on me (I need a second person)

PR #945, opened 2024-08-27
Extend the TfheRustBool Emitter with the Packed API

WoutLegiest:
This PR is a follow up of #842
We implemented a second TfheRustBool Emitter pass, specifically for the packed API (--emit-tfhe-rust-bool-packed)
The former emitter pass (--emit-tfhe-rust-bool) still outputs boolean tfhe-rs ready code.


lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -215,9 +245,25 @@ void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
...
     ::mlir::Value result, ::mlir::Value sks, ::mlir::ValueRange nonSksOperands,
     std::string_view op, SmallVector<std::string> operandTypes) {
-  if (isa<TensorType>(nonSksOperands[0].getType())) {
+  // ToDo: Check again the enum definition
+  if (packedAPI && (op == "and" || op == "nand" || op == "or" || op == "nor" ||
+                    op == "xor" || op == "xnor")) {
+    std::string gateStr = std::string(op);
+    std::transform(
+        gateStr.begin(), gateStr.end(), gateStr.begin(),
+        [](unsigned char c) -> unsigned char { return std::toupper(c); });
```
asraa, line 254: It might be possible to use `llvm::StringRef`'s [upper()](https://github.com/llvm/llvm-project/blob/5b3ba438dfa7815bb0f3be07a300866085a431b9/llvm/lib/Support/StringRef.cpp#L116)

```suggestion
    std::string gateStr = StringRef(op).upper();
```

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -215,9 +245,25 @@ void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
...
+
     auto *opParent = nonSksOperands[0].getDefiningOp();
 
+    size_t numberOfOperands = 0;
+
+    if (isa<TensorType>(result.getType())) {
+      auto resType = mlir::dyn_cast<TensorType>(result.getType());
+      numberOfOperands = resType.getShape()[0];
```
asraa, line 262: *just* in case this ends up being a tensor like tensor<1xN> - maybe use `resType.getNumElements()`

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -215,9 +245,25 @@ void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
...
+    } else {
+      numberOfOperands = 1;
+    }
```
asraa, line 265:
```suggestion
    size_t numberOfOperands = 1;

    if (isa<TensorType>(result.getType())) {
      auto resType = mlir::dyn_cast<TensorType>(result.getType());
      numberOfOperands = resType.getNumElements();
    }
```

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -215,9 +245,25 @@ void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
 LogicalResult TfheRustBoolEmitter::printSksMethod(
     ::mlir::Value result, ::mlir::Value sks, ::mlir::ValueRange nonSksOperands,
     std::string_view op, SmallVector<std::string> operandTypes) {
-  if (isa<TensorType>(nonSksOperands[0].getType())) {
+  // ToDo: Check again the enum definition
```
asraa, line 248: You should be able to check whether the op str can be symbolized as an enum
```
if (packedAPI && symbolizeTfheRustBoolGateEnum(op).has_value())
```

At least, each enum attr should generate `symbolizeENUMNAME` function that returns a `std::optional`

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -215,9 +245,25 @@ void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
...
+  if (packedAPI && (op == "and" || op == "nand" || op == "or" || op == "nor" ||
```
asraa, line 249: could you please add a small comment describing the output expect this branch to take? i finally realized it was writing normal boolean operations using the packed API haha

Something like "If using the packed API, then emit single boolean operations as ...."

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -215,9 +245,25 @@ void TfheRustBoolEmitter::emitReferenceConversion(Value value) {
...
+    std::string gateStr = std::string(op);
+    std::transform(
+        gateStr.begin(), gateStr.end(), gateStr.begin(),
+        [](unsigned char c) -> unsigned char { return std::toupper(c); });
+
     auto *opParent = nonSksOperands[0].getDefiningOp();
 
+    size_t numberOfOperands = 0;
+
+    if (isa<TensorType>(result.getType())) {
```
asraa, line 260:
```suggestion
    // Handle element-wise boolean gate operations with tensor type operands and results.
    if (isa<TensorType>(result.getType())) {
```

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -542,9 +626,30 @@ LogicalResult TfheRustBoolEmitter::emitType(Type type) {
   return success();
 }
 
+std::pair<std::string, std::string> TfheRustBoolEmitter::checkOrigin(
+    Value value) {
+  std::string prefix = "&";
+  std::string suffix = "";
+  // First check if a DefiningOp exists
+  // if not: comes from function definition
```
asraa, line 634: this can also be checked with `isa<BlockArgument>` (just FYI)

asraa (commented):
Thank you! I left a few initial comments, really great work on unifying the emitter and handling the elementwise cases as well for the packed API. 

Could you please rebase the PR to main and then also add a test case for just emitting some packed operation (not end to end tests) just to test the expected syntax of the emitter?

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -171,7 +201,7 @@ LogicalResult TfheRustBoolEmitter::printOperation(func::FuncOp funcOp) {
 
 LogicalResult TfheRustBoolEmitter::printOperation(func::ReturnOp op) {
   std::function<std::string(Value)> valueOrClonedValue = [&](Value value) {
-    auto suffix = "";
+    auto *suffix = "";
```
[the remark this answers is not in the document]
WoutLegiest: The linter gave this as a suggestion. Probably makes the code a fraction faster

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
```diff
@@ -129,6 +155,10 @@ LogicalResult TfheRustBoolEmitter::printOperation(func::FuncOp funcOp) {
       return funcOp.emitOpError()
              << "Failed to emit tfhe-rs bool type " << arg.getType();
     }
+    // Adapt to the Belfort API
+    if (packedAPI && isa<ServerKeyType>(arg.getType())) {
+      os << "Enum";
+    }
```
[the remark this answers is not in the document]
WoutLegiest: Shouldn't we add an extra type to TfheRustBool to model this key?
Maybe this is the stable way

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
WoutLegiest: Thanks! Was looking for something like this!

lib/Target/TfheRustBool/TfheRustBoolEmitter.cpp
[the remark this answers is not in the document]
asraa: Yeah, might be easier to add the new type and add a new case for `emitType`

asraa (commented):
Thank you! Could you check if you can rebase this / squash into a single commit? (I think the first commit has already been merged)

asraa:
Rebased and squashed, will address the key type comment.

asraa:
OK updated and fixed the server key type in the final commit!

j2kun (approved):

PR #946, opened 2024-08-28
Fix warnings with mismatched integral types

ai-mannamalai:
- Fix warnings with mismatched integral types
- make check-heir continues passing locally
- Still few more instances of warnings are present but in next PR perhaps

### DIFF
--- a/lib/Conversion/MemrefToArith/Utils.cpp
+++ b/lib/Conversion/MemrefToArith/Utils.cpp
@@ -23,7 +23,7 @@ std::optional<std::vector<uint64_t>> materialize(
   affine::AffineValueMap thisMap;
   access.getAccessMap(&thisMap);
   std::vector<uint64_t> accessIndices;
-  for (auto i = 0; i < access.getRank(); ++i) {
+  for (size_t i = 0; i < access.getRank(); ++i) {
     // The access indices of the global memref *must* be constant,
     // meaning that they cannot be a variable access (for example, a
     // loop index) or symbolic, for example, an input symbol.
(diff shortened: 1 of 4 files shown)


asraa (approved):
Thank you!

asraa (commented):
FYI i guess copybara merged this commit with the conflicts after I approved it internally, so I'm resolving why it did that and how to fix head

OK I guess it merged the previous commit before the force push update, and we have another internal PR in for after the force push.... I guess that's what's causing the conflict now.

asraa:
Hey @ai-mannamalai this is what happened - but tl;dr, could you please rebase this PR over main?

```
FYI on HEIR maintenance, want to share a mix-up issue I caused: Muthu had this PR out https://github.com/google/heir/pull/946, and since he's external, it goes through manual approval in gerrit before importing any new commit into internal codebase. His first commit (before he force pushed) created INTERNAL_PR_1, which I LGTM'd, and then I approved his second commit in gerrit, resulting in INTERNAL_PR_2.

My accidental LGTM of the first INTERNAL_PR_1 because the second was stalled in gerrit resulted in (1) the first commit being merged at main on github and (2) his PR still being open because the second force pushed commit wasn't submitted. Now his PR has a merge conflict at main because the first commit was merged.

I THINK 
(1) I should not have LGTM'd it, but also copybara probably shouldn't autosubmit something like this, not sure how to prevent this
(2) To resolve: I think I can have Muthu rebase his PR now over main and submit that. Unfortunately, that's work on Muthu :/ I think any thing I do (revert the commit for example) will require Muthu to do something at this point.
```

ai-mannamalai:
updated (part-2) of this commit

asraa (approved):
Thank you so, so much!

PR #963, opened 2024-09-10
Update CMake for new conversion utilities

ai-mannamalai:
Add CMake to **Linalg to Tensor Ext** and **Tosa to Secret Arith** conversions build in repo.


asraa (approved):
thank you as always!

Issue #954, opened 2024-09-04
heir_translate: `Dialect 'memref' not found for custom op 'memref.alloc'`


johnmatter:
I am trying to create a pipeline to get from StableHLO to OpenFHE by way of TOSA. It's possible I'm making a newbie mistake, so I'll write up my approach.

### Environment
I'm on Ubuntu 24.04. My local build of HEIR is with [commit 90bdae81f3ae7320bd00b1e7a2920e63ff0b2ff6](https://github.com/google/heir/commit/90bdae81f3ae7320bd00b1e7a2920e63ff0b2ff6). There's no particular reason for being on this commit—that's just the last time I `git pull`ed

### Lower StableHLO to TOSA with `stablehlo-opt`
Here is a sample function that might be part of a CNN:
```
$ cat in.stablehlo
func.func @main(
  %image: tensor<28x28xf32>,
  %weights: tensor<784x10xf32>,
  %bias: tensor<1x10xf32>
) -> tensor<1x10xf32> {
  %0 = "stablehlo.reshape"(%image) : (tensor<28x28xf32>) -> tensor<1x784xf32>
  %1 = "stablehlo.dot"(%0, %weights) : (tensor<1x784xf32>, tensor<784x10xf32>) -> tensor<1x10xf32>
  %2 = "stablehlo.add"(%1, %bias) : (tensor<1x10xf32>, tensor<1x10xf32>) -> tensor<1x10xf32>
  %3 = "stablehlo.constant"() {value = dense<0.0> : tensor<1x10xf32>} : () -> tensor<1x10xf32>
  %4 = "stablehlo.maximum"(%2, %3) : (tensor<1x10xf32>, tensor<1x10xf32>) -> tensor<1x10xf32>
  "func.return"(%4): (tensor<1x10xf32>) -> ()
}
```
This mostly lowers to TOSA with `stablehlo-opt -stablehlo-prepare-for-tosa in.stablehlo`

```
$ stablehlo-opt -stablehlo-legalize-to-tosa in.stablehlo > in.tosa
$ cat in.tosa
module {
  func.func @main(%arg0: tensor<28x28xf32>, %arg1: tensor<784x10xf32>, %arg2: tensor<1x10xf32>) -> tensor<1x10xf32> {
    %0 = "tosa.const"() <{value = dense<0.000000e+00> : tensor<1x10xf32>}> : () -> tensor<1x10xf32>
[...]

PR #980, opened 2024-09-20
Do not hoist complex op in secret.generic

ZenithalHourlyRate:
From #977. Cc @asraa

Complex op with regions (such as `affine.for`) should not be extracted before secret.generic; it would result in a whole op being eliminated, put as arg and directly yielded (if this complex op is the last one?); then RemoveUnusedGenericArgs would try to remove them, causing an assertion fault seen in #954 and #977.

* the condition `opRange.size() <= 2` stops the problem to occur for simple programs, such as `simple_sum` in `tests/heir_simd_vectorizer/simple_sum.mlir`
* #961 partially fixes the problem as it stops one kind of plaintext, namely tensor.empty from being collapsed.

The test cases provided will crash heir-opt without this fix applied.

### DIFF
--- a/lib/Dialect/Secret/IR/SecretPatterns.cpp
+++ b/lib/Dialect/Secret/IR/SecretPatterns.cpp
@@ -560,6 +560,10 @@ LogicalResult HoistPlaintextOps::matchAndRewrite(
     if (isa<YieldOp>(op)) {
       return false;
     }
+    // complex op
+    if (op.getNumRegions() != 0) {
+      return false;
+    }
     LLVM_DEBUG(llvm::dbgs()
                << "Considering whether " << op << " can be hoisted\n");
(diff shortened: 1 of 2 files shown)


lib/Dialect/Secret/IR/SecretPatterns.cpp
```diff
@@ -560,6 +560,10 @@ LogicalResult HoistPlaintextOps::matchAndRewrite(
     if (isa<YieldOp>(op)) {
       return false;
     }
+    // complex op
```
asraa, line 563: thank you! a very quick suggestion for a comment here
```suggestion
    // Conservatively preserve a complex op with a nested region
    // This could be a replaced with a recursive call to check that all of the regions' operations can be hoisted.
```

asraa (approved):
Thank you!

PR #989, opened 2024-09-26
docs: add ckks to dialects diagram

ZenithalHourlyRate:
As CKKS dialect and its conversion are added recently, the overall architecture diagram should be updated.

Lowering to standard is only supported for bgv for now and it is reflected in the diagram.


asraa (approved):
wonderful, thank you! i appreciate this a lot

Issue #974, opened 2024-09-17
tensor: implement a tensor insertop folder


asraa:
HEIR SIMD vectorizer in the RLWE pipeline fully unrolls programs with tensor semantics. This can result in some sequences of redundant tensor insert operations. For example, in a dot product in a snippet of a program looks like:
```
    %extracted = tensor.extract %arg1[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
    %extracted_255 = tensor.extract %arg2[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
    %0 = openfhe.make_ckks_packed_plaintext %arg0, %cst_254 : (!openfhe.crypto_context, tensor<16xf64>) -> !lwe.rlwe_plaintext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>, underlying_type = f32>
[...]

PR #991, opened 2024-09-26
[pass-status] tensor.extract ops are stil there

ahmedshakill:
implements #974


ahmedshakill:
Hi @asraa 
At this stage the pass is not producing the expected result. Need to improve the forwarding logic.

lib/Transforms/ForwardInsertToExtract/ForwardInsertToExtract.cpp
```diff
@@ -0,0 +1,124 @@
...
+          [&](auto insertOp) { return insertOp.getScalar(); })
+      .Default([&](Operation &) { return failure(); });
+}
+
+LogicalResult ForwardSingleInsertToExtract::matchAndRewrite(
+    tensor::ExtractOp extractOp, PatternRewriter &rewriter) const {
+  LLVM_DEBUG(llvm::dbgs() << "Considering extractOp for replacement: "
+                          << extractOp << "\n");
+
+  for (Operation *use : extractOp.getTensor().getUsers()) {
```
asraa, line 92: Thank you so much for starting this pass! 

I actually think the only issue is from this line. I ran this pass with `--debug` on the testcase you added and noticed that the extract operation that could be replaced wasn't picking up the forwardable insert operation:
```
  * Pattern mlir::heir::ForwardSingleInsertToExtract : 'tensor.extract -> ()' {
Trying to match "mlir::heir::ForwardSingleInsertToExtract"
Considering extractOp for replacement: %extracted_2 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
Considering use for forwarding: %inserted_3 = tensor.insert %5 into %inserted[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
insert op does not dominate extract op
Use is not forwardable: %inserted_3 = tensor.insert %5 into %inserted[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
Considering use for forwarding: %extracted_2 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
insert op does not dominate extract op
Use is not forwardable: %extracted_2 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = <coefficientType = i32, coefficientModulus = 463187969 : i32, polynomialModulus = <1 + x**16>>>, underlying_type = f32>>
"mlir::heir::ForwardSingleInsertToExtract" result 0
```

This is because a `tensor.insert` operation (`%inserted = tensor.insert %2 into %arg2[%c0, %c0] : tensor<1x16x!lwe.rlwe_ciphertext<>>
` produces a *new* tensor `%inserted`, which means that the only operations picked up by this statement are the operations that use the newly inserted tensor. So I think this logic could be very simple - you might just want to check if `extractOp.getTensor().getDefiningOp<tensor::InsertOp>` exists and then replace with the scalar.

ahmedshakill: These **good first issue**s are fun :blush:

ahmedshakill:
Now it gets the job done but not sure if it is ready to handle all cases.

lib/Transforms/ForwardInsertToExtract/ForwardInsertToExtract.cpp
```diff
@@ -0,0 +1,130 @@
...
+        for (auto currentNode = insertOp->getNextNode();
+             currentNode != extractOp.getOperation();
+             currentNode = currentNode->getNextNode()) {
+          if (currentNode->getNumRegions() > 0) {
+            LLVM_DEBUG(llvm::dbgs() << "an op with control flow is between the "
+                                       "insert and extract op\n");
+            return false;
+          }
+
+          if (auto op = dyn_cast<tensor::InsertOp>(currentNode)) {
```
asraa, line 62: I was going to say that `tensor.insert_slice` and `tensor.parallel_insert_slice` should also be guarded here (https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert_slice-tensorinsertsliceop), but I actually realize that since each of these operations returns a *new* tensor, that even this branch of code would never hit. For example, 

```
%inserted = tensor.insert %c0 into %arg[%c0_index]
%inserted_1 = tensor.insert %c1 into %arg[%c0_index]
%extract = tensor.extract %inserted[%c0_index]
```

would still mean that we can forward the %c0 scalar value to replace %extract.

asraa (changes requested):
Thank you! I think it would be good to add simple test cases for:
* Trying to forward a tensor.extract from a function argument (this would hit the `def == nullptr` condition in your code)
* Forwarding with an insert between
* Forwarding with an operation with a region in between (e.g. a for loop)
* Successful forwarding
* Two extracts in a row both forwarded values

lib/Transforms/ForwardInsertToExtract/ForwardInsertToExtract.cpp
asraa: I know! Usually they sound easy and then there's something sneaky in there that I didn't anticipate either :P But at least we all learn

lib/Transforms/ForwardInsertToExtract/ForwardInsertToExtract.cpp
ahmedshakill: > %inserted = tensor.insert %c0 into %arg[%c0_index]
> %inserted_1 = tensor.insert %c1 into %arg[%c0_index]
> %extract = tensor.extract %inserted[%c0_index]

used this as 'Forwarding with an insert between' test

tests/forward_insert_to_extract/forward_insert_to_extract.mlir
```diff
@@ -0,0 +1,162 @@
...
+  %2 = openfhe.add %arg0, %extracted_0, %1 : (!cc, !ct, !ct) -> !ct
+
+  %a2 = openfhe.add %arg0, %extracted_0, %2 : (!cc, !ct, !ct) -> !ct
+
+  %inserted = tensor.insert %2 into %arg2[%c0, %c0] : tensor<1x16x!ct>
+  %inserted_1 = tensor.insert %a2 into %arg1[%c0, %c0] : tensor<1x16x!ct>
+
+    //  CHECK-NOT: tensor.extract %[[INSERTED0]]
+  %extracted_2 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!ct>
+  return %extracted_2 : !ct
```
asraa, line 97: Sorry for the very small nit, but you might want to check which insert value was forwarded here

```suggestion
  // CHECK-NEXT: %[[VALA2:.*]] = openfhe.add %[[ARG0]], %[[EXTRACTED0]], %[[VAL2]]
  %a2 = openfhe.add %arg0, %extracted_0, %2 : (!cc, !ct, !ct) -> !ct

  %inserted = tensor.insert %2 into %arg2[%c0, %c0] : tensor<1x16x!ct>
  %inserted_1 = tensor.insert %a2 into %arg1[%c0, %c0] : tensor<1x16x!ct>

    //  CHECK-NOT: tensor.extract
  %extracted_2 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!ct>
  // CHECK: return %[[VALA2]]
  return %extracted_2 : !ct
```

tests/forward_insert_to_extract/forward_insert_to_extract.mlir
```diff
@@ -0,0 +1,162 @@
...
+  %2 = openfhe.add %arg0, %extracted_0, %1 : (!cc, !ct, !ct) -> !ct
+
+  %inserted = tensor.insert %2 into %arg2[%c0, %c0] : tensor<1x16x!ct>
+
+  //  CHECK-NOT: %[[EXTRACTED1:.*]] = tensor.extract %[[INSERTED0]]
+  %extracted_1 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!ct>
+  //  CHECK-NOT: %[[EXTRACTED2:.*]] = tensor.extract %[[INSERTED0]]
+  %extracted_2 = tensor.extract %inserted[%c0, %c0] : tensor<1x16x!ct>
+
+  %3 = openfhe.add %arg0, %extracted_1, %extracted_2 : (!cc, !ct, !ct) -> !ct
```
asraa, line 160:
```suggestion
  // CHECK: openfhe.add %[[ARG0]], %[[VAL2]], %[[VAL2]]
  %3 = openfhe.add %arg0, %extracted_1, %extracted_2 : (!cc, !ct, !ct) -> !ct
```

to check both uses had the right forward value!

asraa (approved):
Thank you! Just two minor test comments

ai-mannamalai:
@ahmedshakill - could you also add a CMakeLists.txt build file in the pattern of the other sub-folders in lib/Transforms ?

asraa (approved):

asraa:
There's a couple of small test failures that can mostly be fixed by removing undefined substitutions in the LIT test - can you ensure the tests pass and squash? Thank you so so much

asraa (approved):

AlexanderViand-Intel:
> @asraa Sorry I forgot to squash. I see a failed check. Maybe I did something wrong.

If you see "Google internal checks FAILED" on `feedback/copybara` but the main `Build and Test / build-and-test` passes fine, then it's usually something Google needs to fix internally rather than something you did wrong :)

asraa:
> If you see "Google internal checks FAILED" on the feedback/copybara action but the main Build and Test / build-and-test passes fine, then it's usually something Google needs to fix internally rather than something you did wrong :)

Oh yeah! I fixed it internally and just waiting on another +1 internally. (Usually it's a build error incompatible with some extra build requirements internally)

Issue #990, opened 2024-09-26
polynomial-to-standard: error on integer width for poly.add when ring coefficientModulus width is less than coefficient width


ZenithalHourlyRate:
When playing around `tests/bgv/to_polynomial.mlir` with additional `-convert-elementwise-to-affine -polynomial-to-standard`, error happened

```
error: 'arith.extsi' op operand type 'tensor<1024xi32>' and result type 'tensor<1024xi26>' are cast incompatible
    %mul = bgv.mul %x, %y : (!ct1, !ct1) -> !ct2
           ^
note: see current operation: %19 = "arith.extsi"(%13) : (tensor<1024xi32>) -> tensor<1024xi26>
```

Reduced to a minimal working example with argument `-polynomial-to-standard`, where changing the line to `!p = !p2` resolves the problem:

```mlir
#my_poly = #polynomial.int_polynomial<1 + x**1024>
#ring1 = #polynomial.ring<coefficientType = i32, coefficientModulus = 33538049 : i32, polynomialModulus=#my_poly>
#ring2 = #polynomial.ring<coefficientType = i25, coefficientModulus = 33538049 : i25, polynomialModulus=#my_poly>

!p1 = !polynomial.polynomial<ring = #ring1>
!p2 = !polynomial.polynomial<ring = #ring2>

!p = !p1

module {
  func.func @polymul(%x : !p, %y : !p) -> (!p) {
    %add = polynomial.add %x, %y : !p
    return %add : !p
  }
}
```

The related code is the following

https://github.com/google/heir/blob/b79bccf6bf3863083e339bd582b4e06fdc3dd561/lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp#L522-L543

It assumes either coefficientModulus is a power-of-two or its width is the same as coefficientType.

Maybe we should check that coefficientType has the same width as coefficientModulus if it is a prime.

Also, should we consider lowering it to mod_arith so that polynomial-to-standard can be simplified? Currently extsi/extui is everywhere.

asraa:
cc @AlexanderViand @inbelic on this issue

> Also, should we consider lowering it to mod_arith so that polynomial-to-standard can be simplified? 

I think so - but I will need to catch up on this particular pass pipeline.

AlexanderViand-Intel:
Thanks for bringing this up! I’ll add it to the list of “issues that show why we really need formal verification around mod_arith stuff”! 

[...]

Issue #749, opened 2024-06-27
Fix upstream polynomial canonicalization rules


j2kun:
Only apply if there is no coefficient modulus, otherwise we need arith_ext

ZenithalHourlyRate:
If I understand it correctly, is https://github.com/llvm/llvm-project/blob/30f5a3ca150e98d482abc6a4d0e3fe6c12f77695/mlir/lib/Dialect/Polynomial/IR/PolynomialCanonicalization.td#L29-L33 the only issue here?

asraa:
I think the ones introducing arith ops (https://github.com/llvm/llvm-project/blob/30f5a3ca150e98d482abc6a4d0e3fe6c12f77695/mlir/lib/Dialect/Polynomial/IR/PolynomialCanonicalization.td#L59-L64, and the corresponding sub one) are also incorrect in the case where there's a modulus or the coefficients might wrap around the underlying data type. For example, converting `poly.add(intt(p), intt(q))` to `intt(arith.add(p, q))` would be incorrect if `arith.add(p, q)` will overflow mod q or mod the data type, since arith.add doesn't take the modulus into account.

@AlexanderViand has this nice doc! https://docs.google.com/document/d/1aS-rSZP3GJvzemoeAG8ingAyyLUNq2IB5Rgqksilbdk/edit

I think the upstream passes would need to fail to apply when there is a modulus or wrap around, and i suppose we can write our HEIR internal canonicalization with mod_arith in this repo

ZenithalHourlyRate:
After careful inspection, I now think the canonicalization rule SubAsAdd is OK with modulus ; I was thinking that coefficients should in range [0, +mod) (which means -1 needs to be mod-1 instead of 2^32-1), but now I understand it is in (-mod, +mod), so directly `polynomial.mul_scalar -1` is OK and won't overflow; given that the **semantic** of polynomial.mul_scalar is correct.

Checking the lowering of polynomial.mul_scalar confused me; It may result in wrong result; it emits `arith.muli` then `arith.remsi`; `arith.muli` may overflow if the scalar is large enough; `remsi` fortunately avoids the issue in #928. I again think these lowering should be handled in mod_arith.

> Only apply if there is no coefficient modulus, otherwise we need arith_ext

However, for the case of NTT/iNTT, there must be coefficientModulus otherwise primitiveRoot wont exist, so the fix proposed in https://github.com/llvm/llvm-project/pull/110318 is removing them for now in the upstream.

[...]

PR #995, opened 2024-09-27
Lower polynomial to mod_arith for add/sub/mul_scalar

ZenithalHourlyRate:
This is a draft and not cleaned up yet (e.g. BUILD/Cmake), just demoing the concept of lowering polynomial to mod_arith. Some function like mod the polynomial (`PolynomialToStandard::generateOpImplementations`) is not included yet.

Related to #990 and #749

Should we add a separate pass for it (`polynomial-to-mod-arith`, lots of duplicate code) and gradually replace `polynomial-to-standard` or just placing it inside the `polynomial-to-standard` pass?

Extra note: with `mod-arith-to-arith` the result for the test case is elegant

```
  func.func @polysub(%arg0: tensor<4xi32>, %arg1: tensor<4xi32>) -> tensor<4xi32> {
    %c-1_i32 = arith.constant -1 : i32
    %cst = arith.constant dense<-1> : tensor<4xi32>
    %0 = arith.muli %arg1, %cst : tensor<4xi32>
    %cst_0 = arith.constant dense<7681> : tensor<4xi32>
    %1 = arith.remui %0, %cst_0 : tensor<4xi32>
    %2 = arith.addi %arg0, %1 : tensor<4xi32>
    %cst_1 = arith.constant dense<7681> : tensor<4xi32>
    %3 = arith.remui %2, %cst_1 : tensor<4xi32>
    return %3 : tensor<4xi32>
  }
```


ZenithalHourlyRate:
Turned out the lowering above is incorrect. `mod_arith` requires operands to be signless, whereas `polynomial` coefficient is signed

* SubAsAdd canonicalization rule explicitly asks it to be signed (though when mod is a power-of-two > coeffType, signed and unsigned are the same)
* polynomial.FromTensorOp inteprets the input as signed, implicitly. We should define it.
* polynomial-to-standard currently uses remsi, and range keeping is complex (check https://github.com/google/heir/blob/6fd0f1bb23732756f0b7cda8e5be3041b490a9db/lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp#L621-L638)

I agree that any signed integer can be coefficient, for the semantic of polynomial dialect (e.g. -1 or a number inside [mod, 2^31-1)). However we should keep some invariant internally. AFAIK the currently lowering (like ConvertAdd, ConvertMul) tries to keep the invariant that coefficients for polynomial are in (-mod, +mod), however, ConvertFromTensor does not keep the invariant as user may give any input and it did not remsi it.

Or should we keep it in [-mod/2, +mod/2) (for L2/L-infinity norm/noise reason, commonly seen in FHE paper)

In the definition of the polynomial dialect there should be a canonical form of the polynomial, like it has been mod coeffModulus and PolyModulus, then we can have a stable output for ToTensor (actually if it is in (-mod, +mod), we have two possible output for one number) and we can have safely lower to mod_arith in some way.

I'm curious what other FHE libraries deal with this issue.

AlexanderViand-Intel:
> Should we add a separate pass for it (`polynomial-to-mod-arith`, lots of duplicate code) and gradually replace `polynomial-to-standard` or just placing it inside the `polynomial-to-standard` pass?

What we’ve done in the past is to update the existing pipeline “in-place”, so we’d (a) create a pass `polynomial-to-modarith` (b) createa a `modarith-to-standard` pass, and finally (c) switch `polynomial-to-standard` from being a Pass to being a pipeline consisting of  the two passes above.

ZenithalHourlyRate:
Tried to keep the invariant inside FromTensor using mod_arith.reduce, leading to the following behavior if the modulus is large.

```
    %cst = arith.constant dense<2> : tensor<1024xi32>
    %0 = arith.extsi %cst : tensor<1024xi32> to tensor<1024xi33>
    // 2 ** 32 - 1
    %1 = mod_arith.reduce %0 {modulus = 4294967295 : i64} : tensor<1024xi33>
    %2 = arith.trunci %1 : tensor<1024xi33> to tensor<1024xi32>
```

ZenithalHourlyRate:
https://mlir.llvm.org/docs/Dialects/PolynomialDialect/#ringattr

I can not handle the case of coefficient modulus larger than the underlying type (e.g. i32). It has two cases.

* modulus is a power-of-two like `2**33` and `2 ** 64`, which is the same as modulus = 0, then why should we specify a modulus like this. We do have [examples](https://github.com/google/heir/blob/4ea84ba8657f0ecbd2d645f5d5f2794a1e086ff7/tests/polynomial/lower_mul_scalar.mlir#L3) inside heir like this, which is caused by the [polynomial-to-standard implementation](https://github.com/google/heir/blob/4ea84ba8657f0ecbd2d645f5d5f2794a1e086ff7/lib/Conversion/PolynomialToStandard/PolynomialToStandard.cpp#L516-L523) and I think it should be fixed.
* modulus is another number like `2 ** 32 + 1` or a prime of 64 bits. This also has ill semantic as the underlying type can not fully represent all elements and the arithmetic is not what we want because two modulo operations happen (mod 2**32) and (mod p).

Currently the documentation upstream does not define any constraint on this and it does not have a verifier for it.

j2kun:
I think one complication is that MLIR does not specify any overflow semantics by default. So omitting a modulus that equals 2^32 when the type is i32 could result in lowerings that treat overflow differently to how we want. I thought I could handle that by allowing the type of the modulus to differ from the underlying type of the coefficients, and then lowerings could check for these edge cases. But since I'm not working for another month (on baby leave), I won't have the time to dive deep to recommend a fix.

ZenithalHourlyRate:
I think the lowering for add/sub/mul_scalar, along with auxiliary op from_tensor/constant are self-contained enough for review. #990 can be fixed by this PR.

Each of mul, ntt, and intt is large enough for a separate PR, and including them all in this one would make the review process more burdensome.

Things need discussion:

* behavior change of from_tensor/constant as the result tensor would get reduced to [0, cmod)
  + this may makes consequent from_tensor -> to_tensor -> from_tensor -> to_tensor inefficient; maybe canoicalization can handle this
  + or we use other op like from_tensor_raw or from_tensor_reduced
* wierd semantic of the last case of these runner tests involving large prime (note that runner test use `--heir-polynomial-to-llvm`)
* The canonicalization rule SubAsAdd mentioned in #749 is now correct, though we also have lowering for poly.sub now; which way should we go.

ZenithalHourlyRate:
Seems that I need to put `elementwise-to-affine` before `polynomial-to-mod-arith` as mentioned in #769

AlexanderViand-Intel:
> Seems that I need to put `elementwise-to-affine` before `polynomial-to-mod-arith` as mentioned in #769

Yeah, I think that's a good idea. Not many things use the tensor-valued versions of polynomial, but the lowerings from BG/CKKS make use of them as it's very convenient there.

ZenithalHourlyRate:
Rebased to the current directory organization

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/to_mod_arith/lower_add.mlir
AlexanderViand-Intel, line 1: This and it's sibling file are in the wrong location (should be `tests/Dialect/Polynomial/Conversions/polynomial_to_mod_arith`)

tools/heir-opt.cpp
```diff
@@ -205,6 +206,9 @@ void convertToDataObliviousPipelineBuilder(OpPassManager &manager) {
 void polynomialToLLVMPipelineBuilder(OpPassManager &manager) {
   // Poly
   manager.addPass(createElementwiseToAffine());
+  manager.addPass(
+      ::mlir::heir::polynomial::to_mod_arith::createPolynomialToModArith());
+  manager.addPass(mod_arith::createModArithToArith());
   manager.addPass(::mlir::heir::polynomial::createPolynomialToStandard());
```
AlexanderViand-Intel, line 212: Do we still need `PolynomialToStandard` here?

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.td
```diff
@@ -0,0 +1,25 @@
...
+
+include "mlir/Pass/PassBase.td"
+include "lib/Dialect/ModArith/IR/ModArithDialect.td"
+
+def PolynomialToModArith : Pass<"polynomial-to-mod-arith", "ModuleOp"> {
+  let summary = "Lower `polynomial` to `mod-arith`.";
+
+  let description = [{
+    This pass lowers the `polynomial` dialect to `mod-arith`, with
+    affine and tensor.
```
AlexanderViand-Intel, line 12: I think this pass deserves a bit more detailed documentation.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+#include "mlir/include/mlir/IR/Visitors.h"               // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"     // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace polynomial {
+namespace to_mod_arith {
```
AlexanderViand-Intel, line 54: I think there's a few too many namespaces happening here. First, note that that `polynomial` lives upstream, so it'd be just `::mlir::polynomial` (no `heir`), but more importantly, I think I saw a comment from  @asraa on namespace conventions on another PR, and I guess we should decide whether or not conversions live in their starting dialect's namespace?

AlexanderViand-Intel (commented):
Thanks for putting this together! I think the final output looks good, though I'm not a huge fan of how much "modular arithmetic" logic this lowering includes.  For example, the pass should, imho, not care about things such as power-of-two moduli - worrying about how to efficiently realize modular arithmetic is exactly what `ModArith` was introduced for, so this logic should be moved to `ModArithToArith`. 

However, I think working code beats no code, so I'd be in favor of merging this and creating an issue for refactoring this out. Would be interested to hear @inbelic 's opinion on this, too.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
asraa: Weirdly enough I don't remember making a namespace comment on another PR but had a pending review on exactly this point.

In upstream MLIR, conversions like affine-to-standard or arith-to-llvm belong in the general mlir namespace, so either (1) we use the general heir namespace or (2) put it in the starting dialect - i sort of prefer the starting dialect to keep the folder's organization slightly aligned with the namespace.

tools/heir-opt.cpp
ZenithalHourlyRate: `PolynomialToModArith` only handles add/mul_scalar/from_tensor/to_tensor/constant, we still need `PolynomialToStandard` to handle mul/ntt/intt/monomial things

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
ZenithalHourlyRate: > First, note that that polynomial lives upstream, so it'd be just ::mlir::polynomial (no heir)

I followed the `PolynomialToStandard` impl which lives in `::mlir::heir::polynomial`. The further `to_mod_arith` is to avoid name conflict with `PolynomialToStandard` (e.g. both having ConvertFromTensor/ConvertToTensor). I think as the downstream lowering is not upstreamed, living in the `heir` namespace is reasonable; also, putting all them in `::mlir::polynomial` may leads to possible name conflict as the upstream changes (though we can get informed of it)

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/to_mod_arith/lower_add.mlir
ZenithalHourlyRate: Moved. Also moved unit tests related to `polynomial-to-standard` to corresponding folder. `heir_polynoimal_to_llvm` now only contains pipeline tests.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.td
ZenithalHourlyRate: Details added.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+    rewriter.replaceOp(op, constantOp);
+    return success();
+  }
+};
+
+template <typename Op, typename ModArithOp, typename ArithIOp,
+          typename ArithFOp>
+Value modArithBinaryOp(ConvertCommon<Op> &c, ImplicitLocOpBuilder &b, Value lhs,
+                       Value rhs) {
+  if (c.coefficientModulusIsNotNatural) {
```
inbelic, line 357: I think it would be best if we didn't even need to check this in this lowering and it should just always lower to `mod_arith.[add|sub|mul]`. Then we can check all that stuff there. As part of that we should also extend the `mod_arith` ops to include handling of floating point types. I think this is the same as what @AlexanderViand-Intel  is saying. So we could leave this as is and then file the following:

    - Extend `mod_arith` ops to support floating-point values as well.
    - Move this logic to the `mod_arith` to `arith` lowering.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+  LogicalResult matchAndRewrite(
+      MulScalarOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ConvertCommon c(op, adaptor, this->getTypeConverter());
+
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    Value scalar = adaptor.getScalar();
+
+    scalar = modArithReduceOp<MulScalarOp>(c, b, rewriter, scalar);
```
inbelic, line 391:
```suggestion
    Value scalar = modArithReduceOp<MulScalarOp>(c, b, rewriter, adaptor.getScalar());
```

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+    Type reduceType = shape ? Type(reduceTensorType) : Type(reduceIntegerType);
+
+    v = b.create<arith::ExtSIOp>(reduceType, v);
+    v = b.create<mod_arith::ReduceOp>(reduceType, v, c.coefficientModulus);
+    v = b.create<arith::TruncIOp>(resultType, v);
+  } else if (c.coefficientModulusWidth != 0 &&
+             c.resultElementWidth > c.coefficientModulusWidth) {
+    v = b.create<mod_arith::ReduceOp>(resultType, v, c.coefficientModulus);
+  }
+  // else float or natural power of two modulus larger than underlying type.
```
inbelic, line 227: Why don't we need the op if it is a float? Should we really reduce it but can't because it doesn't support floats at the moment?

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+      inputTensorType = cast<RankedTensorType>(inputTypeConverted);
+      inputShape = inputTensorType.getShape()[0];
+      inputElementType = inputTensorType.getElementType();
+      inputElementWidth = inputElementType.getIntOrFloatBitWidth();
+      inputElementIntegerType = dyn_cast<IntegerType>(inputElementType);
+      inputElementFloatType = dyn_cast<FloatType>(inputElementType);
+    }
+  }
+
+  // polynomial type related
```
inbelic, line 165: We don't use some of the following variables outside of the above function. Do we anticipate using them in the future? Otherwise I think it would best to keep them local.

inbelic (commented):
Thanks for contributing this and lots of great work here!
Mostly nits and preferential comments.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
ZenithalHourlyRate: for input* (e.g. inputTensor, inputElementWidth) it is indeed FromTensor specific and now moved to ConvertFromTensor. Some other unused variable are also eliminated, and I think the current variables are essential and may be used in the future.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
ZenithalHourlyRate: The polynomial dialect does not have definition for `coefficientModulus` when `coeffcientType` is a float

```
if the coefficient type is integral, whose coefficients are taken modulo some statically known modulus (coefficientModulus).
```

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
ZenithalHourlyRate: Changed

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
ZenithalHourlyRate: I have now migrated the logic of natural modulus handling inside `mod_arith`. I agree with the idea of handling all the integer stuff in `mod_arith`. But I think handling float point inside `mod_arith` seems a violation of its name.

Also, the `PolynomialToModArith` code must contain some float point specific thing, like `ConvertConstant` where `FloatAttr` is explicitly asked.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+
+    constantOp = modArithReduceOp(c, b, rewriter, constantOp, c.resultShape);
+
+    rewriter.replaceOp(op, constantOp);
+    return success();
+  }
+};
+
+template <typename Op, typename ModArithOp, typename ArithIOp,
+          typename ArithFOp>
```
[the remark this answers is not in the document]
ZenithalHourlyRate: SI will add it in the next round of push.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+  unsigned inputElementWidth;
+  IntegerType inputElementIntegerType;
+  FloatType inputElementFloatType;
+};
+
+template <typename Op>
+Value modArithReduceOp(ConvertCommon<Op> &c, ImplicitLocOpBuilder &b,
+                       ConversionPatternRewriter &rewriter, Value v,
+                       int64_t shape = 0) {
+  // why not c.resultTensorType here
```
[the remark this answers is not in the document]
ZenithalHourlyRate: I rephrased this comment so it is less confusing now.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+  unsigned coefficientTypeWidth;
+  IntegerType coefficientIntegerType;
+  FloatType coefficientFloatType;
+
+  IntegerAttr coefficientModulus;
+  IntPolynomialAttr polynomialModulus;
+
+  APInt coefficientNaturalModulus;
+  APInt coefficientModulusValue;
+  bool coefficientModulusIsNotNatural;
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Removed.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -0,0 +1,456 @@
...
+    addTargetMaterialization([](OpBuilder &builder, Type type,
+                                ValueRange inputs, Location loc) -> Value {
+      return builder.create<ToTensorOp>(loc, type, inputs[0]).getResult();
+    });
+  }
+};
+
+// boilerplate code in one place
+template <typename SourceOp>
+struct ConvertCommon {
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Renamed.

ZenithalHourlyRate (commented):
Talking about moving some logic inside `mod_arith`. I think the handling of `modArithReduceOp` needs some change to `mod_arith.reduce`.

[Note this will interpret x as a signed integer. It is required the bitwidth of q is smaller than that of x](https://heir.dev/docs/dialects/modarith/#mod_arithreduce-heirmod_arithreduceop). The `smaller` requirement makes the lowering pass needs to manually `extsi` the input. Should we allow q (intepreted as unsigned) whose bitwidth is the same as x (signed). This semantic is wierd.

AlexanderViand-Intel:
> Talking about moving some logic inside `mod_arith`. I think the handling of `modArithReduceOp` needs some change to `mod_arith.reduce`.
> 
> [Note this will interpret x as a signed integer. It is required the bitwidth of q is smaller than that of x](https://heir.dev/docs/dialects/modarith/#mod_arithreduce-heirmod_arithreduceop). The `smaller` requirement makes the lowering pass needs to manually `extsi` the input. Should we allow q (intepreted as unsigned) whose bitwidth is the same as x (signed). This semantic is wierd.


I think it'd be a good idea to have a general `mod_arith.reduce` operation that has absolutely no preconditions on the inputs and will fix things as needed as part of the lowering. So I'd be happy with moving the `extsi` to the -to-arith lowering of `mod_arith.reduce`

ZenithalHourlyRate:
> I think it'd be a good idea to have a general `mod_arith.reduce` operation that has absolutely no preconditions on the inputs and will fix things as needed as part of the lowering. So I'd be happy with moving the `extsi` to the -to-arith lowering of `mod_arith.reduce`

Now `mod_arith.reduce` can handle modulus with the same bitwidth as underlying type and the natural modulus. Corresponding modArithReduceOp has been removed.

ZenithalHourlyRate:
> Oh no! I did not realize this PR was in flight when I was working on migrating the polynomial-to-standard pass to use `mod_arith` types (which is now almost done, although I think I still have a lot of cleanup work once tests are passing). Is there a particular part of this PR that you think would be worth incorporating into that work?

I think most of them have been covered by your PR. One thing worth noting is that I used a `CommonConversionInfo` as a helper to simplify type manipulation, such that each ConvertXXX class can care less about them and focus on constructing lowered operations itself.

Issue #817, opened 2024-07-23
Investigate if/how "Verifying Peephole Rewriting In SSA Compiler IRs" can be used in HEIR


AlexanderViand-Intel:
As some of you might be aware a group of researchers around Prof. Tobias Grosser from the University of Edinburgh/Cambridge University have been using (an earlier version of) our `polynomial` dialect as an example for their work on verifying MLIR in [LEAN ](https://lean-lang.org/).

I'm happy to report that their paper[^1] has been accepted at ITP 2024 and is now available: https://doi.org/10.4230/LIPIcs.ITP.2024.9 or https://arxiv.org/abs/2407.03685 🎉 

While this is already a cool result in itself, the natural question would be to see if we can somehow make use of this in HEIR (or MLIR, if it concerns up streamed `polynomial` things). For example, I wonder if something like this could have helped us catch #749 and/or might be useful as we go into more advanced modular arithmetic optimizations and transitions between poly and mod arith. The paper also implements reasoning about bit vectors, so there's a good chance it could capture such lowerings/rewrites.

[^1]: Siddharth Bhat, Alex Keizer, Chris Hughes, Andrés Goens, and Tobias Grosser. 
Verifying Peephole Rewriting in SSA Compiler IRs.
In 15th International Conference on Interactive Theorem Proving (ITP 2024). Leibniz International Proceedings in Informatics (LIPIcs), Volume 309, pp. 9:1-9:20, Schloss Dagstuhl – Leibniz-Zentrum für Informatik (2024)
https://doi.org/10.4230/LIPIcs.ITP.2024.9

AlexanderViand-Intel:
Adding examples of issues/bugs that motivate why formal verification might also be interesting for `polynomial`->(`mod_arith`->)`arith` lowerings:

* PR #731 needed some fixing before merging, because I was not aware of `arith` restrictions on `remui/remsi`, meaning that we cannot fit, e.g., numbers modulo $2^{31}-1$ into `i31` (artificial example) and need an extra bit to ensure things work as expected when lowered to LLVM.
* PR #928 fixes an issue where "the current implementation would have incorrect behaviour if the arith.sub resulted in a negative signed number that was then interpreted as an unsigned int with arith.remui" which "is fixed by adding the modulus to the result before applying remui such that the range after sub shifts from  (-mod, mod) -> [0, mod)"

Maokami:
[...]

PR #998, opened 2024-09-29
mod_arith.reduce: require modulus width to be smaller than underlying type width

ZenithalHourlyRate:
The following code errors when lowering with `mod-arith-to-arith`

```mlir
func.func @test() {
  %x = arith.constant 3 : i8
  %1 = mod_arith.reduce %x { modulus = 217 } : i8
  return
}
```

```
error: 'arith.remsi' op requires the same type for all operands and results
  %1 = mod_arith.reduce %x { modulus = 217 } : i8
```

#927 suggested larger modulus is not liked, and its function implementation also implied that. However, functions like verifier and CastModulusAttributeReduce did not work as intended.

We have another interesting case here

```mlir
func.func @test_lower_rep() {
  %x = arith.constant 3 : i8
  %1 = mod_arith.reduce %x { modulus = 128 } : i8
  return
}
```
gets lowered to
```mlir
module {
  func.func @test_lower_rep() {
    %c3_i8 = arith.constant 3 : i8
    %c-128_i8 = arith.constant -128 : i8
    %0 = arith.remsi %c3_i8, %c-128_i8 : i8
    %1 = arith.addi %0, %c-128_i8 : i8
    %2 = arith.remui %1, %c-128_i8 : i8
    return
  }
}
```
Though `arith.constant -128` seems scary, it is correct as it is -128 when signed and 128 when signed, so `remsi 3, -128` and `remui 3, 128` works.

### Another possiblitiy

I have another version that supports larger modulus (accidentally done when I did not see the reduceOp), though the result is really funny, check this example https://github.com/ZenithalHourlyRate/heir/blob/c186103e156a186bef1adfdd9c7e16e6097c50e9/tests/mod_arith/runner/lower_rep.mlir#L24.


asraa:
cc @inbelic @AlexanderViand-Intel in case you have any feedback - 

The implementation and test runner (thanks!) look good to me, but I just want to double check if there's any implications around this that I'm missing

inbelic:
Thanks for catching this and correcting, these address the review comments that @j2kun had left on the original pr. Apparently the changes addressing the comments are still local to my travel laptop... and I didn't actually push them? That is my bad, apologies.

I will check my local changes when I get home to add what else we can fix while we are. But of the top of my head, I believe we wanted to implement the modulus attribute to be user-typed. Does specifying the modulus type fix the issue? ie. `modulus = 128 : i32`. There should also be a testcase added to properly check that using `CastModulusAttributeReduce` does not increase the element bitwidth in cases that an add would.

ZenithalHourlyRate:
> Does specifying the modulus type fix the issue? ie. `modulus = 128 : i32`

If you are referring to the display issue, no. The lowering [here](https://github.com/google/heir/blob/4ea84ba8657f0ecbd2d645f5d5f2794a1e086ff7/lib/Conversion/ModArithToArith/ModArithToArith.td#L160) uses a bitwidth (note the getActiveBits, so APInt is no help here) the same as the underlying type i8 instead of i32.

inbelic:
Here is the correct reduce op commit that should have been merged: https://github.com/google/heir/compare/main...inbelic:heir:inbelic/reduce-fixed.

I think the only thing missing from that fix is to add a verification that the modulus is not negative, and the corresponding tests. And to not use a redundant negation here: https://github.com/google/heir/commit/6751e094e709a16311cb79760d7c5b61c61fe902#diff-22104e8e4148ba3a95719aafe57f6d258e4a98e464e39cd9ae06ffc22f3c9c5bR24

inbelic:
Perfect thanks. LGTM

AlexanderViand-Intel (approved):
LGTM (primarily since it looks good to @inbelic ;)

(I'm adding this to the list of "reasons why we really need formal verification" in #817 because I wish we didn't have to think about this kind of stuff xD )

asraa (commented):
Thank you!

PR #1001, opened 2024-10-02
adds a devcontainer.json

AlexanderViand-Intel:
Another “quality of life” thing for getting people started with HEIR quickly.
This adds a .devcontainer json, so that code spaces created from HEIR have (almost) everything set-up already, basically saving people the working of going through the getting-started setup.

Currently, this still requires running a full bazel build + hedron_compile_commands after the container is created, as adding that to the container creation would make launching a codespace really slow. If we enable pre-built containers, this would not be an issue, so we could have it come pre-built + clangd-ready, too.

If someone wants to contribute back, they’d of course still need to manually create a fork and add it as a new remote, but I think that’s inevitable.

### DIFF
--- a/.devcontainer/cmake/devcontainer.json
+++ b/.devcontainer/cmake/devcontainer.json
@@ -1,0 +1,56 @@
+{
+  "hostRequirements": {
+    "cpus": 4,
+    "memory": "16gb",
+    "storage": "32gb"
+  },
+  // This devcontainer is based on Ubuntu 24.04 LTS
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
+  "features": {
+    // we must install a non-os-provided python to get around
+    // the "externally managed environment" restrictions
+    // that were added in Ubuntu 23+ and prevent pip global/user
+    "ghcr.io/devcontainers/features/python:1": {
+      "version": "3.12",
+      "installJupyterlab": true,
+      "configureJupyterlabAllowOrigin": "*"
+    },
+    // ruby is used by some of the pre-commit tools
+    "ghcr.io/devcontainers/features/ruby:1": {},
+    // install clang and lld
+    // install cmake and ninja (as CMake generator, as it's faster)
+    "ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
+      "packages": "clang, lld, cmake, ninja-build"
+    }
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "llvm-vs-code-extensions.vscode-mlir",
+        "ms-vscode.cpptools-extension-pack"
+      ],
+      "settings": {
+        "cmake.generator": "Ninja",
+        "cmake.configureArgs": [
+          "-DMLIR_DIR=/workspaces/llvm-project/build/lib/cmake/mlir"
+        ],
+        "mlir.server_path": "/workspaces/heir/build/bin/heir-lsp",
+        "python.analysis.exclude": [
+          "bazel-bin",
+          "bazel-heir",
+          "bazel-out",
+          "bazel-testlogs",
+          "build",
+          ".git",
+          "venv",
+          ".venv"
+        ]
+      }
+    }
+  },
+  // install pre-commit,
+  "onCreateCommand": "pip install --user -r requirements-dev.txt && pre-commit install",
+  // fetch and install LLVM/MLIR
+  // TODO (#1009): uncomment the updateContentCommand line once prebuilds are enabled
+  // "updateContentCommand": "/workspaces/heir/.devcontainer/cmake/setup_mlir.sh"
+}
--- a/.devcontainer/cmake/setup_mlir.sh
+++ b/.devcontainer/cmake/setup_mlir.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# Repository URL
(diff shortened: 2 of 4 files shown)


j2kun (approved):

j2kun:
> pre-built containers

I'm guessing this would incur a hosting cost? Do you know of any services that provide this? I could see about getting Google to pay for it.

AlexanderViand-Intel:
> I'm guessing this would incur a hosting cost? Do you know of any services that provide this? I could see about getting Google to pay for it.

Yep, I think this is down to the repo owner/org. Afaik, the "pre-building" itself is a github action so would incur normal CI costs/CI runner usage (we could sync this to the nightly release rather than every main update, I think), but there's also a "storage" fee for the pre-built codespace. Not sure if it's standard pricing for everyone, or if enterprise accounts have a different setup.

Who actually pays for *running* codespaces is a (org-level) configurable thing, and can be set to either the user who creates it or the org, and I'm guessing it's set to the former.

j2kun:
Ah, so this is a GitHub owned thing end to end. Makes sense.

j2kun:
Reading more about it, I think this is something we can do pretty straightforwardly. @AlexanderViand-Intel  Can you update this config to do the full bazel build + hedron_compile_commands, and I can set up the pre-build to run on a schedule (say, once per change to the LLVM commit?)

AlexanderViand-Intel:
> @AlexanderViand-Intel Can you update this config to do the full bazel build + hedron_compile_commands

Done & running a test-build of the container on my fork right now

AlexanderViand-Intel:
I played with this a bit more, but unfortunately, the default codespaces machines that GH makes available to all users only have 32GB of storage, and `bazel build @heir//...` runs out of space near the end. Any ideas on how to reduce the disk space taken up by the build? IIRC, for MLIR there's a huge difference between building with SHARED_LIBRARIES on or not - which is bazel doing?

j2kun:
Just a diagnostic: when I do a build on my machine and run:

```
 du -sh $(bazel info output_base)
25G     /home/j2kun/.cache/bazel/_bazel_j2kun/b07a48cf9da8c73a462d3608e5c508dd
```

I suppose the remaining 7G is the OS and other installed libraries.

j2kun:
@AlexanderViand-Intel if you increase the disk space as you did in one of your commits, what happens? Does it warn you it will charge you?

AlexanderViand-Intel:
> @AlexanderViand-Intel if you increase the disk space as you did in one of your commits, what happens? Does it warn you it will charge you?

Unfortunately, it will just not let you create a codespace if the requirements are more than what the 2/4 core machines (both have 32GB of storage) provide and your account does not have access to the 8/16/32 core machines (which most accounts do not).

j2kun:
When I just build `@heir//tools:heir-opt`, I get only 6 GiB of disk space. One thing we could do is run a bazel query to get all targets that do not depend on mlir-cpu-runner, and build those.

This should look like this:

`bazel build $(bazel query '@heir//... except rdeps(//..., @llvm-project//mlir:mlir-cpu-runner)')`

But the query fails because apparently there are some projects loaded that we don't have defined

```
preloading transitive closure failed: no such package '@libpfm//': The repository '@libpfm' could not be resolved: Repository '@libpfm' is not defined
```

AlexanderViand-Intel:
Oh, nice @heir-opt! I guess for an initial playground, having `heir-opt` and `heir-lsp` (that also reminds me to add the setting to make the MLIR extension use it!) should be a good start, then we can revisit switching to @heir//... once the build has been "slimmed down" (#847)

j2kun:
From the GitHub docs:

> For organizations on a GitHub Team or GitHub Enterprise Cloud plan, when a user creates a codespace from a repository in the organization, either the user or the organization can pay for the codespace. The user who creates a codespace can't choose who pays for it, but the organization can choose to pay for certain users. In an organization's settings, you can choose for codespaces to be user-owned or organization-owned.
> 
> If an organization chooses for codespaces to be user-owned, a user who creates a codespace from a repository in the organization always pays for the codespace. The user's access to create codespaces depends on the visibility of the repository and your organization's access settings.
> 
> If an organization chooses for codespaces to be organization-owned, the organization will pay for a codespace if all the following things are true:
> 
> The organization has set a non-zero spending limit for GitHub Codespaces.
> The codespace is created from one of the organization's repositories, or from a fork of one of the organization's repositories. This includes both public and private repositories.
> The user creating the codespace is a member or collaborator of the organization, and the organization has enabled GitHub Codespaces for this user. This can include all members and collaborators if the organization has chosen to enable Codespaces for all users. If Codespaces isn't enabled for a user, they can still create codespaces from public repositories in the organization, but the user will pay for these codespaces.

Based on this, I think our best bet is lowering the disk usage, the main reason being that I don't think we should let anyone be a collaborator on the project just to get access to codespaces. That would be necessary even if Google generously enables and pays for org-owned codespaces by default.

j2kun (approved):

AlexanderViand-Intel:
I've swapped the base image from the "universal" GH kitchen sink image to a more basic ubuntu 24.04 LTS image, as this gets us (a) much newer cmake, clang, lld, etc and (b) less unrelated stuff that's just taking up space. This does mean the initial code space setup is a bit slower as it's installing python/etc, but that overhead will disappear once we have prebuilds.

asraa (approved):
Thank you so much!

PR #1002, opened 2024-10-03
Add tips for building MLIR and OpenFHE on low-RAM systems

FlorentCLMichel:
I'm not entirely sure where is the best place for this; let me know if this would better fit elsewhere.


AlexanderViand-Intel (approved):
Thanks, this is great! Good as place as any for now, imho 👍

FlorentCLMichel:
Ah, indeed - seems my local version of `pre-commit` did not run all the checks. I just upgraded it to version 3.8.0 and ran it again.

asraa (approved):
Thank you for these! Much appreciated

PR #1011, opened 2024-10-07
ci: remove unneeded bazel llvm patches

asraa:
thank you @AlexanderViand-Intel for the reminder!

it would be so nice if there was a github action that auto removed them TBH


j2kun:
@asraa what do you think the GH action would look like? I suspect this automation might best be internal to Google so we don't have to worry about pulling it through copybara.

asraa:
> @asraa what do you think the GH action would look like? I suspect this automation might best be internal to Google so we don't have to worry about pulling it through copybara.

So a github action could run on failed build-and-test workflows at main

```
on:
  workflow_run:
    workflows: ["Stage deploy"]
    types:
      - completed
jobs:
  check-if-stale-patch:
    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
    outputs:
        remove-patch: ${{ steps.search-logs.outputs.stale-patch }}
    steps:
      - name: search-logs 
        run: |
           # Search logs of failed jobs if it contains something like "bazel/llvm.patch failed to apply"

  remove-patch:
    if: ${{ jobs.check-if-stale-patch.outputs.remove-patch }}
    steps:
      - name: Create PR with remove bazel/llvm.patch and updated import_llvm.patch 
```

If we did this through copybara, i would imagine this is a lot more difficult to automate though. We could get the PR created from this workflow to auto-add the pull_ready label, but I don't actually know how much copybara would like a PR created by a github actions bot (though I can generate a personal PAT to be used in the create workflow that would make the authorship from a person)

PR #765, opened 2024-07-02
initial template of lazyrelin pass

ahmedshakill:
Implementation of #599 

Here is the current progress status
- [x] The lazy relin pass is visible in heir-opt. 
- [x] Looked into google or-tools tutorial on LP and MIP. 
- [x] Need an in-depth look into Jeremy's ILP blog.  
- [ ] Translate the minimization problem into an ILP one. 
- [ ] Introduce trait/interface

Issue #860, opened 2024-08-01
`--secret-to-<scheme>`: type system challenges for BGV/CKKS modswitch/rescale


AlexanderViand-Intel:
Currently, our `--secret-to-bgv` pass translates an `%r = arith.muli %lhs, %rhs : i32` to the following code:
```llvm
%t = bgv.mul %lhs, %rhs : !ct1, !ct1 -> !ct2 
%r = bgv.relinearize  %t { from_basis = array<i32: 0, 1, 2, 3>, to_basis = array<i32: 0, 1>}: !ct2 -> !ct1
```
As discussed in the context of lazy relinearization (#599, #765), we cannot remove the relin: the DialectConversion would error out, as the return type is not the `!ct1` expected (DialectConversion's TypeConverter simply maps types to types, with no context being taken into account). It's probably fine to ignore this for relinearization, though, because "deviations" from the basic ciphertext type are mostly *temporary*.

However, there's a more serious issue when we consider `mod_switch` or `rescale`. 
These should *also* happen (roughly) after every multiplication in order to manage noise growth, but they represent a *permanent* change in type that must be carried forward through the rest of the program. Right now, I don't really see a way to achieve this with a DialectConversion, so we might have to switch to a more "basic" driver. 

Since this will be a challenge for both BGV and CKKS, it might be a good idea to directly build this as a modular "RLWE.ciphertext-aware conversion" that we then simply instantiate with specific patterns for the schemes.  
As an  example of an existing "custom dialect conversion driver", there's the OneToNTypeConversion that exists because dialect conversion does not properly handle 1:N type mappings.

AlexanderViand:
While the type-system challenges are technically still there, @ZenithalHourlyRate's `mgmt` dialect based approach to this has basically solved/worked around the first issue (deciding placement), while @j2kun's ContextAwareConversion magic allows doing the required type conversions.

PR #1016, opened 2024-10-07
Add ILP to select optimal relinearization points

### LINKED ISSUE #1026: Write documentation for the relinearization ILP

j2kun:
Fixes https://github.com/google/heir/issues/599
Fixes #1026

@ahmedshakill got most of the way in https://github.com/google/heir/pull/765. This PR cherry-picks their commits and finishes the ILP.

For the Cmake build changes in this PR, note that building all the dependencies of or-tools the CMake flag `-DBUILD_DEPS=ON`. I also found that one of the dependencies didn't build properly for me, so I added `-DUSE_COINOR=OFF` since we don't use `CLP` as a solver (just SCIP). That said, even after this PR I get linker errors for `lib/libMLIRHEIRInterfaces.so.19.0`, which seems unrelated to this PR.


j2kun:
Ready for an initial review while I finish up tests

j2kun:
Discovered when writing more tests:

- Rotate ops need to have linearized inputs, so this needs to be added as an additional constraint in the ILP
- For some reason, the smoke test I added isn't solving the way I expect it to. I suspect my solver isn't handling non-add/mul ops correctly.

AlexanderViand-Intel:
> Ready for an initial review while I finish up tests.

Given your most recent comment, do you still want (initial) reviews already, or should we wait until those issues are resolved?

j2kun:
I don't think the remaining issues will drastically change the structure of the PR, so it is ready for a first round of reviews.

ahmedshakill:
I tried to modify `secret-to-bgv` pass to output only `bgv.mul` and no `relin op` for `arith.muli`. But got stuck there like #860. 🙂

BUILD
```diff
@@ -25,7 +25,7 @@ package_group(
 # use by passing `--//:enable_yosys=0` to `bazel build` or `bazel test`
 string_flag(
     name = "enable_yosys",
-    build_setting_default = "1",
+    build_setting_default = "0",
 )
```
AlexanderViand-Intel, line 29: is this just a (lack of) rebasing artifact or an intentional change?

lib/Analysis/LazyRelinAnalysis/LazyRelinAnalysis.cpp
```diff
@@ -0,0 +1,354 @@
...
+#include "llvm/include/llvm/Support/Debug.h"            // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"             // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"                 // from @llvm-project
+#include "ortools/linear_solver/linear_solver.h"  // from @com_google_ortools
+
+// For now, fix the key basis degree bound to 3. Could make it a pass flag
+// later, or introduce costs for high-degree operations and let it be
+// unconstrained.
+constexpr int MAX_KEY_BASIS_DEGEE = 3;
```
AlexanderViand-Intel, line 16: Does a key basis degree of three mean three polynomials per ctxt and keybasis (1, s, s^2) or does it mean the highest degree in the keybasis is three (1,s,s^2,s^3)? I'm guessing the latter, but an example in the comment here probably wouldn't hurt

nit: typo in name
```suggestion
constexpr int MAX_KEY_BASIS_DEGREE = 3;
```

lib/Analysis/LazyRelinAnalysis/BUILD
```diff
@@ -0,0 +1,17 @@
...
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "LazyRelinAnalysis",
+    srcs = ["LazyRelinAnalysis.cpp"],
+    hdrs = ["LazyRelinAnalysis.h"],
+    deps = [
+        "@com_google_ortools//ortools/base",
+        "@com_google_ortools//ortools/linear_solver",
```
AlexanderViand-Intel, line 11: What's the (build time/space) cost of adding these dependencies?

lib/Analysis/LazyRelinAnalysis/LazyRelinAnalysis.h
```diff
@@ -0,0 +1,36 @@
...
+
+  LogicalResult solve();
+
+  // Return true if a relin op should be inserted after the given
+  // operation, according to the solution to the optimization problem.
+  bool shouldInsertRelin(Operation *op) const { return solution.lookup(op); }
+
+  // Return the key basis degree at the given SSA value, as determined
+  // by the solution to the optimization problem.
+  // Don't need an "after relin" version because after relin is always 1.
```
AlexanderViand-Intel, line 23: another place that might benefit from making "key basis degree" meaning more explicit.

```suggestion
  // Don't need an "after relin" version because after relin the keybasis is always (1, s),
  // so the key basis degree is always 1
```

lib/Transforms/LazyRelin/LazyRelin.cpp
```diff
@@ -0,0 +1,220 @@
...
+                                                     beforeBasis, afterBasis);
+        result.replaceAllUsesExcept(reduceOp.getResult(), {reduceOp});
+      }
+    });
+
+    // At this point we need to do some cleanup. The process of removing the
+    // initial relinearize ops and inserting new ones did not update the key
+    // basis sizes on result types. So we walk the IR and update them if
+    // necessary.
+    funcOp->walk([&](Operation *op) {
```
AlexanderViand-Intel, line 72: This seems complicated enough that it should maybe be a separate utils thing to massage types back into consistency?

lib/Transforms/LazyRelin/LazyRelin.td
```diff
@@ -0,0 +1,26 @@
...
+    let summary = "Apply lazy relinearization";
+    let description = [{
+        This pass defers relinearization ops as late as possible in the IR.
+        This is more efficient in cases where multiplication operations are followed by
+        additions, such as in a dot product. Because relinearization also adds error,
+        deferring it can reduce the need for bootstrapping.
+
+        In much of the literature, lazy relinearization is applied manually. See for example
+        Blatt-Gusev-Polyakov-Rohloff-Vaikuntanathan 2019 https://eprint.iacr.org/2019/223
+        and Lee-Lee-Kim-Kim-No-Kang 2020 https://eprint.iacr.org/2020/1549.
```
AlexanderViand-Intel, line 16: I think it'd be good to also mention EVA/EVA Improved and other work that looks at automatically inserting relin.

AlexanderViand-Intel (commented):
LGTM! I quite like the analysis/pass split, and it looks like the "relin-insertion-based-on-analysis" part should be re-usable for other relin insertion approaches, too.

BUILD
j2kun: Pulled in from the cherry pick, will rever.

lib/Transforms/LazyRelin/LazyRelin.td
j2kun: Done in ef66e8eb

This also makes me think we should rename the pass to not be called "lazy", since "lazy" implies a specific kind of heuristic.

lib/Analysis/LazyRelinAnalysis/LazyRelinAnalysis.cpp
j2kun: Done

lib/Transforms/LazyRelin/LazyRelin.cpp
j2kun: The most complicated part are all the guards, so I moved that into a local helper function in 8e657c05. I don't know how much of the result type patching can be reused, because the generic part is the operation cloner/replacer. Maybe we could do connect it to type inference hooks somehow, like "re-infer the result types and replace the op with its correct result type", but that is beyond the scope of this PR.

j2kun:
OK, fixed up all the tests.

lib/Transforms/LazyRelin/LazyRelin.cpp
AlexanderViand-Intel: Thanks, that's a great first step! I agree my full request would be too much work for this PR

AlexanderViand-Intel:
> OK, fixed up all the tests. Last two things
> 
> * Whether we should rename the "lazy" part to something closer to its actual behavior (maybe "optimize-relinearization"?)
> * Where to put model documentation

I like `optimize-relinearization` 👍  (and don't have a good suggestion for the second point)

j2kun:
@asraa this is ready for a final review.

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.td
```diff
@@ -0,0 +1,33 @@
...
+        as possible.
+
+        In this pass, we use an integer linear program to determine the optimal relinearization
+        strategy. This is an adaptation of the ILP described in
+        https://www.jeremykun.com/2023/11/15/mlir-a-global-optimization-and-dataflow-analysis
+
+        TODO(#1026): document the ILP spec
+    }];
+
+    let dependentDialects = ["mlir::heir::bgv::BGVDialect"];
```
asraa, line 30: could you please file a follow-up issue (good first issue!) to generalize this pass to also work with CKKS? I think we could add a RelinearizationOpInterface trait / MulOp trait and have the pass be agnostic of RLWE scheme dialect

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.h
```diff
@@ -0,0 +1,41 @@
...
+  OptimizeRelinearizationAnalysis(Operation *op) : opToRunOn(op){};
+  ~OptimizeRelinearizationAnalysis() = default;
+
+  LogicalResult solve();
+
+  // Return true if a relin op should be inserted after the given
+  // operation, according to the solution to the optimization problem.
+  bool shouldInsertRelin(Operation *op) const { return solution.lookup(op); }
+
+  // Return the key basis degree at the given SSA value, as determined by the
```
asraa, line 21: maybe ciphertext size or key size (minus 1) ? for rotated ciphertexts, whose basis is `(1, s^2)`, this analysis will still perform correctly since the size of the ciphertexts still increases the same way that `(1, s)` canonical ciphertexts do, although their "key basis degree" is probably greater than 1.

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
```diff
@@ -0,0 +1,371 @@
...
+#include "llvm/include/llvm/ADT/DenseMap.h"             // from @llvm-project
+#include "llvm/include/llvm/ADT/TypeSwitch.h"           // from @llvm-project
+#include "llvm/include/llvm/Support/Debug.h"            // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"             // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"                 // from @llvm-project
+#include "ortools/linear_solver/linear_solver.h"  // from @com_google_ortools
+
+// The key basis degree describes the highest degree in the key basis, i.e., a
+// key basis of (1, s, s^2) has degree 2, while (1, s, s^2, s^3) has degree 3.
```
asraa, line 13: Yeah, I think my comment above also applies here. Maybe you can say to ignore rotated ciphertext key bases? Or update per the comment above to size instead of degree

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
```diff
@@ -0,0 +1,371 @@
...
+constexpr int MAX_KEY_BASIS_DEGREE = 3;
+constexpr int INITIAL_KEY_BASIS_DEGREE = 1;
+constexpr int IF_THEN_AUX = 100;
+
+using namespace operations_research;
+
+namespace mlir {
+namespace heir {
+
+#define DEBUG_TYPE "OptimizeRelinearizationAnalysis"
```
asraa, line 27: nit: are we using kebab case for DEBUG_TYPE (here and below)? https://github.com/google/heir/blob/bd75f81f4b1d6214498b70a93dd966ea01609237/lib/Analysis/TargetSlotAnalysis/TargetSlotAnalysis.cpp#L15

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+std::string nameAndLoc(Operation *op) {
+  std::string varName;
+  llvm::raw_string_ostream ss(varName);
+  ss << op->getName() << "_" << op->getLoc();
+  return ss.str();
+}
+
+bool hasCiphertextType(ValueRange range) {
+  return std::any_of(range.begin(), range.end(), [](Value value) {
```
asraa, line 37:
```suggestion
  return llvm::any_of(range, [](Value value) {
```

here and below in the transform (unless there was a reason to avoid the LLVM helpers?)

j2kun:
Hmmm. I think this will also break the CMake build because we don't yet have the ortools dependency in cmake, right?

AlexanderViand-Intel:
> Hmmm. I think this will also break the CMake build because we don't yet have the ortools dependency in cmake, right?

Yes, unfortunately. Though it looks like ortools is available as a binary, so we can just make install it system wide I think (unless you're doing something that would require building it from source?)

j2kun:
Can't use it as a binary because we're depending directly on the C++ API. I'll try FetchContent.

AlexanderViand-Intel:
> Can't use it as a binary because we're depending directly on the C++ API. I'll try FetchContent.

Yep, they even have a nice example of how to do it here: https://github.com/google/or-tools/blob/stable/cmake/README.md#using-fetchcontent

asraa (commented):
Thank you!!! 

I admit I'm glossed over a little bit at the solver's implementation, I'll keep digging in it though.

j2kun:
> > Can't use it as a binary because we're depending directly on the C++ API. I'll try FetchContent.
> 
> Yep, they even have a nice example of how to do it here: https://github.com/google/or-tools/blob/stable/cmake/README.md#using-fetchcontent

Facepalm: there is a CMake config for my mlir-tutorial, which includes or-tools 🤣

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
```diff
@@ -0,0 +1,370 @@
...
+        .Case<bgv::MulOp>([&](auto op) {
+          // result_degree =
+          //   (arg1_degree + arg2_degree) (1 - insert_relin_op)
+          //   + 1 * insert_relin_op,
+          //
+          // but linearized due to the quadratic term input_key_basis_degree *
+          // insert_relin_op, and an extra variable inserted to keep track of
+          // the difference between the before_relin and after_relin degrees:
+          //
+          // before_relin = arg1_degree + arg2_degree
```
asraa, line 202: -1 here? 

i'm confused how the tests verify with https://github.com/google/heir/blob/bd75f81f4b1d6214498b70a93dd966ea01609237/lib/Dialect/FHEHelpers.h#L24

tests/optimize_relinearization/optimize_relinearization.mlir
```diff
@@ -0,0 +1,161 @@
...
+// CHECK-SAME: %[[ARG1:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-SAME: %[[ARG2:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-SAME: %[[ARG3:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-NEXT: %[[MUL0:.*]] = bgv.mul %[[ARG0]], %[[ARG1]]
+// CHECK-NEXT: %[[MUL1:.*]] = bgv.mul %[[ARG2]], %[[ARG3]]
+// CHECK-NEXT: %[[ADD:.*]] = bgv.add %[[MUL0]], %[[MUL1]]
+// CHECK-NEXT: %[[RELINEARIZE0:.*]] = bgv.relinearize %[[ADD]]
+// CHECK-NEXT: return %[[RELINEARIZE0]]
+
+func.func @two_muls_followed_by_add(%arg0: !ct, %arg1: !ct, %arg2: !ct, %arg3: !ct) -> !ct {
```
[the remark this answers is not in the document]
j2kun: The last test in this file as a rotate op.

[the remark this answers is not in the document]
asraa: I guess I meant that rotated inputs that are later multiplied and relinearized work - but it really does since the basis size is the same as any other canonical ciphertext.

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.cpp
```diff
@@ -0,0 +1,231 @@
...
+            // correct by construction, nothing to do.
+            return;
+          })
+          .Case<bgv::MulOp>([&](auto op) {
+            auto lhsDeg = op.getLhs().getType().getRlweParams().getDimension();
+            auto rhsDeg = op.getRhs().getType().getRlweParams().getDimension();
+            auto resultDeg =
+                op.getResult().getType().getRlweParams().getDimension();
+
+            if (lhsDeg + rhsDeg != resultDeg) {
```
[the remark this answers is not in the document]
j2kun: I think you're conflating degree and dimension.

(1, s) * (1, s) -> (1, s, s^2) (1+1=2), but the dimension goes from (2, 2) -> 3 = 2+2-1.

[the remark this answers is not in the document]
j2kun: This also adds some confusion because dimension can be 2 while degree is also 2, for the basis of (1, s^2)

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.h
j2kun: This is a point at which my lack of understanding of the cryptosystem implementation is hurting me. We always start at (1, s), and the relin always goes back to (1, s), but under what circumstances do we get other key bases? 

- Multiplication goes (1, s) -> (1, s, s^2), but can also go (1, s, s^2) (1, s, s^2) -> (1, s, s^2, s^3, s^4), right?
- Does rotation go (1, s) -> (1, s^2)?
- Can multiplication handle arguments in different key bases? (1, s) * (1, s^2) for example?

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.cpp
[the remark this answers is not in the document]
asraa: I think it's true I'm conflating degree and dimension, but don't we define `RlweParams.getDimension` to be 2 for the canonical case? https://github.com/google/heir/blob/780ae9e4957c194c8074ba7083afafa16205a41d/tests/bgv/ops.mlir#L31C1-L32C1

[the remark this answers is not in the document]
asraa: I guess my confusion is why assign `degree = rlweparams.getdimension()`?

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.cpp
```diff
@@ -0,0 +1,231 @@
...
+            op->erase();
+          });
+    });
+  }
+
+  void runOnOperation() override {
+    Operation *module = getOperation();
+    module->walk([&](func::FuncOp op) {
+      processFunc(op);
+      LLVM_DEBUG(op.dump());
```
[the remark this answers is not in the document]
j2kun: Fixed in b96968c0

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.cpp
```diff
@@ -0,0 +1,231 @@
...
+        operandDegrees.push_back(std::nullopt);
+      }
+    }
+
+    if (!std::any_of(
+            resultDegrees.begin(), resultDegrees.end(),
+            [](std::optional<int> degree) { return degree.has_value(); }) ||
+        !std::any_of(
+            operandDegrees.begin(), operandDegrees.end(),
+            [](std::optional<int> degree) { return degree.has_value(); })) {
```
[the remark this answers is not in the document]
j2kun: Done

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
j2kun: Fixed in 140fdbdd

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
j2kun: Done.

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.h
AlexanderViand-Intel: * In theory, yes, multiplication can easily be defined for larger key basis, though implementations of that are, afaik, rare
* Rotation does something weirder. There's still a switch of keys involved, but it's somewhat orthogonal to the key basis[^1]. Since rotation virtually always includes keyswitching, you can think of it as a no-op on the key basis (1,s) in, (1,s) out.
* I'm not actually sure about that, never thought about it. Given that higher-than-(1,s)-mul is already super rare, I guess it might not matter in practice.



[^1]: I think it's something like permuting coefficients but I'm pretty rusty on the details here. I believe one could think of rotation (w/o keyswitching, which is, other than for mul, virtually always considered an implicit aprt of the operation) as going form (1, s) to (1, s') where s' is some "permuted" version).

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.cpp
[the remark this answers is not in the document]
j2kun: Ah, you're right. That's a bug.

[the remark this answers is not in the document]
j2kun: Fixed in 8e792b62

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.h
asraa: > Multiplication goes (1, s) -> (1, s, s^2), but can also go (1, s, s^2) (1, s, s^2) -> (1, s, s^2, s^3, s^4), right?

Correct! Although not many libraries will implement that high dimension multiplication (so I think your max basis constant is useful)

> Does rotation go (1, s) -> (1, s^2)?

Or (1, s^i) -> (1, s^j) more generally. In theory it could also rotate (1, s, s^2) to (1, s^i, s^2^i), but again i doubt anyone is exposing that.

> Can multiplication handle arguments in different key bases? (1, s) * (1, s^2) for example?

Yes, in theory, but again not exposed. I worked it out on paper and unfortunately a mul between (1, s^i) and (1, s^j) will end up being a result of ciphertext with basis (1, s^i, s^j, s^{i+j}). In the case where i and j agree is where we see the subtraction by one since the cross terms combine to (1, s^i, s^{i + i})

AlexanderViand-Intel: > > Does rotation go (1, s) -> (1, s^2)?
> 
> Or (1, s^i) -> (1, s^j) more generally. In theory it could also rotate (1, s, s^2) to (1, s^i, s^2^i), but again i doubt anyone is exposing that.

can you go into more detail on that one? I thought the key switch that happened during rotations was orthogonal to the degrees, but I guess I'm mistaken on that?

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.td
j2kun: Done

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.h
asraa: > can you go into more detail on that one? I thought the key switch that happened during rotations was orthogonal to the degrees, but I guess I'm mistaken on that?

Yeah, the key switch that happens during rotations is orthogonal to the key switching that changes a ciphertext key basis from `(1, s, s^2)` to `(1, s)`. But actually key switching keys are all generated from the same principle. I'm not sure which part to go into, but Galois rotations are just performed by applying the automorphism `X -> X^i` on the polynomial. So that can be applied to a size 3 ciphertext or a size two ciphertext. The same automorphism will apply to the secret key it is encrypted under (so you would map (1, s) to (1, s^i)).

asraa: Here's an implementation in BGV: https://eprint.iacr.org/2011/566.pdf
The key switching procedure on page 29 can be done from any type of key, like from (1, s^i) back to (1, s) or from (1, s, s^2) back to (1, s) or from (1, s^i, s^2^i) to (1, s).

Obviously the client needs to generate all of these though.

tests/optimize_relinearization/optimize_relinearization.mlir
```diff
@@ -0,0 +1,161 @@
...
+!ct1 = !lwe.rlwe_ciphertext<encoding=#encoding, rlwe_params=#params1, underlying_type=i3>
+
+// CHECK-LABEL: func.func @two_muls_followed_by_add
+// CHECK-SAME: %[[ARG0:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-SAME: %[[ARG1:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-SAME: %[[ARG2:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-SAME: %[[ARG3:[^:]*]]: !lwe.rlwe_ciphertext
+// CHECK-NEXT: %[[MUL0:.*]] = bgv.mul %[[ARG0]], %[[ARG1]]
+// CHECK-NEXT: %[[MUL1:.*]] = bgv.mul %[[ARG2]], %[[ARG3]]
+// CHECK-NEXT: %[[ADD:.*]] = bgv.add %[[MUL0]], %[[MUL1]]
```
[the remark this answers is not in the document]
j2kun: Done

j2kun:
> I admit I'm glossed over a little bit at the solver's implementation, I'll keep digging in it though.

@asraa I think it's fair to not devote too much time to this right now, I can write up a better doc and walk you through it after I'm back from leave.

asraa:
> @asraa I think it's fair to not devote too much time to this right now, I can write up a better doc and walk you through it after I'm back from leave.

Thank you! I do understand the simpler constraints like operand1.deg - operand2.deg = 0, but the later parts confuse me :)

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.h
AlexanderViand-Intel: Ohh, I see now what you mean - that's also my understanding of rotate and keyswitch, but I think it's still better to say that "rotation does not / cannot modify the size of the key basis", and to instead talk about (generic) keyswitching separately:  "could also ~~rotate~~ keyswitch (1, s, s^2) to (1, s^i, s^2^i)".

In general, it seems like the (applied) FHE community has slightly inconsistent terminologoy:
`rotate` usually meansto "automorphism + keyswitch" where as  `mul` generally only means "tensor-like-product", and the associated keyswitch is a  separate `relin` operation. (The crypto-y  view of FHE is a bit more consistent, and calls what most libraries call multiply the "pre-multiply", with "multiply" defined as including the keyswitch/relin)

AlexanderViand-Intel: > actually, i did mean rotate - that doesn't change the size either. that meant just permuting the coefficients with the automorphism

Now I'm confused... let me try and make the keybasis notation a bit more explicit:
In my understanding, a rotate-by-two (w/o keyswitch) gives you a change from (1, s(X)) -> (1, s(X^2)) whereas the key basis change in a multiply (w/o keyswitch/relin) is (1, s(X)) -> (1, s(X), s(X)^2)) which is fundamentally different?

asraa: They are! Now take a multiplied ciphertext with basis (1, s(X), s(X)^2) and then apply the X -> X^2 transformation and now you've transformed from (1, S(X), S(X)^2) to (1, S(X^i), S(X^i)^2) (the transform in that comment)

j2kun: I'm trying to figure out what simplification works best for now. Maybe assume rotation is always linear in/out, and constrain it appropriately, and so we always have degree == dimension-1. That would require no changes to this PR, but maybe some additional comments explaining the assumptions

asraa: I think that's reasonable. Most libraries will always assume rotations expect (1, s) in

j2kun:
Squashing the PR.

asraa (approved):
Thank you!!

AlexanderViand-Intel (approved):

asraa:
Getting around to merging this now, thank you for the wait! I'll take care of the rebase internally since it needs a BUILD fix for the dependency.

j2kun:
Due to the internal deprecation of the MPModel API, I migrated the solver API to math_opt and squashed the changes from https://github.com/google/heir/pull/1033 into this PR (since https://github.com/google/heir/pull/1033 contained simplifications of the model building step).

j2kun:
The build failure doesn't occur on my machine, looks like an outdated cstdlib on the GH action runner:

```
external/com_google_ortools/ortools/util/fp_roundtrip_conv.cc:66:7: error: call to 'to_chars' is ambiguous
      std::to_chars(buffer.data(), buffer.data() + buffer.size(), value);
      ^~~~~~~~~~~~~
/usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/charconv:363:1: note: candidate function
_GLIBCXX_TO_CHARS(char)
```

How can we resolve this? Maybe use ubuntu 22.04 or 24.04 instead of 20.04?

docs/content/en/docs/Design/relinearization_ilp.md
```diff
@@ -0,0 +1,199 @@
...
+- Operand agreement: For each operation with operand SSA values $v_1, \\dots,
+  v_k$, $\\textup{KB}_{v_1} = \\dots = \\textup{KB}_{v_k}$, i.e., all key basis
+  inputs must match.
+- Special linearized ops: `bgv.rotate` and `func.return` require linearized
+  inputs, i.e., $\\textup{KB}\_{v_i} = 1$ for all inputs $v_i$ to these
+  operations.
+- Before relinearization key basis: for each operation $o$ with operands $v_1,
+  \\dots, v_k$, constrain $\\textup{KB}^{br}_{\\textup{result}(o)} =
+  f(\\textup{KB}_{v_1}, \\dots, \\textup{KB}\_{v_k})$, where $f$ is a statically
+  known linear function. For multiplication $f$ it addition, and for all other
```
asraa, line 144:
```suggestion
  known linear function. For multiplication $f$ is addition, and for all other
```

asraa (approved):
Wonderful, thank you so much! The design doc looks good to me, and I appreciate the simplifying assumptions spelled out. There's a small typo that I'll fix in the internal patch.

PR #1033, opened 2024-10-10
Add a page on heir.dev for the relinearization ILP model spec

j2kun:
Rebased over https://github.com/google/heir/pulls
Fixes https://github.com/google/heir/issues/1026

For a rendered version of the document, see https://www.overleaf.com/read/sksjnfhjwmnf#e62cb6

While I was writing this doc, I realized some of the code building the constraints could be simplified by leaning more heavily on the "before relin" variables.


docs/content/en/docs/Design/relinearization_ilp.md
```diff
@@ -0,0 +1,199 @@
...
+= (c_0, c_1)$ and $\\mathbf{d} = (d_0, d_1)$ produces a ciphertext $\\mathbf{f}
+= (f_0, f_1, f_2)$. This triple can be decrypted by taking a dot product with
+$(1, s, s^2)$.
+
+With this in mind, each RLWE ciphertext $\\mathbf{c}$ has an associated _key
+basis_, which is the vector $\\mathbf{s_c}$ whose dot product with $\\mathbf{c}$
+decrypts it.
+
+Usually a larger key basis is undesirable. For one, operations in a higher key
+basis are more expensive. Repeated multiplications exponentially increase the
```
AlexanderViand-Intel, line 28: It's probably worth mentioning that it also leads to really bad noise growth behavior.

docs/content/en/docs/Design/relinearization_ilp.md
```diff
@@ -0,0 +1,199 @@
...
+length of the key basis. So to avoid this, an operation called _relinearization_
+was designed that converts a ciphertext from a given key basis back to $(1, s)$.
+Doing this requires a set of _relinearization keys_ to be provided by the client
+and stored by the server.
+
+In general, key bases can be arbitrary. Rotation of an RLWE ciphertext by a
+shift of $k$, for example, first applies the automorphism $x \\mapsto x^k$. This
+converts the key basis from $(1, s)$ to $(1, s^k)$, and more generally maps $(1,
+s, s^2, \\dots, s^d) \\mapsto (1, s^k, s^{2k}, \\dots, s^{kd})$. Most FHE
```
AlexanderViand-Intel, line 37: Do rotations do that, though? There should be a difference between $(1, s(X)^k)$ (what's in the doc but just making the polynomial nature of $s$ explicit)  and $(1, s(X^k))$ (what a rotation by k produces, prior to key-switching)? Unless there's some ring property that makes these two things magically the same?

AlexanderViand-Intel: Also pinging @asraa since I guess this is the continuation of the discussion we had in the PR ;)

docs/content/en/docs/Design/relinearization_ilp.md
```diff
@@ -0,0 +1,199 @@
...
+In much of the literature, lazy relinearization is applied manually. See for
+example
+[Blatt-Gusev-Polyakov-Rohloff-Vaikuntanathan 2019](https://eprint.iacr.org/2019/223)
+and [Lee-Lee-Kim-Kim-No-Kang 2020](https://eprint.iacr.org/2020/1549). In some
+compiler projects, such as the [EVA compiler](https://eprint.iacr.org/2021/1505)
+relinearization is applied automatically via a heuristic, either "eagerly"
+(immediately after each multiplication op) or "lazily," deferred as late as
+possible.
+
+In HEIR, relinearization placement is implemented via a mixed-integer linear
```
AlexanderViand-Intel, line 70: Missing a #heading before this?

AlexanderViand-Intel (approved):
Thanks for putting this together so quickly (and the convenient  rendered version!)

As usual, great exposition/writing, though I'm going to admit I've only skimmed the actual ILP formulization 😅

My only real questions is, once again, about what exactly the key basis after a rotation should be, i.e., (1, s(X^k)) vs (1, s(X)^k))

docs/content/en/docs/Design/relinearization_ilp.md
j2kun: I think you're right and I was being sloppy. Will wait for Asra to confirm.

j2kun: Man someone needs to write a good tutorial about this stuff 😛

docs/content/en/docs/Design/relinearization_ilp.md
j2kun: Fixed in eac36665

docs/content/en/docs/Design/relinearization_ilp.md
j2kun: Fixed in eac36665

Issue #821, opened 2024-07-24
Make successful CMake build & test a CI PR merge requirement 


AlexanderViand-Intel:
Depends on
- [x] #788
- [ ] #818
- [x] #819
- [ ] #820 
- [x] #822 

and potentially on
* #797

j2kun:
Trying to keep the CMake working for some of my PRs, and looking for a canonical set of commands to run for the cmake build. So far I have:

```
# build LLVM
mkdir external/llvm-project/build && cd external/llvm-project/build
cmake -G Ninja ../llvm \                                                                                                                           -DLLVM_ENABLE_PROJECTS="mlir" \
  -DLLVM_TARGETS_TO_BUILD="host" \
  -DLLVM_ENABLE_ASSERTIONS=ON \
  -DLLVM_BUILD_EXAMPLES=OFF \
  -DCMAKE_BUILD_TYPE=Release \
  -DLLVM_USE_SPLIT_DWARF=ON \
  -DLLVM_ENABLE_RTTI=ON \
  -DLLVM_USE_LINKER=lld \
  -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
# is it necessary to actually build MLIR at this step?
cmake --build . --target check-mlir

# navigate back to HEIR root
popd

mkdir build && cd build
cmake -G Ninja .. \
  -DMLIR_DIR=external/llvm-project/build/lib/cmake/mlir \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_CXX_COMPILER=clang++
cmake --build . --target heir-opt
```

AlexanderViand-Intel:
My recommendation would be to take a look at [.devcontainer/cmake/setup_mlir.sh](https://github.com/google/heir/blob/main/.devcontainer/cmake/setup_mlir.sh)  and [.devcontainer/cmake/setup_heir.sh](https://github.com/google/heir/blob/main/.devcontainer/cmake/setup_heir.sh)  which (modulo paths being a bit different) should hopefully reflect "best practice".

@ai-mannamalai let me know if they don't reflect what you're doing and should be updated!

AlexanderViand:
Closing remaining cmake-related issues after we fully dropped cmake support

PR #1035, opened 2024-10-10
Improved CMake build

### LINKED ISSUE #819: Add custom `add_heir_....` CMake macros
We cannot use the existing `add_mlir_{dialect/conversion/transform/analysis}` CMake macros as they assume that source files are organized in a specific pattern, which is different from what we do in HEIR. However, it should be relatively straight-forward to write our own custom CMake macro that does the same thing, just with our directory structure in mind.

AlexanderViand-Intel:
This fixes the CMake build after the recent reorganization.
However, this also cleans up the cmake setup quite a bit and resolves #819  

* I added  custom helper functions (`add_mlir_dialect`, `add_mlir_pass`) that can handle the standard tablegen setup for a dialect/pass.
* (Hopefully) all targets are now using the correct `add_mlir_dialect_library`/`add_mlir_conversion_library`/`add_mlir_library` calls (which means that we can use the lists these functions auto-generate to get all dialect libs, all conversion libs, etc.
* Target names are now (hopefully) consistent with bazel, except that everything is prefixed with "HEIR". Note that CMake doesn't support the "folder structure" approach that bazel has to targets, and target names do not include paths.
* Lots of small cleanup here and there

There's one minor change that also affects the bazel build, and that's moving the Utils in `lib/Target` to `lib/Utils/TargetUtils` (same as was already done for `ConversionUtils` in the Conversion reorg)

There's stil more work to be done towards #821, primarily LLVM/MLIR auto-setup, so I'll post a follow-up PR soon.


AlexanderViand-Intel:
@ai-mannamalai  This should hopefully fix the CMake build, please let me know if it works for you/leave a review of the changes!

ai-mannamalai:
OK checking today

ai-mannamalai:
My CMake configuration and MLIR are correctly specified (-DMLIR_DIR etc.) where LLVM/MLIR is installer:
```
cmake -S . -B buildx -DMLIR_DIR=${MLIR_DIR}
CMake Deprecation Warning at CMakeLists.txt:7 (cmake_policy):
  The OLD behavior for policy CMP0116 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- The C compiler identification is AppleClang 16.0.0.16000026
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for histedit.h
-- Looking for histedit.h - found
-- Found LibEdit: /Library/Developer/CommandLineTools/SDKs/MacOSX15.0.sdk/usr/include (found version "2.11")
-- Found ZLIB: /Library/Developer/CommandLineTools/SDKs/MacOSX15.0.sdk/usr/lib/libz.tbd (found version "1.2.12")
-- Found zstd: /opt/homebrew/lib/libzstd.dylib
-- Found LibXml2: /Library/Developer/CommandLineTools/SDKs/MacOSX15.0.sdk/usr/lib/libxml2.tbd (found version "2.9.13")
-- Using MLIRConfig.cmake in: /usr/local/lib/cmake/mlir
-- Using LLVMConfig.cmake in: /usr/local/lib/cmake/llvm
-- Linker detection: Apple
-- Performing Test LLVM_LINKER_SUPPORTS_NO_EXPORTED_SYMBOLS
-- Performing Test LLVM_LINKER_SUPPORTS_NO_EXPORTED_SYMBOLS - Success
-- Performing Test LLVM_LINKER_SUPPORTS_NO_WARN_DUPLICATE_LIBRARIES
-- Performing Test LLVM_LINKER_SUPPORTS_NO_WARN_DUPLICATE_LIBRARIES - Success
-- Configuring done (2.0s)
-- Generating done (0.9s)
-- Build files have been written to: /Users/mannamalai/devel/heir/buildx
mannamalai@core-mannamalai heir % cd buildx
```

however the build fails,
```
 make                                       
[  0%] Building HEIRInterfaces.h.inc...
[  1%] Building HEIRInterfaces.cpp.inc...
[  1%] Built target HEIRInterfacesOpsIncGen
[  1%] Building OpenfheDialect.h.inc...
[  2%] Building OpenfheDialect.cpp.inc...
[  2%] Built target HEIROpenfheDialectIncGen
[  2%] Building OpenfheOps.h.inc...
[  3%] Building OpenfheOps.cpp.inc...
[  3%] Built target HEIROpenfheOpsIncGen
[  4%] Building OpenfheTypes.h.inc...
[  4%] Building OpenfheTypes.cpp.inc...
[  5%] Built target HEIROpenfheTypesIncGen
[  5%] Built target HEIROpenfheIncGen
[  5%] Built target mlir-headers
[  5%] Built target mlir-generic-headers
[  6%] Building CXX object lib/Dialect/Openfhe/IR/CMakeFiles/obj.HEIROpenfhe.dir/OpenfheDialect.cpp.o
In file included from /Users/mannamalai/devel/heir/lib/Dialect/Openfhe/IR/OpenfheDialect.cpp:1:
/Users/mannamalai/devel/heir/lib/Dialect/Openfhe/IR/OpenfheDialect.h:4:10: fatal error: 'mlir/include/mlir/IR/Builders.h' file not found
    4 | #include "mlir/include/mlir/IR/Builders.h"               // from @llvm-project
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.

```

I thought we were keeping the old style -DENABLE_INSTALLED_MLIR=/path/to/mlir but you have removed it.

AlexanderViand-Intel:
> My CMake configuration and MLIR are correctly specified (-DMLIR_DIR etc.) where LLVM/MLIR is installer:
> [...]
> I thought we were keeping the old style -DENABLE_INSTALLED_MLIR=/path/to/mlir but you have removed it.

Oh, sorry, are you currently using an installed CMake? I can add the if/else back in, though I figured we'd want to move to only supporting non-installed as that's how the FetchContent version will work (I'm on the move today and at HES tomorrow, but still hoping to have that done in time for the WAHC tutorial on Monday). In the future, if people want to specify their own -MLIR_DIR, I'd assume it'd be for in-sync development of things such as `Polynomial`, so they'd probably be working off of a non-installed CMake, too?

lib/Dialect/CGGI/Transforms/CMakeLists.txt
```diff
@@ -1,3 +1,47 @@
+# Since this defines two passes in one *.td file,
```
asraa, line 1: is this something we should mark as a HEIR style thing? i can file a follow-up issue that we should probably make a Developer Best Practices / Style guide.

tools/CMakeLists.txt
```diff
@@ -1,174 +1,131 @@
...
+    # TODO: Create an add_mlir_transform or similar function
+    # to avoid the need to manually list all dialect-specific transforms
+    # HEIRBooleanVectorizer
+    # HEIRLWETransforms
+    # HEIROpenfheTransforms
+    # HEIRPolynomialTransforms
+    # HEIRSecretTransforms
+    # HEIRSetDefaultParameters
+    # HEIRTensorExtTransforms
+    # HEIRTosaToSecretArith
```
asraa, line 85: I think these can just go? heir-translate (and lsp) shouldn't  need any transforms

asraa (approved):
Thank you!

tools/CMakeLists.txt
AlexanderViand-Intel: Ah, yes, I left the commented-out ones for each tool to make it explicit what ISN'T being included, but I'm also happy to remove the comments!

lib/Dialect/CGGI/Transforms/CMakeLists.txt
AlexanderViand-Intel: The transforms (especially the non-dialect-associated ones in `lib/Transforms`) are a bit of a wild west, even in Bazel-land. I do think we should set-up some kind of "HEIR style" - is there a way to have something similar to `add_heir_pass` (in `cmake/AddHEIR.cmake`) for bazel, too? That way, the style becomes self-enforcing ;)

AlexanderViand-Intel: Done, I left the comments in heir-opt, which needs nearly all of them anway, but removed them from heir-translate and heir-lsp

asraa: > is there a way to have something similar to add_heir_pass

I think so! There's bazel macros that can generate the library rules for you (MLIR uses this for the custom tablegen builds), so we could actually create a bazel macro for a heir_dialect_library and so on. I think it would remove some of the boilerplate. I'll file an issue about it for later

j2kun: > for bazel, too? That way, the style becomes self-enforcing ;)

The Google internal processes force a bazel style guide, and Asra and I regularly patch PRs internally to keep it up to standard. I doubt there's a ton of benefit going beyond that for MLIR specific stuff, but I'm open to suggestions. File an issue.

AlexanderViand-Intel:
I've rebased and updated this to include the new relin optimization, but I'm having some issues with getting the `or-tools` dependency to build.

```
[build] In file included from /workspaces/heir/build/_deps/absl-src/absl/strings/internal/str_format/extension.cc:16:
[build] /workspaces/heir/build/_deps/absl-src/absl/strings/internal/str_format/extension.h:34:35: error: unknown type name 'uint8_t'
[build]    34 | enum class FormatConversionChar : uint8_t;
[build]       |       
``` 

My FetchContent for or-tools is a bit different, as I'm grabbing the commit hash from the `WORKSPACE` file - can you try if this builds for you @j2kun? See https://github.com/google/heir/blob/9912bdf9c4968b0bf9f0d67989991d142342697e/lib/Analysis/OptimizeRelinearizationAnalysis/CMakeLists.txt

Note: `main` currently does not even configure successfully for me, but that makes sense as the CMakeLists.txt there does not even `set(BUILD_DEPS ON)` for or-tools and so it fails when it can't `find_package` for its dependencies.

ai-mannamalai:
#include<stdint.h>
or
#include<cstdint>

Depending on compiler.
From: Alexander Viand ***@***.***>
Date: Monday, October 14, 2024 at 9:26 PM
To: google/heir ***@***.***>
Cc: Muthu Annamalai ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [google/heir] Improved CMake build (PR #1035)

I've rebased and updated this to include the new relin optimization, but I'm having some issues with getting the or-tools dependency to build.

[build] In file included from /workspaces/heir/build/_deps/absl-src/absl/strings/internal/str_format/extension.cc:16:

[build] /workspaces/heir/build/_deps/absl-src/absl/strings/internal/str_format/extension.h:34:35: error: unknown type name 'uint8_t'

[build]    34 | enum class FormatConversionChar : uint8_t;

[build]       |

My FetchContent for or-tools is a bit different, as I'm grabbing the commit hash from the WORKSPACE file - can you try if this builds for you @j2kun<https://github.com/j2kun>? See https://github.com/google/heir/blob/9912bdf9c4968b0bf9f0d67989991d142342697e/lib/Analysis/OptimizeRelinearizationAnalysis/CMakeLists.txt<https://github.com/google/heir/blob/9912bdf9c4968b0bf9f0d67989991d142342697e/lib/Analysis/OptimizeRelinearizationAnalysis/CMakeLists.txt>

Note: main currently does not even configure successfully for me, but that makes sense as the CMakeLists.txt there does not even set(BUILD_DEPS ON) for or-tools and so it fails when it can't find_package for its dependencies.

j2kun:
I did notice some older versions of or-tools wasn't working properly with the absl dependency, so I tried pinning to more recent commits.

AlexanderViand-Intel:
Came back from dinner, looked at this with fresh eyes and realized I'm an idiot: I used the same regex for or-tools and rapidjson (with the name = .... part changed) but their WORKSPACE blocks had different numbers of lines between `name` and `commit` :(  As a result, the commit variable was empty and FetchContent wasn't grabbing the right version of or-tools.

I blame mostly myself for not noticing the empty commit hash variable, but also a little bit CMake for still not having a competent regex engine w/ support for non-greedy matching...

I've fixed the REGEX, added checks that the commit hash isn't empty to both rapidjson + ortools and it now errors hard + gives a message telling people how to fix it if anything changes in their WORKSPACE blocks in the future.

The dependencies now build fine, but or-tools complains about missing files, so I still need to investigate that.
```
#include "ortools/math_opt/model.pb.h"
#include "ortools/math_opt/model_update.pb.h"
```

j2kun:
Those are generated protocol buffer headers. So you'll need to figure out how to run the protobuf compiler in cmake I think. I'd imagine it's in the deps already and you just have to find the right existing command or config.

j2kun:
Maybe BUILD_Protobuf? Though I feel BUILD_DEPS  should suffice...

AlexanderViand-Intel:
I've tried a more recent version (9.11) and it does not seem to have reappeared, though I'll need to reconfirm on a fresh codespace to make sure that's not just correlation. Would switching to 9.11 (ed94162b910fa58896db99191378d3b71a5313af) be an issue?

ai-mannamalai:
Ok wow – copy-pasta is very sneaky. I’ve had bugs like this with MLIR TD. Nice catch Alex.

From: Alexander Viand ***@***.***>
Date: Tuesday, October 15, 2024 at 7:56 PM
To: google/heir ***@***.***>
Cc: Muthu Annamalai ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [google/heir] Improved CMake build (PR #1035)

Came back from dinner, looked at this with fresh eyes and realized I'm an idiot: I used the same regex for or-tools and rapidjson (with the name = .... part changed) but their WORKSPACE blocks had different numbers of lines between name and commit :( As a result, the commit variable was empty and FetchContent wasn't grabbing the right version of or-tools.

I blame mostly myself for not noticing the empty commit hash variable, but also a little bit CMake for still not having a competent regex engine w/ support for non-greedy matching...

I've fixed the REGEX, added checks that the commit hash isn't empty to both rapidjson + ortools and it now errors hard + gives a message telling people how to fix it if anything changes in their WORKSPACE blocks in the future.

The dependencies now build fine, but or-tools complains about missing files, so I still need to investigate that.

#include "ortools/math_opt/model.pb.h"

#include "ortools/math_opt/model_update.pb.h"

j2kun:
9.11 should be fine. The newer the better

PR #1037, opened 2024-10-11
Fix tensor type for polynomial-mul-to-ntt

ZenithalHourlyRate:
Formerly polynomial-mul-to-ntt chooses the immediate tensor type to be just the size of the coefficientModulus. For example, when mod = 7681, its width is 13 and the result would be `tensor<nxi13, #ring>`

```
In #ring, coefficientType is i32 
%0 = polynomial.ntt %arg0 : !polynomial.polynomial<ring #ring> -> tensor<4xi13, #ring>
```

However, when further lower `polynomial.ntt` using `polynomial-to-standard`, we would encounter the following error. Obviously, we can not fit `i32` into `i13`.

```
void writeAPIntsToBuffer(size_t, std::vector<char> &, APRangeT &&) [APRangeT = llvm::ArrayRef<llvm::APInt> &]:
 Assertion `(*it).getBitWidth() <= storageWidth' failed.
```

Code snippet (requires #1036, otherwise the mul-to-ntt rewrite itself wont work). This PR itself does not depend on #1036.

Command `--convert-polynomial-mul-to-ntt --mod-arith-to-arith --polynomial-to-standard`

```
#cycl = #polynomial.int_polynomial<1 + x**4>
#root = #polynomial.primitive_root<value=1925:i32, degree=8:i32>
#ring = #polynomial.ring<coefficientType = i32, coefficientModulus = 7681 : i32, polynomialModulus=#cycl, primitiveRoot=#root>
!p = !polynomial.polynomial<ring=#ring>

module {
  func.func @polymul(%x : !p, %y : !p) -> (!p) {
    %add = polynomial.mul %x, %y : !p
    return %add : !p
  }
}
```

Cc @AlexanderViand for #817

### DIFF
--- a/lib/Dialect/Polynomial/Transforms/NTTRewrites.td
+++ b/lib/Dialect/Polynomial/Transforms/NTTRewrites.td
@@ -15,9 +15,7 @@ def GetRingModAttr : NativeCodeCall<
 
 def InputTensorType : NativeCodeCall<
       "RankedTensorType::get({$0.getPolynomialModulus().getPolynomial().getDegree()},"
-      " IntegerType::get($_builder.getContext(),"
-      "  ($0.getCoefficientModulus().getValue() - 1).getActiveBits()),"
-      " $0)">;
+      " $0.getCoefficientType(), $0)">;
 
 def CreateCModConstant : NativeCodeCall<
       "$_builder.create<arith::ConstantOp>($0.getLoc(), $2,"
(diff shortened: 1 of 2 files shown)


AlexanderViand-Intel (approved):
LGTM! I think this change makes a lot of sense - the coefficient type in `!polynomial.polynomial` essentially already specifies the intended "container", so the compiler should just respect that decision.

asraa (approved):

asraa:
Thank you! Could you please rebase this? We had a massive test folder reorganization.

ZenithalHourlyRate:
Rebased

PR #1046, opened 2024-10-13
WIP: create `PISA` dialect (+ emitter and passes)

AlexanderViand-Intel:
WIP, as it needs to be updated to reflect the recent test/src folder re-org properly + has a few outstanding ToDos.

* Adds a `PISA` target/exit dialect based on PICA (https://eprint.iacr.org/2024/707) and the Intel HERACLES SDK (https://github.com/IntelLabs/hec-p-isa-tools/)
* Adds an emitter from the PISA dialect to a `*.csv` format used in the Intel HERACLES SDK.
* Adds a (WIP) conversion from `Polynomial` to `PISA`.


ai-mannamalai:
nice; just curious to know how KERN_GEN and KERN would be positioned (if it were a MLIR dialect) relative to P-ISA dialect ?

AlexanderViand:
> nice; just curious to know how KERN_GEN and KERN would be positioned (if it were a MLIR dialect) relative to P-ISA dialect ?

I think kerngen would be roughly equivalent not to a new dialect but a lowering from ckks directly to pisa

Issue #1045, opened 2024-10-13
build: create build macros


asraa:
for cmake:
- [ ] add_heir_transforms

for bazel: 
- [ ] investigate whether it's worth creating bazel macros to avoid some boilerplate around dialect / pass generation

j2kun:
@AlexanderViand-Intel maybe you can explain what part of the boilerplate seems confusing? Is it the invocation of `gentbl_cc_library`? as in https://github.com/google/heir/blob/90796977be9b7da90f5dd8bd8a9fbfc75039cc33/lib/Dialect/Secret/Transforms/BUILD#L156C1-L156C18

AlexanderViand-Intel:
Less "confusing", just somewhat verbose compared to `add_heir_dialect(DialectName)` ;)

The "convenience factor" that I like is that the CMake macro automatically generates inc files for ops, types, attributes, enums, etc based on whether the corresponding *.td file exits (plus, at least for passes, a few additional rules based on observed use patterns). So there's less changing of build files when you add/remove *.td files.

j2kun:
I think I can do that. Only caveat I can see rn is that the dependencies will have to be specified outside the macro.

asraa:
Should we close this given the new bazel macros (and deprioritized cmake)? @j2kun

PR #1055, opened 2024-10-15
add_heir_transforms build macro for bazel

j2kun:
Part of https://github.com/google/heir/issues/1045#issuecomment-2412100449

The part that I don't like about this is how much it hides from the person using the macro. How are they supposed to know, for example, that the generated target is `pass_inc_gen`? Among all these other "smart defaults", it adds more of a burden on the coder to keep the defaults in the back of their mind, or else re-read the source of the macro (on top of knowing what tblgen does).

### DIFF
--- a/lib/Transforms/transforms.bzl
+++ b/lib/Transforms/transforms.bzl
@@ -1,0 +1,101 @@
+"""Macros to streamline the generation of HEIR transform tablegen-generated
+files."""
+
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library")
+
+def add_heir_transforms(
+        pass_name = None,
+        td_file = None,
+        header_filename = None,
+        registration_name = None,
+        doc_filename = None,
+        generated_target_name = "pass_inc_gen",
+        name = None,
+        deps = None):
+    """A HEIR standard way to invoke tablegen for transforms.
+
+    This macro calls `gentbl_cc_library` with appropriate parameters, generating:
+
+    - A header file, called `{pass_name}.h.inc` by default.
+    - A documentation file, called `${pass_name}Passes.md` by default.
+    - A bazel target containing the generated headers, called `pass_inc_gen` by
+      default.
+
+    Args:
+      td_file: A string containing the path to the tablegen file to process. Uses
+        "{pass_name}.td" by default. This file should contain a `def`
+        inheriting from `Pass` for each pass defined. Multiple passes may be
+        defined in the same file.
+      pass_name: A string containing the name of the pass to generate.
+        This produces:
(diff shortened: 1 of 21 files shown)


AlexanderViand-Intel:
I agree there's a cost here in terms of "smart defaults". I'd be willing to say that for Dialects, we *should* enforce a strict naming pattern and structure/organization. Imho, having a macro that essentially forces you stick to that naming scheme is worth it there. If this works similarily to the cmake one (checking which files exists/etc), we could also simplify the template scripts a bit: simply generate all *.td files and put a `//FIXME: Simply delete this file if you do not want to add <Thing-this-file-is-for>`.  Without the "magic macro" approach, you'd have to either specify ahead of time which files you want, or need to know how to remove/add them to both `BUILD` and `CMakeLists.txt`.

This isn't just theory - while rebasing this PR, I've alread benefitted quite a bit from the macros, as various conflicts due to additions of Traits/Enums/etc to passes could be resolved by simply ignoring those changes as the macro automagically picked up the added files.

I do agree on the fact that the `inc_gen` /`IncGen` (we should unify this across bazel/cmake anyway) suffix is coming out of nowhere, and it's also not super clear that `add_heir_<thing>` is only for tablegen include generation. For now, how about we name it something explicit such as `add_dialect_incgen`? I picked the more generic name in CMake, as I was thinking about merging this functionality with the existing `add_mlir_thing` macros to make a single `add_heir_thing` macro.

For passes (both dialect-associated conversions/transforms and free-floating transforms) I think I agree - the naming conventions there seem a lot less clear cut, so while I still very much like the convenience (and we can use a more explicit name there, too), unless we think we want to enforce a clearer naming/structure convention on asses, I'm not going to fight too hard for that specific macro.

j2kun:
Added explicit `generated_target_name` to al the invocations, and squashed.

asraa (approved):
LGTM!

AlexanderViand-Intel (approved):

lib/Transforms/transforms.bzl
```diff
@@ -0,0 +1,91 @@
+"""Macros to streamline the generation of HEIR transform tablegen-generated
+files."""
+
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library")
+
+def add_heir_transforms(
+        pass_name = None,
```
asraa, line 7: could you do me a huge favor to save time merging this, but can you add a `name` keyword argument to make buildifier happy? All macros need one (it can be unused and optional)

j2kun: Will do later tonight

j2kun: Done!

PR #1062, opened 2024-10-21
docs: use openfhe context helper for dot_product

ZenithalHourlyRate:
Issue #661 should be safe to close now


asraa (approved):
Thank you thank you!

j2kun (approved):

PR #1074, opened 2024-11-05
fix: fix heir magic to capture output for colab

asraa:
previously, using the custom magic command for heir-opt and heir-translate did not print any output from the subprocess command running the binaries in google colab. it did work in external jupyter notebooks.

it turns out that the subprocess outputs weren't explicitly captured - google colab is on an older version of subprocess that may be missing some kinds of standard outs. https://ipykernel.readthedocs.io/en/stable/changelog.html#id268

@AlexanderViand-Intel

### DIFF
--- a/scripts/jupyter/heir_play/heir_opt.py
+++ b/scripts/jupyter/heir_play/heir_opt.py
@@ -28,6 +28,7 @@ class BinaryMagic(Magics):
             [os.path.abspath(self.binary_path)] + shlex.split(line),
             input=cell,
             text=True,
+            capture_output=True,
         )
         if completed_process.returncode != 0:
             print(f"Error running {self.binary_path}")


AlexanderViand-Intel (commented):
Thanks a lot! I can't believe 1 line was all it took xD

scripts/jupyter/heir_play/heir_opt.py
```diff
@@ -28,6 +28,7 @@ def run_binary(self, line, cell):
             [os.path.abspath(self.binary_path)] + shlex.split(line),
             input=cell,
             text=True,
+            capture_output=True,
```
ruidazeng, line 31: Is it really the best to use `capture_output=True` here? Capturing the output here means that the entire stdout and stderr are read into memory,  which might lead to the binary to produce a large amount of output.  Without this we could allow the output to be streamed directly to the console.

j2kun: What console? This is intended for use in a notebook environment, and this change is specifically to work with Google colab, which would be unusable without this change, until they update their dependencies.

AlexanderViand-Intel:
Btw, I think the pypi package needs an update to incorporate this fix? @j2kun

j2kun:
> Btw, I think the pypi package needs an update to incorporate this fix? @j2kun

Done! 0.0.4 is on pypi now.

Issue #1084, opened 2024-11-11
[modarith] Add a modarith type to describe storage type & modulus


asraa:
This would remove the mod arith op attributes tracking the modulus.

We realized this would be a nice "type" to have to describe the coefficient type of a ring. In the case of working with $$\mathbb{Z}_q$$, the coefficient type would be a modarith type like `modarith<type = i32, modulus = q>`. RNS types would be a list of modarith types.

See the type refactor here: https://github.com/google/heir/pull/1083

j2kun:
Reopening for the remaining TODO

PR #1088, opened 2024-11-12
Implement ModArithType for `mod_arith` dialect

ZenithalHourlyRate:
See #1084

This PR is in draft state, opened here for reviewing the `ModArithType` type itself.

It is worth noting is that the type explicitly does not allow modulus value to be too large to fill the entire underlying type as there would be signedness issues. As we are working in FHE situations where modulus width typically goes up to only 60 for 64bit native int, this requirement does not harm the functionality and edge cases for lowering to native int could be avoided.

Later pushes will replace all existing `ModArithOps` to use the `ModArithType` and corresponding `ModArithToArith` lowering. Currently `mreduce` and `madd` serves only as a demo for the type verification and would be renamed to `reduce`/`add` once finished migration.

Syntax now:

```mlir
!Zp1 = !mod_arith.mod_arith<modulus = 65537 : i32>

module {
  func.func @dot_product() -> !Zp1 {
    %c = arith.constant 1 : i32
    %0 = mod_arith.mreduce %c : i32 -> !Zp1
    %1 = mod_arith.madd %0, %0 : !Zp1
    return %1 : !Zp1
  }
}
```


j2kun:
IMO this looks great. I think the use of the `: i64` type specifier for the "storage type" is a nice way to avoid extra verbosity.

lib/Dialect/ModArith/IR/ModArithTypes.td
```diff
@@ -0,0 +1,48 @@
...
+    ```
+    !Zp1 = !mod_arith.mod_arith<modulus = 7> // implicitly being i64
+    !Zp2 = !mod_arith.mod_arith<modulus = 65537 : i32>
+    !Zp3 = !mod_arith.mod_arith<modulus = 536903681 : i64>
+    ```
+  }];
+  let parameters = (ins
+    "::mlir::IntegerAttr":$modulus
+  );
+  let assemblyFormat = "`<` struct(params) `>`";
```
j2kun, line 43: If we don't think we will have additional parameters, we could even try

```suggestion
  let assemblyFormat = "`<` $modulus `>`";
```

And it would skip the `modulus = ` part, I think.

ZenithalHourlyRate:
Encountered issues of integer range when trying to migrate `mod_arith.barrett_reduce` and `subifge` because barett reduction will give a result in `[0, 2p)`, and `[p, 2p)` are not represented in `mod_arith.mod_arith` type.

HEaaN.mlir resolves such issue by using a `modf`

> Both ModArith and Poly operations may optionally carry modulus factor (modf ). If an instruction is tagged with a modulus factor 𝑓𝑖, its output is allowed to fit in [0, 𝑓𝑖· 𝑝) where 𝑝 is the exact modulus. Assigning a larger 𝑓𝑖 generates a faster loop because SubIfGE in Barrett reduction can be removed. We describe our algorithm for tagging modulus factors of instructions at Sec. 6.1.

Should we annotate the `modf` in the op or in the type? I think the former is easier to implement and can be done on the fly by the analysis pass, and lowering can use such information to do a cheaper lowering. We can also verify how many `modf` are availble given the storage type and the modulus in the analysis pass. These are all for later PRs.

For this PR. I think migrating basic ops like add/sub/mul with basic lowering is enough. And as barett_reduce is not used for now (though there was a PR #712 doing such transform), I'll just use the ModArithType for `barrett_reduce` without ensuring its range.

tests/Dialect/ModArith/Conversions/mod_arith_to_arith/runner/lower_add.mlir
```diff
@@ -17,4 +25,4 @@ func.func @test_lower_add() {
   return
 }
 
-// CHECK_TEST_ADD: [1225, 7276, 7645, 0]
+// CHECK_TEST_ADD: [1258, 7276, 7678, 0]
```
ZenithalHourlyRate, line 28: I was actually confused by the original runner result. The original result is not correct for either signed interpretation (this PR's semantic) or unsigned interpretation (original semantic).

For example, `67108862 = -2` and `67108863 = -1` for i26, yet there addition result modulo 7681 should not be 7645.

```python
>>> (67108863 + 67108862) % 7681
7612
>>> (-1 + -2) % 7681
7678
```

Though the first result is correct: we have `36789492 = -30319372` and `29498763`,

```python
>>> (36789492 + 29498763) % 7681
1225
>>> ((-30319372) + 29498763) % 7681
1258
```


Then the first result is incorrect for `lower_mul.mlir`, 

```python
>>> (36789492 * 29498763) % 7681
1637
>>> ((-30319372) * 29498763) % 7681
1600
```

ZenithalHourlyRate: Oh..in the original semantic, mod_arith.add/sub/mul require the input to be in [0, q), and the runner tests violated that constraint.

j2kun:
> Should we annotate the `modf` in the op or in the type? I think the former is easier to implement and can be done on the fly by the analysis pass, and lowering can use such information to do a cheaper lowering. We can also verify how many `modf` are availble given the storage type and the modulus in the analysis pass. These are all for later PRs.
> 
> For this PR. I think migrating basic ops like add/sub/mul with basic lowering is enough. And as barett_reduce is not used for now (though there was a PR #712 doing such transform), ~I'll just use the ModArithType for `barrett_reduce` without ensuring its range~, I'll leave it as is so that we can have a useable ModArithType earlier.

Your reasoning is sound for the decision here. From my perspective, the main trade-off here is the inefficiency loss by incurring an extra field on every `mod_arith` type vs the extra work of lowering that type. The former doesn't seem that bad alone, but the worry is always that 5 or 10 more such optimizations will lead to 5 or 10 more fields on the type, so having a systematic strategy to avoid that is good.

If we put it on the ops, I just want to quickly sketch out the extra work that would require. Say we want to lower `modarith<5 : i3>` and there's a factor of 3 on some ops. The pass ultimately needs to be able to type-convert types without knowing which op it's on. So the pass setup would need to first run an analysis over the IR (or, say, at the func level) to determine the appropriate storage type to lower to. This might be as simple as just a `max` over modulus factors found across all ops, in which case perhaps we could reuse int range analysis from upstream. In this example, the analysis would determine we lower to `i5` (because `fi*p = 3*5 = 15` fits in an `i4` but is too close to the max), and then that width would be given to the TypeConverter's constructor, and it would use that instead of the `: i3` annotation when converting the type (the analysis would default to a width of 3 if no modulus-factors are found among the relevant ops).

lib/Dialect/ModArith/IR/ModArithOps.td
```diff
@@ -15,63 +16,71 @@ class ModArith_Op<string mnemonic, list<Trait> traits = [Pure]> :
...
+  );
+  let results = (outs ModArithLike:$output);
+  let hasVerifier = 1;
+  let assemblyFormat = "operands attr-dict `:` type($input) `->` type($output)";
+}
+
+def ModArith_ExtractOp : ModArith_Op<"extract", [Pure, ElementwiseMappable]> {
+  let summary = "extract the integer value stored";
+
+  let description = [{
```
j2kun, line 36: Please add a description with an example syntax. I think the main goal of this op is to type convert, and it would ultimately be "free" (lower to zero machine instructions).

lib/Dialect/ModArith/IR/ModArithOps.h
```diff
@@ -3,11 +3,44 @@
...
 #define GET_OP_CLASSES
 #include "lib/Dialect/ModArith/IR/ModArithOps.h.inc"
 
+namespace mlir {
+namespace heir {
+namespace mod_arith {
+
+template <typename OpType>
+inline ModArithType getResultModArithType(OpType op) {
+  return llvm::cast<ModArithType>(
```
j2kun, line 20: Nit: I believe you can omit the `llvm::` prefix because the mlir namespace re-imports them under `mlir::`

```suggestion
  return cast<ModArithType>(
```

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -14,23 +16,208 @@ namespace mod_arith {
...
+  APInt modulus = type.getModulus().getValue();
+  return IntegerType::get(type.getContext(), modulus.getBitWidth());
+}
+
+Type convertModArithLikeType(ShapedType type) {
+  if (auto modArithType = llvm::dyn_cast<ModArithType>(type.getElementType())) {
+    return type.cloneWith(type.getShape(), convertModArithType(modArithType));
+  } else {
+    return type;
+  }
```
j2kun, line 29: tiny readability nit:

```suggestion
  }
  
  return type;
```

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -14,23 +16,208 @@ namespace mod_arith {
...
+    addArgumentMaterialization([](OpBuilder &builder, Type type,
+                                  ValueRange inputs, Location loc) -> Value {
+      return builder.create<ReduceOp>(loc, type, inputs[0]).getResult();
+    });
+
+    // Convert from a mod arith type to an integer type : use mextract
+    addTargetMaterialization([](OpBuilder &builder, Type type,
+                                ValueRange inputs, Location loc) -> Value {
+      return builder.create<ExtractOp>(loc, type, inputs[0]).getResult();
+    });
```
j2kun, line 57: I suspect these lines could be deleted without impacting the lowering at all; since all modarith ops are lowered by the patterns, any `unrealized_conversion_cast` ops that would be inserted by the absence of these lines would also be cleaned up.

Also, inserting `reduce` ops suggests nonzero work being applied (though cf. my question in the other comment about the semantics of reduce), and I see this type materialization stuff as temporary workarounds for the chicken-and-egg problem of dialect lowering. If, for some unexpected reason, these materialization ops stick around after the lowering is done, we'd prefer to be notified with an error (unresolved `unrealized_conversion_cast` error at the end of the lowering) than to have them quietly lowered to ops that use actual CPU cycles.

```suggestion
```

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -14,23 +16,208 @@ namespace mod_arith {
...
+  }
+};
+
+template <typename Op>
```
j2kun, line 61:
```suggestion
// A helper function to generate the attribute or type needed to represent
// the result of modarith op as an integer before applying a remainder 
// operation.
template <typename Op>
```

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -14,23 +16,208 @@ namespace mod_arith {
...
+  LogicalResult matchAndRewrite(
+      ReduceOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    auto cmod = b.create<arith::ConstantOp>(modulusAttr(op));
+    // ModArithType ensures cmod can be correctly interpreted as a signed number
+    auto rems = b.create<arith::RemSIOp>(adaptor.getOperands()[0], cmod);
+    auto add = b.create<arith::AddIOp>(rems, cmod);
+    // TODO: better with a subifge
```
j2kun, line 103: Please attach the TODOs to an issue, here and below.

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
j2kun, line 1: I'm curious what was stopping you from continuing to use the tablegen definitions for these patterns? I forget, but perhaps it's because DRR doesn't support generating dialect conversion patterns?

j2kun (changes requested):
Looks fantastic! My main critique is that we should be clear at the start about the operation semantics in the tablegen description fields, as that will be the source of truth when folks (particularly me) forget what we decided on.

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
ZenithalHourlyRate: DRR seems not meant for OpConversionPattern, especially with the TypeConverter/Adaptor thing.

j2kun: Ah yes, now I remember. Thanks!

j2kun (approved):

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -14,23 +16,207 @@ namespace mod_arith {
...
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    auto cmod = b.create<arith::ConstantOp>(modulusAttr(op, true));
+    auto x = b.create<arith::ExtUIOp>(modulusType(op, true),
+                                      adaptor.getOperands()[0]);
+    auto y = b.create<arith::ExtUIOp>(modulusType(op, true),
+                                      adaptor.getOperands()[1]);
+    auto acc = b.create<arith::ExtUIOp>(modulusType(op, true),
+                                        adaptor.getOperands()[2]);
```
asraa, line 209: just for later - the OpAdaptor comes with named accessors for the operands (like adaptor.getLhs(), adaptor.getRhs(), adapator.getAcc())

asraa (approved):

PR #1090, opened 2024-11-13
Add mlir-to-secret-arith pipeline

ZenithalHourlyRate:
Split the scheme-agnostic part of existing arith FHE pipeline into a standalone pipeline so that the user can inspect scheme-agnostic optimization result HEIR has done, instead of inspecting the verbose bgv/openfhe dialect.

Before, to check scheme-agonistic optimizations, the user has to call a chain of passes like `--secretize --wrap-generic -heir-simd-vectorizer`, and with a bunch of `cse/canonicalize`.

BTW, I think some FHE ciphertext management operation can be _annotated_ here, like relinearize/mod reduce operation. For example, the lazy realinearization pass operates on the bgv dialect, it deletes the existing relin op then adds them back according to the analysis, which requires bgv type fixing. If annotated early, the secret-to-bgv lowering can directly produce correct type.

### DIFF
--- a/lib/Pipelines/ArithmeticPipelineRegistration.cpp
+++ b/lib/Pipelines/ArithmeticPipelineRegistration.cpp
@@ -71,9 +71,8 @@ void heirSIMDVectorizerPipelineBuilder(OpPassManager &manager) {
   manager.addPass(createCSEPass());
 }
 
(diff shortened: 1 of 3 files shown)


j2kun (approved):

j2kun:
LGTM! Thanks for this

Issue #1086, opened 2024-11-12
Issues with `--comb-to-cggi` when lowering secret loop-carried dependence


raghav198:
I am trying to lower an `affine.for` loop with a secret loop carried dependence that gets captured by a `secret.generic` block in the loop body as in the example below:

```
  func.func @sum(%arr: !secret.secret<memref<8xi8>>) -> !secret.secret<i8> {

      %c0 = arith.constant 0 : i8
      %s0 = secret.conceal %c0 : i8 -> !secret.secret<i8>

      %ans = affine.for %i = 0 to 8 step 1 iter_args(%acc = %s0) -> (!secret.secret<i8>) {
          %new_acc = secret.generic ins(%arr, %acc: !secret.secret<memref<8xi8>>, !secret.secret<i8>) {
              ^bb0(%ARR: memref<8xi8>, %ACC: i8):
                  %cur = memref.load %ARR[%i] : memref<8xi8>
                  %sum = arith.addi %cur, %ACC : i8
                  secret.yield %sum : i8
          } -> (!secret.secret<i8>)
          affine.yield %new_acc : !secret.secret<i8>
      }
      return %ans : !secret.secret<i8>
  }
```

Currently I'm running the above through the `--yosys-optimizer` and `--secret-distribute-generic` passes, which both work, but when trying to further lower via `--comb-to-cggi` I get the following error:

```
 error: failed to legalize unresolved materialization from ('!secret.secret<i8>') to 'memref<8x!lwe.lwe_ciphertext<encoding = #lwe.unspecified_bit_field_encoding<cleartext_bitwidth = 3>>>' that remained live after conversion
            %new_acc = secret.generic ins(%arr, %acc: !secret.secret<memref<8xi8>>, !secret.secret<i8>) {
```

Digging a bit further, it looks like its unable to transform ops that use the secret iter_arg because it still gets used in the corresponding `affine.yield`, but I'm not sure what I could be doing differently to avoid this, or if this is a ug.

j2kun:
I started from this, ran yosys-optimizer, and cut out a bunch of stuff that seemed irrelevant to the bug, to get a smaller reproducer which fails when running `--secret-distribute-generic --secret-to-cggi`

```mlir
module {
  func.func @sum(%arg0: !secret.secret<memref<8xi8>>) -> !secret.secret<i8> {
    %true = arith.constant true
    %c7 = arith.constant 7 : index
    %c6 = arith.constant 6 : index
[...]

PR #1092, opened 2024-11-13
secret-to-cggi: lower conceal, affine for/yield

j2kun:
Fixes https://github.com/google/heir/issues/1086


asraa (commented):
Ah, thank for making this! The general idea looks good to me - I realize the secret.conceal came about when I was distributing secret generic through the affine for op and was never handled in lowering to secret-to-cggi.

j2kun:
> Ah, thank for making this! The general idea looks good to me - I realize the secret.conceal came about when I was distributing secret generic through the affine for op and was never handled in lowering to secret-to-cggi.

The bigger difficulty here was actually lowering the loop-carried dependence, and without a special rule for lowering affine.for and affine.yield it just won't work.

j2kun:
I was able to simplify this a bit by reusing `ConvertAny` for the affine for/yield type conversions (and it gave me the needed materialization op for free. The main semantic difference now is that the greedy runner runs at the end of `secret-to-cggi`, which runs all folders and caused some test assertions to fail because the code became more efficient by removing some statements of the form `arith.select %cond, %val, %val` (just use `%val` and drop the op).

asraa (approved):

PR #1096, opened 2024-11-16
Convert polynomial coefficient type to use modarith

j2kun:
Major changes in this PR:

- coefficientModulus attribute removed from ring attribute
- coefficientType relaxed to allow AnyType
- polynomial op verifiers enhanced to ensure scalar and polynomial coefficient types agree
- polynomial-to-standard renamed to polynomial-to-mod-arith, and most lowerings emit mod_arith ops
- SubAsAdd canonicalization pattern removed; lowering subtraction directly is more efficient.
- Copied additional upstream polynomial tests that I forgot to in my last PR (e..g, tests/Dialect/Polynomial/IR/errors.mlir)

Shortcomings:

- polynomial-to-mod-arith will fail for basically all polynomial types with non-mod-arith coefficient types. Since we don't have a use case for lowering int/float polynomials, I'm OK with this state of affairs indefinitely.
- The NTT lowering is mostly unchanged, and I just take the mod-arith types, unpack them to their underlying integer type/value, and repack it back up as a mod-arith at the end. I just didn't have it in me to re-write the entire NTT lowering in this already gargantuan PR.


lib/Dialect/Polynomial/IR/PolynomialCanonicalization.td
```diff
@@ -1,25 +1,54 @@
...
+>;
+
+def SubAsAddModArith : Pat<
   (Polynomial_SubOp $f, $g),
   (Polynomial_AddOp $f,
     (Polynomial_MulScalarOp $g,
-      (Arith_ConstantOp (getMinusOne $g))))>;
+        (ModArith_ConstantOp (getMinusOneModArith $g)))),
+  [(isModArithCoeffType $g)]
+>;
```
ZenithalHourlyRate, line 51: I think `sub = add (mulScalar)` is costly as `modarith.sub` is much cheaper than `modarith.add (modarith.mul).` As now the `ConvertAdd` in `PolynomialToStandard` is simple enough, we can have a corresponding `ConvertSub`

j2kun: Removed!

lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
```diff
@@ -203,33 +204,6 @@ Attribute FloatPolynomialAttr::parse(AsmParser &parser, Type type) {
   return FloatPolynomialAttr::get(parser.getContext(), result.value());
 }
 
-LogicalResult RingAttr::verify(
```
WoutLegiest, line 206: Why is this verifier removed? Is this replaced with another one?

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/lower_mul.mlir
```diff
@@ -50,87 +48,45 @@
...
 // CHECK: }
 
 func.func @lower_poly_mul(%poly0: !poly_ty, %poly1: !poly_ty) -> !poly_ty {
   %poly2 = polynomial.mul %poly0, %poly1 : !poly_ty
   return %poly2 : !poly_ty
 }
-
-// -----
-
-// Test a non-machine-word sized modulus to ensure it remsi's correctly
```
WoutLegiest, line 95: Why was this test removed?

WoutLegiest (approved):
LGTM, lot is boilerplate changes of types, namespaces and conversion to the new types

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/lower_mul.mlir
j2kun: There is no longer a remsi expected here, since the output of this pass is in mod_arith, so the mod_arith lowerings assert that remainder ops are inserted correctly. So this was a check for certain behavior that has been removed

PR #1099, opened 2024-11-19
Not packed integration into the BoolVectorizer

WoutLegiest:
Additional code for not gates into the boolvectorizer
Adding debug code; left to use to build the new version of the vectorizer
Small refactors of function names

### DIFF
--- a/lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
+++ b/lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
@@ -1,5 +1,6 @@
 #include "lib/Dialect/CGGI/Transforms/BooleanVectorizer.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cstdint>
 #include <vector>
@@ -50,7 +51,7 @@ bool areCompatibleBool(Operation *lhs, Operation *rhs) {
   return OpTrait::hasElementwiseMappableTraits(lhs);
 }
 
-FailureOr<SmallVector<Attribute>> BuildGateOperands(
+FailureOr<SmallVector<Attribute>> buildGateOperands(
     const SmallVector<Operation *> &bucket, MLIRContext &context) {
   SmallVector<Attribute> vectorizedGateOperands;
   for (auto *op : bucket) {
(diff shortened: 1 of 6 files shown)


lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
```diff
@@ -139,36 +142,58 @@ int bucketSize(const SmallVector<SmallVector<Operation *>> &buckets) {
...
 }
 
+int countNotOps(DenseMap<Operation *, SmallVector<SmallVector<Operation *>>>
+                    compatibleOps) {
+  int count = 0;
+  DenseMap<Operation *, SmallVector<SmallVector<Operation *>>>::iterator it;
+
+  for (it = compatibleOps.begin(); it != compatibleOps.end(); it++) {
+    if (isa<cggi::NotOp>(it->first)) count++;
+  }
```
j2kun, line 152: I would try using `llvm::count_if` if possible, then this entire function might be small enough to inline at the sole call site.

https://github.com/llvm/llvm-project/blob/197e0125c3ed991f11d97d27e529c02af06e9f37/llvm/include/llvm/ADT/STLExtras.h#L1945

lib/Pipelines/BooleanPipelineRegistration.cpp
```diff
@@ -156,8 +157,9 @@ void registerTosaToBooleanFpgaTfhePipeline(const std::string &yosysFilesPath,
                                   /*abcBooleanGates=*/true);
 
         // Vectorize CGGI operations
-        pm.addPass(createStraightLineVectorizer(
-            StraightLineVectorizerOptions{.dialect = "cggi"}));
+        // pm.addPass(createStraightLineVectorizer(
+        //     StraightLineVectorizerOptions{.dialect = "cggi"}));
```
j2kun, line 161:
```suggestion
```

j2kun (approved):
LGTM!

lib/Dialect/CGGI/Transforms/BooleanVectorizer.cpp
WoutLegiest: Changed it. 
Cool didn't know that was a thing!

PR #1114, opened 2024-11-23
RNS: allow mod arith types as basis types

### LINKED ISSUE #1098: [modarith] Implement RNSBasisTypeInterface on mod arith types

ZenithalHourlyRate:
Fixes #1098

As the coprime relationship is not associative, the description is changed accordingly.


lib/Dialect/RNS/IR/RNSTypes.cpp
```diff
@@ -60,11 +82,42 @@ struct PolynomialRNSBasisTypeInterface
...
+    auto other = mlir::dyn_cast<ModArithType>(otherRnsBasisType);
+    if (!other) {
+      return false;
+    }
+
+    APInt thisModulus = thisType.getModulus().getValue();
+    APInt otherModulus = other.getModulus().getValue();
+
+    // require same bitwidth
+    if (thisModulus.getBitWidth() != otherModulus.getBitWidth()) {
```
j2kun, line 103: I'm slightly surprised this suffices to catch when `getModulus().getType()` are different, since I thought the APInt uses a minimum of 64 bits for its storage type, no matter the underlying value. I see the test that checks this, but I would also wonder if checking `getModulus().getType()` equality here would be safest.

tests/Dialect/RNS/IR/syntax.mlir
```diff
@@ -22,3 +22,32 @@ func.func @test_syntax(%arg0: !ty) -> !ty {
...
+!Zp1 = !mod_arith.int<3721063133 : i64>
+!Zp2 = !mod_arith.int<65537 : i32>
+
+// expected-error@+1 {{RNS type has incompatible basis types}}
+!ty_modarith_bad = !rns.rns<!Zp1, !Zp2>
+
+// -----
+
+// expected-error@+1 {{does not have RNSBasisTypeInterface}}
+!ty_int_bad = !rns.rns<i32, i64>
```
j2kun, line 53: Please add a test where it succeeds as well.

j2kun (approved):

tests/Dialect/RNS/IR/syntax.mlir
ZenithalHourlyRate: I think there is no case for i32 to pass the check?

j2kun: Sorry, I meant for mod_arith types. All tests added in this PR, even for mod arith, check error cases.

ZenithalHourlyRate: Added a succeeding case for storage type check.

j2kun (approved):

PR #1118, opened 2024-11-25
Small `Secretness` improvements: `{secret.secret}` annotation awareness and an `--annotate-secretness` pass 

AlexanderViand-Intel:
Adds a simple pass (based on the existing debug code for `--convert-if-to-select`) that adds the secretness to each op as an attribute, primarily for ease of debugging/explanation. This PR also makes `SecretnessAnalysis` aware of the `{secret.secret}` annotations, so that one does not need to run `--wrap-generic` beforehand in order to get correct results.

Example output:
```llvm 
  func.func @example(%arg0: i32 {secret.secret}) -> i32 {
    %c1_i32 = arith.constant {secretness = false} 1 : i32
    %0 = arith.addi %c1_i32, %arg0 {secretness = true} : i32
    return %0 : i32
  }
```


lib/Transforms/AnnotateSecretness/AnnotateSecretness.td
```diff
@@ -0,0 +1,14 @@
+#ifndef LIB_TRANSFORMS_ANNOTATESECRETNESS_ANNOTATESECRETNESS_TD_
+#define LIB_TRANSFORMS_ANNOTATESECRETNESS_ANNOTATESECRETNESS_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def AnnotateSecretness : Pass<"annotate-secretness"> {
+  let summary = "Apply all folding patterns from canonicalize";
```
j2kun, line 7:
```suggestion
  let summary = "Annotate secret SSA values in the IR";
```

lib/Transforms/AnnotateSecretness/AnnotateSecretness.cpp
```diff
@@ -0,0 +1,60 @@
...
+      signalPassFailure();
+      return;
+    }
+
+    // Add an attribute to the operations to show determined secretness
+    OpBuilder builder(context);
+    getOperation()->walk([&](Operation *op) {
+      if (op->getNumResults() == 0) return;
+      auto *secretnessLattice =
+          solver.lookupState<SecretnessLattice>(op->getResult(0));
```
j2kun, line 42: This is a debugging pass, but it might be worth making this a bit more complete with respect to multi-result ops.

Maybe loop over op results and add the result index to the attribute name?

`result_{i}_secretness`

Or maybe only do this if there is more than one result.

j2kun (approved):

lib/Transforms/AnnotateSecretness/AnnotateSecretness.cpp
AlexanderViand-Intel: Ohh, good point - I'll add that tomorrow!

AlexanderViand-Intel: Done :)

AlexanderViand-Intel:
CI fails because `main` fails to build at the moment (@j2kun?):

`ERROR: Skipping '//...': error loading package under directory '': error loading package 'scripts/jupyter': Label '//learning/grp/pynball:build_defs.bzl' is invalid because 'learning/grp/pynball' is not a package; perhaps you meant to put the colon here: '//:learning/grp/pynball/build_defs.bzl'?`

j2kun (approved):

PR #1120, opened 2024-11-25
bump WORKSPACE versions to avoid protobuf errors

j2kun:
Figured out the problem. @AlexanderViand-Intel it was because one of the earlier dependencies (rules_proto) was migrated into the protobuf main tree and deprecated itself, but it also contained a hard-coded a transitive dependency on an older pinned version of protobuf. So removing that dependency fixes the problem, and then to be safe I bumped a bunch of versions to make them all consistent (e.g., update to the protobuf/absl version expected by ortools), and moved the protobuf dep earlier in the workspace file to help prevent other hidden transitive dependencies from overwriting the protobuf dep.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -33,6 +33,46 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
+# Depend on a hermetic python version
+new_git_repository(
+    name = "rules_python",
+    commit = "9ffb1ecd9b4e46d2a0bca838ac80d7128a352f9f",  # v0.23.1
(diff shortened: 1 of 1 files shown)


WoutLegiest (approved):
Nice, I also had found this problem, though it was a problem om my machines protobuf. Only had trouble when running the `bazel run @hedron_compile_commands//:refresh_all` command.

asraa (approved):

AlexanderViand-Intel (approved):
Thanks for figuring this out! I figured out bumping protobuf didn't solve it but gave up when I realized I'd have to bump OR-tools and all its dependencies. Is there some kind of tool to update the commit hashes automatically, or did you have to manually dig them up?

PR #1125, opened 2024-11-27
Arith to ModArith Conversion pass and Mac Transformer pass

WoutLegiest:
- Adding conversion pass to convert arith into mod arith; now limited operations are supported. 
   Conversion is place in the Arith folder, but lives in the heir::arith dialect  
- Transformation pass in mod-arith to find consecutive mul and add operations and replace them with a Mac operation
- Pipeline pass for Tosa to secret/Arith i32

The arith-to-mod-arith pass is required to lower a neural network TOSA model to a CGGI backend. This pass will transform the operations to the mod-arith dialect, where the find-mac pass can be used to convert consecutive multiply addition operations into a single operation. In a later pass (ongoing work), these large precision MAC operations (typically 64 or 32-bit) will be lowered into small precision (8 or 4b) operations that can be mapped to CGGI operations.


j2kun:
Looks like some tests are failing. Could you fix that and then send the PR to me for review?

j2kun:
One thing I think would be helpful is an explanation (perhaps in the PR description, maybe also in the tablegen pass description field) of the broader reason for having this pass. What e2e use case is it intended to support, and what stage of that e2e use case is it managing?

tests/Dialect/Arith/Conversions/ArithToModArith/arith-to-mod-arith.mlir
```diff
@@ -0,0 +1,55 @@
+// RUN: heir-opt --arith-to-mod-arith --split-input-file %s | FileCheck %s --enable-var-scope
+
+// CHECK-LABEL: @test_lower_add
+// CHECK-SAME: (%[[LHS:.*]]: [[T:.*]], %[[RHS:.*]]: [[T]]) -> [[T]] {
```
j2kun, line 4: can you change at least one of these tests to specify exactly what the lowered type should look like? Since arith -> mod_arith is making a nontrivial choice here, it would be good to formalize it in the test.

lib/Dialect/ModArith/Transforms/ConvertToMac.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+struct FindMac : public OpRewritePattern<mod_arith::AddOp> {
+  FindMac(mlir::MLIRContext *context)
+      : OpRewritePattern<mod_arith::AddOp>(context) {}
+
+  LogicalResult matchAndRewrite(mod_arith::AddOp op,
+                                PatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    auto *parent = op.getLhs().getDefiningOp();
```
j2kun, line 25: What stops it from being the Rhs? (Maybe a canonicalization pattern I missed?) You could make a second pattern, or else have something like

`Value otherArg = op.getRhs()`

and have a branch that swaps them if the other parent is a mul.

j2kun, line 25: You can use `getDefiningOp<mod_arith::MulOp>` to cast and fetch at the same time, and a null check is all that is needed

https://github.com/llvm/llvm-project/blob/2dba66b8536506cb2f41a86a01e74819c40c53a4/mlir/include/mlir/IR/Value.h#L148-L151

lib/Dialect/ModArith/Transforms/ConvertToMac.cpp
```diff
@@ -0,0 +1,58 @@
...
+      return failure();
+    }
+
+    auto mul_parent = cast<mod_arith::MulOp>(parent);
+
+    auto result =
+        b.create<MacOp>(mul_parent.getLhs(), mul_parent.getRhs(), op.getRhs());
+
+    rewriter.replaceOp(op, result);
+    rewriter.eraseOp(mul_parent);
```
j2kun, line 37: This is only legal if the parent op has no other uses. Does it still make sense to insert a MAC op if the parent mul result is used in some other location? Then you're just replacing an add op with a MAC op, not strictly reducing op counts.

I think it probably is still useful (in which case do the replacement and conditionally erase), but if not you can alternatively have the match fail if the parent fails `hasSingleUse`.

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace mlir {
+namespace heir {
+namespace arith {
+
+#define GEN_PASS_DEF_ARITHTOMODARITH
+#include "lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.h.inc"
+
+static mod_arith::ModArithType convertArithType(Type type) {
+  auto modulus = type.getIntOrFloatBitWidth();
```
j2kun, line 27: This modulus seems incorrect. An `i32` changes to `mod_arith.int<32 : i32>`? I think it should instead be a modulus of `2**31` (1 + max value of signed int32).

Also note that mod_arith has limitations on the allowed modulus size vs the container type. So if you wanted it to be `2**32` then you'd need to have the container type be at least `i33` or larger.

j2kun (changes requested):

lib/Dialect/ModArith/Transforms/ConvertToMac.cpp
WoutLegiest: Cool, TIL

lib/Dialect/ModArith/Transforms/ConvertToMac.cpp
WoutLegiest: Mhh, the idea of the pass is mostly to find the MAC operations so that they can be lowered more easily. 
For now, I have removed the easeOp, and count that the automatic DCE will remove the multiply operation if it is not required.

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
WoutLegiest: Woops, is indeed quite a big fault. Thanks for catching!

lib/Dialect/ModArith/Transforms/ConvertToMac.cpp
j2kun: It's a common pattern to erase if uses is empty. I think that'd be better here than relying on dce being called later

j2kun:
@WoutLegiest is this ready for a final review?

WoutLegiest:
> @WoutLegiest is this ready for a final review?

Yes it is!

lib/Pipelines/BooleanPipelineRegistration.cpp
```diff
@@ -93,9 +93,9 @@ void tosaToCGGIPipelineBuilder(OpPassManager &pm,
   pm.addPass(createCanonicalizerPass());
 
   // Booleanize and Yosys Optimize
-  pm.addPass(createYosysOptimizer(yosysFilesPath, abcPath, options.abcFast,
-                                  options.unrollFactor, /*useSubmodules=*/true,
-                                  abcBooleanGates ? Mode::Boolean : Mode::LUT));
+  pm.addPass(createYosysOptimizer(
+      yosysFilesPath, abcPath, options.abcFast, options.unrollFactor,
+      /*useSubmodules=*/true, abcBooleanGates ? Mode::Boolean : Mode::LUT));
```
AlexanderViand-Intel, line 98: unrelated to PR? did something change in the formatter?

lib/Utils/ConversionUtils/ConversionUtils.cpp
```diff
@@ -193,7 +193,7 @@ struct ConvertFromElements
         newShape.append(shape.begin(), shape.end());
 
         // Create a dense constant for targetShape
-        auto shapeOp = rewriter.create<arith::ConstantOp>(
+        auto shapeOp = rewriter.create<mlir::arith::ConstantOp>(
```
AlexanderViand-Intel, line 196: Using this instance but there's of course a lot of similar ones: I'm not sure we need to have the `heir::arith` namespace (aka `::mlir::heir::arith`). We already have other passes live in `::mlir::<upstream-dialect>` iirc, so I think we could just put the transform into `::mlir::arith` and avoid a lot of potential for confusion.

tests/Dialect/ModArith/Transforms/find_mac.mlir
```diff
@@ -0,0 +1,23 @@
...
+func.func @double_mac(%arg0: i16) -> i16 {
+  %c1 = arith.constant 1: i16
+  %c2 = arith.constant 2 : i16
+  %c3 = arith.constant 3: i16
+
+  // CHECK: %[[MAC1:.*]] = mod_arith.mac %[[ARG]], %{{.*}}, %{{.*}} : [[T]]
+  // CHECK: %[[ADD:.*]] = mod_arith.add %[[MAC1]], %{{.*}} : [[T]]
+  // CHECK: %[[MAC2:.*]] = mod_arith.mac %[[ADD]], %{{.*}}, %{{.*}} : [[T]]
+  // CHECK: return %[[MAC2:.*]] : [[T]]
+
```
AlexanderViand-Intel, line 14: filecheck test here should assign names to the operands and make sure they are the expected ones :)

AlexanderViand-Intel (commented):
I don't think I fully understand why this TOSA pipeline needs to use `mod_arith` but I already have some ideas for howto (ab)use ArithToModArith so super happy to see it :)

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
```diff
@@ -0,0 +1,122 @@
...
+namespace mlir {
+namespace heir {
+namespace arith {
+
+#define GEN_PASS_DEF_ARITHTOMODARITH
+#include "lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.h.inc"
+
+static mod_arith::ModArithType convertArithType(Type type) {
+  auto modulusBitSize = type.getIntOrFloatBitWidth();
+  auto modulus = (int)(1 << (modulusBitSize - 1)) - 1;
```
j2kun, line 28: I suspect this will overflow in general, and we may want to use APInt to be more precise. But I'm fine merging this as-is for now.

lib/Dialect/ModArith/Transforms/ConvertToMac.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (!parent) {
+      auto parentRhs = op.getRhs().getDefiningOp<mod_arith::MulOp>();
+      if (!parentRhs) {
+        return failure();
+      }
+      parent = parentRhs;
+    }
+
+    auto result =
+        b.create<MacOp>(parent.getLhs(), parent.getRhs(), op.getRhs());
```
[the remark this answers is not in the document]
j2kun: It may be easier to separate it into multiple patterns for each case.

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
j2kun: I think the very last `-1` should be removed. With -1 it's the max value, and without -1 it's the modulus, no?

We may need to use a 64-bit storage type to store the modulus in the IR, but that should be OK...

j2kun (changes requested):

lib/Pipelines/BooleanPipelineRegistration.cpp
WoutLegiest: my offline linter decided that it needed a different alignment

lib/Utils/ConversionUtils/ConversionUtils.cpp
WoutLegiest: It is because the ArithToModArith pass lives in the heir::arith namespace.
It is more to be in the same line as the rest of the conversion passes X-to-Y, they all are in namespace X. 
Do you maybe recall with pass it is that lives in an upstream dialect? I will try to migrate it then.

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
WoutLegiest: The -1 was indeed to keep everuthong stored in a i32 number. 
Changed it, now it stored in 64b

lib/Utils/ConversionUtils/ConversionUtils.cpp
AlexanderViand-Intel: I thought the LinAlg-to-Y pass was in `::mlir::linalg` but that's in `::mlir::heir::linalg`, following the same pattern.
I'm not sure if that's a great pattern to follow though, because it does introduce pretty annoying aliasing issue. @j2kun any thoughts on this?

j2kun: I don't see it as particularly annoying, though I would be open to having more unique names. I don't think it should block this PR

j2kun (approved):

Issue #1091, opened 2024-11-13
Annotate management operations for BGV/CKKS


ZenithalHourlyRate:
I was trying to insert relinearize/modreduce operations using analysis for optimization purpose, and find that after the current `secret-to-<scheme>` lowering it is hard to manipulate the IR (especially we are migrating lwe types; also check #860). Instead we can place these management ops in secret dialect.

For example, the current lazy relinearization pass is implemented in bgv dialect, it can be also applied to ckks. Also, that pass needs to worry about LWE types, which should be handled by lowering passes.

I wrote a demo using the current depth analysis in my [fork](https://github.com/ZenithalHourlyRate/heir/commit/b9a34f58c30a6de8a69ea575990a93ab15a920ec)

```mlir
module {
  func.func @func(%arg0: !secret.secret<i16>, ...) {
    %0 = secret.generic ins(...
    ^bb0(%arg8: i16, ...):
      %1 = arith.muli %arg8, %arg9 {depth = 1 : i64, lhs = 0 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      %2 = arith.muli %1, %arg10 {depth = 2 : i64, lhs = 1 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      %3 = arith.muli %2, %arg11 {depth = 3 : i64, lhs = 2 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      %4 = arith.muli %3, %arg12 {depth = 4 : i64, lhs = 3 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      %5 = arith.muli %4, %arg13 {depth = 5 : i64, lhs = 4 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      %6 = arith.muli %5, %arg14 {depth = 6 : i64, lhs = 5 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      %7 = arith.muli %6, %arg15 {depth = 7 : i64, lhs = 6 : i64, relin = 3 : i64, rhs = 0 : i64} : i16
      secret.yield %7 : i16
    } -> !secret.secret<i16>
    return %0 : !secret.secret<i16>
  }
}
```

If such direction is favored, the following stpes can be done

Relinearize part

* Modify `SecretGenericOpMulConversion` to emit `RelinearizeOp` as needed.
* Modify `OptimizeRelinearization` to work on secret dialect, so that it does not need to worry about LWE types.

Modswitch/Rescale part

[...]

PR #1126, opened 2024-11-28
BGV: place management op and add RNS lowering

ZenithalHourlyRate:
Related to #1091

This is a WIP PR demonstrating the basic implementation form of the FHE management op placement. Requesting for review as there are many design choices that should be made.

This serves as a baseline as we may further implement lazy mod reduce, or for CKKS, microsoft/EVA or other style scale/level management op placement.

Currently we are relying on OpenFHE to do the management op placement for us (inside each op, OpenFHE would try to `AdjustCiphertext`), which the compiler should take hands on itself. This ability would potentially benefit bgv-to-poly lowering path.

For BGV, the very basic constraint is ensuring that the depth/level of FHE arithmetic op operands match (For ciphertext-plaintext op, it means creating an ad-hoc _ciphertext_ with matching level). So this PR adds a very basic placement strategy that when mulDepth mismatches, a mod reduce is inserted.

#### where these management op should be placed

Tries to solve such problem by adding a `mgmt` dialect before going into `BGV/LWE` because the lowering to `BGV/LWE` already needs the depth information (hence the moduli_chain)

#### where the metadata like depth should be stored.

It is discussed not to be in the type as things going on type argument size would explode.

I already annotated the maxMulDepth in `secret::genericOp` attr-dict.

I also wanted to annotate the `mulDepth attribute` on each `Value`, but faced a technical problem that some `Value` are block argument (of `secret::GenericOp`) and we can not do `setArgAttr` like what `secretize` has done to function arguments (where `FunctionOpInterface` comes into play).

### Example

<details>
With input

```mlir
func.func @func(
    %arg0 : i16,
    %arg1 : i16,
    %arg2 : i16
  ) -> i16 {
    %1 = arith.muli %arg0, %arg1 :i16
    %2 = arith.addi %1, %arg2 : i16
  return %2 : i16
}
```
with `--mlir-to-secret-arithmetic="entry-function=func" --operation-balancer --secret-with-mgmt-bgv` we get
```mlir
    %0 = secret.generic ins(%arg0, %arg1, %arg2 : !secret.secret<i16>, !secret.secret<i16>, !secret.secret<i16>) attrs = {mulDepth = 1 : i64} {
    ^bb0(%arg3: i16, %arg4: i16, %arg5: i16):
      %1 = arith.muli %arg3, %arg4 : i16
      %2 = mgmt.modreduce %arg5 : i16
      %3 = arith.addi %1, %2 : i16
      secret.yield %3 : i16
    } -> !secret.secret<i16>
```
</details>
With input

```mlir
    %c = arith.constant 1 : i16
    %1 = arith.muli %arg0, %arg0 :i16
    %2 = arith.addi %1, %arg1 : i16
    %3 = arith.addi %c, %2 : i16
```
We get
```mlir
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<i16>, !secret.secret<i16>) attrs = {arg0 = {level = 1 : i64}, arg1 = {level = 1 : i64}} {
    ^bb0(%arg2: i16, %arg3: i16):
      %1 = arith.muli %arg2, %arg2 {level = 1 : i64} : i16
      %2 = mgmt.relinearize %1 {level = 1 : i64} : i16
      %3 = mgmt.modreduce %2 {level = 0 : i64} : i16
      %4 = mgmt.modreduce %arg3 {level = 0 : i64} : i16
      %5 = arith.addi %3, %4 {level = 0 : i64} : i16
      %6 = arith.addi %5, %c1_i16 {level = 0 : i64} : i16
      secret.yield %6 : i16
    } -> !secret.secret<i16>
```

### TODO

* [x] Fix mod reduce for cipher-plain type op.
* [x] Mgmt to LWE lowering
* [x] relinearize op placement


ZenithalHourlyRate:
Implemented the OpenFHE `FLEXIBLEAUTO` and `FLEXIBLEAUTOEXT` way to handle mod reduce (namely mod reduce before mult) (main difference being using ModReduce for mismatch instead of cross-level adjustment like LevelReduce)

Example:
```mlir
func.func @func(
    %arg0 : i16,
    %arg1 : i16
  ) -> i16 {
    %c = arith.constant 2 : i16
    %1 = arith.muli %arg0, %arg0 :i16
    %2 = arith.addi %1, %arg1 : i16
    %3 = arith.muli %2, %c : i16
  return %3 : i16
}
```

With `--mlir-to-secret-arithmetic="entry-function=func" --operation-balancer --secret-with-mgmt-bgv="include-first=true"` (namely `AUTOEXT`) we get
```mlir
module {
  func.func @func(%arg0: !secret.secret<i16>, %arg1: !secret.secret<i16>) -> !secret.secret<i16> {
    %c2_i16 = arith.constant {level_pt = 1 : i64, level_scheme = 3 : i64} 2 : i16
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<i16>, !secret.secret<i16>) attrs = {arg0 = {dimension = 2 : i64, level = 3 : i64}, arg1 = {dimension = 2 : i64, level = 3 : i64}} {
    ^bb0(%arg2: i16, %arg3: i16):
      %1 = mgmt.modreduce %arg2 {dimension = 2 : i64, level = 2 : i64} : i16
      %2 = arith.muli %1, %1 {dimension = 3 : i64, level = 2 : i64} : i16
      %3 = mgmt.relinearize %2 {dimension = 2 : i64, level = 2 : i64} : i16
      %4 = mgmt.modreduce %arg3 {dimension = 2 : i64, level = 2 : i64} : i16
      %5 = arith.addi %3, %4 {dimension = 2 : i64, level = 2 : i64} : i16
      %6 = mgmt.modreduce %5 {dimension = 2 : i64, level = 1 : i64} : i16
      %7 = arith.muli %6, %c2_i16 {dimension = 3 : i64, level = 1 : i64} : i16
      %8 = mgmt.relinearize %7 {dimension = 2 : i64, level = 1 : i64} : i16
      %9 = mgmt.modreduce %8 {dimension = 2 : i64, level = 0 : i64} : i16
      secret.yield %9 : i16
    } -> !secret.secret<i16>
  }
}
```

I think this serves as a baseline for BGV mgmt. I have implemented part of the the parameter selection logic from OpenFHE, where we can actually select primes! (not in this PR, still need cleanup) and corresponding lowering all the way to an OpenFHE executable (not in this PR, dirty hack).

Screenshot2024-12-04-23:21:13

One thing annoying me is that OpenFHE does not provide some compiler-friendly APIs .
 * Fine-grained prime control, where compiler can manually provide a prime via, e.g. `SetPrimeAt(i, prime)`.
 * For BGV in FLEXIBLEAUTO(EXT) mode, compiler issued `ModReduce` does not work; I assume a FLEXIBLEMANUAL mode is wanted, as there are some constraints for FIXEDMANUAL where we get far less primes avaiable.
 
I'll try to reach out to OpenFHE guys to ask them if they want to provide a compiler friendly interface; I'll at least maintain a version that provides these API, so I can get things truly e2e.

lib/Transforms/SecretWithMgmt/SecretWithMgmtBGV.cpp
```diff
@@ -0,0 +1,385 @@
...
+                  }
+
+                  for (auto operand : mulOp->getOperands()) {
+                    auto secretness = getSecretness(operand);
+                    if (!secretness) {
+                      continue;
+                    }
+                    b.setInsertionPoint(mulOp);
+                    Value managed = operand;
+                    for (auto i = 0; i != levelResult - levelMap.at(operand);
```
j2kun, line 124:
```suggestion
                    for (auto i = 0; i < levelResult - levelMap.at(operand);
```

nit: protect against a potential infinite loop

lib/Transforms/SecretWithMgmt/SecretWithMgmtBGV.cpp
```diff
@@ -0,0 +1,385 @@
...
+                         ++i) {
+                      managed =
+                          b.create<mgmt::ModReduceOp>(mulOp->getLoc(), managed);
```
j2kun, line 127: It looks like this is in a loop to allow for mod-reducing across multiple levels. Would it make more sense to upgrade the `mod_reduce` op at this layer of abstraction to permit jumping across as many levels as desired? I am not an expert at the implementation of BGV, but I suspect you can implement the combined mod_reduce with one formula (combining multiple modulus factors at once) rather than many steps.

Or maybe you did it this way because OpenFHE doesn't permit that?

lib/Transforms/SecretWithMgmt/SecretWithMgmtBGV.cpp
```diff
@@ -0,0 +1,385 @@
...
+  void alwaysModreduceWhenLevelMismatch() {
+    DataFlowSolver solver;
+    solver.load<dataflow::DeadCodeAnalysis>();
+    solver.load<dataflow::SparseConstantPropagation>();
+    solver.load<SecretnessAnalysis>();
+    if (failed(solver.initializeAndRun(getOperation()))) {
+      getOperation()->emitOpError() << "Failed to run the analysis.\n";
+      signalPassFailure();
+      return;
+    }
```
j2kun, line 153: This can probably be refactored into a function since you do it so often:

```cpp
LogicalResult runSecretnessAnalysis(DataFlowSolver &solver, Operation *operation) {
    solver.load<dataflow::DeadCodeAnalysis>();
    solver.load<dataflow::SparseConstantPropagation>();
    solver.load<SecretnessAnalysis>();
    if (failed(solver.initializeAndRun(getOperation()))) {
      operation->emitOpError() << "Failed to run the analysis.\n";
      return failure();
    }
}
```

Then

```cpp
DataFlowSolver solver;
if (failed(runSecretnessAnalysis(solver, getOperation()))) {
  signalPassFailure();
  return;
}
```

lib/Transforms/SecretWithMgmt/SecretWithMgmtBGV.cpp
j2kun, line 1: I think structurally this may benefit from separating some of the work into separate analyses. I suspect you didn't do this because you would be changing the IR (which would invalidate the analysis), and you don't have a way to propagate that change through the dataflow solver so later ops have updated level, etc.

It turns out you can actually propagate that, cf. for example https://github.com/google/heir/blob/ef10aeb15fde7654dd1e8b27decb5bd0bf3566a5/lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp#L95 which modifies the IR but keeps the secretness analysis in sync.

So the structure would be (e.g., for level mismatch rewrite): 

- Analysis to compute the modulus level
- Have a pattern that matches when an op has two inputs with mismatching levels (querying the solver for the SSA value dynamically)
- Insert the mod reduce ops, create new lattice objects for the new SSA values, and call propagateIfChanged on them

What do you think of this? This would be useful if we think that the analyses could be reused or generalized further.

j2kun (changes requested):
Great work! I'm very much in favor of this addition. I made some broad suggestions for cleanup and restructuring which are open to discussion, but I don't have any major objections to the meat of the pass. I wonder if you've tried lowering the `mgmt` ops yet, and if there are any obstacles you hit when doing that.

ZenithalHourlyRate:
Get the BGV RNS support working; still need to fix some minor issues

Here is the result of 8 multiplication with --operation-balancer
```mlir
!Z1005037682689_i64_ = !mod_arith.int<1005037682689 : i64>
!Z1032955396097_i64_ = !mod_arith.int<1032955396097 : i64>
!Z1095233372161_i64_ = !mod_arith.int<1095233372161 : i64>
!Z998595133441_i64_ = !mod_arith.int<998595133441 : i64>
!rns_L1_ = !rns.rns<!Z1095233372161_i64_>
!rns_L2_ = !rns.rns<!Z1095233372161_i64_, !Z1032955396097_i64_>
!rns_L3_ = !rns.rns<!Z1095233372161_i64_, !Z1032955396097_i64_, !Z1005037682689_i64_>
!rns_L4_ = !rns.rns<!Z1095233372161_i64_, !Z1032955396097_i64_, !Z1005037682689_i64_, !Z998595133441_i64_>
#ring_rns_L1_1_x1024_ = #polynomial.ring<coefficientType = !rns_L1_, polynomialModulus = <1 + x**1024>>
#ring_rns_L2_1_x1024_ = #polynomial.ring<coefficientType = !rns_L2_, polynomialModulus = <1 + x**1024>>
#ring_rns_L3_1_x1024_ = #polynomial.ring<coefficientType = !rns_L3_, polynomialModulus = <1 + x**1024>>
#ring_rns_L4_1_x1024_ = #polynomial.ring<coefficientType = !rns_L4_, polynomialModulus = <1 + x**1024>>
!rlwe_ct_L1_D2_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <ring = #ring_rns_L1_1_x1024_>, underlying_type = i16>
!rlwe_ct_L2_D2_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <ring = #ring_rns_L2_1_x1024_>, underlying_type = i16>
!rlwe_ct_L2_D3_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <dimension = 3, ring = #ring_rns_L2_1_x1024_>, underlying_type = i16>
!rlwe_ct_L3_D2_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <ring = #ring_rns_L3_1_x1024_>, underlying_type = i16>
!rlwe_ct_L3_D3_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <dimension = 3, ring = #ring_rns_L3_1_x1024_>, underlying_type = i16>
!rlwe_ct_L4_D2_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <ring = #ring_rns_L4_1_x1024_>, underlying_type = i16>
!rlwe_ct_L4_D3_ = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <dimension = 3, ring = #ring_rns_L4_1_x1024_>, underlying_type = i16>
module {
  func.func @func(%arg0: !rlwe_ct_L4_D2_, %arg1: !rlwe_ct_L4_D2_, %arg2: !rlwe_ct_L4_D2_, %arg3: !rlwe_ct_L4_D2_, %arg4: !rlwe_ct_L4_D2_, %arg5: !rlwe_ct_L4_D2_, %arg6: !rlwe_ct_L4_D2_, %arg7: !rlwe_ct_L4_D2_) -> !rlwe_ct_L1_D2_ {
    %0 = bgv.mul %arg0, %arg1 : (!rlwe_ct_L4_D2_, !rlwe_ct_L4_D2_) -> !rlwe_ct_L4_D3_
    %1 = bgv.relinearize %0 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L4_D3_ -> !rlwe_ct_L4_D2_
    %2 = bgv.mul %arg2, %arg3 : (!rlwe_ct_L4_D2_, !rlwe_ct_L4_D2_) -> !rlwe_ct_L4_D3_
    %3 = bgv.relinearize %2 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L4_D3_ -> !rlwe_ct_L4_D2_
    %4 = bgv.modulus_switch %3 {to_ring = #ring_rns_L3_1_x1024_} : !rlwe_ct_L4_D2_ -> !rlwe_ct_L3_D2_
    %5 = bgv.modulus_switch %1 {to_ring = #ring_rns_L3_1_x1024_} : !rlwe_ct_L4_D2_ -> !rlwe_ct_L3_D2_
    %6 = bgv.mul %5, %4 : (!rlwe_ct_L3_D2_, !rlwe_ct_L3_D2_) -> !rlwe_ct_L3_D3_
    %7 = bgv.relinearize %6 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L3_D3_ -> !rlwe_ct_L3_D2_
    %8 = bgv.mul %arg4, %arg5 : (!rlwe_ct_L4_D2_, !rlwe_ct_L4_D2_) -> !rlwe_ct_L4_D3_
    %9 = bgv.relinearize %8 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L4_D3_ -> !rlwe_ct_L4_D2_
    %10 = bgv.mul %arg6, %arg7 : (!rlwe_ct_L4_D2_, !rlwe_ct_L4_D2_) -> !rlwe_ct_L4_D3_
    %11 = bgv.relinearize %10 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L4_D3_ -> !rlwe_ct_L4_D2_
    %12 = bgv.modulus_switch %9 {to_ring = #ring_rns_L3_1_x1024_} : !rlwe_ct_L4_D2_ -> !rlwe_ct_L3_D2_
    %13 = bgv.modulus_switch %11 {to_ring = #ring_rns_L3_1_x1024_} : !rlwe_ct_L4_D2_ -> !rlwe_ct_L3_D2_
    %14 = bgv.mul %12, %13 : (!rlwe_ct_L3_D2_, !rlwe_ct_L3_D2_) -> !rlwe_ct_L3_D3_
    %15 = bgv.relinearize %14 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L3_D3_ -> !rlwe_ct_L3_D2_
    %16 = bgv.modulus_switch %15 {to_ring = #ring_rns_L2_1_x1024_} : !rlwe_ct_L3_D2_ -> !rlwe_ct_L2_D2_
    %17 = bgv.modulus_switch %7 {to_ring = #ring_rns_L2_1_x1024_} : !rlwe_ct_L3_D2_ -> !rlwe_ct_L2_D2_
    %18 = bgv.mul %17, %16 : (!rlwe_ct_L2_D2_, !rlwe_ct_L2_D2_) -> !rlwe_ct_L2_D3_
    %19 = bgv.relinearize %18 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : !rlwe_ct_L2_D3_ -> !rlwe_ct_L2_D2_
    %20 = bgv.modulus_switch %19 {to_ring = #ring_rns_L1_1_x1024_} : !rlwe_ct_L2_D2_ -> !rlwe_ct_L1_D2_
    return %20 : !rlwe_ct_L1_D2_
  }
}


```

ZenithalHourlyRate:
I think I have cleaned up the BGV RNS lowering path quite a bit, and every BGV mlir input I can find gets correctly through the `secret-to-bgv` now (turned off some LWE type check of cource).

Strongly request for review on this lowering, particularly on `SecretToBGV.cpp` and `ConversionUtils.h`, I think they are insane.

Remaining thing
 * fix LWE type so they can work with RNSType
 * fix tests, they assume ModArithType

lib/Dialect/Secret/Conversions/SecretToBGV/SecretToBGV.cpp
```diff
@@ -41,59 +45,81 @@ namespace mlir::heir {
...
+    ring = rlweRing;
+
+    // actually isLegal/isSignatureLegal will always be true
+    // relying on addIllegalDialect<secret::SecretDialect>
+    // for checking function signature correctness
     addConversion([](Type type) { return type; });
 
     // Convert secret types to BGV ciphertext types
-    addConversion([ctx, this](secret::SecretType type) -> Type {
+    addTypeAttributeConversion([ctx, this](secret::SecretType type,
```
j2kun, line 101: I think this is the only "hacky" part of this PR: that you're implementing this function for something it's not intended to be used for, just because it happens to be a member of TypeConverter already and TypeAttr happens to be a legal return value. It's clever but I think we can do even better.

What if instead we had a little HEIR-specific class hierarchy:

```cpp
// in ConversionUtils.h
class ContextAwareTypeConverter : public TypeConverter {
  // Returns a TypeRange with the converted types of the values in the input range,
  // taking into account the context of the values (e.g., defining ops or uses).
  TypeRange convertValueRangeTypes(ValueRange values);

  // Returns a TypeRange with the converted types of the results of an op,
  // taking into account the context of the op when selecting the new type.
  TypeRange convertOpResultTypes(Operation *op, TypeRange resultTypes);
}

// in SecretToBGV.cpp
class SecretToBGVTypeConverter : ContextAwareTypeConverter {
  Type convertTypeWithAttr(Type type, Attribute *attr) override {
    // your single-type conversion in addTypeAttributeConversion

    // if attr is nullptr, fall back to the base class type converter
  }

  TypeRange convertValueRangeTypes(ValueRange values) override {
    // attribute extraction logic and invocation of convertTypeWithAttr
  }

  TypeRange convertOpResultTypes(Operation *op, TypeRange resultTypes) {
    // similar to convertValueRangeTypes, but the mgmt attr extracted from `op`.
  }
}
```

Then in the patterns in ConversionUtils.h, you can `dyn_cast` the input `TypeConverter` to `ContextAwareTypeConverter`, and if the cast succeeds you'd call its new function instead (otherwise rely on `OpAdaptor` as usual). This has the benefit that `SecretToBGVTypeConverter` could even contain a DataflowSolver instance as a member variable, and reference the dataflow analysis results directly at type conversion time (would that eliminate the need to persist attributes on the ops in the first place?).

Moreover, since `ConvertFuncWithMgmt` isn't used as part of the actual dialect conversion framework (just `walkAndApplyPatterns`), it can inherit from RewritePattern instead of `OpConversionPattern`, directly take a `ContextAwareTypeConverter` in its constructor, and avoid dyn-casts entirely, which would be a little simpler and less confusing to the reader.

ZenithalHourlyRate: > HEIR-specific class hierarchy

I was thinking that SecretToBGVTypeconverter should not appear in ConversionUtils (as CKKS also uses it) so chose to misuse its interface. A middle abstract Typeconverter could help.

> This has the benefit that SecretToBGVTypeConverter could even contain a DataflowSolver instance as a member variable, and reference the dataflow analysis results directly at type conversion time (would that eliminate the need to persist attributes on the ops in the first place?).

I think we still face the problem of DataflowSolver getting invalidated during the conversion and has to re-run it for every possible middle state of the conversion, which could be hard.

The current analysis only works after `mlir-to-secret-arithmetic` but before `secret-distribute-generic` so that things get simple, if we are analysing it at funcOp level things are going to be complex (transfer state for both GenericOp{arith::MulIOP} and bgv::MulOp}), where inside analysis we will check whether secret::GenericOp contains only one op; this is similar to the `persist attribute` technique but put more complexity on to the analysis instead of the conversion pass.

ZenithalHourlyRate: Also the Secretnessanalysis mainly depends on the SecretType to work (so that should be used before `secret-to-scheme`), but the SecretType will be converted away during this conversion and the analysis has to deal with such complexity by analysing LWE types to give

ZenithalHourlyRate:
Finally get all the tests working. Now the BGV pipeline output will be RNS. You can check `getting_started.md` for the new openfhe output, though unfortunately with FLEXIBLEAUTOEXT mode enabled in OpenFHE, openfhe.mod_reduce actually takes no effect.

Docs on the pass/dialect/op/attr are also added.

lib/Transforms/SecretWithMgmt/SecretWithMgmtBGV.cpp
ZenithalHourlyRate: To use an updated mgmt.modreduce op, we have to update bgv.modulus_switch and openfhe.modreduce op accordingly, though OpenFHE's ModReduce do support an arg `levels` describing levels to drop, I think I'll leave it as future refining.

lib/Utils/ConversionUtils/ConversionUtils.h
```diff
@@ -103,14 +161,58 @@ class SecretGenericOpConversion
...
+    // so that we can use it in the next conversion
+    // but as secret-to-ckks/secret-to-cggi is also using this...
+    // does not change the signature for now
+    return matchAndRewriteInner(op, resultTypes, inputs, /*mgmtAttr,*/
+                                innerOp.getAttrs(), rewriter);
+  }
 
-    return matchAndRewriteInner(op, resultTypes, inputs, innerOp.getAttrs(),
-                                rewriter);
+  static void copyMgmtAttr(Operation *op, Operation *newOp) {
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Now MgmtAttr does not need to attach to the newly created BGV ops. The invariant becomes when there is a SecretType there is a MgmtAttr and this function is removed.

lib/Dialect/Secret/Conversions/SecretToBGV/SecretToBGV.cpp
ZenithalHourlyRate: I still maintain the MgmtAttr approach as it is cleaner for debugging (debuging into analysis is quite hard). I think an Analysis based TypeConverter can be done in the future, but that will touch too many analysis to makes things work (SecretAnalysis and all this PR's analyses, handling all the intermediate state of dialect lowering) and worth another PR. It has been already too many thing in this PR.

ZenithalHourlyRate:
Rebased.

ZenithalHourlyRate:
Renamed `secret-with-mgmt-bgv` to `secret-insert-mgmt-bgv`

AlexanderViand-Intel (approved):
Thanks for staying up and walking me through the PR!

ZenithalHourlyRate:
Test fixed.

j2kun (approved):

j2kun:
@ZenithalHourlyRate can you rebase please? Then we can get this merged today.

PR #1127, opened 2024-11-29
openfhe emitter: register mod arith dialect

ZenithalHourlyRate:
`heir-translate --emit-openfhe-pke-header` would complain

```
 error: `!"mod_arith"<"int<463187969 : i32>">` type created with unregistered dialect. If this is intended, please call allowUnregisteredDialects() on the MLIRContext, or use -allow-unregistered-dialect with the MLIR opt tool used
```


j2kun:
I expect any IR that is given to `heir-translate` for openFHE codegen should not have `mod_arith` ops in it. Is lowering `mod_arith` to `arith` not sufficient?

ZenithalHourlyRate:
> I expect any IR that is given to `heir-translate` for openFHE codegen should not have `mod_arith` ops in it

Well, currently openfhe.op has arguments with RLWE type, which includes RingAttr thus ModArithType.

j2kun:
Ah yes, of course. Thanks!

j2kun (approved):

PR #1130, opened 2024-11-29
polynomial: pretty print RingAttr

ZenithalHourlyRate:
This PR tries to reduce the verbosity of polynomial type assembly format.

With this change, we get a much cleaner output

```mlir
module {
  func.func @main(%arg0: !polynomial.polynomial<<!mod_arith.int<65537 : i64>, <1 + x**8192>>>) -> !polynomial.polynomial<<!mod_arith.int<65537 : i64>, <1 + x**8192>>> {
    return %arg0 : !polynomial.polynomial<<!mod_arith.int<65537 : i64>, <1 + x**8192>>>
  }
}
```

Instead of

```mlir
module {
  func.func @main(%arg0: !polynomial.polynomial<ring = <coefficientType = !mod_arith.int<65537 : i64>, polynomialModulus = <1 + x**1024>>>) -> !polynomial.polynomial<ring = <coefficientType = !mod_arith.int<65537 : i64>, polynomialModulus = <1 + x**1024>>> {
    return %arg0 : !polynomial.polynomial<ring = <coefficientType = !mod_arith.int<65537 : i64>, polynomialModulus = <1 + x**1024>>>
  }
}
```

Other examples:

```mlir
module {
  func.func @main(%arg0: !polynomial.polynomial<<!rns.rns<!mod_arith.int<65537 : i64>, !mod_arith.int<17 : i64>>, <1 + x**1024>>>) -> !polynomial.polynomial<<!rns.rns<!mod_arith.int<65537 : i64>, !mod_arith.int<17 : i64>>, <1 + x**1024>>> {
    return %arg0 : !polynomial.polynomial<<!rns.rns<!mod_arith.int<65537 : i64>, !mod_arith.int<17 : i64>>, <1 + x**1024>>>
  }
}
```

If such change is favored, I'll change corresponding tests.

### DIFF
--- a/lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
+++ b/lib/Dialect/Polynomial/IR/PolynomialAttributes.cpp
@@ -204,6 +204,20 @@ Attribute FloatPolynomialAttr::parse(AsmParser &parser, Type type) {
   return FloatPolynomialAttr::get(parser.getContext(), result.value());
 }
 
+// ring
+LogicalResult parseRingIntPolyAttr(AsmParser &parser, IntPolynomialAttr &attr) {
+  auto parsed = IntPolynomialAttr::parse(parser, nullptr);
+  if (parsed) {
+    attr = mlir::dyn_cast<IntPolynomialAttr>(parsed);
+    return success();
+  }
+  return failure();
+}
+
+void printRingIntPolyAttr(AsmPrinter &p, IntPolynomialAttr attr) {
+  attr.print(p);
+}
+
 }  // namespace polynomial
 }  // namespace heir
 }  // namespace mlir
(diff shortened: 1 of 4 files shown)


j2kun:
I like the ability to tell what the attributes mean from the textual IR.

Can we instead try configuring the polynomial dialect so that the pretty-printer uses aliases for polynomial types and attributes by default? According to this link, it is possible and it appears to only require changing tablegen: https://discourse.llvm.org/t/how-to-get-type-alias-printings-in-mlir-ir/78372/4

ZenithalHourlyRate:
> Can we instead try configuring the polynomial dialect so that the pretty-printer uses aliases for polynomial types and attributes by default?

Oh, I did not find the way to create aliases. I'll try to figure this out.

j2kun:
I only learned that this was possible today by trying to track down that old thread to "prove" to myself that it could not be done while trying to formulate a response to your PR :)

Issue #1105, opened 2024-11-22
Add a Python frontend to HEIR


j2kun:
I wanted to post the in-progress design doc for the Python frontend here, as it's getting close to a decent shape.

https://docs.google.com/document/d/1pv7ri7GN4VN2-VUuIXwkfYnJOK3rEg16at1qSuNkU7o/edit?usp=sharing

PR #1137, opened 2024-12-04
Add an MVP python frontend

j2kun:
This is still very much in progress, and I hope to merge in a first working end-to-end version that is still missing many features, but solves all the main problems:

- JIT-compiling python bytecode -> SSA
- SSA -> heir-opt + heir-translate + clang
- Loading the results back into something that can be called from python
- Adding a pybind11 emitter to heir-translate
- Doing the above in a decently platform-independent way


j2kun:
So this PR is at the point where it can feasibly be merged (bazel build is accounted for, though it will need some changes when copybara merges it internally).

It is still missing a huge number of features to be considered useful or worth demoing. A short list of work that needs doing:

- Adding a decorator and assembling the functions into an actually-usable frontend
- Configuring heir-opt/heir-translate paths properly
- Configuring heir-opt/heir-translate pipeline options instead of hard-coding them
- Adding support for `cf` ops in `heir-opt`  (e.g., the commented-out branching test can't run because no pipeline supports secret `cf.br`)
- Requiring/supporting type hints (right now everything is assumed i64) and doing type propagation through the SSA form (I think numba handles this but I have to invoke it, and I'm not sure how it will handle our custom type annotations)
- Finishing the meat of mlir_emitter, as most paths raise NotImplementedError
- Adding a semblance of not being hard-coded to OpenFHE (e.g., splitting it into a separate backend instead of inlining it in pipeline.py)

j2kun:
I started working on the decorator that will wrap this in https://github.com/j2kun/heir/pull/10

heir_py/openfhe_config.py
```diff
@@ -0,0 +1,89 @@
...
+    # The libraries to link against (without lib prefix or .so suffix)
+    link_libs: list[str]
+
+    # The type of include paths to use during codegen. Options are:
+    # - "install-relative": use paths relative to the installed OpenFHE
+    # - "source-relative": relative to the openfhe development repository.
+    include_type: str = "install-relative"
+
+
+DEFAULT_INSTALLED_OPENFHE_CONFIG = OpenFHEConfig(
```
asraa, line 25: just feedback for another time: i think it might be worth adding a warning incase these hardcoded dirs don't exist so that the user gets early feedback before clang tries to compile with these flags

heir_py/pipeline_test.py
```diff
@@ -0,0 +1,50 @@
...
+        kp = cc.KeyGen()
+        _heir_foo.foo__configure_crypto_context(cc, kp.secretKey)
+        arg0_enc = _heir_foo.foo__encrypt__arg0(cc, 7, kp.publicKey)
+        arg1_enc = _heir_foo.foo__encrypt__arg1(cc, 8, kp.publicKey)
+        res_enc = _heir_foo.foo(cc, arg0_enc, arg1_enc)
+        res = _heir_foo.foo__decrypt__result0(cc, res_enc, kp.secretKey)
+
+        self.assertEqual(-15, res)
+
+    # def test_branch(self):
```
asraa, line 24: could you add a quick comment here saying that heir-opt's pipelines don't support cf dialect (and that's why this test is commented out, not because of part of the python compiler pipeline) just in case someone is checking out the frontend tests?

heir_py/mlir_emitter.py
```diff
@@ -0,0 +1,199 @@
...
+class TextualMlirEmitter:
+    def __init__(self, ssa_ir):
+        self.ssa_ir = ssa_ir
+        self.temp_var_id = 0
+        self.numba_names_to_ssa_var_names = {}
+        self.globals_map = {}
+
+    def emit(self):
+        func_name = self.ssa_ir.func_id.func_name
+        # probably should use unique name...
```
asraa, line 18: might be something i don't know about numba but aren't all func names supposed to be unique in a module?

heir_py/mlir_emitter.py
```diff
@@ -0,0 +1,199 @@
...
+        blocks = self.ssa_ir.blocks
+        str_blocks = []
+        first = True
+
+        for block_id, block in sorted(blocks.items()):
+            instructions = []
+            for instr in block.body:
+                result = self.emit_instruction(instr)
+                if result:
+                    # print(f": {result}")
```
asraa, line 45: meant to keep?

asraa (approved):
Thank you so much for starting this! This is a ton of complex work.

heir_py/mlir_emitter.py
```diff
@@ -0,0 +1,199 @@
...
+                name = self.get_or_create_name(assign.target)
+                # FIXME: fix type (somehow the pretty printer on assign.value knows it's
+                # an int???)
+                return f"{name} = arith.constant {assign.value.value} : i64"
+            case ir.Global():
+                self.globals_map[assign.target.name] = assign.value.name
+                return ""
+        raise NotImplementedError()
+
+    def emit_expr(self, expr):
```
WoutLegiest, line 132: Alternative approach: store all of the strings in a set or dict and use something like: 
`if expr.op in set:
    raise NotImplementedError() `

WoutLegiest:
More general questions: 
- Do we expect that bazel has already build heir-opt; heir-translate before using the python pipeline (Or are we using the nightly builds) ? 
- I guess a tutorial will be the next?

WoutLegiest (approved):
LGTM, nice work!

heir_py/clang.py
```diff
@@ -0,0 +1,96 @@
+"""A helper to fing and run clang."""
```
AlexanderViand-Intel, line 1: nit: 
```suggestion
"""A helper to find and run clang."""
```

heir_py/mlir_emitter.py
```diff
@@ -0,0 +1,199 @@
...
+        self.temp_var_id = 0
+        self.numba_names_to_ssa_var_names = {}
+        self.globals_map = {}
+
+    def emit(self):
+        func_name = self.ssa_ir.func_id.func_name
+        # probably should use unique name...
+        # func_name = ssa_ir.func_id.unique_name
+
+        # FIXME: use inferred or explicit types for args
```
AlexanderViand-Intel, line 21: I know the whole feature is WIP, but these should probably still have an issue number for tracking (maybe just a single "fix up python frontend issues" issue for now?)

heir_py/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+    function,
+    openfhe_config: OpenFHEConfig = DEFAULT_INSTALLED_OPENFHE_CONFIG
+):
+    """Run the compiler."""
+    # The temporary workspace dir is so that heir-opt, heir-translate, and
+    # clang can have places to write their output files. It is cleaned up once
+    # the function returns, at which point the compiled python module has been
+    # loaded into memory and the raw files are not needed.
+    #
+    # For debugging, add delete=False to TemporaryDirectory (python3.12+)
```
AlexanderViand-Intel, line 27: Can you expose this via an optional arg on `run_compiler` (maybe something more general such as `debug_mode`)? I'm guessing it'll be something we use a lot in the near- to mid-term

heir_py/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+    # tempfile.mkdtemp()` and manually clean it up.
+    with tempfile.TemporaryDirectory() as workspace_dir:
+        func_id = FunctionIdentity.from_function(function)
+        bytecode = ByteCode(func_id)
+        ssa_ir = Interpreter(func_id).interpret(bytecode)
+        mlir_textual = TextualMlirEmitter(ssa_ir).emit()
+        func_name = func_id.func_name
+        module_name = f"_heir_{func_name}"
+
+        # FIXME: allow user to configure heir-opt path
```
AlexanderViand-Intel, line 39: see other comment on issue tracking above

AlexanderViand-Intel (commented):
Thanks so much for putting this together! Despite being an early protpype, I think this really deserves a small README, not least because I unfortunately seem to be too dumb to get it to run properly. (`"FileNotFoundError: [Errno 2] No such file or directory: '/home/aviand/.cache/bazel/_bazel_aviand/ec6a8a3116c7020336910879e3bce41d/execroot/heir/bazel-out/k8-dbg/bin/heir_py/tools/heir-opt'"`)

UPDATE: nevermind, looks like it just wanted a different working directory. Still in favor of that README, though, as I'm still running into some issues (`In file included from /tmp/tmpvzbaskj7/foo.cpp:2:
/usr/local/include/openfhe/pke/openfhe.h:40:10: fatal error: 'openfhecore.h' file not found
#include "openfhecore.h"`) 😉

heir_py/mlir_emitter.py
j2kun: Done in c2b7faa9

heir_py/clang.py
j2kun: Fixed

heir_py/pipeline_test.py
j2kun: I just removed the test, I will add it back in a future PR

heir_py/mlir_emitter.py
j2kun: it's just a python thing, that this func_name is only the inner-most name, not including the module import path to the function. There is a qualified name, but I haven't looked into using it, or if it is even necessary.

heir_py/mlir_emitter.py
j2kun: Removed

j2kun:
> I think this really deserves a small README

Added a "developers" readme in `heir_py/`

heir_py/__init__.py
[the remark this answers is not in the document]
j2kun: I just kept it since I will be using it in the next PR as the mechanism to `from heir_py import compile` and get straight to the decorator.

PR #1138, opened 2024-12-04
Mod Arith: migrate barrett/subifge to mod arith type

bqeyfukqbukq:
This fix issue #1084 

This is wip

This converts barret reduction and subifge to ModArithType, and added corresponding lowering in ModArithToArith.cpp 

One thing worth discussing is that the input type is still SignlessIntegerLike, otherwise lowering won't work.


j2kun:
Hi @bqeyfukqbukq, what is the status of this PR? I think we would like appreciate this contribution, but it has some syntax errors and given its age it likely needs to be rebased and updated.

If you no longer want to work on this, let me know and I can cherry-pick your commits and finish the work.

ZenithalHourlyRate:
The author is one of my colleagues in our group and I was trying to introduce the project to him. After a few discussion he decided to do other projects so left the PR here.

> I can cherry-pick your commits and finish the work.

One thing that should be handled carefully is the bitwidth of intermediate result in barrett reduction, where the original author used 3B and we (wrongly) used 2B, then realized it should be 3B but did not get a proper type for it.

j2kun:
Tentatively closing, but we may revive this at some point in the future

Issue #1107, opened 2024-11-22
Quality of Life Pass/Pipeline Features


AlexanderViand-Intel:
Small fixes/add-ons that I noticed while drawing up a pass/pipeline overview figure:

- [x] `--secretize`'s `entry-function` parameter should be optional and, if omitted, the pass should apply to all functions in the input (module?). This should then also be reflected in any pipelines that use `--secretize`, making their `entry-function` optional, too.
- [x]  In order to better support mixed ctxt/ptxt examples, and for better integration with frontends, create `--annotated-mlir-to-....` that are the same as the current `--mlir-to-...` pipelines but without the `--secretize` pass.
- [ ] The `ciphertext-degree` parameter in the various passes & pipelines that use it should be optional, with the compiler using the size of (the right-most-non-one-dimension of) the first secret tensor in the function signature if the parameter is omitted (and erroring out if no such secret tensor function parameter exists)
See: #1197
- [x] `--optimize-relinearization` should be added to the RLWE pipeline builder? (EDIT: Not until #1148, and ideally also #1032 are fixed)
- [x] `--operation-balancer` shuld be added to the mlir-to-secret-arithmetic pipeline
- [x] Investigate if the `--bgv-to-openfhe` and `--ckks-to-openfhe` passes can be merged into a single `--rlwe-to-openfhe` pass combining the patterns from both passes.
Also: #1150
- [x] (Optional/TBD): Split the library-specific parts of the pipeline into separate pipelines. While that might require doing something like `--mlir-to-ckks --rlwe-to-openfhe`, it might be nicer to keep backend-specific stuff separate/modular. 
- [ ] Better error messages ( #1143)

AlexanderViand-Intel:
Based on discussion from today's office hours: 

For the `ciphertext-degree` parameter, we should decouple that into the actual ctxt degree (which can be optional and default to something reasoanble such as 16k) and something like `--assume-native-tensor-size` that does what was mentioned above.

In general, these things should not be just stored in a single pass's paramaters but encoded into the IR (at the module level)

ZenithalHourlyRate:
Also I'm wondering why `--operation-balancer` did not get into the `mlir-to-secret-arithmetic` pipeline, though its absense makes me happy when constructing deep circuit for debugging.

ZenithalHourlyRate:
> Do you happen to know which tests you expect to break?

No, haven't tried it.

AlexanderViand-Intel:
[...]

PR #1139, opened 2024-12-04
Secretize cleanup

AlexanderViand-Intel:
Addresses the first two points of #1107 plus some other small cleanup around Secretize.

* `--secretize` now applies to all functions in the module, unless the (renamed) `function=<name>` flag is given, allowing the flag to be omitted most of the time.
* `--mlir-to-secret-arithmetic` (and the `-to-ckks/-to-bgv` pipelines that use it) no longer includes `--secretize` (or its flags) and instead, test files now either include `--secretize` or (preferred) explicit `{secret.secret}` annotations. 
Note that the `to-openfhe` versions still have an `entry-function` flag, as that seems to be required for some of the openFHE stuff.
I did _not_ yet remove secretize from the boolean pipelines since I'm less familiar with them, but I think that would be the right approach there, too (even if we still need to designate an entry function)  with an eye towards making the pipelines more "user/frontend-friendly" rather than "testing-friendly".

And some other stuff thrown in:
* `secretize.cpp` lives in `lib/Transforms/secretize` but the tests lived in `lib/Dialect/Secret/Transforms/secretize`. I moved the tests since that's easier, but we could also consider moving the pass implementation instead.
* `--wrap-generic` now uses the recent [walk-based pattern rewriter](https://github.com/llvm/llvm-project/pull/113825). I'm sure there are other one-shot passes like this that don't need the fixed-point iteration logic of the greedy rewriter.
* I used Hongren's new alias magic to clean up `naive_matmul.mlir`  by simply roundtripping it through heir-opt. Again, I'm sure there's more cleanup opportunities like this we can watch out for as we touch other parts of the code/tests.


j2kun:
> remove secretize from the boolean pipelines

This sounds fine to me, at least for consistency's sake.

tests/Examples/openfhe/naive_matmul.mlir
j2kun, line 1: So much cleaner 😅

j2kun:
Well done. Approving and marking as pull_ready now, unless you want to also remove secretize from the boolean pipelines in this same PR.

j2kun (approved):

AlexanderViand-Intel:
> Well done. Approving and marking as pull_ready now, unless you want to also remove secretize from the boolean pipelines in this same PR.

Squashed :) I'll deal with the boolean stuff another time (or hope @asraa  tackles it instead xD)

asraa (approved):
Thanks! It's true that secretize is more of a testing convenience artifact.

PR #1131, opened 2024-11-30
Generate Alias for RLWE, ModArith, and Polynomial

ZenithalHourlyRate:
Check https://github.com/google/heir/pull/1130#issuecomment-2508866775

Cc @j2kun Definitely worth a blog post. Never seen a tutorial on this.

With this change, `--mlir-to-openfhe-bgv` result would be much cleaner.

<del>Though a lot of tests failed due to this change...</del>Can be mitigated by `--mlir-print-local-scope`

```mlir
!Z463187969_i32_ = !mod_arith.int<463187969 : i32>
#ring_Z463187969_i32_1_x8_ = #polynomial.ring<coefficientType = !Z463187969_i32_, polynomialModulus = <1 + x**8>>
!rlwe_pt = !lwe.rlwe_plaintext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, ring = #ring_Z463187969_i32_1_x8_, underlying_type = i16>
!rlwe_ct = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <ring = #ring_Z463187969_i32_1_x8_>, underlying_type = i16>
!rlwe_ct1 = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <dimension = 3, ring = #ring_Z463187969_i32_1_x8_>, underlying_type = i16>
module {
  func.func @func(%arg0: !openfhe.crypto_context, %arg1: !rlwe_ct, %arg2: !rlwe_ct) -> !rlwe_ct {
    %0 = openfhe.mul_no_relin %arg0, %arg1, %arg1 : (!openfhe.crypto_context, !rlwe_ct, !rlwe_ct) -> !rlwe_ct1
    %1 = openfhe.relin %arg0, %0 : (!openfhe.crypto_context, !rlwe_ct1) -> !rlwe_ct
    return %1 : !rlwe_ct
  }
  func.func @func__encrypt__arg0(%arg0: !openfhe.crypto_context, %arg1: i16, %arg2: !openfhe.public_key) -> !rlwe_ct {
    %splat = tensor.splat %arg1 : tensor<8xi16>
    %0 = arith.extsi %splat : tensor<8xi16> to tensor<8xi64>
[...]

PR #1141, opened 2024-12-04
Fix RingAttr alias and Add RNSType alias

ZenithalHourlyRate:
Cc @AlexanderViand-Intel

With this change, the example in https://github.com/google/heir/pull/1131#issuecomment-2517969073 becomes

```mlir
!Z463187969_i32_ = !mod_arith.int<463187969 : i32>
!Z463187970_i32_ = !mod_arith.int<463187970 : i32>
!Z463187971_i32_ = !mod_arith.int<463187971 : i32>
#ring__1_x8_ = #polynomial.ring<coefficientType = !rns.rns<!Z463187969_i32_, !Z463187970_i32_, !Z463187971_i32_>, polynomialModulus = <1 + x**8>>
!rlwe_ct = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 16, cleartext_bitwidth = 16>, rlwe_params = <ring = #ring__1_x8_>, underlying_type = tensor<8xi16>>
module {
  func.func @foo(%arg0: !rlwe_ct, %arg1: !rlwe_ct) -> !rlwe_ct {
    %0 = bgv.add %arg0, %arg1 : !rlwe_ct
    return %0 : !rlwe_ct
  }
}
``` 

The reason is that it is assumed that when `AliasResult::NoAlias` is returned `raw_ostream` does not have any content. MLIR reused this buffer so ModArithType accidentally gets the `ring_` prefix.

This fix is safe in that although `#ring__1_x8_` is a `FinalResult`, MLIR can still assign trailing number to it if found duplicate.

Note that it is tricky to have a `std::string` as temp buffer as Type/Attribute often does not have the toString method; they only have the interface to raw_stream.

For example:
```mlir
#ring__1_x8_ = #polynomial.ring<coefficientType = !rns.rns<!Z463187969_i32_, !Z463187970_i32_, !Z463187971_i32_>, polynomialModulus = <1 + x**8>>
#ring__1_x8_1 = #polynomial.ring<coefficientType = !rns.rns<!Z463187969_i32_, !Z463187970_i32_>, polynomialModulus = <1 + x**8>>
```

### DIFF
--- a/lib/Dialect/Polynomial/IR/PolynomialDialect.cpp
+++ b/lib/Dialect/Polynomial/IR/PolynomialDialect.cpp
@@ -42,26 +42,15 @@ struct PolynomialOpAsmDialectInterface : public OpAsmDialectInterface {
                    .Case<RingAttr>([&](auto &ringAttr) {
                      os << "ring_";
 
-                     auto coefficientType = ringAttr.getCoefficientType();
-                     auto res =
-                         llvm::TypeSwitch<Type, AliasResult>(coefficientType)
-                             .Case<IntegerType>([&](auto &integerType) {
-                               os << integerType;
-                               return AliasResult::FinalAlias;
-                             })
-                             .template Case<mod_arith::ModArithType>(
-                                 [&](auto &modArithType) {
-                                   os << "Z";
-                                   os << modArithType.getModulus().getValue();
-                                   os << "_";
-                                   os << modArithType.getModulus().getType();
-                                   return AliasResult::FinalAlias;
-                                 })
-                             .Default([&](auto &type) {
-                               return AliasResult::NoAlias;
-                             });
+                     auto type = ringAttr.getCoefficientType();
+                     auto *interface =
+                         dyn_cast<OpAsmDialectInterface>(&type.getDialect());
+
+                     auto res = interface->getAlias(type, os);
                      if (res == AliasResult::NoAlias) {
-                       return res;
+                       // always safe as MLIR will sanitize it into an
+                       // identifier
+                       os << type;
                      }
 
                      auto polynomialModulus = ringAttr.getPolynomialModulus();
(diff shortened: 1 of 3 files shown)


j2kun (approved):

j2kun:
This change seems fine, I'll approve. An alternative would have been to write to a temporary stream and then only do a single write to `os`.

ZenithalHourlyRate:
Got a better way to handle this.

ZenithalHourlyRate:
Ready now.

`os << type` is always safe as MLIR will sanitize it into an identifier.

```mlir
#ring_tensor3C4xi323E_1_x8_ = #polynomial.ring<coefficientType = tensor<4xi32>, polynomialModulus = <1 + x**8>>
```

lib/Dialect/Polynomial/IR/PolynomialDialect.cpp
```diff
@@ -42,26 +42,13 @@ struct PolynomialOpAsmDialectInterface : public OpAsmDialectInterface {
...
-                               return AliasResult::NoAlias;
-                             });
+                     auto type = ringAttr.getCoefficientType();
+                     auto *interface =
+                         dyn_cast<OpAsmDialectInterface>(&type.getDialect());
+
+                     auto res = interface->getAlias(type, os);
                      if (res == AliasResult::NoAlias) {
-                       return res;
+                       os << type;
```
AlexanderViand-Intel, line 51: > `os << type` is always safe as MLIR will sanitize it into an identifier.

Maybe add a comment about this in the actual code?

AlexanderViand-Intel (approved):
Thanks for the quick fix!

j2kun (approved):

lib/Dialect/Polynomial/IR/PolynomialDialect.cpp
ZenithalHourlyRate: Done.

asraa (approved):

PR #1146, opened 2024-12-05
Extend `--lwe-add-client-interface` to CKKS (among other fixes)

AlexanderViand-Intel:
Interestingly, no CKKS-specific change in `--lwe-add-client-interface` itself was necessary (which I think is a good sign for HEIR's architecture ✨), so this is mostly various enabling fixes/features:

* `--secret-to-ckks`: this now uses the `inverse canonical embedding` encoding attribute for its ptxts and ctxts
* `--lwe-add-client-interface`: this can now find RLWEParams from a Ctxt-typed function argument if if that argument is a tensor/etc
* `lwe` Dialect:  the encode/decode ops now also allow Float-like and not just Integer-like
* OpenFHE Emitter: 
  *  it allows f8/f16 types but since they don't exist in C++ (pre C++23), it just uses float (and emits a corresponding warning). More generally, the type emitter system can now generate warnings with MLIR Locations.
  * It now also accepts "1D-like" tensors (e.g., `tensor<1x1x16x...>`)
  * Multi-dim tensors are no longer translated to `std::vector<std::vector<...>>` nests but instead a single flattened vector.  This uses pre-existing logic (from the binary part of HEIR, afaik) for translating indices to flattened form.
  * When the CKKS scheme flag is given, it uses `getCKKSPackedValue` instead of just `getPackedValue` (and also converts from `std::complex` by discarding the imaginary part).
* move `--lwe-add-client-interface` from the OpenFHE specific pipeline to the general RLWE pipeline
* remove the custom client functions from the CKKS examples in `tests/Examples/openfhe` (except `naive_matmul` which is a bit unusual)


lib/Dialect/Secret/Conversions/SecretToCKKS/SecretToCKKS.cpp
```diff
@@ -105,7 +105,8 @@ class SecretToCKKSTypeConverter : public TypeConverter {
       // TODO(#785): Set a scaling parameter for floating point values.
       auto ciphertext = lwe::RLWECiphertextType::get(
           ctx,
-          lwe::PolynomialEvaluationEncodingAttr::get(ctx, bitWidth, bitWidth),
+          lwe::InverseCanonicalEmbeddingEncodingAttr::get(ctx, bitWidth,
```
asraa, line 108: nice fix, thanks!

tests/Examples/openfhe/naive_matmul.mlir
```diff
@@ -9,7 +9,7 @@
 !rlwe_pt = !lwe.rlwe_plaintext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, ring = #ring_Z463187969_i32_1_x1024_, underlying_type = f32>
 !rlwe_ct = !lwe.rlwe_ciphertext<encoding = #lwe.polynomial_evaluation_encoding<cleartext_start = 32, cleartext_bitwidth = 32>, rlwe_params = <ring = #ring_Z463187969_i32_1_x1024_>, underlying_type = f32>
 module {
-  func.func @matmul(%arg0: tensor<1x16x!rlwe_ct>, %arg1: tensor<1x16x!rlwe_ct> {secret.secret}) -> tensor<1x16x!rlwe_ct> {
+  func.func @matmul(%arg0: tensor<1x16x!rlwe_ct>, %arg1: tensor<1x16x!rlwe_ct>) -> tensor<1x16x!rlwe_ct> {
```
asraa, line 12: hah, thank you! my bad

asraa (commented):
Thank you so much! So many cleanups - there's just one FIXME in there for the TODO link :)

tests/Examples/openfhe/naive_matmul.mlir
AlexanderViand-Intel: Actually, I think I might have added that in my last secretize-cleanup PR 🙈

AlexanderViand-Intel:
> there's just one FIXME in there for the TODO link :)
Oops, thought I'd gotten them all.

On that note: If you search for `FIXME` in the current codebase there's actually a few... Some are in the jinja templates for new passes/etc, so that makes sense, but there's also a bunch in `CGGITo{Jaxite,TfheRust,TfheBool}` and a few other files that should probably be linked to issues, too.

lib/Dialect/LWE/IR/LWEOps.td
```diff
@@ -85,7 +85,7 @@ def LWE_EncodeOp : LWE_Op<"encode", [HasEncoding<"output", "encoding", "LWEPlain
   }];
 
   let arguments = (ins
-    SignlessIntegerLike:$plaintext,
+    SignlessIntegerOrFloatLike:$plaintext,
```
j2kun, line 88: Perhaps this should come with a verifier that ensures the int/float is compatible with the chosen encoding.

j2kun (changes requested):

lib/Dialect/LWE/IR/LWEOps.td
AlexanderViand-Intel: Can we make that judgment at this point, though? We could meaningfully want to encode cleartext floats into BGV style plain texts (fixes point representation), pack scalars into SIMD plaintexts, integers into CKKS, etc. I guess the only one that *really* doesn't work is putting a vector into a non-packed encoding - I'll add a verifier for that tomorrow :)

j2kun: Yeah I was thinking mostly that the float input type is only compatible with CKKS. A fixed-point type might be compatible with BGV, but since there's no fixed-point type in SignlessIntegerOrFloatLike (you'd need the quantization dialect I believe), that would default to being an integer type with fixed-point semantics.

tests/Examples/openfhe/dot_product_8f_test.cpp
```diff
@@ -43,9 +19,9 @@ TEST(DotProduct8FTest, RunTest) {
...
       dot_product__configure_crypto_context(cryptoContext, secretKey);
 
-  std::vector<double> arg0 = {0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8};
-  std::vector<double> arg1 = {0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9};
-  double expected = 2.4 + 0.1;
+  std::vector<float> arg0 = {0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8};
+  std::vector<float> arg1 = {0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9};
+  float expected = 2.4 + 0.1;
 
   // TODO(#891): support other schemes besides BGV in add-client-interface
```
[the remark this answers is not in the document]
AlexanderViand-Intel: Thanks, done!

lib/Dialect/LWE/IR/LWEOps.td
AlexanderViand-Intel: Done!

j2kun (approved):

AlexanderViand-Intel:
> @AlexanderViand-Intel unfortunately we haven't figured out how to auto-squash PRs yet, so you'll have to manually squash.

Done!

j2kun:
I think I did the internal patch on Friday and it's just awaiting internal review

PR #1152, opened 2024-12-06
wrap-generic: skip secretless function body

### LINKED ISSUE #1151: CollapseSecretlessGeneric does not fix function signature when all secret::Generic completely collapsed
For input

```mlir
  func.func @func(%arg0: i16) -> i16 {
    %c1_i16 = arith.constant 1 : i16
    return %c1_i16 : i16
  }
```

with `--secretize --wrap-generic --canonicalize`, we get the error

```
error: type of return operand 0 ('i16') doesn't match function result type ('!secret.secret<i16>') in function @func
func.func @func(
^
```

From the debug message, it shows that `CollapseSecretlessGeneric` canonicalization pattern collapsed the `GenericOp`, altered its Use (`func::ReturnOp`), then the function body becomes

```mlir
"func.func"() <{function_type = (!secret.secret<i16>) -> !secret.secret<i16>, sym_name = "func"}> ({
^bb0(%arg0: !secret.secret<i16>):                                         
  %0 = "arith.constant"() <{value = 1 : i16}> : () -> i16
  "func.return"(%0) : (i16) -> ()                                         
}) : () -> ()                                                             
```

We should stop such secretless code to `wrap-generic`, which can be done by adding `--cse --canoicalize` before `--wrap-generic` then let `--wrap-generic` check if any argument has any Uses.

ZenithalHourlyRate:
Fixes #1151 

Call cse/canonicalize in this pass because there might be some function body that is dead but uses the argument

### DIFF
--- a/lib/Transforms/Secretize/WrapGeneric.cpp
+++ b/lib/Transforms/Secretize/WrapGeneric.cpp
@@ -20,13 +20,43 @@ struct WrapWithGeneric : public OpRewritePattern<func::FuncOp> {
   LogicalResult matchAndRewrite(func::FuncOp op,
                                 PatternRewriter &rewriter) const override {
     bool hasSecrets = false;
+    bool secretHasUse = false;
 
+    // check if any of the arguments are secret
+    // and if any of the secret arguments are used
+    for (unsigned i = 0; i < op.getNumArguments(); i++) {
+      if (op.getArgAttr(i, secret::SecretDialect::kArgSecretAttrName) !=
+          nullptr) {
+        hasSecrets = true;
+
+        if (!op.getArgument(i).getUses().empty()) {
+          secretHasUse = true;
+        }
+      }
+    }
+
+    if (!hasSecrets) {
+      // Match failure, no secret inputs.
+      return failure();
+    }
+    if (!secretHasUse) {
+      // No need to wrap the function, no secret inputs are used.
+      // remove secret attribute
+      for (unsigned i = 0; i < op.getNumArguments(); i++) {
+        if (op.getArgAttr(i, secret::SecretDialect::kArgSecretAttrName) !=
+            nullptr) {
+          op.removeArgAttr(i, secret::SecretDialect::kArgSecretAttrName);
+        }
+      }
+      return failure();
+    }
+
+    // begin transform
     SmallVector<Type, 4> newInputs;
     for (unsigned i = 0; i < op.getNumArguments(); i++) {
       auto argTy = op.getArgument(i).getType();
       if (op.getArgAttr(i, secret::SecretDialect::kArgSecretAttrName) !=
           nullptr) {
-        hasSecrets = true;
         op.removeArgAttr(i, secret::SecretDialect::kArgSecretAttrName);
 
         auto newTy = secret::SecretType::get(argTy);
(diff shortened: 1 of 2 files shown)


lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -20,13 +20,43 @@ struct WrapWithGeneric : public OpRewritePattern<func::FuncOp> {
   LogicalResult matchAndRewrite(func::FuncOp op,
                                 PatternRewriter &rewriter) const override {
     bool hasSecrets = false;
+    bool secretHasUse = false;
 
+    // check if any of the arguments are secret
+    // and if any of the secret arguments are used
+    for (unsigned i = 0; i < op.getNumArguments(); i++) {
```
j2kun, line 27: If we proceed with this PR, I think this could be made slightly cleaner by keeping the "!hasSecrets" check from the original code, and replacing this loop with

``` 
for (unsigned i = 0; i < op.getNumArguments(); i++) {
  if (op.getArgAttr(i, secret::SecretDialect::kArgSecretAttrName) !=
      nullptr && !op.getArgument(i).getUses().empty()) {
    op.removeArgAttr(i, secret::SecretDialect::kArgSecretAttrName);
  }
}
```

That is, always remove secret attributes for unused values, and then the original check for "no secret values" would work as normal.

asraa (commented):
Wouldn't the CollapseSecretlessGeneric / RemoveNonSecretGenericArgs solve this? I wonder if you just add the `populateGenericCanonicalizers` onto this pass patterns?

j2kun (commented):
My only concern with this PR is that this changes the front-end interface of the compiler. If someone says "this function argument is secret," then they expect the compiler to convert the type signature to an encrypted version of that argument's type. It would be surprising to then try to call the function with an encrypted value and encounter a type error.

This is a risk even though you're right, that annotating an unused func arg as secret is almost certainly a bug on the user's part.

The main problem is that after all the folding, the return value is not secret. As an alternative, in that case we could emit a warning and insert a `secret.conceal` op which would lower to a trivial encryption to ensure the function signature remains secret. Maybe this materializes as a pattern that looks for a `func.return` op whose argument is not secret when the containing function has the matching secret return type. Or maybe we can add an optional attribute to the secret type to force its preservation, and the relevant folders/patterns that would remove secret types can check for it and fail if it is found.

What do you think?

ZenithalHourlyRate:
> My only concern with this PR is that this changes the front-end interface of the compiler. If someone says "this function argument is secret," then they expect the compiler to convert the type signature to an encrypted version of that argument's type.

Also addressing https://github.com/google/heir/issues/1151#issuecomment-2523855612, if we go with the way with fixing `CollapseSecretlessGeneric`, user could still experience this surprise. Doing it later with `CollapseSecretlessGeneric` might be cleaner, but that means type conversion back then why should not stop at first.

We should discuss the semantic of `{secret.secret}`; is that an annotation mainly for compiler and the compiler can discard it at will, or is that the responsiblity of the user and compiler would reject it when it found false secretness.

j2kun:
> Since the annotations are attributes and could be discard

While I agree with the conclusion that `--wrap-generic` should always be first, in most pipelines function op attributes happen to be preserved because there are so few passes that wholesale replace function ops, and the ones that do are careful to copy over all function attributes.

Issue #987, opened 2024-09-25
test: secret_inserts.mlir is flakey 


asraa:
At head in CI I see that `tests/convert_secret_insert_to_static_insert/secret_inserts.mlir` flakily fails (maybe 2/10 runs).

Here's the two outputs for the @insert_and_sum file:

```
  func.func @insert_and_sum(%arg0: !secret.secret<tensor<32xi16>>, %arg1: !secret.secret<index>) -> !secret.secret<tensor<32xi16>> {
    %c0_i16 = arith.constant 0 : i16
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<tensor<32xi16>>, !secret.secret<index>) {
    ^bb0(%arg2: tensor<32xi16>, %arg3: index):
      %1:2 = affine.for %arg4 = 0 to 32 iter_args(%arg5 = %c0_i16, %arg6 = %arg2) -> (i16, tensor<32xi16>) {
        %extracted = tensor.extract %arg6[%arg4] : tensor<32xi16>
        %3 = arith.addi %arg5, %extracted : i16
        %4 = affine.for %arg7 = 0 to 32 iter_args(%arg8 = %arg6) -> (tensor<32xi16>) {
          %5 = arith.cmpi eq, %arg7, %arg3 : index
          %6 = affine.for %arg9 = 0 to 32 iter_args(%arg10 = %arg8) -> (tensor<32xi16>) {
            %8 = arith.cmpi eq, %arg9, %arg7 : index
            %inserted = tensor.insert %3 into %arg10[%arg9] : tensor<32xi16>
            %9 = scf.if %8 -> (tensor<32xi16>) {
              scf.yield %inserted : tensor<32xi16>
            } else {
              scf.yield %arg10 : tensor<32xi16>
            }
            affine.yield %9 : tensor<32xi16>
          }
          %7 = scf.if %5 -> (tensor<32xi16>) {
            scf.yield %6 : tensor<32xi16>
          } else {
            scf.yield %arg8 : tensor<32xi16>
          }
          affine.yield %7 : tensor<32xi16>
        }
        affine.yield %3, %4 : i16, tensor<32xi16>
      }
      %2 = affine.for %arg4 = 0 to 32 iter_args(%arg5 = %1#1) -> (tensor<32xi16>) {
        %3 = arith.cmpi eq, %arg4, %arg3 : index
        %inserted = tensor.insert %1#0 into %arg5[%arg4] : tensor<32xi16>
[...]

PR #1154, opened 2024-12-06
transforms: re-run analysis when needed by subsequent rewrite

### LINKED ISSUE #1153: Bogus solver->propagateIfChanged in transform passes
In several passes like `convert-if-to-select` and `convert-secret-insert-to-static-insert`, the transformation   will replace ops, thus creating new `Value`, invalidating the `SecretAnalysis` as we can not get the `SecretLattice` for that `Value`.

To address such problem, they used `solver->propagateIfChanged` inside the RewritePattern, when the Analysis has stopped running, such as
https://github.com/google/heir/blob/8f222b27ab7f2cc03369c4ea037bdfc79d505cbb/lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp#L91-L96

Actually, `solver->propagateIfChanged` only adds the work item that "other Lattice depending on this Lattice should update", but not actually executing it. This can be checked by these lines (collapsed below)

<summary>MLIR source codes
<details>

DataFlowSolver propagateIfChanged:

https://github.com/llvm/llvm-project/blob/c91ba04328e1ded6f284469a7828d181324d4e30/mlir/lib/Analysis/DataFlowFramework.cpp#L135-L143

For generic Lattice:

https://github.com/llvm/llvm-project/blob/c91ba04328e1ded6f284469a7828d181324d4e30/mlir/include/mlir/Analysis/DataFlowFramework.h#L467-L474

For SparseLattice (sparse lattice use a Value's useDef to maintain dependency):

https://github.com/llvm/llvm-project/blob/c91ba04328e1ded6f284469a7828d181324d4e30/mlir/lib/Analysis/DataFlow/SparseAnalysis.cpp#L33-L40

However, the worklist is only consumed by solver->initializeAndRun

https://github.com/llvm/llvm-project/blob/c91ba04328e1ded6f284469a7828d181324d4e30/mlir/lib/Analysis/DataFlowFramework.cpp#L106-L133
</details>
</summary>

We also have an example below:

```mlir
func.func @secret_condition_with_non_secret_int(%inp: i16, %cond: !secret.secret<i1>) -> !secret.secret<i16> {
  %0 = secret.generic ins(%inp, %cond : i16, !secret.secret<i1>) {
  ^bb0(%copy_inp: i16, %secret_cond: i1):
    %1 = scf.if %secret_cond -> (i16) {
      %2 = arith.addi %copy_inp, %copy_inp : i16
      scf.yield %2 : i16
    } else {
      scf.yield %copy_inp : i16
    }
    %2 = arith.addi %1, %1 : i16
    %3 = arith.muli %2, %2 : i16
    secret.yield %3 : i16
  } -> !secret.secret<i16>
}
```

After running `--convert-if-to-select`, with proper LLVM_DEBUG we get the following log, which shows incorrect analysis result

```mlir
Running ConvertIfToSelect pass
SecretnessAnalysis: Visiting operation ... (many lines)
Before tranformation:
module {
  func.func @secret_condition_with_non_secret_int(%arg0: i16, %arg1: !secret.secret<i1>) -> !secret.secret<i16> {
    %0 = secret.generic ins(%arg0, %arg1 : i16, !secret.secret<i1>) attrs = {secretness = true} {
    ^bb0(%arg2: i16, %arg3: i1):
      %1 = scf.if %arg3 -> (i16) {
        %4 = arith.addi %arg2, %arg2 {secretness = false} : i16
        scf.yield %4 : i16
      } else {
        scf.yield %arg2 : i16
      } {secretness = false}
      %2 = arith.addi %1, %1 {secretness = false} : i16
      %3 = arith.muli %2, %2 {secretness = false} : i16
      secret.yield %3 : i16
    } -> !secret.secret<i16>
  }
}
(incorrect analysis for %2/%3, expected)
Before apply pattern
(NOTE: no VisitOperation called!)
After apply pattern
module {
  func.func @secret_condition_with_non_secret_int(%arg0: i16, %arg1: !secret.secret<i1>) -> !secret.secret<i16> {
    %0 = secret.generic ins(%arg0, %arg1 : i16, !secret.secret<i1>) attrs = {secretness = true} {
    ^bb0(%arg2: i16, %arg3: i1):
      %1 = arith.addi %arg2, %arg2 {secretness = false} : i16
      %2 = arith.select %arg3, %1, %arg2 {secretness = true} : i16
      %3 = arith.addi %2, %2 {secretness = false} : i16
      %4 = arith.muli %3, %3 {secretness = false} : i16
      secret.yield %4 : i16
    } -> !secret.secret<i16>
  }
}

(the secretness of %3 and %4 did not get properly propagated, even if %2 get the correct SecretLattice as the transformation manually set it up)
```

The solution to it is to call `solver->initializeAndRun` after each PatternRewrite.

ZenithalHourlyRate:
Fixes #1153 

Analysis should be re-run across different match of RewritePattern, otherwise even if the Lattice exists, it does not propagate and would result in incorrect result.

Also cleanup of annotate-secretness used in multiple places.

### DIFF
--- a/lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
+++ b/lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
@@ -102,5 +102,30 @@ LogicalResult SecretnessAnalysis::visitOperation(
   return mlir::success();
 }
 
+void annotateSecretness(Operation *top, DataFlowSolver *solver) {
+  // Add an attribute to the operations to show determined secretness
+  top->walk([&](Operation *op) {
+    for (unsigned i = 0; i < op->getNumResults(); ++i) {
+      std::string name = op->getNumResults() == 1
+                             ? "secretness"
+                             : "result_" + std::to_string(i) + "_secretness";
+      auto *secretnessLattice =
+          solver->lookupState<SecretnessLattice>(op->getOpResult(i));
+      if (!secretnessLattice) {
+        op->setAttr(name, StringAttr::get(op->getContext(), "null"));
+        return;
+      }
+      if (!secretnessLattice->getValue().isInitialized()) {
+        op->setAttr(name, StringAttr::get(op->getContext(), ("unknown")));
+        return;
+      }
+      op->setAttr(name,
+                  BoolAttr::get(op->getContext(),
+                                secretnessLattice->getValue().getSecretness()));
+    }
+    return;
+  });
+}
+
 }  // namespace heir
 }  // namespace mlir
--- a/lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
+++ b/lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
@@ -33,8 +33,9 @@ struct IfToSelectConversion : OpRewritePattern<scf::IfOp> {
   using OpRewritePattern<scf::IfOp>::OpRewritePattern;
 
  public:
-  IfToSelectConversion(DataFlowSolver *solver, MLIRContext *context)
-      : OpRewritePattern(context), solver(solver) {}
+  IfToSelectConversion(Operation *top, DataFlowSolver *solver,
+                       MLIRContext *context)
+      : OpRewritePattern(context), top(top), solver(solver) {}
 
   LogicalResult matchAndRewrite(scf::IfOp ifOp,
                                 PatternRewriter &rewriter) const override {
@@ -88,20 +89,17 @@ struct IfToSelectConversion : OpRewritePattern<scf::IfOp> {
             ifOp.getLoc(), cond, trueVal, falseVal);
       }
 
-      // Update the secretness of the new results, using the "secretness" of
-      // the condition which could have been either "secret" or "uninitialized"
-      for (auto &r : newResults) {
-        auto *lattice = solver->getOrCreateState<SecretnessLattice>(r);
-        solver->propagateIfChanged(lattice, lattice->join(secretness));
-      }
-
       rewriter.replaceOp(ifOp, newResults);
     }
 
-    return success();
+    // re-run the analysis propagate secretness across
+    // newly created Values
+    return solver->initializeAndRun(top);
   }
 
  private:
+  // root operation the pass is on, should never be altered hence never null
+  Operation *top;
   DataFlowSolver *solver;
 };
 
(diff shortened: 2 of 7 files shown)


lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
```diff
@@ -88,20 +89,16 @@ struct IfToSelectConversion : OpRewritePattern<scf::IfOp> {
...
     }
 
-    return success();
+    // re-run the analysis propagate secretness across
+    // newly created Values
+    return solver->initializeAndRun(top);
   }
 
  private:
+  Operation *top;
```
AlexanderViand-Intel, line 101: Is there a potential for this to become null during a pass, e.g., because a rewrite replaces the main function operation the pass acts on?

lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
```diff
@@ -102,5 +102,30 @@ LogicalResult SecretnessAnalysis::visitOperation(
   return mlir::success();
 }
 
+void annotateSecretness(Operation *top, DataFlowSolver *solver) {
```
AlexanderViand-Intel, line 105: Nice cleanup!

AlexanderViand-Intel (approved):
Nice spot,thanks! My only concern is whether keeping a pointer to the original operation around is always safe?
EDIT: Addressed, LGTM!

lib/Transforms/ConvertIfToSelect/ConvertIfToSelect.cpp
ZenithalHourlyRate: this is often the ModuleOp, which the pass takes as the root operation (runOnOperation) and is not allowed to alter it.

https://mlir.llvm.org/docs/PassManagement/

> All passes in MLIR derive from OperationPass often (ModuleOp) and adhere to the following restrictions
>
> Must not modify the state of operations other than the operations that are nested under the current operation.

AlexanderViand-Intel: Oh, you're right - since you set this to `getOperation()` in the `runOnOperation` function, it would indeed be a bug for a pass to delete that op! Can you add a comment about that? Thanks!

j2kun (approved):

asraa (approved):
Nice!! I just realized, maybe this is related? https://github.com/google/heir/issues/987

I'll try to revert the check once this is submitted! https://github.com/google/heir/commit/96185d91acba9726459677703eea9a004f05b7d5

ZenithalHourlyRate:
> maybe this is related? #987

I was also thinking of this issue

Issue #1148, opened 2024-12-05
`--optimize-relinearization` crashing


AlexanderViand-Intel:
@j2kun  While trying to add the `--optimize-relinearization` pass to the default RLWE pipeline, I've unfortunately hit some kind of issue:

`bazel run //tools:heir-opt -- --mlir-to-bgv=ciphertext-degree=8  --optimize-relinearization  $PWD/tests/Examples/openfhe/dot_product_8.mlir` 
says 
````
Error solving the problem: INVALID_ARGUMENT: duplicate name inserted: 
InsertRelin_bgv.rotate_loc("/home/aviand/code/heir/tests/Examples/openfhe/dot_product_8.mlir":1:1); ModelProto.variables are invalid
/home/aviand/code/heir/tests/Examples/openfhe/dot_product_8.mlir:1:1: error: Failed to solve the optimization problem```
then dumps the IR before actually crashing with 
```
heir-opt: external/llvm-project/llvm/include/llvm/ADT/DenseMap.h:204: const ValueT &llvm::DenseMapBase<llvm::DenseMap<mlir::Value, operations_research::math_opt::Variable>, mlir::Value, operations_research::math_opt::Variable, llvm::DenseMapInfo<mlir::Value>, llvm::detail::DenseMapPair<mlir::Value, operations_research::math_opt::Variable>>::at(const_arg_type_t<KeyT>) const [DerivedT = llvm::DenseMap<mlir::Value, operations_research::math_opt::Variable>, KeyT = mlir::Value, ValueT = operations_research::math_opt::Variable, KeyInfoT = llvm::DenseMapInfo<mlir::Value>, BucketT = llvm::detail::DenseMapPair<mlir::Value, operations_research::math_opt::Variable>]: Assertion `Iter != this->end() && "DenseMap::at failed due to a missing key"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
````
and dumping the stack.

In fact, if I dump the output of `--mlir-to-bgv=ciphertext-degree=8` into a file and then call heir-opt with `--optimize-relinearization` on that, it'll jump straight to the crashing/stack dump, not even emitting the error and IR...

**Comment by j2kun** [comment]
Acknowledged and reproduced. I will investigate tomorrow

**Comment by j2kun** [comment]
Another smell that is bizarre: when I dump the IR after `-secret-to-bgv` I get

```
!Z463187969_i32_ = !mod_arith.int<463187969 : i32>
[...]

PR #1155, opened 2024-12-06
ensure optimize-relinearization variable names are unique

j2kun:
Fixes https://github.com/google/heir/issues/1148

### DIFF
--- a/lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
+++ b/lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
@@ -51,13 +51,6 @@ namespace heir {
 
 #define DEBUG_TYPE "optimize-relinearization-analysis"
 
-std::string nameAndLoc(Operation *op) {
-  std::string varName;
-  llvm::raw_string_ostream ss(varName);
-  ss << op->getName() << "_" << op->getLoc();
-  return ss.str();
-}
-
 bool hasCiphertextType(ValueRange range) {
   return llvm::any_of(range, [](Value value) {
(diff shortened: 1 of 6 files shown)


lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
```diff
@@ -67,6 +61,32 @@ bool hasCiphertextType(ValueRange range) {
...
+  int resultFreeVarNameIndex = 0;
+  auto uniqueName = [&](Operation *op) {
+    std::string varName;
+    llvm::raw_string_ostream ss(varName);
+    ss << op->getName() << "_";
+    if (useLocBasedVariableNames) {
+      ss << op->getLoc();
+    } else if (op->getNumResults() == 0) {
+      ss << resultFreeVarNameIndex++;
+    } else {
```
asraa, line 84: QQ - why not only use the free var name index if you have to use it for the 0 result case? Can you just omit SelectVariableNames in this pass completely if you already have a free var name counter?
(I guess for simplicity why not apply this to all ops, regardless of how many results they have).

asraa (commented):
Nice fix!

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
j2kun: Yeah, I could just give up and use completely opaque identifiers when the pass option is not set. I think I will do that.

asraa (approved):

AlexanderViand-Intel (approved):
LGTM

PR #1163, opened 2024-12-09
Fix/Workaround for type mismatch after CollapseSecretlessGeneric

### LINKED ISSUE #1151: CollapseSecretlessGeneric does not fix function signature when all secret::Generic completely collapsed

AlexanderViand-Intel:
Fixes #1151

### DIFF
--- a/lib/Dialect/Secret/IR/SecretOps.cpp
+++ b/lib/Dialect/Secret/IR/SecretOps.cpp
@@ -467,7 +467,7 @@ void populateGenericCanonicalizers(RewritePatternSet &patterns,
                                    MLIRContext *ctx) {
   patterns.add<CollapseSecretlessGeneric, RemoveUnusedYieldedValues,
                RemoveUnusedGenericArgs, RemoveNonSecretGenericArgs,
-               HoistPlaintextOps>(ctx);
+               HoistPlaintextOps, FixSecretInFunctionType>(ctx);
 }
 
 // When replacing a generic op with a new one, and given an op in the original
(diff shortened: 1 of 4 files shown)


j2kun (approved):

asraa:
I wasn't there over office hours, but can I ask why not do the following:
* In wrap generic, incorporate secretness analysis to update the result types of the function in this line: https://github.com/google/heir/blob/97d62eec208923f41989df782d147339d9fda287/lib/Transforms/Secretize/WrapGeneric.cpp#L45-L47

Then even if we wrap the generic body, at least the result types are correct for the function? The the collapse pattern won't error out.

asraa:
OK I see that comment was something you described in the issue and is blocked on: https://github.com/google/heir/issues/888

Could you add a comment or a link to the issue in that pattern?

asraa (approved):

PR #1164, opened 2024-12-10
fix crash in RotateAndReduce

### LINKED ISSUE #870: `--heir-simd-vectorizer` crash on tensor size 1
While playing around, I ran into this fun little edge case: 
```llvm
!t = tensor<1xi16>
func.func @foo(%x: !t , %y: !t ) -> !t {
  %r0 = tensor.empty() : !t
  %r = affine.for %i = 0 to 1 iter_args(%r = %r0) -> !t {
    %xi = tensor.extract %x[%i] : !t
    %yi = tensor.extract %y[%i] : !t
    %si = arith.addi %xi, %yi : i16
    %ri = tensor.insert %si into %r[%i] : !t
    affine.yield %ri : !t
  }
  return %r : !t
}
```
Runing `--heir-simd-vectorizer` on the snippet above crashes the compiler with a ValueRange assert (the loop unroll works fine, so it's really one of the vectorizer passes, but I didn't have time to go explore further where in the pipeline it crashes). Not high priority, just wanted to note it down.

AlexanderViand-Intel:
Fixes #870 

`--rotate-and-reduce`  crashed for tensors of size one due to a bug in `tryReplaceOperations`, which effectively did the following (simplified snippet):
```C++
// finalOp pointer is created
Operation *finalOp;
...
// for loop that would set finalOp, but is skipped when size is 1
for (int64_t shiftSize = tensorShape[0] / 2; shiftSize > 0;
         shiftSize /= 2) {
      auto rotatedTensor = b.create<tensor_ext::RotateOp>(
          tensor, b.create<arith::ConstantOp>(b.getIndexAttr(shiftSize)));
      auto addOp = b.create<ArithOp>(tensor, rotatedTensor);
      finalOp = addOp;
      tensor = addOp->getResult(0);
  }
...
// crashes with a surprisingly cryptic message 
op->replaceAllUsesWith(finalOp);
```

The fix is pretty simple: explicitly initialize finalOp as nullptr and only do the replace at the end if finalOp isn't null.

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
+++ b/lib/Dialect/TensorExt/Transforms/RotateAndReduce.cpp
@@ -46,7 +46,7 @@ struct RotateAndReduce : impl::RotateAndReduceBase<RotateAndReduce> {
                << "Trying to replace rotations ending in " << *op << "\n");
     auto b = ImplicitLocOpBuilder(op->getLoc(), op);
     auto tensor = reduction.getTensor();
-    Operation *finalOp;
+    Operation *finalOp = nullptr;
     auto tensorShape =
         mlir::cast<RankedTensorType>(tensor.getType()).getShape();
     for (int64_t shiftSize = tensorShape[0] / 2; shiftSize > 0;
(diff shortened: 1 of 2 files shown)


asraa (approved):
thanks!

j2kun (approved):

PR #1165, opened 2024-12-10
[poly][mod_arith] Update `fastNTT` to emit mod_arith ops

### LINKED ISSUE #1113: Convert fastNTT to emit mod_arith ops directly
In https://github.com/google/heir/pull/1067 I did something slightly hacky while migrating polynomial-to-standard to use mod_arith ops: for the more complicated parts of the NTT lowerings, instead of lowering to mod_arith ops and types, I inserted conversion ops into and out of mod_arith at the start and end of ntt/intt. This allowed me to complete the polynomial type migration more quickly, but leaves that pass in a strange state where some conversions lower to mod_arith and others convert to arith.

This issue is to finish converting fastNTT to mod_arith.

inbelic:
```- update computeReverseBitOrder to work on mod_arith type
- update butterfly operations to use mod_arith ops
- restrict the types of the ntt/intt ops to only allow the ModArith attr
  and add verification of the operations

- updates testcases lower_ntt.mlir and lower_intt.mlir to check for new
  emitted ops and the types of operand/results
- updates testcases lower_ntt_runner.mlir and lower_intt_runner.mlir to
  use the newly required types
- updates ntt_benchmakr.mlir and lower_ntt_perf_runner.mlir to use newly
required types
- updates canonicalization.mlir and mod_arith_coefficients.mlir to use
newly required types

- let tensor.cast be a legal op in the mod-arith-to-arith pass to stop a
failure to resolve materialization error
review comments:

- fix ElementwiseToAffine losing embedding info by passing it down to
the empty tensor create op
```

Resolves #1113

### DIFF
--- a/lib/Dialect/Polynomial/IR/PolynomialOps.td
+++ b/lib/Dialect/Polynomial/IR/PolynomialOps.td
@@ -1,6 +1,7 @@
 #ifndef LIB_DIALECT_POLYNOMIAL_IR_POLYNOMIALOPS_TD_
 #define LIB_DIALECT_POLYNOMIAL_IR_POLYNOMIALOPS_TD_
 
+include "lib/Dialect/ModArith/IR/ModArithTypes.td"
 include "lib/Dialect/Polynomial/IR/PolynomialAttributes.td"
 include "lib/Dialect/Polynomial/IR/PolynomialDialect.td"
 include "lib/Dialect/Polynomial/IR/PolynomialTypes.td"
@@ -309,7 +310,7 @@ def Polynomial_NTTOp : Polynomial_Op<"ntt", [Pure]> {
     Polynomial_PolynomialType:$input,
     OptionalAttr<Polynomial_PrimitiveRootAttr>:$root
   );
-  let results = (outs RankedTensorOf<[AnyInteger]>:$output);
+  let results = (outs RankedTensorOf<[ModArith_ModArithType]>:$output);
   let assemblyFormat = "$input attr-dict `:` qualified(type($input)) `->` type($output)";
   let hasCanonicalizer = 1;
   let hasVerifier = 1;
@@ -330,7 +331,7 @@ def Polynomial_INTTOp : Polynomial_Op<"intt", [Pure]> {
     The choice of primitive root may be optionally specified.
   }];
   let arguments = (
-    ins RankedTensorOf<[AnyInteger]>:$input,
+    ins RankedTensorOf<[ModArith_ModArithType]>:$input,
     OptionalAttr<Polynomial_PrimitiveRootAttr>:$root
   );
   let results = (outs Polynomial_PolynomialType:$output);
--- a/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
+++ b/lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
@@ -51,8 +51,8 @@ struct ConvertAnyElementwiseMappableOpOnRankedTensors : public RewritePattern {
     auto ip = rewriter.saveInsertionPoint();
 
     // Create an empty tensor as initial value of the iter_args
-    Value target =
-        rewriter.create<tensor::EmptyOp>(op->getLoc(), shape, elementType);
+    Value target = rewriter.create<tensor::EmptyOp>(
+        op->getLoc(), shape, elementType, resultType.getEncoding());
 
     llvm::SmallVector<Value, 1> indices;
 
--- a/tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_ntt_perf_runner.mlir
+++ b/tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_ntt_perf_runner.mlir
@@ -19,14 +19,14 @@ func.func @test_poly_ntt() {
   %insert_rand1 = tensor.insert_slice %rand_coeffs into %insert_rand0[65280] [256] [1] : tensor<256xi32> into tensor<65536xi32>
   %rand1_enc = mod_arith.encapsulate %insert_rand1 : tensor<65536xi32> -> tensor<65536x!coeff_ty>
   %poly = polynomial.from_tensor %rand1_enc : tensor<65536x!coeff_ty> -> !poly_ty
-  %0 = polynomial.ntt %poly {root=#root} : !poly_ty -> tensor<65536xi32, #ring>
+  %0 = polynomial.ntt %poly {root=#root} : !poly_ty -> tensor<65536x!coeff_ty, #ring>
 
   // Insert casts so that intt(ntt()) does not get folded away during polynomial
   // canonicalization
-  %cast = tensor.cast %0 : tensor<65536xi32, #ring> to tensor<65536xi32>
-  %cast_back = tensor.cast %cast : tensor<65536xi32> to tensor<65536xi32, #ring>
+  %cast = tensor.cast %0 : tensor<65536x!coeff_ty, #ring> to tensor<65536x!coeff_ty>
+  %cast_back = tensor.cast %cast : tensor<65536x!coeff_ty> to tensor<65536x!coeff_ty, #ring>
 
-  %1 = polynomial.intt %cast_back {root=#root} : tensor<65536xi32, #ring> -> !poly_ty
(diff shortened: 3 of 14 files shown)


inbelic:
Having refactored to using the `mod_arith` within the tensor type I get the [following error](https://gist.github.com/inbelic/9b3fc9c9793578dc117ee9dd6b6bba96) during the `mod-arith-to-arith` pass when testing the `lower_ntt_runner` and `lower_intt_runner` tests.

I poked around with the type-converter to see if was because the `RankedTensorType` encoding gets passed down when we clone the type, but I wasn't able to resolve the issue.

ZenithalHourlyRate:
> Having refactored to using the `mod_arith` within the tensor type I get the [following error](https://gist.github.com/inbelic/9b3fc9c9793578dc117ee9dd6b6bba96) during the `mod-arith-to-arith` pass when testing the `lower_ntt_runner` and `lower_intt_runner` tests.

It was that `tensor::CastOp` did not get through the TypeConverter. With the following diff I can get the `lower_ntt_runner` through the `poly-to-mod-arith` and `mod-arith-to-arith`

```diff
--- a/lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
+++ b/lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
@@ -337,9 +337,9 @@ void ModArithToArith::runOnOperation() {
   addStructuralConversionPatterns(typeConverter, patterns, target);
 
   target.addDynamicallyLegalOp<
-      tensor::EmptyOp, tensor::ExtractOp, tensor::InsertOp, affine::AffineForOp,
-      affine::AffineYieldOp, linalg::GenericOp, linalg::YieldOp,
-      tensor::ExtractSliceOp, tensor::InsertSliceOp>(
+      tensor::EmptyOp, tensor::ExtractOp, tensor::InsertOp, tensor::CastOp,
+      affine::AffineForOp, affine::AffineYieldOp, linalg::GenericOp,
+      linalg::YieldOp, tensor::ExtractSliceOp, tensor::InsertSliceOp>(
       [&](auto op) { return typeConverter.isLegal(op); });
 
 
```

inbelic:
Thanks, good catch. It works as you described when manually applying the passes, but the same error occurs if we try to directly use the `--heir-polynomial-to-llvm` pass.

I updated all the testcases affected by the change but they require to manually denote the `polynomial-to-mod-arith` and `mod-arith-to-arith` passes before `--heir-polynomial-to-llvm` to lower successfully. But it feels like this isn't the ideal case...

j2kun:
Planning to review this tomorrow.

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -309,7 +310,7 @@ def Polynomial_NTTOp : Polynomial_Op<"ntt", [Pure]> {
     Polynomial_PolynomialType:$input,
     OptionalAttr<Polynomial_PrimitiveRootAttr>:$root
   );
-  let results = (outs RankedTensorOf<[AnyInteger]>:$output);
+  let results = (outs RankedTensorOf<[ModArith_ModArithType]>:$output);
```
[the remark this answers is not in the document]
j2kun: IMO it's fine to leave it until we have a need for a more general type here (i.e., we don't have an immediate need for a vector/memref of mod_arith types)

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_ntt_perf_runner.mlir
```diff
@@ -1,4 +1,5 @@
-// RUN: heir-opt %s --heir-polynomial-to-llvm \
+// RUN: heir-opt %s --polynomial-to-mod-arith \
+// RUN:  --mod-arith-to-arith --heir-polynomial-to-llvm \
```
j2kun, line 2: I cloned your PR to my local machine and I am able to run `heir-polynomial-to-llvm` on this file without the extra pass flags. What error are you seeing that makes you think you should add this?

To be more specific, I ran

```
bazel run --noallow_analysis_cache_discard //tools:heir-opt -- --heir-polynomial-to-llvm $PWD/tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_ntt_perf_runner.mlir
```

It may help to add the flags `--mlir-print-ir-after-all --mlir-print-ir-tree-dir=/tmp/mlir` and that will print the IR after each pass to a file so you can inspect it more easily than `--debug`.

j2kun (approved):
LGTM with 1 minor change and a question

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_ntt_perf_runner.mlir
j2kun: FWIW, I also removed the flags and the tests still pass.

inbelic: Just to confirm that you only mean this specific testcase?

I was able to remove the flags on this test but not any of the others locally. I had just been a bit overzealous when adding the flags, so I don't have any failures locally for this testcase either.

j2kun: In principle this shouldn't be a problem because those passes are in the broader pipeline. We can submit as-is, but please put a TODO where it is strictly necessary (and if possible, post the error you see in the issue).

j2kun (approved):

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_ntt_perf_runner.mlir
inbelic: Narrowing down on a fix to the `ElementwiseToAffine` pass, need to step away for now but will update the pr later today.

inbelic:
Okay, I added the commit to resolve the issue. TLDR, was that the encoding attribute was lost when we constructed the tensor type.

It is rather unrelated to this pr so if you'd prefer to create a separate issue with its own pr I can do that. But it might be small enough to include in here?

LMK, and I will squash with/without the commit accordingly.

j2kun:
> Okay, I added the commit to resolve the issue. TLDR, was that the encoding attribute was lost when we re-constructed the tensor type during the `ElementwiseToAffine` pass.
> 
> It is rather unrelated to this pr so if you'd prefer to create a separate issue with its own pr I can do that. But it might be small enough to include in here?
> 
> LMK, and I will squash with/without the commit accordingly.

I think it's fine to include here. Good catch, and I always warn people about attributes getting lost across passes!

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
```diff
@@ -70,7 +70,11 @@ struct ConvertAnyElementwiseMappableOpOnRankedTensors : public RewritePattern {
 
       // If first loop: replace scalar op
       if (i == 0) {
-        rewriter.replaceOp(op, loop);
+        // We will cast back to the return type to retain an embedding if
+        // included
+        auto cast = rewriter.create<tensor::CastOp>(op->getLoc(), resultType,
+                                                    loop.getResults());
```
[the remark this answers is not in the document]
inbelic: Nice, yes there is! Changed to set the type directly in the `tensor.empty` op and removed the changes to the elementwise testcases.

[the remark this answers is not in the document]
j2kun: A nice improvement. I did check up if this would have been handled by canonicalization: `FoldEmptyTensorWithCastOp` is upstream, but it requires the encoding attributes to be the same to apply the folder.

j2kun (approved):

lib/Transforms/ElementwiseToAffine/ElementwiseToAffine.cpp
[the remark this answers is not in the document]
AlexanderViand-Intel: > Nice, yes there is! Changed to set the type directly in the `tensor.empty` op and removed the changes to the elementwise testcases.

Awesome, thanks!

PR #1173, opened 2024-12-11
[Draft] Convert the ExtSI operation to mod_arith

WoutLegiest:
Stuck on the conversion of the arith.extsi to the new mod_arith.mod_switch


WoutLegiest:
Debug output: 

```
ImplicitTypeIDRegistry::lookupOrInsert(mlir::OpTrait::VariadicResults<Empty>)
'mod_arith.mod_switch' op operand #0 must be Integer type with modular arithmetic, but got 'i8'
mlir-asm-printer: 'func.func' failed to verify and will be printed in generic form
ImplicitTypeIDRegistry::lookupOrInsert(mlir::AttributeTrait::IsLocation<Empty>)
/usr/local/google/home/legiest/heir/ml_affine.mlir:2:49: error: failed to legalize unresolved materialization from ('!mod_arith.int<127 : i8>') to ('i8') that remained live after conversion
  func.func @for_9097789534307714736(%arg0: i8, %arg1: i8, %arg2: i32) -> i32 {
                                                ^
/usr/local/google/home/legiest/heir/ml_affine.mlir:2:49: note: see current operation: %0 = "builtin.unrealized_conversion_cast"(%arg1) : (!mod_arith.int<127 : i8>) -> i8
/usr/local/google/home/legiest/heir/ml_affine.mlir:4:10: note: see existing live user here: %3 = "mod_arith.mod_switch"(%0) : (i8) -> !mod_arith.int<2147483647 : i32>
    %1 = arith.extsi %arg1 : i8 to i32
         ^
ImplicitTypeIDRegistry::lookupOrInsert(mlir::detail::PreservedAnalyses::AllAnalysesType)
```

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
```diff
@@ -108,6 +75,24 @@ struct ConvertConstant : public OpConversionPattern<mlir::arith::ConstantOp> {
...
+
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      ::mlir::arith::ExtSIOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    auto result = b.create<mod_arith::ModSwitchOp>(
+        op.getLoc(), convertArithType(op.getType()), op.getIn());
```
j2kun, line 90: I see the problem: you should be using `adaptor.getIn()` instead of `op.getIn()`.

The `OpAdaptor` has the type-converted operands and result types stored on it, but when you pass the original types from `ExtSIOp`, it leaves them as integers, and the op fails to verify and it adds that unrealized conversion cast op to try to compensate.

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
```diff
@@ -160,6 +145,29 @@ struct GenericOpPattern : public OpConversionPattern<Op> {
...
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      mlir::memref::LoadOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    // auto *parent = op.getMemRef().getDefiningOp();
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    SmallVector<Type> retTypes;
+    if (failed(this->getTypeConverter()->convertTypes(op->getResultTypes(),
```
j2kun, line 161: again, this should use the adaptor which already has the type-converted results.

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
WoutLegiest: Damn; that's the difference between adaptor and op. Thanks very much for the help!

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
asraa: Adaptors actually don't have the result types that are type converted. I forgot the reasoning behind this.

j2kun: Ah, thanks for the reminder!

Issue #1174, opened 2024-12-11
Use tensor_ext::ExtractOp for extracting an element out of a slot of a single ciphertext


ZenithalHourlyRate:
Related to #913 

In the secretized IR, there are two semantics of tensor::ExtractOp
  * One that does normal tensor extract, from a tensor of ciphertext
  * Another needs to extract a slot from a simd ciphertext

And this depends on the vectorizer to decide a tensor::ExtractOp bears which semantic. We can also check [the lowering code](https://github.com/google/heir/blob/016735e9f5bd652920206b454bb13c6ce799f6f2/lib/Dialect/Secret/Conversions/SecretToCKKS/SecretToCKKS.cpp#L151-L162) on how to decide which semantic one tensor::ExtractOp is.

For example, in matmul, tensor::extract is decided to be extracting from tensor of ciphertext

```mlir
  func.func @matmul(%arg0: tensor<1x2xf32> {secret.secret}, %arg1: tensor<1x2xf32> {secret.secret}) -> tensor<1x2xf32> {
    %0 = "tosa.const"() <{value = dense<[[2.0, 3.0], [4.0, 6.0]]> : tensor<2x2xf32>}> : () -> tensor<2x2xf32>
    %1 = affine.for %arg2 = 0 to 1 iter_args(%arg3 = %arg1) -> (tensor<1x2xf32>) {
      %2 = affine.for %arg4 = 0 to 2 iter_args(%arg5 = %arg3) -> (tensor<1x2xf32>) {
        %3 = affine.for %arg6 = 0 to 2 iter_args(%arg7 = %arg5) -> (tensor<1x2xf32>) {
          %extracted = tensor.extract %arg0[%arg2, %arg6] : tensor<1x2xf32>
          %extracted_0 = tensor.extract %0[%arg6, %arg4] : tensor<2x2xf32>
          %extracted_1 = tensor.extract %arg7[%arg2, %arg4] : tensor<1x2xf32>
          %4 = arith.mulf %extracted, %extracted_0 : f32
          %5 = arith.addf %extracted_1, %4 : f32
          %inserted = tensor.insert %5 into %arg7[%arg2, %arg4] : tensor<1x2xf32>
          affine.yield %inserted : tensor<1x2xf32>
        }
        affine.yield %3 : tensor<1x2xf32>
      }
      affine.yield %2 : tensor<1x2xf32>
    }
    return %1 : tensor<1x2xf32>
  }
```

[...]

PR #1176, opened 2024-12-11
CKKS: place mgmt ops and add RNS lowering

ZenithalHourlyRate:
This is a mirror of #1126.

However, the complicacy is that in the CKKS pipeline we support _tensor of ciphertext_, which renders the following difficulties

* tensor::ExtractOp has two semantic (mentioned in #1174)
* ckks::RescaleOp takes `tensor of ciphertext` as input, which only accepts RLWECiphertext but not RLWECiphertextLike.
* In `tests/Dialect/Secret/Conversions/secret_to_ckks/matmul_loop.mlir`, `affine` is not unrolled so `LevelAnalysis` can not work. It should not work as we do not have loop invariant kept here. For example, for the loop to carry out correctly in FHE, the invariant is `the level is the same`, which heavily depends on the body.

  More case analysis here. The body is a mul_plain + add, which in different mgmt op placement strategy there will be different level analysis result. For example, if we mod reduce immediately after mul_plain, we get the level reduced by one hence the invariant is not kept. For the current strategy we fortunately kept this invariant.

My current workarounds over these are

* Copy the secret-to-ckks code on deciding whether to pack tensor into ciphertext, then annotate these tensor::ExtractOp to distinguish semantic so that various analysis could work.
* Fully unroll the tensor, namely for each element we generate a ckks.rescale.
* I prefer to disable `matmul_loop.mlir` for now, asking the user to full-loop-unroll. In the future, I should make the analysis establishing invariant and error out when user wrote some invariant-broken code, so that we can have correct level analysis for loops.


j2kun (approved):
LGTM! Generally ok with the workarounds for now.

j2kun:
You will need to rebase to fix the build error I unfortunately pushed to main earlier today.

ZenithalHourlyRate:
Rebased and squashed.

ZenithalHourlyRate:
Fixed the building. Seems that mlir::arith and mlir::heir::arith (added by arith to mod arith) has some name conflict.

AlexanderViand-Intel:
This just got copybara-merged, but things look a bit funky?

asraa:
I have a feeling [9315298](https://github.com/google/heir/commit/931529889653bd1201cdfef5383f30a942bf18ac) was merged and that's why github didn't recognize this as merged. (I think jeremy cloned this and fixed the internal build issues at that commit?)

PR #1182, opened 2024-12-12
Lattigo: BGV: add relinearize/mod reduce/rotate

ZenithalHourlyRate:
This introduces the keyswitching/modulus switching ability thus we are able to do relinearize/rotate.

Check `tests/Dialect/Lattigo/Emitters/emit_lattigo.mlir` for detail usage of them. It runs and get a correct result.

Also, the following input can be used with current `secret-insert-mgmt-bgv` to get a correct management

```mlir
func.func @add(%arg0: tensor<4xi16>, %arg1: tensor<4xi16>) -> tensor<4xi16> {
  %0 = arith.addi %arg0, %arg1 : tensor<4xi16>
  %1 = arith.muli %0, %arg1 : tensor<4xi16>
  %2 = tensor_ext.rotate %1, %c1 : tensor<4xi16>, index
  return %2 : tensor<4xi16>
}
```

With `--secretize --mlir-to-secret-arithmetic --secret-insert-mgmt-bgv --secret-distribute-generic --secret-to-bgv=poly-mod-degree=4 --bgv-to-lattigo --cse` we get

```mlir
  func.func @add(%arg0: !lattigo.bgv.evaluator, %arg1: !lattigo.rlwe.ciphertext, %arg2: !lattigo.rlwe.ciphertext) -> !lattigo.rlwe.ciphertext {
    %0 = lattigo.bgv.add %arg0, %arg1, %arg2 : (!lattigo.bgv.evaluator, !lattigo.rlwe.ciphertext, !lattigo.rlwe.ciphertext) -> !lattigo.rlwe.ciphertext
    %1 = lattigo.bgv.mul %arg0, %0, %arg2 : (!lattigo.bgv.evaluator, !lattigo.rlwe.ciphertext, !lattigo.rlwe.ciphertext) -> !lattigo.rlwe.ciphertext
    %2 = lattigo.bgv.relinearize %arg0, %1 : (!lattigo.bgv.evaluator, !lattigo.rlwe.ciphertext) -> !lattigo.rlwe.ciphertext
    %3 = lattigo.bgv.rotate_columns %arg0, %2 {offset = 1 : index} : (!lattigo.bgv.evaluator, !lattigo.rlwe.ciphertext) -> !lattigo.rlwe.ciphertext
    %4 = lattigo.bgv.rescale %arg0, %3 : (!lattigo.bgv.evaluator, !lattigo.rlwe.ciphertext) -> !lattigo.rlwe.ciphertext
    return %4 : !lattigo.rlwe.ciphertext
  }
```

It can correctly translate to GO code and runs with correct result with a main function.

Note that `--lwe-add-client-interface` still does not work as I do not have a lowering for that. Thus we do not have `--mlir-to-lattigo-bgv` for now. Will do so in future PR.

### Discussion

* Rotation for BGV in lattigo is exposed as `RotateColumns` and `RotateRows`. If we want to get a correct rotation for any index (currently only works for index 1), we have to compute the correct way to use these API.
* GaloisKey are generated using the `x -> x^i` (called `galEl`), instead of `index`. As we can see in the code, I generated the GaloisKey for galEl 5 but then for rotation I used the index 1. If not generating with correct galois key, we get the following

```
panic: cannot apply Automorphism: GaloisKey[5] is nil: key for galEl 5 = 5^{1} key is missing
```

### DIFF
--- a/lib/Dialect/BGV/Conversions/BGVToLattigo/BGVToLattigo.cpp
+++ b/lib/Dialect/BGV/Conversions/BGVToLattigo/BGVToLattigo.cpp
@@ -33,6 +33,16 @@ using ConvertSubOp =
     ConvertRlweBinOp<lattigo::BGVEvaluatorType, SubOp, lattigo::BGVSubOp>;
 using ConvertMulOp =
     ConvertRlweBinOp<lattigo::BGVEvaluatorType, MulOp, lattigo::BGVMulOp>;
+using ConvertRelinOp =
+    ConvertRlweUnaryOp<lattigo::BGVEvaluatorType, RelinearizeOp,
+                       lattigo::BGVRelinearizeOp>;
+using ConvertModulusSwitchOp =
+    ConvertRlweUnaryOp<lattigo::BGVEvaluatorType, ModulusSwitchOp,
+                       lattigo::BGVRescaleOp>;
+
+// TODO(#1186): figure out generic rotating using BGVRotateColumns/RowsOp
+using ConvertRotateOp = ConvertRlweRotateOp<lattigo::BGVEvaluatorType, RotateOp,
+                                            lattigo::BGVRotateColumnsOp>;
 
 struct BGVToLattigo : public impl::BGVToLattigoBase<BGVToLattigo> {
   void runOnOperation() override {
@@ -60,8 +70,9 @@ struct BGVToLattigo : public impl::BGVToLattigoBase<BGVToLattigo> {
     });
 
     patterns.add<AddEvaluatorArg<bgv::BGVDialect, lattigo::BGVEvaluatorType>,
-                 ConvertAddOp, ConvertSubOp, ConvertMulOp>(typeConverter,
-                                                           context);
+                 ConvertAddOp, ConvertSubOp, ConvertMulOp, ConvertRelinOp,
+                 ConvertModulusSwitchOp, ConvertRotateOp>(typeConverter,
+                                                          context);
 
     if (failed(applyPartialConversion(module, target, std::move(patterns)))) {
       return signalPassFailure();
--- a/lib/Dialect/Lattigo/IR/LattigoBGVOps.td
+++ b/lib/Dialect/Lattigo/IR/LattigoBGVOps.td
@@ -76,9 +76,16 @@ def Lattigo_BGVNewEvaluatorOp : Lattigo_BGVOp<"new_evaluator"> {
   let summary = "Create a new evaluator in the Lattigo BGV dialect";
   let description = [{
     This operation creates a new evaluator for performing operations on ciphertexts in the Lattigo BGV dialect.
+
+    By default, the evaluator is created with the provided parameters and could execute
+    operations which does not relying on evaluation keys.
+
+    To support operations that require evaluation keys,
+    the optional evaluation key set should be provided.
   }];
   let arguments = (ins
-    Lattigo_BGVParameter:$params
+    Lattigo_BGVParameter:$params,
+    Optional<Lattigo_RLWEEvaluationKeySet>:$evaluationKeySet
   );
   let results = (outs Lattigo_BGVEvaluator:$evaluator);
 }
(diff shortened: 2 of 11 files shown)


lib/Dialect/BGV/Conversions/BGVToLattigo/BGVToLattigo.cpp
```diff
@@ -33,6 +33,16 @@ using ConvertSubOp =
...
 using ConvertMulOp =
     ConvertRlweBinOp<lattigo::BGVEvaluatorType, MulOp, lattigo::BGVMulOp>;
+using ConvertRelinOp =
+    ConvertRlweUnaryOp<lattigo::BGVEvaluatorType, RelinearizeOp,
+                       lattigo::BGVRelinearizeOp>;
+using ConvertModulusSwitchOp =
+    ConvertRlweUnaryOp<lattigo::BGVEvaluatorType, ModulusSwitchOp,
+                       lattigo::BGVRescaleOp>;
+
+// TODO: figure out generic rotating using BGVRotateColumns/RowsOp
```
j2kun, line 43: Please attach this to an issue.

lib/Dialect/Lattigo/IR/LattigoBGVOps.td
```diff
@@ -78,7 +78,8 @@ def Lattigo_BGVNewEvaluatorOp : Lattigo_BGVOp<"new_evaluator"> {
     This operation creates a new evaluator for performing operations on ciphertexts in the Lattigo BGV dialect.
   }];
   let arguments = (ins
-    Lattigo_BGVParameter:$params
+    Lattigo_BGVParameter:$params,
+    Optional<Lattigo_RLWEEvaluationKeySet>:$evaluationKeySet
```
j2kun, line 82: Please document how the semantics change when this is included.

lib/Dialect/Lattigo/IR/LattigoBGVOps.td
```diff
@@ -116,4 +117,47 @@ def Lattigo_BGVMulOp : Lattigo_BGVBinaryOp<"mul"> {
...
+  let summary = "Rescale a ciphertext in the Lattigo BGV dialect";
+  let description = [{
+    This operation rescales a ciphertext value in the Lattigo BGV dialect.
+  }];
+}
+
+def Lattigo_BGVRotateColumnsOp : Lattigo_BGVOp<"rotate_columns"> {
+  let summary = "Rotate columns of a ciphertext in the Lattigo BGV dialect";
+  let description = [{
+    This operation rotates the columns of a ciphertext value in the Lattigo BGV dialect.
```
j2kun, line 146: The lattigo code comments don't= explain the difference between rows and columns. Do you know? I expect "columns" corresponds to slots in the standard sense, but what does "rows" mean? It might be worth documenting here, or putting a link to where it is explained.

lib/Dialect/Lattigo/IR/LattigoRLWEOps.td
```diff
@@ -31,6 +31,43 @@ def Lattigo_RLWEGenKeyPairOp : Lattigo_RLWEOp<"gen_key_pair"> {
...
+
+def Lattigo_RLWEGenGaloisKeyOp : Lattigo_RLWEOp<"gen_galois_key"> {
+  let summary = "Generates a new RLWE Galois key";
+  let description = [{
+    This operation generates a new RLWE Galois key
+  }];
+  let arguments = (ins
+    Lattigo_RLWEKeyGenerator:$keyGenerator,
+    Lattigo_RLWESecretKey:$secretKey,
+    Builtin_IntegerAttr:$galoisElement
```
j2kun, line 54: The semantics of this galoisElement should be documented. Is it the `i` in `x -> x^i`?

lib/Target/Lattigo/LattigoEmitter.cpp
```diff
@@ -303,7 +361,9 @@ LogicalResult LattigoEmitter::printNewMethod(::mlir::Value result,
 LogicalResult LattigoEmitter::printEvalInplaceMethod(
     ::mlir::Value result, ::mlir::Value evaluator, ::mlir::Value operand,
     ::mlir::Value operandInplace, std::string_view op, bool err) {
-  std::string errResult = err ? ", err := " : "";
+  // TODO: err := or err = depends on the context...
```
j2kun, line 364: you could avoid this by generating new variable names for each err. Just have an incrementing counter class global.

j2kun (approved):
Very clean! Just a few requests for more docs in the tablegen.

> Rotation for BGV in lattigo is exposed as RotateColumns and RotateRows. If we want to get a correct rotation for any index (currently only works for index 1), we have to compute the correct way to use these API.
>
> GaloisKey are generated using the x -> x^i (called galEl), instead of index. As we can see in the code, I generated the GaloisKey for galEl 5 but then for rotation I used the index 1.

I'd like to know more about the problems here. I read through some of the lattigo docs and I still don't understand the organization or semantics of the rotation API :) Do you have anything I can read that explains it?

FWIW, it looks like the CKKS rotation API is much simpler, and only has one rotate method.

lib/Dialect/Lattigo/IR/LattigoBGVOps.td
ZenithalHourlyRate: I do have a reference comment in SEAL

https://github.com/microsoft/SEAL/blob/119dc32e135cb89c1062076a69310d4413ebc824/native/examples/2_encoders.cpp#L37-L106

But I can not immediately refer to a paper to explain this. Will investigate later.

lib/Dialect/Lattigo/IR/LattigoRLWEOps.td
ZenithalHourlyRate: yes. Will add a comment

ZenithalHourlyRate:
Addressed the review comments.

j2kun (approved):

ZenithalHourlyRate:
Test fixed.

asraa (commented):
@j2kun if you were asking about some references about how the automorphism index (power of the generator (5 in the openfhe ring, and probably lattigo) of the rotation/galois group) relates to the value of the rotational shift I think there is some discussion here: https://eprint.iacr.org/2011/566.pdf

PR #1184, opened 2024-12-12
Add decorator to Python frontend

j2kun:

This PR adds a decorator layer for the Python frontend.

For example, (from `e2e_test.py`)

```python
@compile(backend="openfhe")
def foo(a, b):
    return a * a - b * b

# Test the e2e path
self.assertEqual(-15, foo(7, 8))

# Also test the manual way with crypto_context and keys threaded in
# automatically
foo.setup()
enc_a = foo.encrypt_a(7)
enc_b = foo.encrypt_b(8)
result_enc = foo.foo(enc_a, enc_b)
result = foo.decrypt_result(result_enc)
self.assertEqual(-15, result)
```

Note the crypto context and private/public keys can also be accessed as member variables of `foo`, though which subset of variables are present would be backend-dependent. Also the allowed kwargs on the various functions would also be backend-dependent, as some backends may take the parameters/key args in different combinations.

### DIFF
--- a/heir_py/__init__.py
+++ b/heir_py/__init__.py
@@ -1,0 +1,3 @@
+from decorator import compile
+
+__all__ = ["compile"]
--- a/heir_py/BUILD
+++ b/heir_py/BUILD
@@ -6,6 +6,21 @@ package(
     default_visibility = ["//visibility:public"],
 )
 
+DATA_DEPS = [
+    "@cereal//:headers",
+    "@heir//tools:heir-opt",
+    "@heir//tools:heir-translate",
+    "@openfhe//:headers",
+    "@openfhe//:binfhe",
+    # copybara: openfhe_binfhe_headers
+    "@openfhe//:core",
+    # copybara: openfhe_core_headers
+    "@openfhe//:pke",
+    # copybara: openfhe_pke_headers
+    # copybara: python_runtime_headers
+    "@rapidjson//:headers",
+]
+
 # a single-source build dependency that gives the whole (non-test) source tree;
 # note we chose the style of putting all test rules below, because glob does
 # not recurse into subdirectories with BUILD files in them.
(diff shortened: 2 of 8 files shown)


j2kun:
Ready for review!

asraa (approved):

AlexanderViand-Intel (approved):
Thank you! I love how much this already looks like the design doc from just a few weeks ago! ❤️ 

I'm running into this error message when following the README steps, so maybe that needs updating?

```
(venv) aviand@a4bf018d43cf:~/code/heir$ python -m heir_py.example
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/home/aviand/code/heir/heir_py/__init__.py", line 1, in <module>
    from decorator import compile
ModuleNotFoundError: No module named 'decorator'
```

heir_py/__init__.py
```diff
@@ -0,0 +1,3 @@
+from decorator import compile
```
AlexanderViand-Intel, line 1:
```suggestion
from .decorator import compile
```

README instructions work for me if making the relative path explicit here

AlexanderViand-Intel:
I had a bit more time to play with this and restored + added some debugging info:
* #1188

Also, I noticed that there's not really a "nice" way to get to the default openfhe/heir configs you've defined, as the decorator will default to `from_os_env` config if the options aren't provided.  That means the only way to use the default configs seems to be this, which is a bit ugly (though I guess it's OK if we'll have a more automagic solution soon)
```python
from heir_py import heir_config
from heir_py import openfhe_config

# ....

@compile(backend="openfhe", backend_config=openfhe_config.DEFAULT_INSTALLED_OPENFHE_CONFIG, heir_config=heir_config.DEVELOPMENT_HEIR_CONFIG)
```

j2kun:
Yeah I think I ran into some issues as well with a merge conflict resolution. Let me merge this one and revisit the defaults on monday.

Issue #1145, opened 2024-12-05
Generalize OpenFHE Param/Context/KeyGen functionality to work at scheme/LWE level


AlexanderViand-Intel:
Currently, `--lwe-add-client-interface` generates encode/encrypt and decrypt/decode functions in an backend-agnostic away at the scheme/LWE level.
However, the `ConfigureCryptoContext` pass (See #661 and mentioned PRs) that adds the keygen helper is OpenFHE specific.

ZenithalHourlyRate:
After several PR it seems that LWE level configuration seems unfeasible for now. I think polynomial backend (if we have it one day) should also have its own implementation instead of a generic one.

PR #1189, opened 2024-12-14
Openfhe: get mulDepth from LWE type

ZenithalHourlyRate:
Related to #1145. This is a step towards a generic configuring crypto context pass, as the Lattigo backend also needs one.

Since we have RNS lowering for secret-to-<scheme>, we can get mulDepth from LWE type.

Still marked as draft as the CKKS RNS lowering has not landed. All tests working when merged with #1176

Several changes

* hasMulOp becomes hasRelinOp as we actually does not emit `openfhe.mul` (implicit relin)
* the `configure_crypto_context_complex.mlir` removed scf.if and used arith.add instead, as
  + we can not lower correctly through a plaintext scf.if for now
  + for oblivious if, `arith.select` also does not have a lowering


ZenithalHourlyRate:
Rebased and Marked as ready since #1176 merged.

tests/Dialect/Openfhe/Transforms/configure_crypto_context_complex.mlir
```diff
@@ -1,38 +1,25 @@
...
+  %7 = arith.muli %5, %cst : !ty
+
+  %ret = arith.addi %1, %7 : !ty
+  return %ret : !ty
 }
 
 // CHECK: @complex_func
 // CHECK: @complex_func__generate_crypto_context
-// CHECK: mulDepth = 3
+// CHECK: mulDepth = 2
```
j2kun, line 22: Just for confirmation: this is 2 now because of the inclusion of the optimal relin insertion pass (or rather, the maximum constraint within that pass that prevents going to depth 3).

j2kun (approved):

ZenithalHourlyRate:
Oh this one should wait until #1190

tests/Dialect/Openfhe/Transforms/configure_crypto_context_complex.mlir
ZenithalHourlyRate: It is due to `operation-balancer`.

Without them we get mulDepth 3 but with them (`--secretize --wrap-generic --cse --canonicalize --operation-balancer --secret-insert-mgmt-bgv --secret-distribute-generic --secret-to-bgv=poly-mod-degree=32 --openfhe-configure-crypto-context=entry-function=complex_func`) we get mulDepth 2.

ZenithalHourlyRate: I put the detailed flag in the test so we still get `mulDepth = 3`. Otherwise it is too confusing.

ZenithalHourlyRate:
Rebased and ready now.

tests/Dialect/Openfhe/Transforms/configure_crypto_context_complex.mlir
```diff
@@ -1,46 +1,20 @@
...
+  %0 = arith.muli %arg0, %arg1 : !ty
+  %1 = arith.addi %0, %arg0 : !ty
+  %2 = arith.muli %0, %cst : !ty
+
+  %4 = arith.muli %arg0, %cst : !ty
+  %5 = arith.muli %4, %arg1 : !ty
+  %6 = arith.subi %5, %4 : !ty
+  %7 = arith.muli %5, %cst : !ty
 
-  %ret = scf.if %cond -> !out_ty {
```
AlexanderViand-Intel, line 29: I know why we can't lower an oblivious `scf.if` yet (#825, another context-aware-type-conversion issue) but can you remind me what breaks for plaintext `scf.if`?

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -132,16 +132,30 @@ LogicalResult convertFunc(func::FuncOp op, int64_t mulDepth) {
   ImplicitLocOpBuilder builder =
       ImplicitLocOpBuilder::atBlockEnd(module.getLoc(), module.getBody());
 
+  // get mulDepth from function argument ciphertext type
```
AlexanderViand-Intel, line 135: Can be done as a follow-up issue/PR, but at this point it's not really a "mulDepth" anymore, more of "number of RNS limbs", so it might make sense to rename it (and also in the corresponding OpenFHE dialect op?)

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
AlexanderViand-Intel, line 1: Do you think the MulDepthAnalysis was solid/robust enough to be worth keeping around/re-introducing for other potential uses (e.g., quick-and-dirty cost modelling) or should any such things use the new Mgmt system?

AlexanderViand-Intel (approved):
👍

tests/Dialect/Openfhe/Transforms/configure_crypto_context_complex.mlir
ZenithalHourlyRate: plaintext `scf.if` persists, so it is still a big op and `secret-distribute-generic` can not divide it, hence the `secret-to-<scheme>` could not recognize such pattern.

```mlir
func.func @add(%arg0: tensor<4xi16> {secret.secret}, %arg1: tensor<4xi16> {secret.secret}, %cond : i1) -> tensor<4xi16> {
  %0 = scf.if %cond -> tensor<4xi16> {
    %1 = arith.muli %arg0, %arg1 : tensor<4xi16>
    scf.yield %1 : tensor<4xi16>
  } else {
    %2 = arith.addi %arg0, %arg1 : tensor<4xi16>
    scf.yield %2 : tensor<4xi16>
  }
  return %0 : tensor<4xi16>
}
```

You could try this with `--mlir-to-secret-arithmetic --secret-insert-mgmt-bgv --secret-distribute-generic --secret-to-bgv=poly-mod-degree=4`, and will see a `failed to legalize operation 'secret.generic'` containing `scf.if`

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
ZenithalHourlyRate: OpenFHE API used the name `SetMultiplicativeDepth` so I think the naming here is in accordance with that.

lib/Analysis/MulDepthAnalysis/MulDepthAnalysis.cpp
ZenithalHourlyRate: `LevelAnalysis` (introduced alone side mgmt dialect) is a very similar analysis, it works at the `arith.muli/mulf` level, I think that can be used for other purposes.

Issue #1166, opened 2024-12-10
Investigate if/how to integrate Saturn into HEIR


AlexanderViand-Intel:
Saturn is a introduced in ["Level Aware Bootstrapping Placement for Fully Homomorphic Encryption Using MaxSAT"](http://reports-archive.adm.cs.cmu.edu/anon/2024/CMU-CS-24-134.pdf), a MSc thesis by William Seo from CMU, supervised by Wenting Zheng and Fraser Brown. 

> On average Saturn achieves a 5.25× speedup in bootstrapping placement
selection time compared to DaCapo [7], the state-of-the-art automatic bootstrapping method.
Additionally, the benchmarks generally have faster estimated inference time for a CIFAR-10
image, when bootstrapped by Saturn versus when bootstrapped by DaCapo [7].

j2kun:
FWIW, it appears that OR-tools has a MAX-SAT solver, so this may be easy to replicate: https://github.com/google/or-tools/blob/8edc858e5cbe8902801d846899dc0de9be748b2c/ortools/sat/sat_parameters.proto#L701

Issue #289, opened 2023-11-28
Investigate what parts of DaCapo can be ported to HEIR


j2kun:
DaCapo: Automatic Bootstrapping Management for Efficient Fully Homomorphic Encryption
Seonyoung Cheon, Yongwoo Lee, Dongkwan Kim, Ju Min Lee, Sunchul Jung, Taekyung Kim, Dongyoon Lee, Hanjun Kim

https://www.usenix.org/system/files/sec24summer-prepub-336-cheon.pdf

AlexanderViand-Intel:
It looks like there's a new entry in the [HECATE](https://ieeexplore.ieee.org/document/9741265)/[ELASM](https://www.usenix.org/conference/usenixsecurity23/presentation/lee-yongwoo)/[DaCapo](https://www.usenix.org/conference/usenixsecurity24/presentation/cheon) series:

**HALO: Loop-aware Bootstrapping Management for Fully Homomorphic Encryption**
*Seonyoung Cheon, Yongwoo Lee, Hoyun Youm, Dongkwan Kim, Sungwoo Yun, Kunmo Jeong, Dongyoon Lee, and Hanjun Kim*
To Appear: Proceedings of the 30th ACM International Conference on Architectural Support for Programming Languages and Operating Systems 2025 (ASPLOS), April 2025.
https://dl.acm.org/doi/10.1145/3669940.3707275

Going by the abstract (I don't see a full-text link anywhere?), this might not be 100% applicable to all situations (e.g., packing all loop variables in a single ctxt), but there's some other ideas (e.g., unrolling the loop a bit to avoid wasting levels at the end) that sound like they'd be generally applicable.

asraa:
I took some detailed notes about the papers (request for access, since I didn't port this to my personal gmail): https://docs.google.com/document/d/1exeP95PZfKrxE36kbIBwS2dJSZcFb2HYCe5QaIPxoX0/edit?tab=t.0

But happy to help guide how to port these to HEIR

j2kun:
@asraa if you have the time, could you port your notes to a new issue that is under the general umbrella of supporting scheme operations within loops in HEIR (that would include https://github.com/google/heir/issues/1181 as a sub-task).

We could solicit help from the community more directly, and then I could do it in December if we get no takers in the mean time.

PR #1195, opened 2024-12-15
openfhe: add bootstrap op

ZenithalHourlyRate:
This follows the setup step in [simple-ckks-bootstrapping example in OpenFHE](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/examples/simple-ckks-bootstrapping.cpp).

With input

```mlir
  func.func @func(%arg0: !openfhe.crypto_context, %arg1: !rlwe_ct_L0_) -> !rlwe_ct_L0_ {
    %0 = openfhe.bootstrap %arg0, %arg1 : (!openfhe.crypto_context, !rlwe_ct_L0_) -> !rlwe_ct_L0_
    return %0 : !rlwe_ct_L0_
  }
```

with `--openfhe-configure-crypto-context` we get
```mlir
  func.func @func__generate_crypto_context() -> !openfhe.crypto_context {
    %0 = openfhe.gen_params  {mulDepth = 20 : i64, plainMod = 4295294977 : i64} : () -> !openfhe.cc_params
    %1 = openfhe.gen_context %0 {supportFHE = true} : (!openfhe.cc_params) -> !openfhe.crypto_context
    return %1 : !openfhe.crypto_context
  }
  func.func @func__configure_crypto_context(%arg0: !openfhe.crypto_context, %arg1: !openfhe.private_key) -> !openfhe.crypto_context {
    openfhe.gen_mulkey %arg0, %arg1 : (!openfhe.crypto_context, !openfhe.private_key) -> ()
    openfhe.setup_bootstrap %arg0 {levelBudgetDecode = 3 : index, levelBudgetEncode = 3 : index} : (!openfhe.crypto_context) -> ()
    openfhe.gen_bootstrap_key %arg0, %arg1 : (!openfhe.crypto_context, !openfhe.private_key) -> ()
    return %arg0 : !openfhe.crypto_context
  }
```

Then the translated code can run with a main function similar to `simple-ckks-bootstrap`.

To make it run faster, the following setup step is needed
```
  v3.SetSecurityLevel(HEStd_NotSet);
  v3.SetRingDim(1 << 12);
```

Link the generated code against an OpenFHE with multi-thread support and fast math backend (HEIR internal one is single-thread, slow math backend). It could run within 3 second (96 core machine). Link against HEIR internal one would require about 2 minutes to run.

### Discussion

* mulDepth is now interwined with bootstrap, hence all the RNS instantiation need to change. and #1176 way could not handle this. I think that would need introducing a `mgmt.bootstrap` op and implement a bootstrap management pass  (possibly related to #289, #1166).
* Currently, the `bootstrapDepth` (level consumed by bootstrap) is hard-encoded with value taken from openfhe. This depth will affect logN selection hence slot num, which is needed by the `EvalBootstrapKeyGen`. Now I acquire this time in runtime (by the emitter) from `cryptoContext` instead of putting it in `openfhe.gen_bootstrap_key` op.

### DIFF
--- a/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
@@ -56,9 +56,23 @@ SmallVector<int64_t> findAllRotIndices(func::FuncOp op) {
   return rotIndicesResult;
 }
 
+// Helper function to check if the function has BootstrapOp
+bool hasBootstrapOp(func::FuncOp op) {
+  bool result = false;
+  op.walk<WalkOrder::PreOrder>([&](Operation *op) {
+    if (isa<openfhe::BootstrapOp>(op)) {
+      result = true;
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  return result;
+}
+
 // function that generates the crypto context with proper parameters
 LogicalResult generateGenFunc(func::FuncOp op, const std::string &genFuncName,
-                              int64_t mulDepth, ImplicitLocOpBuilder &builder) {
+                              int64_t mulDepth, bool hasBootstrapOp,
+                              bool insecure, ImplicitLocOpBuilder &builder) {
   Type openfheContextType =
       openfhe::CryptoContextType::get(builder.getContext());
   SmallVector<Type> funcArgTypes;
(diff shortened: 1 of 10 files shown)


ZenithalHourlyRate:
Rebased to new lwe type. Tests will be added later.

AlexanderViand-Intel:
> Link the generated code against an OpenFHE with multi-thread support and fast math backend (HEIR internal one is single-thread, slow math backend). It could run within 3 second (96 core machine). Link against HEIR internal one would require about 2 minutes to run.

Are there any downsides to building OpenFHE with OpenMP and a fast math backend (apart from adding additional dependencies, I guess?). If no, then we should consider doing so as the internal default, especially if people indeed end up using HEIR for benchmarking.  

>  I think that would need introducing a mgmt.bootstrap op
In general, I think it's fine to add low-level operations that the high-level parts of the compiler don't know how to generate (a lot better than the reverse, at least). However, I feel like it'd be nice to also add `ckks.bootstrap` and the lowering to openfhe, so that one could at least write "hand-managed" code at the CKKS level (would probably also make for nicer tests)

ZenithalHourlyRate:
> I feel like it'd be nice to also add `ckks.bootstrap` and the lowering to openfhe,

Definitely.

Now the e2e test is added. Check `tests/Examples/openfhe/simple_ckks_bootstrapping_test.cpp`. To make it run faster, I used insecure parameters, which is enabled by `--openfhe-configure-crypto-context=entry-function=simple_ckks_bootstrapping insecure=true` and it will `SetRingDim(128)`. It now finishes in 3s.

> building OpenFHE with OpenMP and a fast math backend (apart from adding additional dependencies, I guess?

I do agree we should integrate a fastest openfhe for proper evaluation, but that involves a lot of bazel (possibly needs jeremy). 

Also, github actions CI is not super powerful so it won't benefit much from it. Instead I choose to use a smaller parameter for the testing pipeline and added a warning on the test on the insecurity for other people wanting to use it.

j2kun:
> Link the generated code against an OpenFHE with multi-thread support and fast math backend (HEIR internal one is single-thread, slow math backend). It could run within 3 second (96 core machine). Link against HEIR internal one would require about 2 minutes to run.

When I first set this up, the obstacle was the cmake indirection layer. I will file an issue.

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -56,9 +56,23 @@ SmallVector<int64_t> findAllRotIndices(func::FuncOp op) {
...
+  bool result = false;
+  op.walk<WalkOrder::PreOrder>([&](Operation *op) {
+    if (isa<openfhe::BootstrapOp>(op)) {
+      result = true;
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  return result;
+}
```
j2kun, line 70: I think I will clean these up after this PR goes in, since we have a few implementations of "walk until you find fn(op)" in the codebase.

j2kun (approved):
LGTM!!

j2kun:
There are some merge conflicts. If you don't have time to fix them I can do an internal patch tomorrow to resolve it.

ZenithalHourlyRate:
Rebased.

Issue #1193, opened 2024-12-15
Split `--<scheme>-to-lwe` into two passes


AlexanderViand-Intel:
In my `--<scheme>-to-openfhe` to `--lwe-to-openfhe` merger (PR #TBD), I'm using `--<scheme>-to-lwe` in its current form to lower only the "common" operations we already have in the `lwe` dialect (e.g., add, mul, etc.) to reduce the amount of scheme specific ops that the openFHE pipeline needs to deal with. The current form of this pass leaves many operations (e.g., `<scheme>.rotate`, `bgv.mod_switch`, etc) unmodified, which is fine for the OpenFHE pipeline.

However, we will eventually want to extend `--<scheme>-to-lwe` to declare all `<scheme>` ops illegal and lower such operations to `lwe` (necessary for lowering to `polynomial`). There's a good chance that this lowering will be directly to "building blocks", e.g., a `rotate` might lower to an explicit permutation/automorphism + keyswitch operation, rather than `lwe.rotate`. As a result, the output of this extended pass would no longer be suitable for the OpenFHE piepeline.

This issue exists as a reminder to turn `--<scheme>-to-lwe` into a pipeline with two passes: first, a `--<scheme>-common-to-lwe` pass with the current behavior that lowers only trivial/common ops, and a `--<scheme>-specific-to-lwe` pass that lowers the remaining ops, potentially "below" the library backend abstraction level.

asraa:
Nice! I wonder if that behavior can just be controlled by a difference in what the ConversionTarget declares is legal/illegal.

AlexanderViand-Intel:
> Nice! I wonder if that behavior can just be controlled by a difference in what the ConversionTarget declares is legal/illegal.

Ohh, I like the idea! That might mean this could even just be controlled by a pass flag, so I updated the issue name to be more generic.

PR #1196, opened 2024-12-15
Merge `--<bgv/ckks>-to-openfhe` into `--lwe-to-openfhe`

AlexanderViand-Intel:
This removes (most of) the redundancy in the BGV/CKKS to OpenFHE pipeline, merging `BGVToOpenfhe`, `CKKSToOpenfhe`, `LWEToOpenfhe` and `RLWEToOpenfhe`. Originally noted in #1107.

This restructures the `--<rlwe-scheme>-to-openfhe` into `--<scheme>-to-lwe` and `--lwe-to-openfhe`. Despite the naming, there's still scheme-specific ops (e.g., `bgv.mod_switch`) that get ignored by the first pass and are handled directly by the latter pass (see #1193).

There's the two main changes:
*  [OpenFHE pipeline now converts to common LWE first](https://github.com/google/heir/commit/dfa89e2636fe53281628ac50f9c1d32a686d6b16)  
   * Since this does `applyPatternsAndFoldGreedily`, I've also updated various tests so they no longer rely on dead code.
   * Note that `--bgv-to-lwe` and `--ckks-to-lwe` are currently pretty much carbon-copies of each other, but this should eventually be changed/be factored out by future work (see #1193)
* [merge --<bgv/ckks>-to-openfhe into --lwe-to-openfhe](https://github.com/google/heir/commit/93bc66027fa6d800b01a1b6ba93a8e37d1018ef7)
  * This uses mostly the same patterns as before, everything is just (imho) a lot better organized now

and a bunch of smaller improvements/prepation steps:

* The OpenFHE pipeline now correctly uses the encoding attribute in the IR, rather than the old `ckks=False/True` flag on a pattern to decide which encoding op to use. Also fixes a few CKKS tests that were using the wrong encoding attribute.
I was able to remove some encoding-related checks since we now have verifiers for the lwe encode op that take care of this.
* I've generalized the `containsLWEOrDialect<Dialect>(funcOp op)` helper template into   `containsDialects<Dialect...>(funcOp op)` (Yay for C++11 template packs)
* Moved the rewrite of `<scheme>.extract` in terms of `<scheme>.mul_plain` and `<scheme>.rotate` to `--<scheme>-to-lwe` as it doesn't need to be OpenFHE/backend specific

Not squashed yet for hopefully slightly easier reviewing. 

~~This will also probably need to be rebased/conflicts merged if @ZenithalHourlyRate related PR(s) are merged first (or vice versa): #1176~~ Done!


AlexanderViand-Intel:
Rebased after #1176 was merged (with surprisingly few conflicts - yay for HEIR's abstraction layers working 👍)

AlexanderViand-Intel:
@j2kun this is rebased and should be good to review

lib/Dialect/CKKS/Conversions/CKKSToLWE/CKKSToLWE.cpp
```diff
@@ -0,0 +1,49 @@
...
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    auto *module = getOperation();
+
+    RewritePatternSet patterns(context);
+    patterns.add<Convert<AddOp, lwe::RAddOp>, Convert<SubOp, lwe::RSubOp>,
+                 Convert<NegateOp, lwe::RNegateOp>, Convert<MulOp, lwe::RMulOp>,
+                 lwe::ConvertExtract<ExtractOp, MulPlainOp, RotateOp> >(
+        context);
+    if (failed(applyPatternsAndFoldGreedily(module, std::move(patterns)))) {
```
j2kun, line 43: use `walkAndApplyPatterns` here for better performance, since no fixedpoint iteration is required.

lib/Dialect/CKKS/Conversions/CKKSToLWE/CKKSToLWE.td
```diff
@@ -0,0 +1,20 @@
+#ifndef LIB_DIALECT_CKKS_CONVERSIONS_CKKSTOLWE_CKKSTOLWE_TD_
+#define LIB_DIALECT_CKKS_CONVERSIONS_CKKSTOLWE_CKKSTOLWE_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def CKKSToLWE : Pass<"ckks-to-lwe"> {
+  let summary = "Lower `ckks` to `lwe` dialect.";
+
+  let description = [{
+    This pass lowers the `ckks` dialect to `lwe` dialect.
```
j2kun, line 10: here and for BGV, please mention that some scheme-specific ops that have no analogue in `lwe` are not lowered, and must be handled directly by the later pass that lowers to an exit dialect.

lib/Dialect/CKKS/Conversions/CKKSToLWE/CKKSToLWE.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"     // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+
+namespace mlir::heir::ckks {
+
+#define GEN_PASS_DEF_CKKSTOLWE
+#include "lib/Dialect/CKKS/Conversions/CKKSToLWE/CKKSToLWE.h.inc"
+
+template <typename CKKSOp, typename LWEOp>
```
j2kun, line 22: This is a direct copy of the template class from BGV. Maybe put the common thing in `LWEPatterns.h` (or maybe a new RewriteUtils under lib/Utils?)

lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
```diff
@@ -22,125 +30,286 @@
...
-  } else {
-    rewriter.replaceOpWithNewOp<openfhe::MakePackedPlaintextOp>(
-        op, plaintextType, cryptoContext, input);
+    lwe::NewLWEPlaintextType plaintextType = lwe::NewLWEPlaintextType::get(
+        op.getContext(),
+        lwe::ApplicationDataAttr::get(adaptor.getInput().getType(),
+                                      lwe::NoOverflowAttr::get(getContext())),
+        lwe::PlaintextSpaceAttr::get(getContext(), op.getRing(),
+                                     op.getEncoding()));
+    if (isa<lwe::InverseCanonicalEncodingAttr>(op.getEncoding()) ||
```
j2kun, line 209: nit: this could be simplified with a type switch.

lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
```diff
@@ -22,125 +30,286 @@
...
+        ConvertBinOp<lwe::RAddOp, openfhe::AddOp>,
+        ConvertBinOp<lwe::RSubOp, openfhe::SubOp>,
+        ConvertBinOp<lwe::RMulOp, openfhe::MulNoRelinOp>,
+        ConvertUnaryOp<lwe::RNegateOp, openfhe::NegateOp>,
+
+        ///////////////////////////////////
+        // Scheme-Specific Op Patterns   //
+        ///////////////////////////////////
+
+        // AddPlain (//FIXME: Not really all that scheme-specific)
```
j2kun, line 287: Here and below, remove fixmes or resolve them.

You want an `lwe.add_plain/sub_plain/mul_plain` op?

j2kun (changes requested):

j2kun:
Looks really great! Just a few nitpicks and we're good to merge.

lib/Dialect/CKKS/Conversions/CKKSToLWE/CKKSToLWE.cpp
AlexanderViand-Intel: Oh, you're right - I thought we'd still need it because `ConvertExtract` generates ops that need other patterns applied to them, but I guess the walk also walks the freshly inserted ones!

lib/Dialect/CKKS/Conversions/CKKSToLWE/CKKSToLWE.cpp
AlexanderViand-Intel: I thought about putting it in ConversionUtils but didn't because it's not a conversion pattern, so I like the idea of RewriteUtils!

lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
AlexanderViand-Intel: Happy to change it if you want, but I don't think there's a nice way to handle the OR in that case? (I.e.., I don't think llvm::TypeSwitch supports fall-through?)

lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
AlexanderViand-Intel: I think eventually it'd be nicer to just support ptxt types on the existing lwe ops, but I'll leave that as a TODO for #1193

lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
j2kun: TypeSwitch supports variadic template arguments in the match which function as a logical OR, see for example https://github.com/llvm/llvm-project/blob/a957cedea9657addbe8b860852cc98306aa437e7/mlir/lib/Dialect/LLVMIR/IR/LLVMTypes.cpp#L894-L898

The caveat is that you have to use `auto` for the callback's argument type and the template expansion has to compile. (or use a more generic type that encapsulates all options). I don't think that should be a problem for this code.

AlexanderViand-Intel: Ohh, cool! Thanks for showing me the example "in the wild" - I actually briefly wondered about whether it does (because I knew isa<> does variadiac) but couldn't find any info on it in the docs (probably should have just tried but it was late...)

I'll add this and then squash/rebase :)

AlexanderViand-Intel:
> Looks really great! Just a few nitpicks and we're good to merge.

Should all be addressed :)

j2kun (approved):

Issue #1103, opened 2024-11-20
Docs for some dialects include upstream attributes


j2kun:
The website docs for some dialects include generated attributes from upstream, e.g., polynomial dialect currently shows

```
AffineMapAttr
ArrayAttr
DenseArrayAttr
DenseIntOrFPElementsAttr
DenseResourceElementsAttr
DenseStringElementsAttr
DictionaryAttr
FloatAttr
IntegerAttr
IntegerSetAttr
OpaqueAttr
SparseElementsAttr
StringAttr
SymbolRefAttr
TypeAttr
UnitAttr
FloatPolynomialAttr
IntPolynomialAttr
PrimitiveRootAttr
RingAttr
TypedFloatPolynomialAttr
TypedIntPolynomialAttr
StridedLayoutAttr
```

When it should just show

```
FloatPolynomialAttr
IntPolynomialAttr
PrimitiveRootAttr
RingAttr
TypedFloatPolynomialAttr
TypedIntPolynomialAttr
```

Same for the mod-arith dialect. I think this is just a tablegen separation issue.

PR #1204, opened 2024-12-16
Attributes: stop including BuiltinAttributes

ZenithalHourlyRate:
Check https://heir.dev/docs/dialects/polynomial/ and other pages like `Mgmt` or `ModArith`. They also includes builtin attributes from MLIR.

The reason is that their `Attributes.td` includes `BuiltinAttributes.td` from MLIR. If we switch to `OpBase.td` then it is fine.

Including `BuiltinAttributes` for `Ops.td` is fine as the build process will filter out attribute definitions.


AlexanderViand-Intel (approved):
❤️

j2kun:
Fixes https://github.com/google/heir/issues/1103

j2kun (approved):

PR #1214, opened 2024-12-18
Mgmt: add annotate-mgmt pass

ZenithalHourlyRate:
Currently `secret-insert-mgmt-bgv/ckks` and `optimize-relinearization` all need to update the `MgmtAttr` when they changed the IR. Further passes operating at secret/mgmt level (e.g. bootstrap placement) may also need this functionality.

This common pattern could be extracted into a single pass to reduce redundancy.


lib/Dialect/Mgmt/Transforms/BUILD
```diff
@@ -0,0 +1,56 @@
...
+            "MgmtPasses.md",
+        ),
+    ],
+    tblgen = "@llvm-project//mlir:mlir-tblgen",
+    td_file = "Passes.td",
+    deps = [
+        "@llvm-project//mlir:OpBaseTdFiles",
+        "@llvm-project//mlir:PassBaseTdFiles",
+    ],
+)
```
j2kun, line 56:
```suggestion
load("@heir//lib/Transforms:transforms.bzl", "add_heir_transforms")

package(
    default_applicable_licenses = ["@heir//:license"],
    default_visibility = ["//visibility:public"],
)

cc_library(
    name = "Transforms",
    hdrs = ["Passes.h"],
    deps = [
        ":AnnotateMgmt",
        ":pass_inc_gen",
        "@heir//lib/Dialect/Mgmt/IR:Dialect",
        "@llvm-project//mlir:IR",
    ],
)

cc_library(
    name = "AnnotateMgmt",
    srcs = ["AnnotateMgmt.cpp"],
    hdrs = ["AnnotateMgmt.h"],
    deps = [
        ":pass_inc_gen",
        "@heir//lib/Analysis/DimensionAnalysis",
        "@heir//lib/Analysis/LevelAnalysis",
        "@heir//lib/Analysis/SecretnessAnalysis",
        "@heir//lib/Dialect/Mgmt/IR:Dialect",
        "@llvm-project//mlir:IR",
        "@llvm-project//mlir:Pass",
        "@llvm-project//mlir:Transforms",
    ],
)

add_heir_transforms(
    generated_target_name = "pass_inc_gen",
    pass_name = "Mgmt",
    td_file = "Passes.td",
)
```

In an effort to simplify some of the bazel stuff, we added this add_heir_transforms macro. I think this usage will work as a drop-in replacement. (Also, did you use the template generator to generate this? If so, I should update that to use the macro).

tests/Dialect/Secret/Conversions/secret_to_ckks/type_converter.mlir
```diff
@@ -1,7 +1,7 @@
 // Tests that the type converter handles both converting tensors to SIMD slots
 // if aligned, or to tensors of ciphertext.
 
-// RUN: heir-opt --mlir-print-local-scope --secret-insert-mgmt-ckks=include-first-mul=false --secret-distribute-generic --canonicalize --secret-to-ckks --split-input-file %s | FileCheck %s
+// RUN: heir-opt --mlir-print-local-scope --secret-insert-mgmt-ckks=include-first-mul=false --annotate-mgmt --secret-distribute-generic --canonicalize --secret-to-ckks --split-input-file %s | FileCheck %s
```
j2kun, line 4: This is OK, but if you want the annotations to be an invariant of the passes that require them, you could run them as a sub-pipeline (similar to what you did running CSE at the end of SecretInsertMgmtBGV) either before a pass that requires them, or at the end of a pass that invalidates them. Then there is no risk of someone forgetting to insert the annotate-mgmt pass.

j2kun (approved):

lib/Dialect/Mgmt/Transforms/BUILD
ZenithalHourlyRate: > Also, did you use the template generator to generate this? If so, I should update that to use the macro

I used the template generator. That should be updated.

j2kun (approved):

PR #1220, opened 2024-12-21
Cleanup deprecated applyPatternsAndFoldGreedily

ZenithalHourlyRate:
Check https://github.com/llvm/llvm-project/pull/104649

### DIFF
--- a/lib/Dialect/CGGI/Transforms/ExpandLUT.cpp
+++ b/lib/Dialect/CGGI/Transforms/ExpandLUT.cpp
@@ -68,7 +68,9 @@ struct ExpandLUT : impl::ExpandLUTBase<ExpandLUT> {
     alignment::populateWithGenerated(patterns);
     patterns.add<ExpandLutLinComb>(context);
 
-    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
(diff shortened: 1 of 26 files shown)


lib/Dialect/CGGI/Transforms/ExpandLUT.cpp
```diff
@@ -68,7 +68,7 @@ struct ExpandLUT : impl::ExpandLUTBase<ExpandLUT> {
     alignment::populateWithGenerated(patterns);
     patterns.add<ExpandLutLinComb>(context);
 
-    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
+    (void)applyPatternsGreedily(getOperation(), std::move(patterns));
```
AlexanderViand-Intel, line 71: something like this:
```suggestion
    // TODO (#1221): Investigate whether folding (default: on) can be skipped here.
    (void)applyPatternsGreedily(getOperation(), std::move(patterns));
```

AlexanderViand-Intel (changes requested):
Can you do me a favor and add a TODO for #1221 on each change?

AlexanderViand-Intel (approved):
Thanks! With the holidays, I'm not sure when this will be merged, though

j2kun (approved):

PR #1223, opened 2024-12-21
doc: add graphviz output for dot-product example

ZenithalHourlyRate:
I was hanging around in the MLIR codebase then found the pass `--view-op-graph`. Thinking that it might be quite helpful in debugging/demonstrating, I intend to add it to the doc so users can get more benefit from it.

When the IR becomes complex (tree structure or complex graph structure), reading the linear SSA IR becomes hard.

As for the complex bazel commandline, check https://github.com/bazelbuild/bazel/issues/4867, where until bazel 8.0 does it support `bazel --quiet`


AlexanderViand-Intel (approved):
Nice write-up, thanks! Not sure it necessarily neeeds to be in "getting started", but that's already a bit of a catch-all page anyway.

docs/content/en/docs/getting_started.md
```diff
@@ -432,3 +432,22 @@ bazel run --noallow_analysis_cache_discard //tools:heir-opt -- \
...
+
+### Optional: Graphviz visualization of the IR
+
+Getting a visualization of the IR during optimization/transformation might help
+you understand what is going on more easily.
+
+Still taking the `dot_product_8.mlir` as an example:
+
+```bash
+bazel run --ui_event_filters=-info,-debug,-warning,-stderr,-stdout --noshow_progress --logging=0 //tools:heir-opt -- --wrap-generic --heir-simd-vectorizer $PWD/tests/Examples/openfhe/dot_product_8.mlir --view-op-graph 2> dot_product_8.dot
```
j2kun, line 444: (review from my phone since I'm not working 😁) To make this simpler, I'm guessing you can use `-o` to write the output to a file, which would remove the need for all the bazel quieting commands.

j2kun (approved):

docs/content/en/docs/getting_started.md
ZenithalHourlyRate: There was no `-o`-like option for `bazel run`. There was a `--output-file` for `bazel query` but that only redirects the stdout instead of stderr.

Links
 * https://bazel.build/reference/command-line-reference
 * https://github.com/bazelbuild/bazel/issues/5511

Issue #481, opened 2024-03-05
Simplify secret.generic syntax


j2kun:
It's currently

```mlir
secret.generic ins(%arg0, %arg1 : secret<ty0>, secret<ty1>) {
^bb0(%arg0: ty0, %arg1: ty1):
  ...ops...
}
```

The discrepancy between how the `ins` part is parsed/printed vs the basic block is just confusing. We should probably change it to something like this

```mlir
secret.generic (%arg0: secret<ty0>, %arg1 : secret<ty1>) {
^bb0(%arg0: ty0, %arg1: ty1):
  ...ops...
}
```

ZenithalHourlyRate:
Fixed by #1793

PR #1224, opened 2024-12-21
ASM: suggest block arg name for secret.generic

ZenithalHourlyRate:
Similar to #1219 but does not depend on that.

In the past, we have seen secret.generic op like this
```mlir
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<tensor<8xi16>>, !secret.secret<tensor<8xi16>>) {
    ^bb0(%arg2: tensor<8xi16>, %arg3: tensor<8xi16>):
      %1 = arith.muli %arg2, %arg3 : tensor<8xi16>
      %2 = tensor_ext.rotate %1, %c4 : tensor<8xi16>, index
      %3 = arith.addi %1, %2 : tensor<8xi16>
      %4 = tensor_ext.rotate %3, %c2 : tensor<8xi16>, index
      %5 = arith.addi %3, %4 : tensor<8xi16>
      %6 = tensor_ext.rotate %5, %c1 : tensor<8xi16>, index
      %7 = arith.addi %5, %6 : tensor<8xi16>
      %extracted = tensor.extract %7[%c7] : tensor<8xi16>
      secret.yield %extracted : i16
    } -> !secret.secret<i16>
```
where the input and block argument are both named `argX`, which will cause confusion.

This PR intends to distinguish them by setting block argument to `inputX`.

By `OpAsmOpInterface`, there could be more readablity improvement, like change `^bb0` to `^body` and `secret.yield` to `yield`, yet there are ~30 tests failing (heavily depends on the `argX`/`bb0` behavior).

The full version is

```mlir
    %computed = secret.generic ins(%arg0, %arg1 : !secret.secret<tensor<8xi16>>, !secret.secret<tensor<8xi16>>) {
    ^body(%input0: tensor<8xi16>, %input1: tensor<8xi16>):
      %0 = arith.muli %input0, %input1 : tensor<8xi16>
      %1 = tensor_ext.rotate %0, %c4 : tensor<8xi16>, index
      %2 = arith.addi %0, %1 : tensor<8xi16>
      %3 = tensor_ext.rotate %2, %c2 : tensor<8xi16>, index
      %4 = arith.addi %2, %3 : tensor<8xi16>
      %5 = tensor_ext.rotate %4, %c1 : tensor<8xi16>, index
      %6 = arith.addi %4, %5 : tensor<8xi16>
      %extracted = tensor.extract %6[%c7] : tensor<8xi16>
      yield %extracted : i16
    } -> !secret.secret<i16>
```

I was also thinking about simplifying the asm format like the below (as we have setArgAttr now), but does not know if such direction is favorable (definitely many changes to the test). If it is acceptable, I'll add an issue on this and work on it.

```mlir
secret.generic (%arg0 -> input0 : tensor<8xi16> {attr}, %arg1 -> input1 : tensor<8xi16> {attr}) attrs = {} {
  %0 = arith...
} -> !secret.secret<i16>
```

### DIFF
--- a/lib/Dialect/Secret/IR/SecretOps.td
+++ b/lib/Dialect/Secret/IR/SecretOps.td
@@ -4,6 +4,7 @@
 include "SecretDialect.td"
 include "SecretTypes.td"
 include "mlir/IR/BuiltinAttributeInterfaces.td"
+include "mlir/IR/OpAsmInterface.td"
 include "mlir/Interfaces/ControlFlowInterfaces.td"
 include "mlir/Interfaces/InferTypeOpInterface.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
@@ -110,7 +111,8 @@ def Secret_YieldOp : Secret_Op<"yield", [
 
 def Secret_GenericOp : Secret_Op<"generic", [
   SingleBlock,
-  SingleBlockImplicitTerminator<"YieldOp">
+  SingleBlockImplicitTerminator<"YieldOp">,
+  OpAsmOpInterface
 ]> {
   let summary = "Lift a plaintext computation to operate on secrets.";
   let description = [{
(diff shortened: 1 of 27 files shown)


j2kun:
I had an old issue to do a marginal improvement to the syntax: https://github.com/google/heir/issues/481

Part of the difficulty with your suggested syntax is that the operand for `generic` need not be secret (though it usually is). So the syntax actually needs to have both types. IMO the less custom we can make the syntax, the simpler our lives will be when migrating from upstream changes.

The other suggested changes are fine with me.

j2kun (approved):

ZenithalHourlyRate:
Closed by https://github.com/google/heir/commit/5a4742601da66868bf3fad0cd700aaff61693fb9

PR #1225, opened 2024-12-21
SelectVariableNames: use suggested name from dialect

ZenithalHourlyRate:
This is inspired by #1219 and shares some code with #1219. This will generate more friendly backend code, for example, openfhe-pke emitter output becomes

```cpp
CiphertextT dot_product(CryptoContextT cc, CiphertextT ct, CiphertextT ct1) {
  std::vector<int64_t> v = {0, 0, 0, 0, 0, 0, 0, 1};
  const auto& ct2 = cc->EvalMultNoRelin(ct, ct1);
  const auto& ct3 = cc->Relinearize(ct2);
  const auto& ct4 = cc->EvalRotate(ct3, 4);
  ...
}

CryptoContextT dot_product__generate_crypto_context() {
  CCParamsT params;
  params.SetMultiplicativeDepth(2);
  params.SetPlaintextModulus(4295294977);
  CryptoContextT cc = GenCryptoContext(params);
  cc->Enable(PKE);
  cc->Enable(KEYSWITCH);
  cc->Enable(LEVELEDSHE);
  return cc;
}
```

And all `@heir//tests/Examples/openfhe/...` pass with this change. However, other backend like Jaxite/TFHE-RS are relying on `getIntForValue` to properly function. I am not so familiar with them so mark as draft now.


ZenithalHourlyRate:
> However, other backend like Jaxite/TFHE-RS are relying on `getIntForValue` to properly function. I am not so familiar with them so mark as draft now.

Tweaked the `selectVariableName` process for these backend tests regarding the default prefix `"v"`.
Namely these tests assume `v0` for the first appearance then `v1`/`v2`, but for simplicity like for the openfhe backend, the `crypto_context` only appears once in one function so I intend to use `cc` instead of `cc0` for it.

Ready for review now.

j2kun (approved):

PR #1226, opened 2024-12-22
bgv-to-lattigo: lower client-interface and plain op

ZenithalHourlyRate:
With this PR we can lower the dot_product_8 example with client interface. The emitter/tests has not been added as there are hacks in this pass that needs discussion.

`lwe-add-client-interface` will add helper function for encryption/decryption, and the function signature is `__encrypt(value, sk/pk)`. However, this is only for openfhe backend, as openfhe needs `CryptoContext` and `sk/pk` to encrypt/decrypt, and openfhe lowering will add its own CryptoContext arg.

For lattigo, things are different in that we have `evaluator/encryptor/decryptor/encoder/decoder` instead of one context, and all of them should be passed on-demand.

This brings the complicacy that a lot of predicating is needed to make sure things are added _in correct order_, otherwise the function signature is not stable.

Currently I add evaluator in a specified order with predicate, so we can have more predictable function signature.

With `--mlir-to-secret-arithmetic --secret-insert-mgmt-bgv --secret-distribute-generic --secret-to-bgv="poly-mod-degree=8" --lwe-add-client-interface --bgv-to-lwe --bgv-to-lattigo` we can get the following IR (of course with some ASM trick locally #1219)

```mlir
  func.func @dot_product(%evaluator: !evaluator, %params: !params, %encoder: !encoder, %ct: !ct, %ct_0: !ct) -> !ct
  
  func.func @dot_product__encrypt(%evaluator: !evaluator, %params: !params, %encoder: !encoder, %encryptor: !encryptor, %arg0: tensor<8xi16>, %arg1: tensor<8xi16>, %skey: !skey) -> (!ct, !ct) {
    %pt = lattigo.bgv.new_plaintext %params : (!params) -> !pt
    %pt_0 = lattigo.bgv.encode %encoder, %arg0, %pt : (!encoder, tensor<8xi16>, !pt) -> !pt
    %ct = lattigo.rlwe.encrypt %encryptor, %pt_0 : (!encryptor, !pt) -> !ct
    %pt_1 = lattigo.bgv.new_plaintext %params : (!params) -> !pt
    %pt_2 = lattigo.bgv.encode %encoder, %arg1, %pt_1 : (!encoder, tensor<8xi16>, !pt) -> !pt
    %ct_3 = lattigo.rlwe.encrypt %encryptor, %pt_2 : (!encryptor, !pt) -> !ct
    return %ct, %ct_3 : !ct, !ct
  }

    func.func @dot_product__decrypt(%evaluator: !evaluator, %params: !params, %encoder: !encoder, %decryptor: !decryptor, %ct: !ct, %skey: !skey) -> i16 {
    %pt = lattigo.rlwe.decrypt %decryptor, %ct : (!decryptor, !ct) -> !pt
    %cst = arith.constant dense<0> : tensor<1xi16>
    %0 = lattigo.bgv.decode %encoder, %pt, %cst : (!encoder, !pt, tensor<1xi16>) -> tensor<1xi16>
    %extracted = tensor.extract %0[%c0] : tensor<1xi16>
    return %extracted : i16
  }

```

### Discussion
* in `__encrypt/decrypt` the `skey` is not needed. This is produced by `lwe-add-client-interface` but not used by furthur backend pass, so should we either not produce it or detele it here. (involves some ordering/validity check of function signature update)
* This should be re-organized in similar style to #1196 as a `--lwe-to-lattigo`.
* `lwe.reinterpret_underlying_type` should be moved earlier (packing pipeline). Backend should not care about them (only care about ct types but not pt types). It used to work because we forward lwe type to openfhe dialect, but not this way for lattigo dialect. Similarly `bgv.extract` should be moved earlier, which I used `bgv-to-lwe` to handle it (#1174).

### DIFF
--- a/lib/Dialect/Lattigo/IR/LattigoBGVOps.td
+++ b/lib/Dialect/Lattigo/IR/LattigoBGVOps.td
@@ -97,7 +97,7 @@ class Lattigo_BGVBinaryOp<string mnemonic> :
   let arguments = (ins
     Lattigo_BGVEvaluator:$evaluator,
     Lattigo_RLWECiphertext:$lhs,
-    Lattigo_RLWECiphertext:$rhs
+    Lattigo_RLWECiphertextOrPlaintext:$rhs
   );
   let results = (outs Lattigo_RLWECiphertext:$output);
 }
--- a/lib/Dialect/BGV/Conversions/BGVToLattigo/BGVToLattigo.cpp
+++ b/lib/Dialect/BGV/Conversions/BGVToLattigo/BGVToLattigo.cpp
@@ -21,6 +21,7 @@
 #include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
 #include "mlir/include/mlir/Support/LogicalResult.h"     // from @llvm-project
 #include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h"  // from @llvm-project
 
 namespace mlir::heir::bgv {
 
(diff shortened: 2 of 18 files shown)


ZenithalHourlyRate:
Ready for review now. This PR has been big enough.

Added `mlir-to-lattigo-bgv` pipeline, and we are able to e2e test dot_product_8 example for lattigo.

> in __encrypt/decrypt the skey is not needed. This is produced by lwe-add-client-interface but not used by furthur backend pass, so should we either not produce it or detele it here. (involves some ordering/validity check of function signature update)

Cleaned it up in conversion.

> This should be re-organized in similar style to https://github.com/google/heir/pull/1196 as a --lwe-to-lattigo.

Will leave it for future PR.

lib/Dialect/Lattigo/IR/LattigoBGVOps.td
```diff
@@ -97,7 +97,7 @@ class Lattigo_BGVBinaryOp<string mnemonic> :
   let arguments = (ins
     Lattigo_BGVEvaluator:$evaluator,
     Lattigo_RLWECiphertext:$lhs,
-    Lattigo_RLWECiphertext:$rhs
+    AnyType:$rhs
```
j2kun, line 100: Could we make this slightly more constrained than AnyType? I think this was changed because it's a union of ciphertext or plaintext type?

j2kun (approved):

ZenithalHourlyRate:
Rebased as #1257 merged. 

> Could we make this slightly more constrained than AnyType? I think this was changed because it's a union of ciphertext or plaintext type?

Added a `Lattigo_RLWECiphertextOrPlaintext` constraint.

j2kun (approved):

PR #1227, opened 2024-12-22
Fix default option for elementwise-to-affine

j2kun:
Cf. https://github.com/google/heir/commit/ccba176a7f2b29f226f9e032ec56fd86ba8fee64#diff-6f8c01b4d228f97d2a800c5b5deca07240dd1203a0063e775b6a9f8e49cc9b9c for the change that introduced this, due to a failure caused by upstream changes.

https://github.com/llvm/llvm-project/pull/118877#issuecomment-2552273446 for the suggested correct form.


j2kun:
This is still not fixed, but it's working as-is. After looking into this I think the right thing to do is actually not try to set the options at all (since unset == empty string for a string option). So I think this will probably involve just removing some tests?

Issue #1207, opened 2024-12-17
Support bootstrap in Mgmt/CKKS dialect and correctly determine MgmtAttr / RNS level


ZenithalHourlyRate:
Mainly from #1195, where a low-level `openfhe.bootstrap` is added but higher level of the compiler is not aware of how to manage it.

* mulDepth is now interwined with bootstrap, hence all the RNS instantiation need to change. and #1176 way could not handle this. Should introduce a `mgmt.bootstrap` op and implement a bootstrap management pass  (possibly related to #289, #1166).
  + I think the naive one is to greedily insert it when all level consumed, to get the pipeline running, then we will take time to optimize it instead of starting from complex solutions.

* ckks does not have a bootstrap op for now.

j2kun:
> I think the naive one is to greedily insert it when all level consumed, to get the pipeline running, then we will take time to optimize it instead of starting from complex solutions.

+1, get something working first.

ZenithalHourlyRate:
For recording purpose: The computation of ckks bootstrap depth highly depends on the internal implementation. For OPENFHE, we recorded the following value

```
FHECKKSRNS::GetBootstrapDepth: UNIFORM_TERNARY with levelBudget {1, 1}: 16
FHECKKSRNS::GetBootstrapDepth: SPARSE_TERNARY with levelBudget {1, 1}: 12
FHECKKSRNS::GetBootstrapDepth: SPARSE_ENCAPSULATED with levelBudget {1, 1}: 12
```

Which means `EvalSine` itself consumes 14 or 10 levels while linear transform like C2S and S2C each takes one level. Note that in ordinary bootstrapping we usually do not select level budget like this! Openfhe suggests a selection of `{4, 4}`, which means the total level consumed can be 22 and 18 levels.

We may just assume an usual parameter, using `SPARSE_TERNARY` key, using level budget of `{4, 4}` and consumes 18 levels. The level remaining is 10 so the bootstrap waterline is 28. See https://github.com/openfheorg/openfhe-development/blob/aa391988d354d4360f390f223a90e0d1b98839d7/src/pke/examples/simple-ckks-bootstrapping.cpp#L100C42-L100C46

PR #1228, opened 2024-12-22
Implement greedy bootstrap op insertion for mgmt/ckks

ZenithalHourlyRate:
Refer to #1207.

A greedy policy is used where when all levels are consumed then a bootstrap is inserted.

Example:

```mlir
func.func @func(
    %x : f16
  ) -> f16 {
    %0 = arith.addf %x, %x : f16
    %r0 = mgmt.modreduce %0 : f16
    %1 = arith.addf %r0, %r0 : f16
    %r1 = mgmt.modreduce %1 : f16
    %2 = arith.addf %r1, %r1 : f16
    %r2 = mgmt.modreduce %2 : f16
    %3 = arith.addf %r2, %r2 : f16
    %r3 = mgmt.modreduce %3 : f16
    %4 = arith.addf %r3, %r3 : f16
    %r4 = mgmt.modreduce %4 : f16
    %5 = arith.addf %r4, %r4 : f16
    // cross level op
    %mixed0 = arith.addf %5, %x : f16
  return %mixed0 : f16
}
```

With `--secret-insert-mgmt-ckks="bootstrap-waterline=3"` we get

```mlir
    %0 = secret.generic ins(%arg0 : !secret.secret<f16>) attrs = {arg0 = {mgmt.mgmt = #mgmt.mgmt<level = 3>}} {
    ^bb0(%arg1: f16):
      %1 = arith.addf %arg1, %arg1 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : f16
      %2 = mgmt.modreduce %1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : f16
      %3 = arith.addf %2, %2 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : f16
      %4 = mgmt.modreduce %3 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : f16
      %5 = arith.addf %4, %4 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : f16
      %6 = mgmt.modreduce %5 {mgmt.mgmt = #mgmt.mgmt<level = 0>} : f16
      %7 = mgmt.bootstrap %6 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : f16
      %8 = arith.addf %7, %7 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : f16
      %9 = mgmt.modreduce %8 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : f16
      %10 = arith.addf %9, %9 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : f16
      %11 = mgmt.modreduce %10 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : f16
      %12 = arith.addf %11, %11 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : f16
      // cross level handling
      %13 = mgmt.modreduce %arg1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : f16
      %14 = mgmt.modreduce %13 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : f16
      %15 = arith.addf %12, %14 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : f16
      secret.yield %15 : f16
    } -> !secret.secret<f16>

```

### Discussion

* greedy policy may get undeterministic output when traversing order is different. Currently using walkAndApplyPattern, it is deterministic, but when use other driver things can be differnt.
* bootstrapDepth is not modeled yet. It it closely related to parameter generation. Check #1195 for detailed break down of bootstrapDepth
* The monotonicity of Analysis is annoying and not manageable when multiple pattern needs to re-run the analysis and their monotonicity requirement is different. Analysis is supposed to be run once then read-only, otherwise the propagation property / monotonicity property is hard to maintain. For example, bootstrap will break LevelAnalysis monotonicity as the `LevelLattice` is supposed to be monotonically increasing, so we need to use a new `DataflowSolver` to get uninitialized Lattice. Note that reuse the old solver means reuse old lattices. It concludes that when IR partially change, old lattice should all be invalidated, yet I did not find the API for it. Now I think the treatment in #1154 is incorrect, because analysis result might be incorrect due to monotonicity.


ZenithalHourlyRate:
> * The monotonicity of Analysis is annoying and not manageable when multiple pattern needs to re-run the analysis and their monotonicity requirement is different.

Fixed by using `solver.eraseAllStates` (introduced in https://github.com/llvm/llvm-project/pull/120881) to allow re-run the solver.

j2kun (approved):

PR #1236, opened 2024-12-30
add_heir_dialect_library build macro for bazel

ZenithalHourlyRate:
Related to #1045, similar to #1055 and see also #1235 where the verbose `gentbl_cc_library` is error-prone.

This PR demos the usability of such macro for LWE/Openfhe dialect. If such change is favored, further PR will update other BUILD file and the template generator.


j2kun (approved):
LGTM!

j2kun:
So I am patching this internally to fix the conflicts, and I will merge it, but I have a request to make for any followups to this: the generated target name and the file being read should always be provided as input to the macro. This will require more typing, but will be more readable, otherwise when looking at a target with a dependency like `dialect_inc_gen`, you will not easily be able to tell where it came from (it fails [the grep test](https://jamie-wong.com/2013/07/12/grep-test/)), and if you have a filename you will not easily be able to see which target depends on it.

Filed https://github.com/google/heir/issues/1242 to remind us

PR #1248, opened 2025-01-08
arith-to-cggi conversion pass

WoutLegiest:
First conversion pass to lower the (TOSA orginated) Arith dialect to CGGI

### DIFF
--- a/lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
+++ b/lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
@@ -1,0 +1,221 @@
+#include "lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.h"
+
+#include "lib/Dialect/CGGI/IR/CGGIDialect.h"
+#include "lib/Dialect/CGGI/IR/CGGIOps.h"
+#include "lib/Dialect/LWE/IR/LWEOps.h"
+#include "lib/Dialect/LWE/IR/LWETypes.h"
+#include "lib/Utils/ConversionUtils.h"
+#include "llvm/include/llvm/Support/Debug.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Affine/IR/AffineOps.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/MemRef/IR/MemRef.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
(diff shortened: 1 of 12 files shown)


lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+static Type convertArithLikeToCGGIType(ShapedType type, MLIRContext *ctx) {
+  if (auto arithType = llvm::dyn_cast<IntegerType>(type.getElementType())) {
+    return type.cloneWith(type.getShape(),
+                          convertArithToCGGIType(arithType, ctx));
+  }
+  return type;
+}
+
+// Remove this class if no type conversions are necessary
```
j2kun, line 36: Remove stray comment

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -0,0 +1,232 @@
...
+      return convertArithToCGGIType(type, ctx);
+    });
+
+    addConversion([ctx](ShapedType type) -> Type {
+      return convertArithLikeToCGGIType(type, ctx);
+    });
+  }
+};
+
+class SecretTypeConverter : public TypeConverter {
```
j2kun, line 53: This looks like it was leftover code, doesn't appear to do anything. Delete?

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -0,0 +1,232 @@
...
+    ArithToCGGITypeConverter typeConverter(context);
+
+    RewritePatternSet patterns(context);
+    ConversionTarget target(*context);
+    target.addLegalDialect<cggi::CGGIDialect>();
+    target.addIllegalDialect<mlir::arith::ArithDialect>();
+
+    target.addDynamicallyLegalOp<mlir::arith::ConstantOp>(
+        [](mlir::arith::ConstantOp op) {
+          // Allow use of constant if it is used to denote the size of a shift
```
j2kun, line 197: (optional): This is not too bad, but one way we could avoid it is by making the CGGI shift op take an attribute as its shift instead of a value. Is there any time when the shift amount is not statically known?

j2kun (approved):
LGTM! (Reviewed from my phone)

j2kun (approved):

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
WoutLegiest: Damn, that sneaked back in during squash

Issue #1252, opened 2025-01-09
Decide on (Cleartext) Integer Semantics for HEIR's Python Frontend


AlexanderViand-Intel:
Python famously has arbitrary precision integers, though a lot of the time "interesting" programs use `np.int64`  or something similar anyway. This raises the question how a basic `a + b` statement between two integers should be typed.

This is of course assuming we know the type of `a` and `b`, either from an annotation (ahead of time compilation) or because we have their actual runtime values (jit compilation). 

There are, afaik, basically three different approaches we can take:

1. "Pythonic Ideal": Track the bitwidth required to represent the result perfectly: 16-bit + 16-bit is 17-bit (or 32-bit, if rounded up to next power of two), 16-bit * 16-bit is 32-bit, etc. 
2. "Numba NBEP1": Numba at some point made the decision to basically upcast smaller types to the machine type, so `int16 + int16` is actually `int64` (at least on a 64-bit machine).  See https://numba.readthedocs.io/en/stable/proposals/integer-typing.html. Note that Numba does _not_ do this for arrays, so `array(int32) + array(int32)` is still `array(int32)`.
3. "MLIR  /  overflow=none": This is how we currently use the `arith` dialect. Adding/Multiplying two values of type `i32` still results in an `i32`. Overflow is essentially considered "Undefined Behavior" and the compiler simply says "not my problem".

In discussions about this so far, we've always gone with Option 3, as it's by far the easiest to deal with for arithmetic FHE, where we need to impose a fixed plaintext modulus. While Option 1 has some appeal, I think the costs of trying to handle this far outweigh the benefits. Finally, while I understand why Numba chose to "snap to pointer size", I don't think adapting this makes sense for us.

If there's consensus on this (or at least no active outcries against), I propose we go for option 3. However, that poses a bit of an issue with just using Numba's Type Inference out of the box:

<details>
<summary><b>Python code to see Numba Type Inference in action</b></summary>

```python
from numba.core.registry import cpu_target
from numba.core import compiler, sigutils
from numba.core.typed_passes import type_inference_stage

# Define a test function
def example_function(x, y):
    z = x + y
    return z

sig_string = "int16(int16, int16)"

test_ir = compiler.run_frontend(example_function)
typingctx = cpu_target.typing_context
targetctx = cpu_target.target_context
typingctx.refresh()
targetctx.refresh()

fn_args, fn_retty = sigutils.normalize_signature(sig_string)
[...]

PR #1253, opened 2025-01-09
Python Frontend: Use Numba Type Inference

AlexanderViand-Intel:
This does a few things:
* Enable Numba type inference and uses the inferred types to decide the MLIR type that gets emitted
* Add an `@mlir` decorator that is a slightly modified version of numba's `@intrinsic`, with the main difference being that one can provide a pure python implementation and run it under python. This allows us to define, e.g., `linalg.matmul`  and assign it `np.matmul` as its python implementation.
* Changes a few defaults/adds a few options to the pipeline here and there. Not really related to type inference, might pull out into another PR
* Adds dummy classes for MLR types (e.g., `Secret[...]`), in preparation for grabbing signature types from type annotations (rathter than from a string passed into the decorator, as is Numba's approach for ahead-of-time compilation)
* TODO: lots of cleanup, but most importantly we need to define MLIR-style integers and Tensor types (see also #1252)

PS: fun fact - in numba/numpy world, `i4` is not a 4-bit integer (as in MLIR), but shorthand for the 4-*byte* type `int32`.


AlexanderViand-Intel:
Sorry for the lack of clean-up/updates on this one, I was feeling a bit under the weather. 

The big ToDo, as discussed last week, to move from Numba-style inference to something that supports statically shaped tensors is still open, but I did do some other smaller cleanup

* You can now select between schemes (only, `bgv` and `ckks` so far, but it'd be interesting to see how much effort it would be to support the boolean pipeline (at least to the heir-translate output)).
* The emitter now supports floats (i.e., emits arith.addf instead of arith.addi)
* You can now use Python type annotations to specify types (including ranked tensors, though the sizes will be converted to "?")
* The frontend now respects `Secret[...]` annotations and no longer relies on `--secretize`

Here's a current example:
```python
from heir_py import compile
from heir_py.mlir import *
from heir_py.types import *

@compile() # defaults to scheme="bgv"", backend="openfhe", debug=False
def foo(x : Secret[I16], y : Secret[I16]):
  sum = x + y
  diff = x - y
  mul = x * y
  expression = sum * diff + mul
  return expression

foo.setup() # runs keygen/etc
enc_x = foo.encrypt_x(7)
enc_y = foo.encrypt_y(8)
result_enc = foo.eval(enc_x, enc_y)
result = foo.decrypt_result(result_enc)
print(f"Expected result for `foo`: {foo(7,8)}, decrypted result: {result}")
```

AlexanderViand-Intel:
As mentioned in yesterday's meeting, there's been a bit more progress, in addition to what I mentioned above:

* Lots of debugging improvements, including
   * Support for custom pipelines , making the frontend a convenient tool for debugging/testing HEIR itself
   ```python
   @compile(heir_opt_options=["--mlir-to-bgv", "--canonicalize", "--cse"], backend=None, debug=True)
   ```

   * Debug mode now also creates `--view-op-graph` Graphviz/*.dot files so you can inspect pre/post HEIR IR visually
   * Python source locations are now correctly forwarded to MLIR, e.g., you'll now get 
   `/home/aviand/code/heir/heir_py/example.py:19:0: error: failed to legalize operation 'secret.generic' that was explicitly marked illegal` , and most decent IDEs should let you click on the path:line:col to jump to it directly.
   
 * Initial work on supporting different backends, but there's really only `openfhe` or `none` at the moment, with the latter simply meaning that no `heir-translate` happens and that no actual fhe eval and helper functions get generated.
 
Since this PR is getting a bit large already, I'm leaning towards cleaning this up (once #1302 is merged, allowing for full scheme/backend separation) and then making a new PR for the tensor shape inference stuff (which will probably be an MLIR pass after all).

j2kun:
> Python source locations are now correctly forwarded to MLIR

❤️ !!

Feel free to clean/split it up as you see fit. I have no problems reviewing a large PR for this one.

j2kun:

AlexanderViand-Intel:
This is essentially ready to review, I just need some help getting the test setup working properly, as the e2e test currenlty fails if run via bazel.

Since the last update, major changes include
* Some heavy re-factoring/cleanup, including lots and lots of new checks and error messages (now in color 🌈!)
* You can now `pip install` the frontend (well, locally, as it's not yet on pypi)
* OpenFHE is now no longer special, and instead there are pluggable backends, which can do whatever heir-opt, heir-translate, clang/etc calls they want to do, as long as they wire up everything to the expected python client functions. The `DummyBackend` for use with custom `heir_opt_options` simply makes all of these no-ops/plain python.
* Probably some other stuff I forgot 😅

AlexanderViand-Intel:
> I have no problems reviewing a large PR for this one.

@j2kun : let me now if this is *too* large, and I'll figure out how to split it into more sensible chunks

j2kun:
The testing issue is because you put the python package one step further nested under `frontend/heir`, and bazel's working directory is the base directory of the project. So all the imports would need to be changed from `import heir` to `import frontend.heir`. Ideally a published python package would not have that more verbose import path. But we'd need to figure out how to make it nice for development and publishing.

What do you think about leaving the paths as more verbose now, and having a followup PR focus on the PyPI publishing part? I think what a lot of people do in this case (e.g., jax, tensorflow) is call a bunch of `mv` commands from bazel to make the python import paths work out. It would be nice though, if instead we could find a way to tell bazel to set its working directory to `frontend/` for these tests.

j2kun:
Ohh! Researching this I found on https://bazel.build/reference/be/python

```
imports	
List of strings; default is []

List of import directories to be added to the PYTHONPATH. Subject to "Make variable" substitution. These import directories will be added for this rule and all rules that depend on it (note: not the rules this rule depends on. Each directory will be added to `PYTHONPATH` by `py_binary` rules that depend on this rule. The strings are repo-runfiles-root relative, Absolute paths (paths that start with `/`) and paths that references a path above the execution root are not allowed and will result in an error.
```

So with this patch I can get it to run in bazel, although the test fails

```patch
diff --git a/frontend/BUILD b/frontend/BUILD
index 190aee59..342bf248 100644
--- a/frontend/BUILD
+++ b/frontend/BUILD
@@ -34,6 +34,7 @@ py_library(
     ),
     data = DATA_DEPS,
     deps = [
+        "@frontend_pip_deps_colorama//:pkg",
         "@frontend_pip_deps_numba//:pkg",
         "@frontend_pip_deps_pybind11//:pkg",
         "@frontend_pip_deps_pybind11_global//:pkg",
diff --git a/frontend/testing.bzl b/frontend/testing.bzl
index 3d24a318..902f7dd6 100644
--- a/frontend/testing.bzl
+++ b/frontend/testing.bzl
@@ -29,6 +29,7 @@ def frontend_test(name, srcs, deps = [], data = [], tags = []):
             ":frontend",
             "@com_google_absl_py//absl/testing:absltest",
         ],
+        imports = ["."],
         data = data,
         tags = tags,
         env = {
```

AlexanderViand-Intel:
Now the bazel tests work locally but fail on the CI, probably because I messed something up with openfhe_config...

```
ValueError: Error running /usr/bin/clang. stderr was:

/tmp/tmpm1pni9oh/foo.cpp:2:10: fatal error: 'openfhe/pke/openfhe.h' file not found
    2 | #include "openfhe/pke/openfhe.h"  // from @openfhe
      |          ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.


options were:

/usr/bin/clang /tmp/tmpm1pni9oh/foo.cpp -o /tmp/tmpm1pni9oh/foo.so -O3 -fPIC -shared -std=c++17 -I/usr/local/include/openfhe -I/usr/local/include/openfhe/binfhe -I/usr/local/include/openfhe/core -I/usr/local/include/openfhe/pke -L/usr/local/lib -lOPENFHEbinfhe -lOPENFHEcore -lOPENFHEpke 
```
I'll look into it tomorrow!

frontend/heir/mlir/types.py
```diff
@@ -0,0 +1,86 @@
...
+class I64(MLIRTypeAnnotation):
+ def numba_str():
+   return "int64"
+
+
+
+# Helper functions
+
+def to_numba_str(type) -> str:
+  if(get_origin(type) == Secret):
```
j2kun, line 58: Not major, but I wonder if it would be better to enforce this at the time the type is constructed. Something like

```python
class NonSecret(Protocol):
    pass
    
T = TypeVar('T', bound=NonSecret)

class Secret(Generic[T], MLIRTypeAnnotation):
  ...
  
class I32(NonSecret, MLIRTypeAnnotation):
  ...
```

I am not an expert in Python typing, so this may be a bad way to do it.

frontend/heir/mlir/mlir_decorator.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if not kwargs:
+        # No option is given
+        return _mlir(*args)
+    else:
+        # options are given, create a new callable to recv the
+        # definition function
+        def wrapper(func):
+            return _mlir(func)
+        return wrapper
```
j2kun, line 116:
```suggestion
    name = getattr(func, '__name__', str(func))
    llc = _MLIR(name, func, **kwargs)
    llc._register()
    return llc
```

Since the only usage we have doesn't need the wrapper, let's leave the rest of this stuff out.

Also, what is `llc`? Looks like copypaste, and I can't figure out what it is supposed to mean from the upstream numba source. Weird that it conflicts with `llc` the llvm compiler, but maybe `c` is for cache?

frontend/e2e_test.py
```diff
@@ -1,20 +1,18 @@
...
 
   def test_simple_arithmetic(self):
-    @compile(backend="openfhe")
-    def foo(a, b):
+    @compile() # defaults to BGV and OpenFHE
+    def foo(a : Secret[I64], b : Secret[I64]):
       return a * a - b * b
 
-    # Test the e2e path
+    # Test plaintext functionality
```
j2kun, line 12: I'm slightly against having the default callable run the cleartext function. I think it is mildly helpful to be able to call the cleartext function no matter which backend you compile with. On the other hand, it's surprising to compile a function and then, by an accident of not fully understanding the API, run something other than the compiled function.

It would be uniform to use the `DummyBackend` (or `CleartextBackend`) to do the cleartext computation. But if we really want a cleartext implementation attached to every backend, we should add it as a distinguished name, such as `eval_cleartext` or similar.

frontend/heir/mlir/mlir_decorator.py
```diff
@@ -0,0 +1,116 @@
...
+import weakref
+import collections
+import functools
+
+from numba.core import types, config
+from numba.core.typing.templates import infer
+from numba.core.serialize import ReduceMixin
+
+
+class _MLIR(ReduceMixin):
```
j2kun, line 13: Wow what a mess. I'm hoping this doesn't leak complexity into the rest of the frontend :(

frontend/heir/backends/openfhe/backend.py
```diff
@@ -0,0 +1,227 @@
...
+
+  def __getattr__(self, key):
+    if key == "crypto_context":
+       msg = f"HEIR Error: Please call {self.compilation_result.func_name}.setup() before calling {self.compilation_result.func_name}.encrypt/eval/decrypt";
+       init(autoreset=True)
+       print(Fore.RED + Style.BRIGHT + msg)
+       raise RuntimeError(msg)
+
+    if key.startswith("encrypt_"):
+      # TODO (#1162): Prevent people from doing a = enc_x, b = enc_y for foo(x,y) but then calling foo(b,a)!
```
j2kun, line 51: So long as the types match up, I don't see a particular reason to prevent this. Maybe we could extend the wrappers to allow for passing keyword arguments so that people can be careful if they want to be. Generally I don't think the compiled code can enforce this, so it would require the wrapper to check the kwargs.

frontend/heir/backends/openfhe/backend.py
```diff
@@ -0,0 +1,227 @@
...
+      # TODO (#1162): Consider hardening the eval function against handcrafted ciphertexts with wrong metadata being passed in?
+      fn = self.compilation_result.main_func
+
+      def wrapper(*args, crypto_context=None):
+        return fn(crypto_context or self.crypto_context, *args)
+
+      return wrapper
+
+    # if key == "func":
+    #   return self.compilation_result.func
```
j2kun, line 75:
```suggestion

```

frontend/heir/backends/openfhe/backend.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class OpenFHEBackend(BackendInterface):
+    def __init__(self, openfhe_config: OpenFHEConfig):
+        self.openfhe_config = openfhe_config
+        self.heir_opt_options = []
+
+    def run_backend(self, workspace_dir, heir_opt, heir_translate, func_name, arg_names, secret_args, heir_opt_output, debug):
+        # Initialize Colorama for error and debug messages
+        init(autoreset=True)
```
j2kun, line 90: nit: The comment would not be needed (more self-documenting) if instead you did this:

```
import colorama
from colorama import Fore, Style

...

    colorama.init(...)
```

Also internal Google style guide does not allow importing barewords, so I will likely have to patch this internally to be

```
import colorama

Fore = colorama.Fore
Style = colorama.Style
```

frontend/heir/backends/openfhe/backend.py
```diff
@@ -0,0 +1,227 @@
...
+        so_filepath = Path(workspace_dir) / f"{func_name}.so"
+        linker_search_paths = [self.openfhe_config.lib_dir]
+        if(debug):
+            args = clang_backend.clang_arg_helper(cpp_source_filepath=cpp_filepath,
+            shared_object_output_filepath=so_filepath,
+            include_paths=self.openfhe_config.include_dirs,
+            linker_search_paths=linker_search_paths,
+            link_libs=self.openfhe_config.link_libs)
+            print(
+            f"HEIRpy Debug (OpenFHE Backend):\033[1m Running clang {' '.join(str(arg) for arg in args)}\033[0m"
```
j2kun, line 162:
```suggestion
            f"HEIRpy Debug (OpenFHE Backend):" + Style.BRIGHT + f" Running clang {' '.join(str(arg) for arg in args)}"
```

Guessing this is what you had before using colorama?

j2kun: Ditto below on line 186

frontend/heir/backends/util/clang.py
```diff
@@ -30,6 +30,46 @@ def _find_clang(self):
       return clang
     return shutil.which("clang")
 
+
+  def clang_arg_helper(
```
j2kun, line 34: This is currently only used in debug mode. Could you also use it in the actual compilation func below, so that we avoid code drift between the two implementations?

You could also just pass a debug mode (or a debug callback) to the clang helper and have it print the args there.

j2kun (changes requested):
Amazing work!

j2kun:
> Now the bazel tests work locally but fail on the CI, probably because I messed something up with openfhe_config...
> 
> ```
> ValueError: Error running /usr/bin/clang. stderr was:
> 
> /tmp/tmpm1pni9oh/foo.cpp:2:10: fatal error: 'openfhe/pke/openfhe.h' file not found
>     2 | #include "openfhe/pke/openfhe.h"  // from @openfhe
>       |          ^~~~~~~~~~~~~~~~~~~~~~~
> 1 error generated.
> 
> 
> options were:
> 
> /usr/bin/clang /tmp/tmpm1pni9oh/foo.cpp -o /tmp/tmpm1pni9oh/foo.so -O3 -fPIC -shared -std=c++17 -I/usr/local/include/openfhe -I/usr/local/include/openfhe/binfhe -I/usr/local/include/openfhe/core -I/usr/local/include/openfhe/pke -L/usr/local/lib -lOPENFHEbinfhe -lOPENFHEcore -lOPENFHEpke 
> ```
> 
> I'll look into it tomorrow!

The command is using `-I/usr/local/include/openfhe` when it should be using a bazel-ified path. I think this is because `frontend/heir/pipeline.py:compile` uses `DEFAULT_INSTALLED_OPENFHE_CONFIG` as its kwarg default, rather than calling `from_os_env()`. I think we should always use `from_os_env` since it internally defaults to the `DEFAULT_INSTALLED_OPENFHE_CONFIG` when env vars are missing.

That said, I am trying to wrap your `example.py` in a `py_binary` and for whatever reason that _doesn't_ set a `RUNFILES_DIR` env var... so I will have to improve that in future work.

j2kun:
> That said, I am trying to wrap your example.py in a py_binary and for whatever reason that doesn't set a RUNFILES_DIR env var... so I will have to improve that in future work.

I was able to figure out why this is: bazel doesn't set the same environment variables for binary targets vs test targets, so I need to add another env var check for `BINDIR`. This should be easy but I will wait for this PR to be submitted.

AlexanderViand-Intel:
Thanks for the detailed review and all the bazel-debugging! I'll try and get this pull-ready later today :)

j2kun:
FYI @AlexanderViand-Intel  Asra's loop support is in review, so you may have a big rebase ahead of you.

j2kun:
> > FYI @AlexanderViand-Intel Asra's loop support is in review, so you may have a big rebase ahead of you.
> 
> Sorry for letting this sit for so long, as you've probably noticed I've not had a lot of bandwidth the last few weeks. I guess the best move is to wait for #1307 to be merged, then rebase and split this PR into two, one with the various QoL features and another one for the actual type inference?

SGTM. I'd give it maybe two weeks before we have to take it over and get it merged so we can have the issues ironed out before the conferences :)

AlexanderViand-Intel:
I just checked, and there's actually not much to do for the rebase, just needed to replace a hardcoded `(i64)` with a type lookup.  However, there is a more subtle issue, in that the iterator variable will have HW-bitwidth, which we probably don't want, since that might force a large ptxt modulus to be selected in the future. I guess we want to round up to the next power of two bitwidth that can fit `range_end`? 

More generally, HEIR doesn't currently support doing things such as `i8 + i16` (more specifically, I don't think our lowerings handle the `arith.truncate`/`arith.extui` this would require. I feel like this will be more and more relevant as we get code not from hand-written MLIR but autogenerated from frontend(s).

frontend/heir/pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            f.write(graph)
+
+
+    # Run heir_opt
+    if(debug):
+        heir_opt_options.append("--view-op-graph")
+        print("HEIR Debug: " + Style.BRIGHT + f"Running heir-opt {' '.join(heir_opt_options)}")
+    heir_opt_output, graph = heir_opt.run_binary_stderr(
+        input=mlir_textual,
+        options=(heir_opt_options),
```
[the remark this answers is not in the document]
AlexanderViand-Intel: Interestingly enough, it doesn't seem like pyink got rid of this automatically.

frontend/heir/core.py
```diff
@@ -0,0 +1,55 @@
...
+    - {func_name}
+    """
+    ...
+
+  def __call__(self, *args, **kwargs):
+    """Forwards to the original function."""
+    return self.func(*args, **kwargs)
+
+class BackendInterface(ABC):
+  ...
```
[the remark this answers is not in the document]
AlexanderViand-Intel: Thanks, I think I had pylance disabled accidentally - I've fixed this and a bunch of other typing issues.

Of course, any pure-python use of a `@compile` decorated function `foo` results in
`Argument of type "Literal[7]" cannot be assigned to parameter "a" of type "Secret[I64]" in function "foo"
  "Literal[7]" is not assignable to "Secret[I64]"`  so I had to do `def compile(...) -> Any` which makes the type of `foo` Unknown and avoids the issue (at least for my version of pyright. Mypy didn't seem to spot the issue either way).
  
I went down the rabbit hole of trying to get the type-checker-determined type of foo to be `(int, int) -> int` but unfortunately didn't find a satisfying solution. Someone with more knowledge of Python type checkers might be able to do it, though.

frontend/heir/backends/dummy.py
```diff
@@ -0,0 +1,52 @@
...
+
+      def wrapper(arg):
+        print("HEIR Warning (Dummy Backend): " + Fore.YELLOW + Style.BRIGHT + f"{self.compilation_result.func_name}.{key}() is a no-op in the Dummy Backend")
+        return arg
+
+      return wrapper
+
+    if key == self.compilation_result.func_name or key == "eval":
+      print("HEIR Warning (Dummy Backend): " + Fore.YELLOW + Style.BRIGHT + f"{self.compilation_result.func_name}.eval() is the same as {self.compilation_result.func_name}() in the Dummy Backend.")
+      return self.func
```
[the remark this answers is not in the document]
AlexanderViand-Intel: I think that's a good point! Hopefully, renaming this to "Cleartext Backend" makes it a bit more obvious, especially once we add a "Plaintext Backend" that actually goes through HEIR, does arithmetization, etc. That requires the mod-arith based plaintext execution from #1266, though

frontend/heir/mlir_emitter.py
```diff
@@ -110,6 +152,36 @@ def emit_assign(self, assign):
           # nothing to do, forward the name to the arg of bool()
           self.forward_name(from_var=assign.target, to_var=assign.value.args[0])
           return ""
+        if self.globals_map[func.name] == "matmul":
```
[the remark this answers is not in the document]
AlexanderViand-Intel: This probably shouldn't even be in the PR, since there's no real support for tensor types anyway 🙈

frontend/heir/mlir/types.py
AlexanderViand-Intel: Oh, neat! Unfortunately, this seems to give much less useful error message than the simple runtime check:

`TypeError: Cannot create a consistent method resolution order (MRO) for bases Generic, NonSecret, MLIRTypeAnnotation`

frontend/heir/mlir/mlir_decorator.py
AlexanderViand-Intel: I'm generally not yet happy with how the "intrinsic" thing works, as I've struggled to get something that has (a) produces a nice func call in the bytecode (b) executes normally in pure python and (c) works with type inference.

SInce this is only for matmul, which I think makes sense to drop from the PR, I'll just remove this, too.

frontend/e2e_test.py
AlexanderViand-Intel: I think I might agree on not using foo.__call__ for the pure python implementation. I think in the original design, foo(a,b) was essentially foo.setup, foo.eval(foo.enc_a(a),foo.enc_b(b)) and I'll probably revert it to that behavior.

I don't think of CleartextBackend as an alternative, though. That's really more a debugging tool for when you want to run compiler pipelines that don't actually lead to any backend. How about we expose the pure python function as `foo.original(...)`?

frontend/heir/mlir/mlir_decorator.py
AlexanderViand-Intel: Yeah, see also the comment above. I'm really not happy with how the "intrinsic" thing works and I think I need to dig a bit deeper into numba here. I've gone down the rabbit hole a bit on this and tried various ways Numba suggests you do this (overload, intrinsic, etc) but the best I can do so far is this kind of "franken-intrisic". If we don't want to use numba's type inference, things might get a bit easier.

frontend/heir/backends/openfhe/backend.py
AlexanderViand-Intel: This was a legitimate concern I got when demoing the frontend internally. One reason we should prevent this is that people might think it's OK to swap, e.g., enc_a(a) and enc_b(b) because they are both ciphertexts, but with a suitably clever HEIR pipeline, they might actually differ on things such as packing, RNS elements, etc. A backend like OpenFHE would then either just produce garbage or maybe throw a cryptic runtime error about incompatible ciphertexts.

I was thinking about having `encrpyt_<arg_name>` return tuple[arg_name, self.compilation_result.arg_enc_funcs[arg_name]]` instead of the just the latter, and then having eval check that they match.

frontend/e2e_test.py
j2kun: Any name that requires opt-in syntax is fine with me.

frontend/heir/heir_cli/heir_cli_config.py
```diff
@@ -12,12 +13,13 @@ class HEIRConfig:
   heir_opt_path: str
   heir_translate_path: str
 
-
+repo_root = pathlib.Path(__file__).resolve().parents[3]
```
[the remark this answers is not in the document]
AlexanderViand-Intel: Oh, nice! I had to modify it slightly because bazel-bin might also be a sibling of one of the parents rather than a parent directly (if running directly from the source dir, rather than via bazel) but this should be way more robust!

AlexanderViand-Intel:
I've integrated the review feedback and removed all the matmul/intrinsic related things for now, as I need to dig deeper into numba for that. 

I've also found a workaround for the issue mentioned in #1252 (numba will upcast, e.g., int8 + int8 to int32/int64). It's not very pretty, so there's  a good chance some internal tool will not accept it (I recall some C++ linking hack also getting stopped by internal checks), but I guess loading a modified version of a module via importlib is *technically* "normal" python?

frontend/heir/backends/openfhe/backend.py
j2kun: I see the problem now. I suppose forcing it by adding a layer of indirection in the API is a good approach. Another weaker option is to use a Python feature that allows you to force the user to pass values by named keyword arguments. At least then it will be easy to see when someone is violating the order of operands.

AlexanderViand-Intel:
@j2kun: I think everything I was planning to do (and a lot of semi-related QoL things...feature creep I guess 😓)  are in now.|

There is one more "QoL" thing I'd love to tackle, and that's the inability to have multiple instances of the frontend/HEIR (the "PublicKey registerd" error), but I think I might need your assistance to tackle that one as it's less about the frontend and more about the bindings.

Next steps would be 
* Enabling "intrinsics" (e.g., `linalg.matmul`) while preserving pure-python functionality,
* Tensor type/shape inference (either in Python or MLIR)

j2kun (approved):

Issue #1168, opened 2024-12-10
Analyse the EvalAddCount and KeySwitchCount and Populate N/moduli chain for BGV


ZenithalHourlyRate:
Main reference: ["Revisiting Homomorphic Encryption Schemes for Finite Fields"](https://eprint.iacr.org/2021/204) and [OpenFHE pke examples readme](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/examples/README.md)

With the modreduce/relinearize strategy introduced in #1126, we can analyse the EvalAddCount and KeySwitchCount, hence determining the noise growth and enabling HEIR to select ring dimension `N` and the moduli chain `qi` according to the paper.

The first step could be using the analysis to determine these _counts_ and when configuring the OpenFHE CryptoContext we can pass these parameters and let OpenFHE determine `N` and `qi`.

Further step could introduce a pass to use these counts to populate `N`/`qi` inside HEIR, which allows more information inside HEIR, allows us to SetFirstModSize/SetScalingModSize for OpenFHE and could benefit the new Lattigo backend (#1159) which allows the user/compiler to specify `N`/`qi` in a more fine grained way.

Further more, instead of OpenFHE specifying only the first/scaling modulus size, the compiler can specify each level modulus size because we can get the AddCount/KeySwitchCount for each level (instead of a global one). This often means a smaller `Q`.

Also, we can optimize the size of special moduli `pi`, and may possibly reduce the ring dimension `N` as the total `QP` is smaller. This mainly refers to Section 6 of ["Optimisations and tradeoffs for HElib"](https://eprint.iacr.org/2023/104)

PR #1254, opened 2025-01-11
BGV: Count Add/Keyswitch in one level for OpenFHE

ZenithalHourlyRate:
Part of #1168.

This is a baseline of parameter selection for BGV. While concrete param (like moduli) is still selected by the OpenFHE library, the information of the circuit needed by the OpenFHE is now provided by the compiler.

Note that providing `EvalAddCount/KeySwitchCount` is crucial under the [Application-Aware security model](https://ia.cr/2024/203), where the compiler/library should reject circuit whose parameter can not be generated, or generate larger parameters. Cite

> In the case of the attack in [11] the proper OpenFHE use of
BGV/BFV for this scenario would require the user to supply
the number of additions before generating the parameters
using SETEVALADDCOUNT, or an equivalent multiplicative
depth using SETMULTIPLICATIVEDEPTH. One can check
that, when this is done correctly, a larger parameter set is
generated by OpenFHE than the one used for the attack.

When further optimization on parameter selection is landed, this can be kept as a comparable baseline for evaluation purpose. For end-user, they could then select different param generation mechanism based on their needs.

### Example

For the `dot_product_8.mlir`, we get the following counting result.

```mlir
module {
  func.func @func(%arg0: !secret.secret<tensor<8xi16>>, %arg1: !secret.secret<tensor<8xi16>>) -> !secret.secret<i16> attributes {mgmt.openfhe_params = #mgmt.openfhe_params<evalAddCount = 8, keySwitchCount = 15>} {
    %c2 = arith.constant 2 : index
    %c4 = arith.constant 4 : index
    %c7 = arith.constant 7 : index
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<tensor<8xi16>>, !secret.secret<tensor<8xi16>>) attrs = {arg0 = {addCount = 1 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>}, arg1 = {addCount = 1 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>}} {
    ^bb0(%input0: tensor<8xi16>, %input1: tensor<8xi16>):
      %1 = arith.muli %input0, %input1 {addCount = 1 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2, dimension = 3>} : tensor<8xi16>
      %2 = mgmt.relinearize %1 {addCount = 1 : i64, keySwitchCount = 1 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
      %3 = tensor_ext.rotate %2, %c4 {addCount = 1 : i64, keySwitchCount = 2 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>, index
      %4 = arith.addi %2, %3 {addCount = 2 : i64, keySwitchCount = 3 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
      %5 = tensor_ext.rotate %4, %c2 {addCount = 2 : i64, keySwitchCount = 4 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>, index
      %6 = arith.addi %4, %5 {addCount = 4 : i64, keySwitchCount = 7 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
      %7 = tensor_ext.rotate %6, %c1 {addCount = 4 : i64, keySwitchCount = 8 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>, index
      %8 = arith.addi %6, %7 {addCount = 8 : i64, keySwitchCount = 15 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
      %9 = mgmt.modreduce %8 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<8xi16>
      %extracted = tensor.extract %9[%c7] {addCount = 1 : i64, keySwitchCount = 1 : i64, mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<8xi16>
      %10 = mgmt.modreduce %extracted {mgmt.mgmt = #mgmt.mgmt<level = 0>} : i16
      secret.yield %10 : i16
    } -> !secret.secret<i16>
  }
}
```

A very counter-intuitive result is that instead of `KeySwitchCount` being 4 (1 mul-relin + 3 rotate), it is 15 as it has been accumulated via additions. Detailed definition of KeySwitchCount could be found in [KPZ21, section 4](https://ia.cr/2021/204). Though I want to ask for review of relevent people whether I understand them correctly.

As a result, the generated param becomes

```
Element Parameters: ILDCRTParams [m=32768 n=16384 q=748300257136160271476183612726042708089465225445377 ru=0 bigq=0 bigru=0]
  m_params:
    0: ILParams [m=32768 n=16384 q=140737488486401 ru=3152167261 bigq=0 bigru=0]
    1: ILParams [m=32768 n=16384 q=9007199255560193 ru=523433461118 bigq=0 bigru=0]
    2: ILParams [m=32768 n=16384 q=9007199255658497 ru=785944929279 bigq=0 bigru=0]
    3: ILParams [m=32768 n=16384 q=65537 ru=9 bigq=0 bigru=0]
```

where the moduli chain has `[47, 53, 53, 16]` bits for each level and 169 bits in total.

When they are not specified, the generated param is

```
Element Parameters: ILDCRTParams [m=32768 n=16384 q=187075064281998384050658167829053521773423705948161 ru=0 bigq=0 bigru=0]
  m_params:
    0: ILParams [m=32768 n=16384 q=140737488486401 ru=3152167261 bigq=0 bigru=0]
    1: ILParams [m=32768 n=16384 q=4503599627763713 ru=51902047037 bigq=0 bigru=0]
    2: ILParams [m=32768 n=16384 q=4503599627796481 ru=38602712989 bigq=0 bigru=0]
    3: ILParams [m=32768 n=16384 q=65537 ru=9 bigq=0 bigru=0]
  m_originalModulus: 0
```

With moduli chain having `[47, 52, 52, 16]` bits and 167 bits in total.

### DIFF
--- a/lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
+++ b/lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
@@ -1,0 +1,159 @@
+#include "lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.h"
+
+#include "lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h"
+#include "lib/Dialect/Mgmt/IR/MgmtAttributes.h"
+#include "lib/Dialect/Mgmt/IR/MgmtOps.h"
+#include "lib/Dialect/Secret/IR/SecretOps.h"
+#include "lib/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "llvm/include/llvm/ADT/TypeSwitch.h"              // from @llvm-project
+#include "mlir/include/mlir/Analysis/DataFlowFramework.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"      // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"    // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"                // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"                    // from @llvm-project
+#include "mlir/include/mlir/IR/Visitors.h"                 // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"                // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+LogicalResult CountAnalysis::visitOperation(
+    Operation *op, ArrayRef<const CountLattice *> operands,
+    ArrayRef<CountLattice *> results) {
+  auto propagate = [&](Value value, const CountState &state) {
+    auto *lattice = getLatticeElement(value);
+    ChangeResult changed = lattice->join(state);
+    propagateIfChanged(lattice, changed);
+  };
+
+  llvm::TypeSwitch<Operation &>(*op)
+      .Case<secret::GenericOp>([&](auto genericOp) {
+        Block *body = genericOp.getBody();
+        for (auto i = 0; i != body->getNumArguments(); ++i) {
+          auto blockArg = body->getArgument(i);
+          // one Vfresh
+          propagate(blockArg, CountState(1, 0));
+        }
+      })
+      .Case<arith::AddIOp, arith::SubIOp, arith::AddFOp, arith::SubFOp>(
+          [&](auto &op) {
+            // condition on result secretness
+            SmallVector<OpResult> secretResults;
+            getSecretResults(op, secretResults);
+            if (secretResults.empty()) {
+              return;
+            }
+
+            CountState zeroState(0, 0);
+            SmallVector<OpOperand *> secretOperands;
+            getSecretOperands(op, secretOperands);
+            for (auto *operand : secretOperands) {
+              auto countState =
+                  operands[operand->getOperandNumber()]->getValue();
+              zeroState = zeroState + countState;
+            }
+
+            for (auto result : secretResults) {
+              propagate(result, zeroState);
+            }
+          })
+      .Case<arith::MulIOp, arith::MulFOp>([&](auto &op) {
+        SmallVector<OpResult> secretResults;
+        getSecretResults(op, secretResults);
+        if (secretResults.empty()) {
+          return;
+        }
+
+        // now noise is Vmult
+        // TODO(#1168): we can actually do a more fine grained analysis here
+        // distinguishing ct-ct and ct-pt
+        propagate(op.getResult(), CountState(1, 0));
+      })
+      .Case<mgmt::RelinearizeOp, tensor_ext::RotateOp>([&](auto &op) {
+        SmallVector<OpOperand *> secretOperands;
+        getSecretOperands(op, secretOperands);
+        if (secretOperands.empty()) {
+          return;
+        }
+
+        auto state = operands[0]->getValue();
+        if (!state.isInitialized()) {
+          return;
+        }
+
+        propagate(op.getResult(), state.keySwitch());
+      })
+      // TODO(#1174): in BGV tensor::ExtractOp is assumed to be always
+      // mul+const
+      .Case<tensor::ExtractOp>([&](auto &op) {
+        SmallVector<OpResult> secretResults;
+        getSecretResults(op, secretResults);
+        if (secretResults.empty()) {
+          return;
+        }
+
+        // now noise is Vmult + one Vks
+        propagate(op.getResult(), CountState(1, 1));
+      })
+      .Case<mgmt::ModReduceOp>([&](auto modReduceOp) {
+        // implicitly ensure that the operand is secret
+
+        propagate(modReduceOp.getResult(), CountState(0, 0));
+      });
+  // should not propagate through mgmt::ModReduceOp
+  return success();
+}
+
+void annotateCount(Operation *top, DataFlowSolver *solver) {
+  auto getIntegerAttr = [&](int level) {
+    return IntegerAttr::get(IntegerType::get(top->getContext(), 64), level);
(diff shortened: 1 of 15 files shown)


j2kun:
Still taking a look at this, but I noticed in your `dot_product_8.mlir` example the `addCount` parameter is doubling. Is that the right thing to do? I'd expect 3 adds, not `8 = 2^3`.

lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
```diff
@@ -0,0 +1,157 @@
...
+    propagateIfChanged(lattice, changed);
+  };
+
+  llvm::TypeSwitch<Operation &>(*op)
+      .Case<secret::GenericOp>([&](auto genericOp) {
+        Block *body = genericOp.getBody();
+        for (auto i = 0; i != body->getNumArguments(); ++i) {
+          auto blockArg = body->getArgument(i);
+          // one Vfresh
+          propagate(blockArg, CountState(1, 0));
```
j2kun, line 35: Related to my other comment, naively I would expect zero adds here, not 1, and below (line 53) it appears you're adding the countState once per operand, rather than per op.

ZenithalHourlyRate:
> Still taking a look at this, but I noticed in your `dot_product_8.mlir` example the `addCount` parameter is doubling. Is that the right thing to do? I'd expect 3 adds, not `8 = 2^3`.

This indeed is quite a less understood thing until recently. KPZ21 does not illustrate the difference but [Application Aware, Introduction, Attacks by Guo et al.](https://eprint.iacr.org/2024/203) points out the difference:

Let's say we have the following two circuit

```mlir
%0 = add %arg0, %arg1
%1 = add %0, %arg2
%2 = add %1, %arg3
```

And

```mlir
%0 = add %arg0, %arg0
%1 = add %0, %0
%2 = add %1, %1
```

Suppose for each `%argi` the noise is `ei`, the noise for `%2` in the first circuit is `e0 + e1 + e2 + e3` whereas in the second circuit the noise is `8 * e0`.

Note that in KPZ21 section 4, we have the formula `(n_add + 1) || v_fresh || + n_ks || v_ks ||`. From this equation we know that for the first circuit `n_add` is indeed 3, but for the second circuit `n_add` is `7`, which I then use `8` as a close estimation.

> Related to my other comment, naively I would expect zero adds here, not 1, and below (line 53) it appears you're adding the countState once per operand, rather than per op.

This answers why we count on operands (counting `v_fresh`) instead of ops, as different combination of op gives different result.

j2kun:
> Note that in KPZ21 section 4, we have the formula `(n_add + 1) || v_fresh || + n_ks || v_ks ||`. From this equation we know that for the first circuit `n_add` is indeed 3, but for the second circuit `n_add` is `7`, which I then use `8` as a close estimation.

I get that repeated addition blows up the noise faster. But neither KPZ21 nor KBMP24 seems to give a precise definition of `n_add`. Both just call it "the number of additions" without specifying the distinction of what is being added. From that I would infer (perhaps wrongly) that this formula using `n_add` actually gives the worst-case guarantee in the sense that it assumes repeated additions.

My main concerns are:

1. We have code comment docs that allows someone to verify the implementation by referring to something that gives a clear definition (a paper or source code). 
2. The parameter should be named more directly to what it is doing. If it's called "add count" then it should count adds (and a different transformation should be made later if necessary to pass it along to openFHE's API); and if it's doing a sort of coarse noise estimate, then it should be named something that reflects that.

[This link](https://github.com/openfheorg/openfhe-development/blob/7b8346f4eac27121543e36c17237b919e03ec058/src/pke/examples/README.md?plain=1#L178) still only says "max number of additions" and browsing through their code I don't see anywhere that is more specific...

ZenithalHourlyRate:
> 2\. The parameter should be named more directly to what it is doing. If it's called "add count" then it should count adds (and a different transformation should be made later if necessary to pass it along to openFHE's API); and if it's doing a sort of coarse noise estimate, then it should be named something that reflects that.

Then it should be better called `AddNoiseFactor` and `KeySwitchNoiseFactor`, as these are coefficients for (`|| v_fresh ||` / `|| v_mult ||`) and `|| v_ks ||`.

> From that I would infer (perhaps wrongly) that this formula using n_add actually gives the worst-case guarantee in the sense that it assumes repeated additions.

Typically, noise analysis starts with numerial input, e.g. initial noise level being `2^16`. Then for the analysis to carry out, each op is visited, propagating the noise to each ciphertext. For the first circuit, the result noise is `4 * 2^16` and for the second circuit it is `8 * 2^16`. Either worst-case or average-case will give similar result.

The trick for KPZ21 is that instead of detailed numerical analysis, they view each level as composition of add and keyswitch, thus they only need to count the _equivalent_ addition number (not the same as add operation number) and _equivalent_ key switch number.

> [Edit]: We should be able to clear this up by asking Yuriy and the OpenFHE devs on their discourse.

I'll open a thread on openfhe discourse.

j2kun:
Based on https://openfhe.discourse.group/t/question-on-definition-of-evaladdcount-and-keyswitchcount/1835/3 let's go ahead with what you have here and cite that thread for followup. I still suggest that we don't need the attributes and can compute the desired values on the fly, using op interfaces to support the same code applying to multiple dialects.

j2kun:
If you still want to go forward with this PR, let's rebase and I can merge it in

j2kun (approved):

Issue #1258, opened 2025-01-14
Move cyclic repetition code to earlier pipeline instead of in emitter


ZenithalHourlyRate:
Subissue for #1191.

Currently for heir-simd-vectorizer/tensor_ext::rotate to work (i.e. HECO-style optimization like rotate-and-reduce), we assumed a cyclic repetition packing in the ciphertext slot (also assumes the plaintext slot num divides ct slot num).

However, this is handled in emitter (i.e. #1129 and #1226), the very last stage. As this is enforced by the heir-simd-vectorizer, it is clearly error-prone when we handle the assumption in two places (early pipeline and emitter).

j2kun:
So this should be trivial now that the new layout system is in place. I will give it a shot

PR #1264, opened 2025-01-14
Add initial layout selection framework

j2kun:
This PR adds an initial ciphertext layout selection pass.

This involves the following:

- Assigning to each secret tensor-typed SSA value a layout attribute as an AffineMap (see the design doc https://docs.google.com/document/d/1TTj0NDZkn0T9NNQj6idbTk2g65tAb7VxsYpt1hKjK_Y/edit?tab=t.0#heading=h.hcusv1ynbv9n)
- Assumes each function input has a row-major layout, as done by Fhelipe.
- Forward-propagating the layout attributes through the IR, with transformations when passing through ops that affect the layout.
- Inserting layout_conversion ops when the operands of an op have incompatible layouts.
- Persisting layout information as attributes on the IR:
  - func.func persists using the func op's "argAttrs" dictionary, with a new tensor_ext.layout dialect attribute as the key
  - secret.generic persists using our own copy of the func's argAttrs to mark the layouts of operands and block arguments.
  - All other ops persist the layout attributes as an ArrayAttr for the op results (usually just one).

This pass does not optimize for the selected layouts. Instead, I expect to define additional passes that optimize, while allowing this pass to be doubly used as an analysis pass (persisting layout information to ops as attributes) as well as an initial layout selection pass. As such, the logic in this pass is to convert all secret+tensor op operands to have the layout of the op's first operand.

This PR also does not yet lower vecmat/matvec using Halevi-Shoup, nor does it lower convert_layout (which will require its own effort to implement a shift network optimization algorithm). I wanted to get feedback from the community on the direction as it relates to the layout attribute and propagation before doing too much additional work.

## Example commands

An IR that requires layout conversion ops to be inserted

```bash
bazel run --noallow_analysis_cache_discard //tools:heir-opt -- --layout-propagation $PWD/tests/Transforms/layout_propagation/insert_conversion.mlir 
```

Result: https://gist.github.com/j2kun/eef84273a83893a992694c219b4dde3f

Layout assignments for Halevi-Shoup matvec mul:

```bash
bazel run //tools:heir-opt -- --secretize --wrap-generic --drop-unit-dims --canonicalize --cse --linalg-canonicalizations --layout-propagation $PWD/tests/Examples/openfhe/halevi_shoup_matmul.mlir
```

Result: https://gist.github.com/j2kun/75b676fe90d5ba1f2c67ad1a2409406b

## Open questions

- Should we have the layout persisted on the secret type? I think we may be able to get away with  running the pass to persist attributes per op, and use the op-aware type conversion established in Hongren's mgmt pass work. I don't actually use the added secret.layout annotation in this PR at all.
- There will be cases where the layout propagation cannot be known without assuming how the op will be lowered. For example, the tensor_ext.sum op assumes rotate-and-reduce is performed in the obvious way, but it could be done a different way. More complicated examples would be linalg.matmul where both inputs are encrypted, which has a dozen or so academic papers with varying methods. While I can get reproductions of Fhelipe or Orion by assuming fixed implementations of linalg ops, the big open research question is: how can we jointly optimize for the layout and the lowering kernel of an op? I suspect it simply won't be able to use `layout-propagation` as defined in this PR, and will have to do a global optimization from scratch.

## Notes

- Currently I added a tensor_ext.sum op, which I think I will remove in favor of the linalg linalg.reduce op. I wanted to have at least one op in this PR that actually modifies the layout attribute, and sum is probably the easiest one to understand while doing something meaningful.
- Currently a matvec/vecmat op doesn't change the layout of the input vector, since halevi-shoup/squat packing methods maintain layout invariance of the vector. However, if the input vector has a non-row-major encoding (e.g., strided), then we would need to lower the operation in a special way, either by customizing the diagonal packing of the plaintext matrix to match, or else by inserting a compactification op (layout conversion) on the vector.
- I wanted to rework the halevi-shoup matmul example, but noticed it's using the linalg.matmul op instead of matvec/vecmat. I looked into it, and found it was very easy to add a drop-unit-dims transform that would convert matmul to matvec/vecmat when the right dimensions were 1. I think this would be better long-term and I will port the halevi-shoup lowerings to use matvec/vecmat instead of matmul.


tests/Transforms/layout_propagation/insert_conversion.mlir
```diff
@@ -0,0 +1,51 @@
...
+  // CHECK-SAME: ins(%[[arg0]], %[[arg1]]
+  // CHECK-SAME: [[arg0]] = {layout = [[input_map]]}
+  // CHECK-SAME: [[arg1]] = {layout = [[input_map]]}
+  // Note this one denotes the layout of the result of the generic op
+  // CHECK-SAME: layout = [
+  // CHECK-SAME: [[row_reduced_map]]]
+  %0 = secret.generic ins(%arg0, %arg1: !stensor, !stensor) {
+  ^body(%pt_arg0: !tensor, %pt_arg1: !tensor):
+    // result of sum has row-major layout, i.e., with don't cares at the end
+    // (1, 2, ..., 32, x, x, x, x, ..., x)
```
ZenithalHourlyRate, line 29: Related to #1258

rotate-and-reduce code (for example, sum tensor<32x32xi16> to tensor<32xi16> then sum to i16) would assume cyclic repetition for the middle layout (d0 -> d0), namely the tail is not "dont care" but meaningful data. Could we describe that

j2kun: This is a good question. `affine_map` can't natively represent the semantics of what outside the image of the layout map. It seems reasonable to me that we could require cyclic repetition by default for all layouts that don't fill up their ciphertexts. 

I can think of a few obstacles to doing this:

 - It requires the data to have dimensions divisible by a power of 2 (so that we can get at least one full repetition, a partial repetition would be invalid), though I think most compilers do padding on non-power-of-two-sized inputs anyway.
 - Some op may have a kernel that doesn't produce cyclic repetition in its output layout, in which case we would have to insert an op (during the lowering, likely) to restore this invariant. The overhead could add up to unacceptable performance.

But I'm willing to forge ahead with this assumption anyway, since if/when it becomes a problem, we can wrap the layout attribute in a larger attribute that includes an enum specifying the semantics.

j2kun:
Did a bit of cleanup and left some TODOs so we can get the first big piece of this PR in.

New stuff since last comment:

- Added tensor_ext.assign_layout op.
- Removed tensor_ext.sum and replaced it with linalg.reduce
- Completed layout propagation pass for linalg.reduce

Followup work:

- Finish the parts of the layout propagation pass required for matvec/vecmat
- Enable drop-unit-dims and lower the newly-emitted tensor ops like collapse_shape and expand_shape
- Add a new lowering that converts data tensor types to "ciphertext tensor" types that align with the image of the layout map
- Lower convert_layout (https://github.com/google/heir/pull/1286)

ZenithalHourlyRate (approved):
Happy to see this! Just some comments on the MLIR side, others LGTM!

lib/Transforms/LayoutPropagation/LayoutPropagation.td
```diff
@@ -0,0 +1,114 @@
...
+  - Ops with a nested region and block arguments use a dictionary attribute to
+    mark the layout of each block argument. `func.func` in particular uses the
+    `tensor_ext.layout` dialect attribute, while others use an affine map
+    attribute.
+  - Other ops annotate their results with layouts as an ArrayAttr of affine
+    maps. The order of the affine maps corresponds to the order of results.
+
+  When a plaintext op like `arith.constant` is encountered, a
+  `tensor_ext.assign_layout` op is inserted that assigns it a default layout.
+  This semantically corresponds to a plaintext packing operation.
```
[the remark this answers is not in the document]
j2kun: > replacing all uses

Is your suggestion then to just replace the one use that was encountered, and rely on CSE to de-dupe any identically packed plaintexts?

lib/Dialect/TensorExt/IR/TensorExtAttributes.td
```diff
@@ -64,4 +64,20 @@ def SIMDPacking_Attr : TensorExt_Attr<"SIMDPacking", "simd_packing",
...
 
+def TensorExt_LayoutAttr : TensorExt_Attr<"Layout", "layout"> {
+  let summary = "Attribute denoting the layout of a tensor in a set of ciphertexts";
+  let description = [{
+    This attribute contains an affine map that describes the layout of a tensor
+    in a set of ciphertexts. The affine map is a function that maps tensor indices
+    to ciphertext indices (possibly with a ciphertext-selecting index).
+
+    This attribute exists primarily to provide a "dialect attribute" which is
+    required to annotate the arguments of `func.func` arguments.
```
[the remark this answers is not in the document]
j2kun: Removing the tablegen def

lib/Transforms/LayoutPropagation/LayoutPropagation.td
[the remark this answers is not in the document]
ZenithalHourlyRate: > > replacing all uses

> Is your suggestion then to just replace the one use that was encountered, and rely on CSE to de-dupe any identically packed plaintexts?

Yes.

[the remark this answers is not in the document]
j2kun: Fixed.

lib/Transforms/LayoutPropagation/LayoutPropagation.td
```diff
@@ -0,0 +1,114 @@
+#ifndef LIB_TRANSFORMS_LAYOUTPROPAGATION_LAYOUTPROPAGATION_TD_
```
asraa, line 1: Could you place this transform under lib/Dialect/TensorExt/Transforms? If a transform is pretty dialect specific I think we'd like to put it there, to avoid making lib/Transforms/ such a mixed bag (@AlexanderViand-Intel and I still had leftover un-categorized transforms there, but this one seems pretty clearly under TensorExt)

lib/Transforms/LayoutPropagation/LayoutPropagation.td
```diff
@@ -0,0 +1,114 @@
...
+
+  - Ops with a nested region and block arguments use a dictionary attribute to
+    mark the layout of each block argument. `func.func` in particular uses the
+    `tensor_ext.layout` dialect attribute, while others use an affine map
+    attribute.
+  - Other ops annotate their results with layouts as an ArrayAttr of affine
+    maps. The order of the affine maps corresponds to the order of results.
+
+  When a plaintext op like `arith.constant` is encountered, a
+  `tensor_ext.assign_layout` op is inserted that assigns it a default layout.
```
asraa, line 23: correct my understanding - it seems like after the review before about the ptxt's with potentially different layout uses, now each constant op is assigned a default layout. then when an arithmetic op is  encountered using it, it already has a default layout assigned, so then the args go through the compatibility check and possible rectification. Since we skip that block of code that checks if an arg isn't assigned yet (we always assign the constant before we hit the arithmetic op), we don't actually do the replaceUsesWithIf that was suggested below. We actually call the rectifier just chooses the first operand, doesn't that mean that we might not do the thing you wanted to do below and might end up converting the layout of the other arg? Maybe in a follow up we can assign a hierarchy/ordering to layouts and always choose to convert the complex, higher order one to the lower order (default) one?

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
```diff
@@ -0,0 +1,685 @@
...
+        convertLayoutForReduce(inputLayout, op.getDimensions());
+
+    if (reducedInputLayout != initLayout) {
+      ConvertLayoutOp convertOp = builder.create<ConvertLayoutOp>(
+          op->getLoc(), init, AffineMapAttr::get(initLayout),
+          AffineMapAttr::get(reducedInputLayout));
+      Value toReplace = convertOp.getResult();
+      // I'd like to just call op.setOperand(i, toReplace) here, but I can't
+      // figure out how the i'th entry in `getInits` corresponds to the general
+      // OpOperand index.
```
asraa, line 590: it's a bit of a hack but you can use `op.getODSOperandIndexAndLength(1).first + i`

asraa (commented):
Thanks! Just a few very qq and I'll mark as approved for you to submit

lib/Transforms/LayoutPropagation/LayoutPropagation.td
j2kun: I actually had it there and decided, because of the dependence on ops from `tensor`, `tensor_ext` and `linalg`, that it should be considered a general transform not specific to one dialect. There will definitely be more `linalg` ops coming (convolution, matmul, matvec) and I could also see it supporting layout assignment for higher-level ML ops if they are present.

lib/Transforms/LayoutPropagation/LayoutPropagation.td
j2kun: The first lines of `LayoutPropagation::visitOperation` skips ops without secret operands or results, so `arith.constant` and `tensor.empty` actually don't get a layout until they're used in a secret computation. I think actually I could remove them from the visitors and it would not change the behavior of the pass (I went back and forth between giving everything layouts, even plaintexts that are never secret, and limiting it to secret ops only; open to thoughts on that).

You're still right though, that the rectification can result in a poorly-chosen conversion based on operand order. I intend for a second pass to be responsible for optimizing the choice of layout conversions, so that this pass can be reused later in the pipeline when all we need to do is materialize layouts as op attributes. This second pass would be the "backprop" step from Fhelipe, and it would compare the cost of converting operand layouts vs pushing layout conversions backwards through an op, and in the case of a default layout assignment this would result in assigning a different (compatible) layout because it would be the cheapest option.

Updated the PR to make this clearer by removing visitOperation for Constant/Empty op

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
j2kun: I removed the comment; I think the replacement method is safer.

asraa (approved):
thanks!

PR #1271, opened 2025-01-16
Add polynomial.eval op for evaluation of a static polynomial

zmbekdemir:
This is the definition for polynomial.eval op. This operation builds without error but I am not sure if it is the final version. Insights and recommendations are appreciated. #1217

### DIFF
--- a/lib/Dialect/Polynomial/IR/PolynomialOps.td
+++ b/lib/Dialect/Polynomial/IR/PolynomialOps.td
@@ -340,4 +340,30 @@ def Polynomial_INTTOp : Polynomial_Op<"intt", [Pure]> {
   let hasVerifier = 1;
 }
 
+def Polynomial_EvalOp : Polynomial_Op<"eval"> {
+    let summary = "Evaluate a static polynomial attribute at a given SSA value.";
+    let description = [{
+        Evaluates the result of a polynomial specified as a static attribute at a given SSA value.
+        The result represents the evaluation of the polynomial at the input value and produces
+        a corresponding constant value.
+        
+        Example:
+
+        ```mlir
+        #poly = #polynomial.int_polynomial<x**2 + 2*x + 1>
+        %x = arith.constant 5 : i32
+        %result = polynomial.eval #poly, %x : !arith.constant
+        ```
+    }];
+    
+    let arguments = (ins 
+        Polynomial_AnyTypedPolynomialAttr:$polynomial,  // Static polynomial attribute
+        AnyType:$value                         // SSA value
+    );
+
+    let results = (outs AnyType:$result);
+    let assemblyFormat = "$polynomial attr-dict `polynomial` $value type($value) `->` type($result)";
+}
+
+
 #endif  // LIB_DIALECT_POLYNOMIAL_IR_POLYNOMIALOPS_TD_


lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -340,4 +340,30 @@ def Polynomial_INTTOp : Polynomial_Op<"intt", [Pure]> {
   let hasVerifier = 1;
 }
 
+def Polynomial_EvalOp : Polynomial_Op<"eval"> {
+    let summary = "Evaluate a static polynomial attribute at a given SSA value.";
+    let description = [{
+        Evaluates the result of a polynomial specified as a static attribute at a given SSA value.
+        The result represents the evaluation of the polynomial at the input value and produces
+        a corresponding constant value.
```
j2kun, line 348:
```suggestion
        a corresponding scalar value.
```

"constant" would imply statically known, and if the input is an SSA value then it's not in general.

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -340,4 +340,30 @@ def Polynomial_INTTOp : Polynomial_Op<"intt", [Pure]> {
...
+        ```
+    }];
+    
+    let arguments = (ins 
+        Polynomial_AnyTypedPolynomialAttr:$polynomial,  // Static polynomial attribute
+        AnyType:$value                         // SSA value
+    );
+
+    let results = (outs AnyType:$result);
+    let assemblyFormat = "$polynomial attr-dict `polynomial` $value type($value) `->` type($result)";
```
j2kun, line 365:
```suggestion
    let assemblyFormat = "operands attr-dict `:` type($value)";
```

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -340,4 +340,30 @@ def Polynomial_INTTOp : Polynomial_Op<"intt", [Pure]> {
...
+        Evaluates the result of a polynomial specified as a static attribute at a given SSA value.
+        The result represents the evaluation of the polynomial at the input value and produces
+        a corresponding constant value.
+        
+        Example:
+
+        ```mlir
+        #poly = #polynomial.int_polynomial<x**2 + 2*x + 1>
+        %x = arith.constant 5 : i32
+        %result = polynomial.eval #poly, %x : !arith.constant
```
j2kun, line 355:
```suggestion
        %result = polynomial.eval #poly, %x : i32
```

j2kun (changes requested):
So the tablegen looks mostly good to me! A few small nits, and then I'd suggest adding a test using this new syntax in `tests/Dialect/Polynomial/IR/` (could be in, say, `int_coefficients.mlir` or a new file if you prefer). This will ensure that `heir-opt` can run, parse+print the syntax you've chosen, as well as run any IR verifiers.

j2kun:
@zmbekdemir if it's alright with you I can go ahead and cherry-pick your change and add my patches on top to submit it.

zmbekdemir:
@j2kun yes, go ahead please.

j2kun:
Marking as pull_ready so I can patch it internally. The commit should still be attributed to you when it lands.

PR #1274, opened 2025-01-17
Common configure-crypto-context for relin/rotate

### LINKED ISSUE #1145: Generalize OpenFHE Param/Context/KeyGen functionality to work at scheme/LWE level

ZenithalHourlyRate:
This PR deals with the relin/rotation setup, which has been in `--openfhe-configure-crypto-context`.

For now, we add `lwe.relinearize_degree` and `lwe.rotation_indices` attribute on the main function, and expect further backend specific lowering to use them to configure their context.

The exact definition of the lwe attribute should be discussed.

### Example

`--bgv-to-lwe --lwe-configure-crypto-context=entry-function=dot_product`:

```mlir
func.func @dot_product(%ct: !ct_L2_, %ct_0: !ct_L2_) -> !ct_L0_ attributes
 {lwe.relinearize_degree = #lwe.relinearize_degree<basis = [2]>,
 lwe.rotation_indices = #lwe.rotation_indices<indices = [1, 2, 4, 7]>}
```

### DIFF
--- a/lib/Dialect/BUILD
+++ b/lib/Dialect/BUILD
@@ -13,6 +13,7 @@ cc_library(
     hdrs = ["HEIRInterfaces.h"],
     deps = [
         ":interfaces_inc_gen",
+        "@llvm-project//mlir:FuncDialect",
         "@llvm-project//mlir:IR",
     ],
 )
--- a/lib/Dialect/HEIRInterfaces.cpp
+++ b/lib/Dialect/HEIRInterfaces.cpp
@@ -6,5 +6,32 @@
 namespace mlir {
 namespace heir {
 #include "lib/Dialect/HEIRInterfaces.cpp.inc"
+
+// Helper function to find all the relin basis in the function
+SmallVector<int32_t> findAllRelinBasis(func::FuncOp op) {
+  std::set<int32_t> distinctRelinBasis;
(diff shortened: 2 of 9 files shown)


j2kun:
Based on https://openfhe.discourse.group/t/question-on-definition-of-evaladdcount-and-keyswitchcount/1835/3 let's go ahead with what you have here and cite that thread for followup. I still suggest that we don't need the attributes and can compute the desired values on the fly, using op interfaces to support the same code applying to multiple dialects.

ZenithalHourlyRate:
After discussion with @j2kun, we may add `RelinearizeOpInterface` and `RotateOpInterface` with helpers to extract the common patterns instead of extra lwe attributes / lwe ops.

In this way, we may not have a `--lwe-configure-crypto-context` pass (proposed in #1145) that populates some extra func like `@func__configure` with lwe op/attr inside and then backend pass to lower them; instead backend specific `configure-crypto-context` needs to create their own configuration func, which is reasonable as different backend has drastically different ways to configure them.

Issue #1275, opened 2025-01-17
CI: bazel build failure on recent PRs - no such file clang


asraa:
I'm noticiing PRs with:

```
[8,465 / 12,740] [Prepa] TdGenerate lib/Dialect/BGV/IR/BGVOps.cpp.inc [for tool] ... (2 actions, 0 running)
ERROR: /home/runner/work/heir/heir/lib/Dialect/BGV/IR/BUILD:11:11: Compiling lib/Dialect/BGV/IR/BGVDialect.cpp [for tool] failed: (Exit 1): clang failed: error executing command (from target //lib/Dialect/BGV/IR:Dialect) /usr/lib/llvm-14/bin/clang -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer -g0 ... (remaining 95 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
src/main/tools/process-wrapper-legacy.cc:80: "execvp(/usr/lib/llvm-14/bin/clang, ...)": No such file or directory
INFO: Elapsed time: 18.094s, Critical Path: 0.26s
INFO: 80 processes: 18 internal, 62 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
```

I'll investigate and see if I can fix.

asraa:
The runner's don't seem to have changed, e.g. from 
https://github.com/google/heir/actions/runs/12747567920/job/35525904809
to
https://github.com/google/heir/actions/runs/12822845640/job/35756413432

since my first thought is maybe the runners don't have llvm-14. But that doesn't seem to be the case.

asraa:
Even though the runner name is ubuntu-22.04 in the failing link above toggling the runner image seems like its ubuntu 24

https://github.com/actions/runner-images/blob/ubuntu24/20250105.1/images/ubuntu/Ubuntu2404-Readme.mdhttps://github.com/actions/runner-images/blob/ubuntu24/20250105.1/images/ubuntu/Ubuntu2404-Readme.md

Operating System
  Ubuntu
  24.04.1
  LTS
Runner Image
  Image: ubuntu-24.04
  Version: 20250[10](https://github.com/google/heir/actions/runs/12822845640/job/35756413432#step:1:11)5.1.0
[...]

PR #1276, opened 2025-01-17
ci: update runner images to 24.04

asraa:
Test for https://github.com/google/heir/issues/1275


j2kun (approved):

.log
j2kun, line 1: Is this file needd?

asraa: removed, on a side i commented on the actions issue that the 24.04 runners have a long queue - maybe i should just add the runner label to the cache key?

Issue #465, opened 2024-02-27
Find a way to connect HEIR parameters to openFHE parameters


j2kun:
Right now when `heir-translate` emits code for OpenFHE and tfhe-rs, the parameters expressed in the HEIR IR are ignored in the generated code.

E.g., in [binops.mlir](
https://github.com/google/heir/blob/main/tests/openfhe/end_to_end/binops.mlir), I have set arbitrarily

```
#encoding = #lwe.polynomial_evaluation_encoding<cleartext_start=30, cleartext_bitwidth=3>
#params = #lwe.rlwe_params<cmod=7917, dimension=1, polyDegree=16384>
```

And the generated code has none of these values. Instead the test writer must manually set parameters as in [binops_test.cpp](https://github.com/google/heir/blob/main/tests/openfhe/end_to_end/binops_test.cpp#L27-L32).

For tests this is merely annoying, but for real production use cases we will need a way for the compiler to emit code that the client and server can use to configure their "CryptoContext" objects (or similar for other library backends) so that they use the right parameters.

j2kun:
Another example specific to OpenFHE, we hard-code `Ciphertext<DCRTPoly>` regardless of what encoding the mlir specifies.

Maokami:
One concern I have regarding this issue is how to safely set the HEIR parameter when it cannot be represented through a library API.

For example, in OpenFHE, the ciphertext moduli (referred to as cmod in rlwe_params) are internally and automatically calculated based on various parameters. As far as I know, it may not be feasible to explicitly set them to the 'cmod' values specified by HEIR.

In this case, should we simply ignore the cmod and proceed with code generation, or might there be a more suitable solution?

j2kun:
I think in the very long term we would bypass the libraries. In the medium term we could work with the OpenFHE developers to support overriding their calculations (perhaps by exposing a lower level API), and in the short term, yes I think we would just ignore our manual calculations and hope for the best.

j2kun:
Also, I have heard that other backends like lattigo support more fine-tuned parameter settings, so choosing a different backend is also a workaround.

j2kun:
I recently discussed with the OpenFHE devs and @ZenithalHourlyRate, and the OpenFHE folks seem open to this, if we wanted to do the work (with guidance) to update their API.

Meanwhile, we're adding a Lattigo backend which supports custom parameter settings already.

PR #1279, opened 2025-01-20
Lattigo: Add configure-crypto-context pass

ZenithalHourlyRate:
Similar to `--openfhe-configure-crypto-context`.

The helper function signature is
```mlir
func.func @my_func__configure() -> (!lattigo.bgv.evaluator, !lattigo.bgv.parameter, !lattigo.bgv.encoder, !lattigo.rlwe.encryptor, !lattigo.rlwe.decryptor)
```

This is somewhat a violation of #1145. I do not find a scheme/LWE level configure function obvious and if we go on that direction, the design of the configure function needs discussion.

### DIFF
--- a/lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
@@ -1,0 +1,161 @@
+#include "lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.h"
+
+#include "lib/Dialect/Lattigo/IR/LattigoOps.h"
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"  // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace lattigo {
+
+#define GEN_PASS_DEF_CONFIGURECRYPTOCONTEXT
+#include "lib/Dialect/Lattigo/Transforms/Passes.h.inc"
+
+// Helper function to check if the function has RelinearizeOp
+bool hasRelinOp(func::FuncOp op) {
+  bool result = false;
+  op.walk<WalkOrder::PreOrder>([&](Operation *op) {
+    if (isa<BGVRelinearizeOp>(op)) {
+      result = true;
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  return result;
+}
+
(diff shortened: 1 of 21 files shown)


lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -0,0 +1,139 @@
...
+  SmallVector<Type> keypairTypes = {skType, pkType};
+  auto keypair = builder.create<RLWEGenKeyPairOp>(keypairTypes, kgen);
+  auto sk = keypair.getResult(0);
+  auto pk = keypair.getResult(1);
+  auto encryptor =
+      builder.create<RLWENewEncryptorOp>(encryptorType, params, pk);
+  auto decryptor =
+      builder.create<RLWENewDecryptorOp>(decryptorType, params, sk);
+
+  // TODO: do not always generate Relinearization Key
```
asraa, line 89: Would you just be able to check for multiplications? Or is the issue that the ops (evalkeyset, etc) require the relinearization key?

ZenithalHourlyRate: > Or is the issue that the ops (evalkeyset, etc) require the relinearization key?

Yes currently NewEvaluationKeySetOp requires a relinearization key.

```tablegen
    Lattigo_RLWERelinearizationKey:$relinearizationKey,
    Variadic<Lattigo_RLWEGaloisKey>:$galoisKeys
```
I could not make `$relinearizationKey` optional, as MLIR would then complain

```
error: op has multiple variadic operands but no specification over their sizes
```

lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -0,0 +1,139 @@
...
+  funcResultTypes.push_back(decryptorType);
+
+  FunctionType configFuncType =
+      FunctionType::get(builder.getContext(), funcArgTypes, funcResultTypes);
+
+  auto configFuncOp =
+      builder.create<func::FuncOp>(configFuncName, configFuncType);
+  builder.setInsertionPointToEnd(configFuncOp.addEntryBlock());
+
+  // TODO: Allow custom params
```
asraa, line 59: nit: attach https://github.com/google/heir/issues/465

lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
asraa: I see, thank you! Fine with me for now, I think a trait like AttrSizedOperandSegments /may/ help (attaches an operand size attribute for each size) later

asraa (approved):

asraa (approved):
appreciate all the tests!

PR #1283, opened 2025-01-21
tensor_ext: add convert_layout op

j2kun:
This PR adds an op that represents the conversion of a ciphertext layout from one form to another, using the affine_map layout notation defined in https://docs.google.com/document/d/1TTj0NDZkn0T9NNQj6idbTk2g65tAb7VxsYpt1hKjK_Y/edit?tab=t.0#heading=h.hcusv1ynbv9n

### DIFF
--- a/lib/Dialect/TensorExt/IR/TensorExtOps.td
+++ b/lib/Dialect/TensorExt/IR/TensorExtOps.td
@@ -12,6 +12,7 @@ include "mlir/Interfaces/SideEffectInterfaces.td"
 class TensorExt_Op<string mnemonic, list<Trait> traits = []> :
         Op<TensorExt_Dialect, mnemonic, traits> {
   let cppNamespace = "::mlir::heir::tensor_ext";
+  let assemblyFormat = "operands attr-dict `:` functional-type(operands, results)";
 }
 
 def TensorExt_RotateOp : TensorExt_Op<"rotate", [Pure, AllTypesMatch<["tensor", "output"]>]> {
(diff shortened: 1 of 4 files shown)


lib/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -48,4 +49,19 @@ def TensorExt_RotateOp : TensorExt_Op<"rotate", [Pure, AllTypesMatch<["tensor",
...
+  let summary = "Convert from one layout to another.";
+  let description = [{
+    This op represents the conversion of a tensor from one packed layout to
+    another. This is implemented via a "shift network" of ciphertext rotations,
+    plaintext masks (ciphertext-plaintext multiplications), and additions.
+
+    This op is inserted by layout selection passes.
+  }];
+  let assemblyFormat = "operands attr-dict `:` type($output)";
+  let arguments = (ins AnyTensor:$tensor, Builtin_AffineMapAttr:$from_layout, Builtin_AffineMapAttr:$to_layout);
```
asraa, line 62:
```suggestion
  let arguments = (ins AnyRankedTensor:$tensor, Builtin_AffineMapAttr:$from_layout, Builtin_AffineMapAttr:$to_layout);
```

since it's assumed to be so (in the verifier as well)

lib/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -48,4 +49,19 @@ def TensorExt_RotateOp : TensorExt_Op<"rotate", [Pure, AllTypesMatch<["tensor",
...
+  let results = (outs AnyTensor:$output);
```
asraa, line 63:
```suggestion
  let results = (outs AnyRankedTensor:$output);
```

asraa (approved):

PR #1267, opened 2025-01-15
Arith-to-cggi-quart pass lowering to TFHE-rs HL emitter

WoutLegiest:
Follow up of #1265 
- Change the Scalar Right Shift ops definition of CGGI and TFHE-rs to use an index attr to define the shift amount
- Extended some TFHE-rs ops to allow vectors as operands
- Further integration of the arith-to-cggi-quart pass lowering to TFHE-rs HL emitter

PR #1285, opened 2025-01-22
Adding ptxt-ctxt support to arith-to-cggi and cggi ops

WoutLegiest:
Follow-up of #1267 
Making it possible to add/mul a plaintext to a ciphertext in CGGI dialect. 
arith-to-cggi conversion pass update

### DIFF
--- a/lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
+++ b/lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
@@ -22,7 +22,6 @@ static lwe::LWECiphertextType convertArithToCGGIType(IntegerType type,
                                      lwe::UnspecifiedBitFieldEncodingAttr::get(
                                          ctx, type.getIntOrFloatBitWidth()),
                                      lwe::LWEParamsAttr());
-  ;
 }
 
 static Type convertArithLikeToCGGIType(ShapedType type, MLIRContext *ctx) {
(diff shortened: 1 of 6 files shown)


lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -186,6 +185,43 @@ struct ConvertShRUIOp : public OpConversionPattern<mlir::arith::ShRUIOp> {
...
+
+  using OpConversionPattern<SourceArithOp>::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      SourceArithOp op, typename SourceArithOp::Adaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    if (auto lhsDefOp = op.getLhs().getDefiningOp()) {
+      if (isa<mlir::arith::ConstantOp>(lhsDefOp)) {
```
j2kun, line 201: Just a note: in the pipelines you may want to include a `-sccp` pass to ensure statically-inferrable constants are actually folded down to constant ops, or this may match on fewer things.

j2kun (approved):

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
WoutLegiest: Thanks! I'll later will define an pipeline again

Issue #919, opened 2024-08-19
Determine what parts of Vos-Vos-Erkin 22 can be ported to HEIR


j2kun:
Efficient Circuits for Permuting and Mapping Packed Values Across Leveled Homomorphic Ciphertexts.
Jelle Vos, Daniël Vos, and Zekeriya Erkin. 2022.
http://dx.doi.org/10.1007/978-3-031-17140-6_20

This would be an improvement over the baseline in https://github.com/google/heir/issues/914

j2kun:
I read through this one and it's very straightforward and well written.

They decompose each shift required into its constituent bits and shift by the corresponding powers of 2 (e.g., a shift of 5 is 101 requires a shift by 1 and by 4). Up to an exception mentioned later, they always do the rotations in order of bits (shift by 1, then by 2, then by 4, etc.). This can cause conflicts if two intermediate shifted values end up colliding, adding together two values that shouldn't be added together. To avoid this, they construct a graph coloring problem by adding edges when there would be such a conflict, color the graph with the minimal possible colors (using a standard solver) and each color corresponds to a separate copy of the total set of rotations. E.g., three colors means three sets of all log(N) rotations. Then they can run the different sets in parallel. Some minor simplifications are included, like skipping a rotation when all assigned values have that bit zero.

This is still a logarithmic-size construction, like Benes networks, but they claim empirically it does much better for small permutations than HElib's log-size construction.

They have a C++ implementation here: https://github.com/jellevos/perm_map_circuits

j2kun:
I also think it's not necessarily better than the baseline approach in all cases, so we'd want to do both and pick the best for each input program.

j2kun:
I added an implementation of the basic graph coloring approach in https://github.com/j2kun/rotate-solver/blob/main/vos_vos_erkin.py (though I only added one simple test, may still be buggy). I computed the graph to partition the rotation indices, and the rest of the needed code is quite simple: convert a list of index groups to an set of iterated mask+shift+add operations and a final addition.

j2kun:
I also wrote a short blog post explaining the algorithm https://www.jeremykun.com/2024/09/02/shift-networks/

j2kun:
Going to start porting this

PR #1286, opened 2025-01-22
Lower tensor_ext.permute via Vos-Vos-Erkin shift networks

j2kun:
Fixes https://github.com/google/heir/issues/919

This PR implements the Vos-Vos-Erkin shift network technique described in https://doi.org/10.1007/978-3-031-17140-6_20

In support of this, this PR adds:

- A new FrozenVector data type to represent an explicit permutation and use it as a hash map key.
- New utility functions for working with permutations; in particular expanding an affine map to an explicit permutation
- A new op `tensor_ext.permute` to represent the input to the pass. I expect the layout materialization pass to lower `convert_layout` to permute.
- A bug fix in the graph coloring routine.

Implementation notes:

- This PR only supports the baseline "power of two" shifting strategy, however, the strategy itself is extracted into its own class so that later work can generalize. Some ways I can see generalizing include:
   - Trying other bases like negative powers of two (right rotations instead of left rotations).
   - Trying random reorderings of the basis elements (e.g., shift first by 4 then by 16 then by 1 then by...) which can produce a graph with fewer conflicts and hence fewer colors.
   - Evaluate many such strategies and use the one that produces the smallest number of rotation groups.
- This requires writing out the permutation explicitly in memory, which for large ciphertexts (N = 65536) means each `tensor_ext.permute` lowering will require ~0.5 MiB of RAM. If we had ~1k of these ops to lower, it would take ~1 GiB of RAM. This seems OK to me, since if we are doing 1k layout conversions we probably have bigger problems.

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
+++ b/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
@@ -1,0 +1,390 @@
+#include "lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h"
+
+#include <utility>
+
+#include "lib/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "lib/Utils/ADT/FrozenVector.h"
+#include "lib/Utils/AffineMapUtils.h"
+#include "lib/Utils/Graph/Graph.h"
+#include "llvm/include/llvm/Support/Debug.h"            // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"   // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinAttributes.h"     // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypes.h"          // from @llvm-project
+#include "mlir/include/mlir/IR/Diagnostics.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"          // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"             // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"    // from @llvm-project
+
+#define DEBUG_TYPE "implement-shift-network"
+
+namespace mlir {
+namespace heir {
+namespace tensor_ext {
+
+#define GEN_PASS_DEF_IMPLEMENTSHIFTNETWORK
+#include "lib/Dialect/TensorExt/Transforms/Passes.h.inc"
+
+// A permutation of 0..n-1. This vector should always have size n and contain
+// each integer from 0 to n-1 exactly once.
+using Permutation = FrozenVector<int64_t>;
+
+// A group of indices to rotate together
+using RotationGroup = DenseSet<int64_t>;
+
+// Convert an input->output index mapping to a canonical left-shift amount for
+// a given tensor size.
+// Example: 1 -> 13 with a 64-size tensor should produce a rotation of 52
+// Example: 13 -> 1 with a 64-size tensor should produce a rotation of 12
+inline int64_t normalizeShift(int64_t input, int64_t output,
+                              int64_t tensorSize) {
+  int64_t shift = (output - input) % tensorSize;
+  shift = -shift;  // Account for leftward rotations
+  if (shift < 0) {
+    shift += tensorSize;
+  }
+  return shift;
+}
+
+// The ShiftStrategy class applies power-of-two shifts to each set bit in
+// LSB-to-MSB order, 1, 2, 4, 8, .... Each shift amount is considered a "round"
+// in which a group of indices are attempted to be shifted together. This can be
+// used both to identify conflicts for the graph coloring technique of
+// Vos-Vos-Erkin, and also to construct the concrete shift network after a
+// partition has been decided by Vos-Vos-Erkin.
+struct ShiftRound {
+  // Maps the index of the original input to its current position in the
+  // tensor. This may contain multiple indices mapping to the same slot due to
+  // conflicts in the shifting strategy.
+  SmallVector<int64_t> positions;
+  // The set of indices that are rotated left in this round. This can be used
+  // to generate a mask to select the indices that need rotating.
+  SmallVector<int64_t> rotatedIndices;
+  // The amount rotated left;
+  int64_t rotationAmount;
+};
+
+class ShiftStrategy {
+ public:
+  ShiftStrategy() = default;
+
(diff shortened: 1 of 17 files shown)


j2kun:
CC @jellevos for comments on the implementation or ideas for extension.

lib/Utils/AffineMapUtils.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace mlir {
+namespace heir {
+
+// Construct an identity permutation of size n
+inline SmallVector<int64_t> identity(int64_t n) {
+  SmallVector<int64_t> permutation;
+  for (int64_t i = 0; i < n; i++) {
+    permutation.push_back(i);
+  }
```
asraa, line 15: i learned about std::iota recently which you could use here but i don't mind because clear code is clear.

lib/Dialect/TensorExt/Transforms/Passes.td
```diff
@@ -157,4 +157,114 @@ def AlignTensorSizes : Pass<"align-tensor-sizes"> {
...
+  }
+  ```
+  }];
+
+  let dependentDialects = ["mlir::heir::tensor_ext::TensorExtDialect"];
+
+  let options = [
+    Option<
+      "ciphertextSize",
+      "ciphertext-size",
```
asraa, line 262: ugh our naming here is diverging. could you file a placeholder issue to resolve the names ciphertext-degree, tiling-size, and ciphertext-size?

lib/Utils/AffineMapUtils.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+  OpBuilder b(map.getContext());
+  result.resize(rank);
+  for (size_t permInput = 0; permInput < rank; ++permInput) {
+    SmallVector<Attribute> operandConstants{b.getIndexAttr(permInput)};
+    SmallVector<Attribute> results;
+    if (failed(map.constantFold(operandConstants, results))) {
+      assert(false && "constant folding should never fail here");
+      return failure();
+    }
```
asraa, line 25: so each loop iteration is calculating the constantFold at operandConstant? can you assemble the a whole list of the permInput index attrs and feed that to `constantFold` and then the result vector is exactly the result (assuming you `map_vector` from attribute to int64_t after)

tests/Dialect/TensorExt/Transforms/implement_shift_network_example.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: heir-opt --implement-shift-network=ciphertext-size=16 %s | FileCheck %s
```
asraa, line 1: what's being checked here beyond validating the pass runs?

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+  LLVM_DEBUG({
+    llvm::dbgs() << "PermuteOp produces underlying permutation: ";
+    for (int i = 0; i < permutation.size(); i++) {
+      llvm::dbgs() << i << " -> " << permutation[i] << ", ";
+      if (i % 10 == 9) {
+        llvm::dbgs() << "\n";
+      }
+    }
+    llvm::dbgs() << "\n";
```
asraa, line 359: maybe you can have a printer for the permutation to reuse here and above?

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -0,0 +1,409 @@
...
+// Create a tensor with zeros everywhere except for the indices specified in
+// the input `indices` vector.
+Value createMask(TypedValue<RankedTensorType> tensor,
+                 const SmallVector<int64_t> &indices, IRRewriter &rewriter) {
+  auto elementType = tensor.getType().getElementType();
+  SmallVector<Attribute> maskAttrs;
+
+  for (int64_t i = 0; i < tensor.getType().getDimSize(0); i++) {
+    maskAttrs.push_back(rewriter.getIntegerAttr(elementType, 0));
+  }
```
asraa, line 246:
```    
SmallVector<Attribute> maskAttrs(tensor.getType().getDimSize(0), rewriter.getIntegerAttr(elementType, 0));
```

you should just be able to replace with

asraa (approved):
thanks! i admit i have actually glossed over the actual algorithm details for correctness in this review.

lib/Utils/AffineMapUtils.h
j2kun: TIL!

lib/Utils/AffineMapUtils.cpp
j2kun: `constantFold` here is just evaluating the affine map at a given input. I changed it to be more functional, but I'm not convinced it's significantly better.

tests/Dialect/TensorExt/Transforms/implement_shift_network_example.mlir
j2kun: Nothing. This was the example I put in the documentation, so I felt it should run, but I didn't have the steam to add a detailed list of assertions like the other test.

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    LLVM_DEBUG(llvm::dbgs()
+               << "Expanding permutation from " << permutationMap << "\n");
+    if (failed(makeExplicit1DMapping(permutationMap, tensorTy.getNumElements(),
+                                     permutation)))
+      return failure();
+  } else {
+    auto denseElementsAttr =
+        dyn_cast<DenseIntElementsAttr>(op.getPermutation());
```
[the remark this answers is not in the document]
j2kun: I think actually I can remove the check for identity. If the permutation is the identity then the result of this pass should not produce any rotations at all. If we want to fix it later, we can add its as a canonicalization pattern/folder independent of this pass.

lib/Dialect/TensorExt/Transforms/Passes.td
j2kun: Done

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
j2kun: Done

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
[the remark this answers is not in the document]
j2kun: Added an extra test for this

asraa (approved):

PR #1287, opened 2025-01-22
Adding ctxt-ptxt to TFHE rs dialect + Conversion cggi-to-tfhe-rust

WoutLegiest:
Follow up of #1285

### DIFF
--- a/lib/Dialect/TfheRust/IR/TfheRustOps.td
+++ b/lib/Dialect/TfheRust/IR/TfheRustOps.td
@@ -6,6 +6,7 @@ include "TfheRustTypes.td"
 
 include "mlir/IR/BuiltinAttributes.td"
 include "mlir/IR/CommonTypeConstraints.td"
+include "mlir/IR/BuiltinTypes.td"
 include "mlir/IR/OpBase.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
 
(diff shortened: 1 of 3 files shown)


lib/Dialect/TfheRust/IR/TfheRustOps.td
```diff
@@ -30,16 +32,43 @@ class TfheRust_BinaryOp<string mnemonic>
...
+class TfheRust_ScalarBinaryOp<string mnemonic>
+  : TfheRust_Op<mnemonic, [
+    Pure,
+    Commutative,
+    ElementwiseMappable,
+]> {
+  let arguments = (ins
+    TfheRust_ServerKey:$serverKey,
+    TfheRust_CiphertextType:$lhs,
+    AnyTypeOf<[Builtin_Integer, TfheRust_CiphertextType]>:$rhs
```
j2kun, line 44: If you want it to be `ElementwiseMappable`, these and the output type should be the `Like` variants that allow tensors. Maybe simpler to remove `ElementwiseMappable` for now

Same below for SubOp

lib/Target/TfheRustHL/TfheRustHLEmitter.cpp
```diff
@@ -573,6 +573,7 @@ LogicalResult TfheRustHLEmitter::printOperation(tensor::FromElementsOp op) {
   return success();
 }
 
+// Need to produce a
```
j2kun, line 576: Stray comment

lib/Utils/ConversionUtils.h
```diff
@@ -35,6 +35,8 @@
 #include "mlir/include/mlir/Support/LogicalResult.h"  // from @llvm-project
 #include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
 
+#define DEBUG_TYPE "cggi-to-tfhe-rust"
```
j2kun, line 38:
```suggestion
```

j2kun (approved):
Approved with minor nitpicks.

PR #1272, opened 2025-01-16
Tosa compatibility to TFHE-rs HL emitter

WoutLegiest:
Follow up of #1267 
Smaller tweaks and additions to have an initial full TOSA-to-tfhe-rs pipeline. 
Note that this pipeline only supports the `tests/Transforms/tosa_to_boolean_tfhe/hello_world_clean.mlir` test right now.

PR #1288, opened 2025-01-22
Ctxt - Ptxt HL Emitter

WoutLegiest:
Follow up of #1272, #1287


j2kun:
@WoutLegiest looks like encrypt_trivial needs a cast

```
# | error[E0277]: the trait bound `i32: UnsignedNumeric` is not satisfied
# |   --> src/fn_under_test.rs:12:12
# |    |
# | 12 |   let v3 = FheUint32::try_encrypt_trivial(v2).unwrap();
# |    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `UnsignedNumeric` is not implemented for `i32`
# |    |
# |    = help: the following other types implement trait `UnsignedNumeric`:
# |              u128
# |              u16
# |              u32
# |              u64
# |              u8
# |              usize
# |    = note: required for `FheUint<FheUint32Id>` to implement `tfhe::prelude::FheTryTrivialEncrypt<i32>`
```

lib/Dialect/CGGI/Conversions/CGGIToTfheRust/CGGIToTfheRust.cpp
```diff
@@ -175,9 +175,12 @@ struct ConvertLut3Op : public OpConversionPattern<cggi::Lut3Op> {
         serverKey, adaptor.getC(), b.getIndexAttr(2));
     auto shiftedB = b.create<tfhe_rust::ScalarLeftShiftOp>(
         serverKey, adaptor.getB(), b.getIndexAttr(1));
-    auto summedBC = b.create<tfhe_rust::AddOp>(serverKey, shiftedC, shiftedB);
-    auto summedABC =
-        b.create<tfhe_rust::AddOp>(serverKey, summedBC, adaptor.getA());
+    auto outputType =
+        getTypeConverter()->convertType(shiftedB.getResult().getType());
```
asraa, line 179: interesting, just curious - why is this not the same type as `adaptor.getB()` (even though the all types match constraint is removed, should this whole sequence of ops be the same type as adaptor.getC() / getB() / getA())?

lib/Dialect/CGGI/Conversions/CGGIToTfheRust/CGGIToTfheRust.cpp
```diff
@@ -248,9 +257,11 @@ struct ConvertCGGITRBinOp : public OpConversionPattern<BinOp> {
     if (failed(result)) return result;
 
     Value serverKey = result.value();
-
-    rewriter.replaceOp(op, b.create<TfheRustBinOp>(serverKey, adaptor.getLhs(),
-                                                   adaptor.getRhs()));
+    CGGIToTfheRustTypeConverter typeConverter(op->getContext());
```
asraa, line 260: I think you should just be able to use the class member here (`getTypeConverter()`)?

lib/Dialect/TfheRust/IR/TfheRustOps.td
```diff
@@ -30,16 +32,43 @@ class TfheRust_BinaryOp<string mnemonic>
...
+    Pure,
+    Commutative,
+    ElementwiseMappable,
+]> {
+  let arguments = (ins
+    TfheRust_ServerKey:$serverKey,
+    TfheRust_CiphertextType:$lhs,
+    AnyTypeOf<[Builtin_Integer, TfheRust_CiphertextType]>:$rhs
+  );
+  let results = (outs TfheRust_CiphertextType:$output);
```
asraa, line 46: Do the lhs and output types have to match?

lib/Target/TfheRustHL/TfheRustHLEmitter.cpp
```diff
@@ -359,7 +359,12 @@ LogicalResult TfheRustHLEmitter::printOperation(arith::ConstantOp op) {
   // By default, it emits an unsigned integer.
   emitAssignPrefix(op.getResult());
   if (auto intAttr = dyn_cast<IntegerAttr>(valueAttr)) {
-    os << intAttr.getValue().abs() << "u64;\n";
+    os << intAttr.getValue().abs();
+    if (isa<IndexType>(op.getType())) {
+      os << "usize;\n";
```
asraa, line 364: can you add IndexType to the type switch in convertType and remove this branch?

https://github.com/google/heir/blob/d9379845e0b0aee3980775823ff4b44926706855/lib/Target/TfheRustHL/TfheRustHLEmitter.cpp#L655

lib/Target/TfheRustHL/TfheRustHLEmitter.cpp
```diff
@@ -383,6 +388,15 @@ LogicalResult TfheRustHLEmitter::printBinaryOp(::mlir::Value result,
                                                std::string_view op) {
   emitAssignPrefix(result);
 
+  if (auto cteOp = dyn_cast<mlir::arith::ConstantOp>(rhs.getDefiningOp())) {
+    auto intValue =
+        cast<IntegerAttr>(cteOp.getValue()).getValue().getZExtValue();
+    os << checkOrigin(lhs) << variableNames->getNameForValue(lhs) << " " << op
+       << " " << intValue << "u" << cteOp.getType().getIntOrFloatBitWidth()
+       << ";\n";
```
asraa, line 396:
```suggestion
    // Note: arith.constant op requires signless integer types, but here we manually emit an unsigned integer type.
    os << checkOrigin(lhs) << variableNames->getNameForValue(lhs) << " " << op
       << " " << intValue << "u" << cteOp.getType().getIntOrFloatBitWidth()
       << ";\n";
```

since that's why we can't use convertType, I think!

lib/Utils/ConversionUtils.h
```diff
@@ -35,6 +35,8 @@
 #include "mlir/include/mlir/Support/LogicalResult.h"  // from @llvm-project
 #include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
 
+#define DEBUG_TYPE "cggi-to-tfhe-rust"
```
asraa, line 38: hmmm - is there something more generic we can put here / remove?

lib/Dialect/CGGI/Conversions/CGGIToTfheRust/CGGIToTfheRust.cpp
WoutLegiest: Because it is a template function, the getTypeConverter does not directly works

lib/Utils/ConversionUtils.h
WoutLegiest: Yeah, snook back in during rebase/squash

lib/Target/TfheRustHL/TfheRustHLEmitter.cpp
WoutLegiest: Thanks, cool found!

asraa (approved):
Thank you!

Issue #1069, opened 2024-10-23
Reduce reliane on main "canonicalize" pattern


AlexanderViand-Intel:
![Image](https://github.com/user-attachments/assets/ae9cfef8-704e-47dd-a4b0-bbe25d5ed0f9)

I think the correct thing here is to 

- where we rely on specific canonicalization patterns as an integral part of the pass pipeline, put those patterns in a specific name passed, so that if they ever get removed/etc upstream, we don't get sudden failures without warning)
- Verify that all pipelines still produce code succesfully if the canonicalization is removed/replaced with no-op
     - how to efficiently set all canonicalize passes to no-op? (Maybe overwrite the definition of addCanonicalizerPass to be no-op?)
     - how to efficiently check that the output is still "correct" even though it's now syntactically different (i.e., FileCheck tests will fail?)

j2kun:
The only place I know that we do this involves HEIR-owned patterns being added to the canonicalizer (for the HECO-ported pipeline). We could easily extract those into their own pass. This could also fail if patterns removed from the canonicalizer result in code that we haven't code-genned things for yet (e.g., a leftover `tensor.extract` that isn't yet code-genned as a manual mask+rotate)

For other cases, the proposed solution (an automated test to check if they work when canonicalize is replaced by a no-op) would be ideal, but in my opinion a relatively low priority considering everything else we have to work on. By the same token, should we check that the pipelines still work when all folders are replaced by no-ops (in every single application of the greedy rewrite engine)?

If we had to do it, I'd say the best approach would be to add a CLI flag for `no-op-canonicalize`, replace the `heir-opt.cpp` invocations of `createCanonicalizerPass` with a helper function that branches on "no-op" vs actually creating the canonicalizer pass, and then be deliberate about the chosen end-to-end pipeline tests to replicate with the flag enabled. Mainly the tests whose assertions only involve running the final code and asserting an expected output.

PR #1293, opened 2025-01-22
Prettyify and extend `--annotate-secretness`

AlexanderViand-Intel:
While showing HEIR to some colleagues at Intel, they wanted to see the type inference results and were confused by the fact that function return types and terminator operations (secret.yield, func.return, etc) did not have secretness annotations, making them appear to be public.

The current pass was mostly designed as an internal debugging aid for the secretness analysis, but since it seems like it'd also be a very helpful tool for users of HEIR to debug/inspect their programs, I extended and prettified the annotations a bit.

* The annotations now use `{secret.secret}` and `{public}` instead of `{secretness = true}`/`{secretness = false}
* Function results are now annotated in the function signature
* Terminators are now treated as if their operands were results and are therefore annotated
* Multi-result operations are now annotated with `{secretness = [{secret.secret}, {public}]} : i32, i32` rather than `{result_0_secretness = true, result_1_secretness = false} : i32, i32` 

Here's what the test file now looks like once annotated:

```llvm
module {
  func.func @annotated_secretness(%arg0: i32 {secret.secret}, %arg1: i32) {
    %0 = arith.addi %arg1, %arg1 {public} : i32
    %1 = arith.addi %arg0, %arg1 {secret.secret} : i32
    return
  }
  func.func @typed_secretness(%arg0: !secret.secret<i32>, %arg1: i32) {
    %0 = secret.generic ins(%arg0, %arg1 : !secret.secret<i32>, i32) {
    ^body(%input0: i32, %input1: i32):
      %1 = arith.addi %input1, %input1 {public} : i32
      %2 = arith.addi %input0, %input1 {secret.secret} : i32
      secret.yield %2 {secret.secret} : i32
    } -> !secret.secret<i32>
    return
  }
  func.func @multi_result_secretness(%arg0: i32 {secret.secret}, %arg1: i32) {
    %sum, %overflow = arith.addui_extended %arg1, %arg1 {secretness = [{public}, {public}]} : i32, i1
    %sum_0, %overflow_1 = arith.addui_extended %arg0, %arg1 {secretness = [{secret.secret}, {secret.secret}]} : i32, i1
    return
  }
  func.func @return_secretness(%arg0: i32 {secret.secret}, %arg1: i32) -> (i32 {secret.secret}) {
    %0 = arith.addi %arg0, %arg1 {secret.secret} : i32
    return {secret.secret} %0 : i32
  }
}
````

### CODE REVIEW

**Comment by ZenithalHourlyRate** [comment]
> {secret.secret} and {public}

It is kind of asymmetric.

In terms of readability, if you annotate immediate result with `{public}`, the public func argument should also be annotated.

**Comment by j2kun** [comment]
> Terminators are now treated as if their operands were results and are therefore annotated

secret.yield produces something secret by definition, no? Wouldn't we instead want to annotate the results of the `generic` op as secret instead of the operands of the yield? I did a similar thing in my draft tensor layout PR. Perhaps we can combine it as "setResultAttr" for `secret.generic`, similar to func.func.

**Comment by AlexanderViand-Intel** [comment]
> > {secret.secret} and {public}
> 
> It is kind of asymmetric.

Yeah, and I'm also not sure it's the right term, since "not secret" does not necessarily mean "public" (a plaintext value can still be something private to the server). I thought about `{non-secret}` or even just not annotating anything for non-secret values (though that might be confusing if more passes are applied after annotation).

> In terms of readability, if you annotate immediate result with `{public}`, the public func argument should also be annotated.
Indeed, I spotted that just after creating the PR 🙈 

I'm actually thinking about introducing a pass option "explicit" which defaults to false, in which case "non-secret" things aren't annotated. If set to true, non-secret things would be annotated as `{non-secret}` or similar.

**Comment by AlexanderViand-Intel** [comment]
> secret.yield produces something secret by definition, no? 
In canonicalized code, yes - but (afaik) in general code it might return both secret and non-secret values.

> Perhaps we can combine it as "setResultAttr" for `secret.generic`, similar to func.func.
I think that'd be great! One of the reasons I currently explicitly exclude `secret.generic` from annotation because it looked so bad with default attribute printing (the other being that it's generally already pretty explicit about its types).

**Comment by asraa** [comment]
> In canonicalized code, yes - but (afaik) in general code it might return both secret and non-secret values.

I think you may be confusing the args / returns of the generic. The args of uncanonicalized generic may be secret or non-secret (used so that the generic captures its enclosing scope), but the yielded values that correspond to the return types must always be secret: https://github.com/google/heir/blob/96c9775d9ce5e486600d7b0156570fc6a035e3ab/lib/Dialect/Secret/IR/SecretOps.cpp#L64

**Comment by AlexanderViand-Intel** [comment]
> I think you may be confusing the args / returns of the generic. The args of uncanonicalized generic may be secret or non-secret (used so that the generic captures its enclosing scope), but the yielded values that correspond to the return types must always be secret:

Oh, I see! I think my intuition was expecting something like
```llvm
func.func @foo(%s : i16 {secret.secret}, %p : i16 ) -> (i16,i 16)  {
   %0 = arith.addi %s, %p : i16
   %1 = arith.addi %p, %p : i16
   return %0, %1 : i16, i16
}
````
to get `--wrap-generic`'d into 
```llvm
func.func @foo(%s : !secret.secret<i16>, %p : i16 ) -> (i16,i16) {
  %0:2 = secret.generic ins(%s, %p : !secret.secret<i16>, i16) {
    ^body(%ss: i16, %pp: i16):
      %a = arith.addi %ss, %pp : i16
      %b = arith.addi %pp, %pp : i16
      secret.yield %a, %b  : i16, i16
    } -> (!secret.secret<i16>, i16)
 return %0#0, %0#1 : !secret.secret<i16>, i16
}
```

When, in reality, it first gets `--wrap-generic`'d into this (imho, strictly speaking *incorrectly* typed) form:
```llvm
func.func @foo(%s: !secret.secret<i16>, %p: i16) -> (!secret.secret<i16>, !secret.secret<i16>) {
    %0:2 = secret.generic ins(%s, %p : !secret.secret<i16>, i16) {
    ^body(%ss: i16, %pp: i16):
        %1 = arith.addi %ss, %pp : i16
        %2 = arith.addi %ss, %pp : i16
        secret.yield %1, %2 : i16, i16
    } -> (!secret.secret<i16>, !secret.secret<i16>)
    return %0#0, %0#1 : !secret.secret<i16>, !secret.secret<i16>
}
```
which is only corrected after canonicalization (Side note: I think this means I found another case of #1069)
```llvm
func.func @foo(%s: !secret.secret<i16>, %p: i16) -> (!secret.secret<i16>, i16) {
    %0 = arith.addi %p, %p : i16
    %1 = secret.generic ins(%s : !secret.secret<i16>) {
    ^body(%ss: i16):
        %2 = arith.addi %ss, %p : i16
        secret.yield %2 : i16
    } -> !secret.secret<i16>
    return %1, %0 : !secret.secret<i16>, i16
}
````

Either way, I think even with the annotation being technically redundant with the current `secret.generic` semantics, it might make sense to keep it since not everyone (apparently not even everyone working on HEIR 🙈 ) will know that `secret.yield` always implies the result is secret.

**Comment by j2kun** [comment]
Beyond people remembering or not remembering stuff, the code will just be cleaner if the effect of the pass is uniform: _all_ ops get their results annotated, and _all_ block arguments get annotated (this is the part that isn't consistent IIUC), then we won't have to do as much type switching for special cases.

File: lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp [hunk 1]
```diff
@@ -109,26 +109,85 @@ LogicalResult SecretnessAnalysis::visitOperation(
...
+        // arguments and results, so we must use the {secret.secret}
+        // style:
+        if (secretnessLattice &&
+            secretnessLattice->getValue().isInitialized() &&
+            secretnessLattice->getValue().getSecretness()) {
+          func.setResultAttr(
+              i,
+              StringAttr::get(func->getContext(),
+                              secret::SecretDialect::kArgSecretAttrName),
+              UnitAttr::get(func->getContext()));
```
j2kun, line 158:
```suggestion
          func.setResultAttr(
              i,
              secret::SecretDialect::kArgSecretAttrName,
              UnitAttr::get(func->getContext()));
```

setResultAttr has an overload that accepts a StringRef, so you shouldn't need to wrap it in a StringAttr.

https://github.com/llvm/llvm-project/blob/a976036a100b7dd459b6cabac96159875fcd513d/mlir/include/mlir/Interfaces/FunctionInterfaces.td#L530

lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
```diff
@@ -109,26 +109,85 @@ LogicalResult SecretnessAnalysis::visitOperation(
...
-      if (!secretnessLattice->getValue().isInitialized()) {
-        op->setAttr(name, StringAttr::get(op->getContext(), ("unknown")));
-        return;
+    } else if (!llvm::isa<secret::GenericOp>(op)) {
+      // We do not need to add info to secret.generic op returns,
+      // as they already are explicitly typed as !secret.secret<..>
+      SmallVector<NamedAttribute, 1> attributes;
+      if (op->hasTrait<OpTrait::IsTerminator>()) {
+        // Terminators (e.g., func.return, affine.yield) do not have mlir
+        // op results, but do still have logical "results" (mlir operands)
```
j2kun, line 167:
```suggestion
        // Terminators don't have results, but it is helpful for debugging
        // to show an annotation here. Program analysis will generally not
        // read this, but instead query the op that creates an SSA value
        // for its secretness attribute.
```

IMO Since this is duplicate information that is not primarily intended to be be used by HEIR passes, this comment should reflect that.

j2kun: More generally, I think we should put sufficient documentation to express that this entire annotation routine is meant for debugging. I would prefer we don't have later passes that rely on the annotations, rather than relying on (freshly computed) dataflow analyses. In which case my suggested change to this comment is unnecessary.

j2kun (changes requested):

lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.cpp
AlexanderViand-Intel: How about this docstring for the pass
>  Debugging helper that runs the secretness analysis and annotates the IR with the results,
>  extending the `{secret.secret}` annotation to all operation results that are secret. 
>
>  In addition to annotating operation results, the pass also annotates arguments
>  and return types in `func.func` operations, as well as any terminators (e.g. `return`)
> 
>  In `verbose` mode, all results are annotated, including public ones with `{secret.public}`,
>  and values for which the secretness analysis is missing are annotated with `{secret.missing}`,
>  while values where the secretness analysis is inconclusive are annotated with `{secret.unknown}`.

I'd obviously adjust the pass behavior to match

AlexanderViand-Intel: Oh, and should the pass also annotate a function arg in the case its type is already `!secret.secret<T>`? Seems a bit redundant.

j2kun: In the case of a pure debugging tool, I don't think it matters. If there were downstream automation relying on the annotations, having a uniform interface would be best. (so no, don't need it)

AlexanderViand-Intel:
Thanks for all the feedback, I think both the implementation and the resulting annotations look a lot nicer now!

j2kun (approved):

Issue #1284, opened 2025-01-21
`operation-rebalancer` pass modifies IR so as to require additional relinearization ops


AlexanderViand-Intel:
The optimizer generally does a good job of pulling relins through adds, but I found a very odd case here where it will work on `(p*x) + ((x*x)+(y*y))` but not on `((x*x)+(y*y)) + (p*x)`

```llvm
// heir-opt --mlir-to-secret-arithmetic --secret-insert-mgmt-bgv --optimize-relinearization
func.func @bar(%x: i16 {secret.secret}, %y: i16 {secret.secret}, %p: i16) -> (i16) {
    %xx = arith.muli %x, %x : i16
    %yy = arith.muli %y, %y : i16
    %0 = arith.addi %xx, %yy : i16
    %xp = arith.muli %x, %p : i16
    %1 = arith.addi %0, %xp : i16
    func.return %1 : i16
}
///produces a single relin
```

```llvm
// heir-opt --mlir-to-secret-arithmetic --secret-insert-mgmt-bgv --optimize-relinearization
func.func @bar(%x: i16 {secret.secret}, %y: i16 {secret.secret}, %p: i16) -> (i16) {
    %xx = arith.muli %x, %x : i16
    %yy = arith.muli %y, %y : i16
    %0 = arith.addi %xx, %yy : i16
    %xp = arith.muli %x, %p : i16
    %1 = arith.addi %xp, %0 : i16
    func.return %1 : i16
}
///produces two relins, despite the only change being the order of operands for %1
```

j2kun:
Repro'd. Here's the ILP printed in debug mode: https://gist.github.com/j2kun/5ddf0abe4673014e716e37d59611c743

j2kun:
I realized while studying the ILP that this is actually the fault of the operation rebalancer pass. It transforms the IR from

```
      %1 = arith.muli %input0, %input0 {secretness = true} : i16
      %2 = arith.muli %input1, %input1 {secretness = true} : i16
      %3 = arith.addi %1, %2 {secretness = true} : i16
      %4 = arith.muli %input0, %arg2 {secretness = true} : i16
      %5 = arith.addi %4, %3 {secretness = true} : i16
```

To 

```
      %1 = arith.muli %input0, %input0 {secretness = true} : i16
      %2 = arith.muli %input1, %input1 {secretness = true} : i16
      %3 = arith.muli %input0, %arg2 {secretness = true} : i16
      %4 = arith.addi %3, %1 : i16
[...]

Issue #1018, opened 2024-10-08
Add costs for RLWE operations performed in over a large key basis degree


j2kun:
Pre-filing issue for https://github.com/google/heir/pull/1016

The optimization problem in that PR just minimizes the total number of relin ops (with a hard-bound on the maximum key basis size allowed in the program). We could relax this by having a cost per (operation, degree) tuple, and charging the solver for doing ops with a larger key basis. It's not clear how much this could improve efficiency/noise growth, but some papers (like Blatt-Gusev-Polyakov-Rohloff-Vaikuntanathan 2019 https://eprint.iacr.org/2019/223) mention that they might NEVER do relinearization.

It's also not clear to what extent various libraries we export to would support high-basis-degree ops.

But to implement this feature, we would mainly need a way to estimate costs of the ops, and the changes to the ILP would be relatively small.

Issue #1169, opened 2024-12-10
Track scaling factor in encoding attr for BGV (GHS variant) and handle cross-level op


ZenithalHourlyRate:
Related to #785.

In short, the GHS variant encodes `[mQl]_t` for level `l` instead of `[m]_t`. This allows more prime candidates, but requires tracking the scaling factor.

OpenFHE assumes same scaling factor for the same level, and adjust the ciphertext when mismatch so we always get a correct lowering for now.

For the compiler, when we get our own lowering (e.g. `bgv-to-poly`), we need to track them to ensure they are the same or handle them correctly when mismatch, if we are using parameter selection process for GHS variant (related to #1168).

This also has influence on level mgmt strategy when level mismatch (#1126), where we currently use multiple `mgmt.modreduce` to ensure same scaling factor, where a more efficient impl would introduce `mgmt.levelreduce` and adjust scaling factor.

Note that this is different from the current scaling factor definition in e.g. `LWE_FullCRTPackingEncoding` which is described as `A scaling_factor is optionally applied on the scalar when converting from a rounded floating point to an integer.`.

References:

> OpenFHE implements both the original BGV scheme (with unscaled messages) [17]
and the Gentry-Halevi-Smart (GHS) variant (with scaled messages) [35]. The main advantage
of the GHS variant is that the RNS moduli qi do not need to satisfy the congruence relation
qi ≡ 1 mod t to perform modulus switching.

https://eprint.iacr.org/2022/915

> Brakerski et al. proposed to choose
the moduli Qi = 1 mod t. This approach, although very convenient in theory, becomes challenging
in practice when using a large t since it reduces significantly the range of possible moduli for the Qi. As a consequence, Gentry, Halevi, and Smart proposed to keep track of the scaling factor for each ciphertext instead [22]. In particular, they suggested to encrypt [QLm]t
instead of [m]t in Equation (1), which provides natural downscaling to [Qim]t as modulus switching operations are applied. However in this case, one has to pay attention when adding two ciphertexts with different scaling factors.

> Handling crosslevel operations and scaling factors. The GHS variant implemented in HELib uses ciphertext-specific scaling factors. In our BGV variant, we chose a simpler approach where the same scaling factor is used for all ciphertexts at a specific level, which reduces the number of scalar multiplications.

https://eprint.iacr.org/2021/204

AlexanderViand-Intel:
I think this same logic would also be required for CKKS-with-per-level-scaling-factors, as in
[...]

PR #1295, opened 2025-01-23
optimize-relinearization: Remove requirement that arguments have equal key-basis degrees

j2kun:
This PR modifies the ILP for optimize-relinearization to permit the possibility that the operands to an op like `arith.addi` have mixed key-basis degrees. E.g., you can add a ciphertext with basis (1, s) to another with basis (1, s, s^2).

Based on the discussion in https://github.com/google/heir/issues/1284, Lattigo and OpenFHE both support this for BGV.

To hedge against the possibility a future backend does not support this, I made the relaxation optional via a new pass option (`allow-mixed-level-ops`).

The impact of this relaxation on the ILP is not as trivial as I had hoped. When removing the constraints on key basis degree for operands, I had to add new constraints that the output degree is at least as big as each input degree. _Ideally_ this comes paired with an objective term that makes the output degree as small as possible, which would implement a `result_degree = max(arg0_degree, ..., argN_degree)` function. However, adding this to the objective causes the solver to "cheat" by inserting relinearization terms earlier than necessary to as to reduce cost of the new terms.

The "right" solution here is to properly cost model the solver's choices (https://github.com/google/heir/issues/1018). Instead of minimizing the total number of relin ops, have a cost for each op at each level, as well as the cost of doing a relin op from each starting level.

In the interim, we simply ignore the fact that the solver may assign a larger key-basis-degree for the output of an add op than necessary. It doesn't seem to affect the correctness of any results; hallucinating larger key-basis degree doesn't change whether or not you need to relinearize, just the mechanics of how to relinearize, which is handled opaquely by the backend anyway. In particular, if the output degree of an op were truly 1, but the solver decided to make it 2, this would require inserting a relin op when it's not necessary, so the solver naturally avoids this situation. But if the degree were 2 and the solver hallucinated it should be 3, we still relin.

Fixes https://github.com/google/heir/issues/1284

### DIFF
--- a/docs/content/en/docs/Design/relinearization_ilp.md
+++ b/docs/content/en/docs/Design/relinearization_ilp.md
@@ -50,13 +50,15 @@ rotation has key-switching built in, and multiplication relinearizes by default.
 That said, many FHE implementations do allow for the relinearization operation
 to be deferred. A useful such situation is when a series of independent
 multiplications are performed, and the results are added together. Addition can
-operate in any key basis (though all inputs must have the same key basis), and
-so the relinearization op that follows each multiplication can be deferred until
-after the additions are complete, at which point there is only one
-relinearization to perform. This technique is usually called _lazy
-relinearization_. It has the benefit of avoiding expensive relinearization
-operations, as well as reducing noise growth, as relinearization adds noise to
-the ciphertext, which can further reduce the need for bootstrapping.
+operate in any key basis (though depending on the backend FHE implementation's
+details, all inputs may require the same key basis, cf.
+[Optional operand agreement](#optional-operand-agreement)), and so the
+relinearization op that follows each multiplication can be deferred until after
+the additions are complete, at which point there is only one relinearization to
+perform. This technique is usually called _lazy relinearization_. It has the
+benefit of avoiding expensive relinearization operations, as well as reducing
+noise growth, as relinearization adds noise to the ciphertext, which can further
+reduce the need for bootstrapping.
 
 In much of the literature, lazy relinearization is applied manually. See for
 example
@@ -128,13 +130,12 @@ TODO(#1018): update docs when objective is generalized.
 
 ### Constraints
 
+#### Simple constraints
+
 The simple constraints are as follows:
 
 - Initial key basis degree: For each block argument, $\\textup{KB}\_v$ is fixed
   to equal the `dimension` parameter on the RLWE ciphertext type.
-- Operand agreement: For each operation with operand SSA values $v_1, \\dots,
-  v_k$, $\\textup{KB}\_{v_1} = \\dots = \\textup{KB}\_{v_k}$, i.e., all key
-  basis inputs must match.
 - Special linearized ops: `bgv.rotate` and `func.return` require linearized
   inputs, i.e., $\\textup{KB}\_{v_i} = 1$ for all inputs $v_i$ to these
   operations.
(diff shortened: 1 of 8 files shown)


lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.td
```diff
@@ -38,6 +38,13 @@ def OptimizeRelinearization : Pass <"optimize-relinearization"> {
            /*default=*/"false",
            "When true, the ILP uses op source locations in variable names, "
            "which can help debug ILP model bugs.">,
+    Option<"allowMixedLevelOps",
+           "allow-mixed-level-ops",
+           "bool",
+           /*default=*/"true",
+           "When true, allow ops to have mixed-level ciphertexts as inputs, e.g., "
+           "adding two ciphertexts with different key bases; this is supported by "
+           "many FHE backends">,
```
ZenithalHourlyRate, line 47: Since we extensively use the term "degree" in ILP model and doc, the name "allow-mixed-degree-ops" could be more consistent, as pointed out by Alex in https://github.com/google/heir/issues/1284#issuecomment-2609796156

mixed-level/cross-level ct op often refers to different RNS level in the literature.

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
```diff
@@ -242,6 +238,12 @@ LogicalResult OptimizeRelinearizationAnalysis::solve() {
         });
   });
 
+  // When mixed-level ops are enabled, the default result level of an op is the
+  // max of the operand levels. This next block of code adds inequality
+  // constraints to ensure the result is larger than the arguments, but it will
+  // not be constrained by the model to be equal to that max value.
```
ZenithalHourlyRate, line 244: We may be able to correctly model the `max` behavior by again using IF_THEN_AUX, check https://math.stackexchange.com/questions/2446606/linear-programming-set-a-variable-the-max-between-two-another-variables/2447498

> In the interim, we simply ignore the fact that the solver may assign a larger key-basis-degree for the output of an add op than necessary. It doesn't seem to affect the correctness of any results

In my understanding, this works because we have the objective `Min(InsertRelin)` and `InsertRelin` is a _somewhat linear_ function of `Degree_result_br`, and we have `Degree_result_br >= Degree_operand`, which happends to be a minimizing of max, so solver will optimize in our desired way.

One concern is that if we change our objective, how will things play out. Though that would then be a positively weighted combination of `InsertRelin` and things are still a minimizing max, for the ease of debugging (when unfamiliar user dumps the model and sees the hallucinating decision) and understanding, I still think we should accurately model the max.

Or we could at least add a comment here or at the objective saying that the unconstrained max behavior here will be constrained by the `min` objective in some implicit way.

ZenithalHourlyRate (commented):
also remember to update the `relinearization_ilp.md`.

lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
j2kun: I agree that things will change when we change the objective. But in that case operations at a higher key basis degree will naturally be more expensive (and I think a relin op from a higher basis degree would also be more expensive than a relin op from a lower basis degree), so I suspect it would work out naturally according to the actual cost of the operations. And that's mainly why I feel we don't need to do the max modeling trick in the interim, since that work will in my view become obsolete. Plus, having an extra binary variable for each op would have an impact on performance that we should avoid if possible.

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.td
j2kun: Done

j2kun:
Updated the doc as well

ZenithalHourlyRate (approved):
For `relinearization_ilp.md` there is one more place mentioning the _same degree_ requirement, i.e.

> Addition can operate in any key basis (though all inputs must have the same key basis),

AlexanderViand-Intel:
> This PR modifies the ILP for optimize-relinearization to permit the possibility that the operands to an op like `arith.addi` have mixed key-basis degrees. E.g., you can add a ciphertext with basis (1, s) to another with basis (1, s, s^2).
> 
> Based on the discussion in #1284, Lattigo and OpenFHE both support this for BGV.
> 
> To hedge against the possibility a future backend does not support this, I made the relaxation optional via a new pass option (`allow-mixed-level-ops`).

I think the main issue here is for CKKS and the GHS variant of BGV, where the messages are scaled. In those settings, we can't just do $(c_0, c_1) + (c'_0, c'_1, c'_2) = (c_0 + c'_0, c_1 + c'_1, c'_2)$ (as we can, afaik, in traditional BGV and in BFV), because the message in ctxt `c` will be scaled by a different factor than that in ctxt `c'`. While being able to toggle mixed operations off is good, I think it's not actually sufficient to prevent similar scaling-related issues all the time. 

So maybe we don't even need the toggle, but just a solution to #1169 (tracking the message scale) + making the modswitch mgmt aware of the need to adjust the scales?

docs/content/en/docs/Design/relinearization_ilp.md
```diff
@@ -146,6 +147,36 @@ The simple constraints are as follows:
...
+input key basis degrees of each operand to be equal, and one where differing key
+basis degrees are allowed.
+
+This is an option because the model was originally implemented under the
+incorrect assumption that CPU backends like OpenFHE and Lattigo require the key
+basis degree operands to be equal for ops like ciphertext addition. When we
+discovered this was not the case, we generalized the model to support both
+cases, in case other backends do have this requirement.
+
+When operands must have the same key basis degree, Then for each operation with
```
asraa, line 162:
```suggestion
When operands must have the same key basis degree, then for each operation with
```

lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.td
```diff
@@ -38,6 +38,13 @@ def OptimizeRelinearization : Pass <"optimize-relinearization"> {
            /*default=*/"false",
            "When true, the ILP uses op source locations in variable names, "
            "which can help debug ILP model bugs.">,
+    Option<"allowMixedDegreeOperands",
+           "allow-mixed-degree-operands",
+           "bool",
+           /*default=*/"true",
+           "When true, allow ops to have mixed-level ciphertexts as inputs, e.g., "
+           "adding two ciphertexts with different key bases; this is supported by "
+           "many FHE backends">,
```
asraa, line 47:
```suggestion
           "many FHE backends, like OpenFHE and Lattigo">,
```

just for that to propogate to the user in the CLI

asraa (approved):

j2kun:
Squashed and marking pull_ready

PR #1297, opened 2025-01-23
Enable clangd to warn on unused include

ZenithalHourlyRate:
Check #1296 and https://clangd.llvm.org/guides/include-cleaner

Note that project config is in https://clangd.llvm.org/config#files and after clangd-17 this option is on by default (https://clangd.llvm.org/config#unusedincludes). I am using clangd-14.

One worrying false positive warning is that, for `Dialect.cpp` (e.g. `LattigoDialect.cpp`), clangd would say `TypeSwtich.h` and `DialectImplementation.h` are not used, whilst they are used in some `cpp.inc`.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+Diagnostics:
+  UnusedIncludes: Strict


asraa:
> One worrying false positive warning is that, for Dialect.cpp (e.g. LattigoDialect.cpp), clangd would say TypeSwtich.h and DialectImplementation.h are not used, whilst they are used in some cpp.inc.

Hah, yes, I see the same warning arise internally too since it doesn't work well with the generated headers/source files.

j2kun:
I recall asking about this internally on the MLIR group, how to avoid these clangd issues, and not getting any responses. Since it appears it's not just an internal issue, maybe we can ask more broadly in the LLVM community.

j2kun:
The recommendation from MLIR community is to mark these includes with `// IWYU pragma: keep`

https://clangd.llvm.org/design/include-cleaner#iwyu-pragmas

We have this other mechanism (the from @llvm-project) that interferes with this for copybara sake, so I will see if I can come up with something that handles both

j2kun (approved):

PR #1302, opened 2025-01-23
remove scheme flag from openfhe emitter

AlexanderViand-Intel:
As mentioned in today's meeting, no longer needing to specify the scheme as a flag makes building the compilation/translation pipeline much nicer, especially in the Python frontend.

The scheme flag was really only used for two decisions:
* in `getModulePrelude`, to decide whether to emit `CryptoContextBGVRNS` or `CryptoContextCKKSRNS`
* around encode/decode, to decide whether to use `getPackedPlaintext` or `getCKKSPackedPlaintext` and to check consistency

The latter use was entirely redundant, as the encoding info on the op (or on the plaintext type of the result/operand) already clearly indicates what should be used. Ensuring consistency is now achieved via verifiers on `openfhe.make_packed_plaintext` and `openfhe.make_ckks_packed_plaintext`.

Unfortunately, the first use wasn't quite as easy to get rid of and required introducing a (hopefully somewhat temporary) helper attribute: the `--secret-to-<scheme>` passes now add an attribute to the module (e.g., `{bgv.scheme}`) which the openfhe emitter inspects.


lib/Dialect/Secret/Conversions/SecretToBGV/SecretToBGV.cpp
```diff
@@ -164,6 +164,9 @@ struct SecretToBGV : public impl::SecretToBGVBase<SecretToBGV> {
     MLIRContext *context = &getContext();
     auto *module = getOperation();
 
+    // Helper for future lowerings that want to know what scheme was used
+    module->setAttr("bgv.scheme", UnitAttr::get(context));
```
j2kun, line 168: Can we put these strings somewhere common, maybe `lib/Dialect/ModuleAttributes.h`, and I'd suggest, reverse the order so `scheme` comes first in all these particular keys.

lib/Target/OpenFhePke/OpenFhePkeHeaderEmitter.cpp
```diff
@@ -45,7 +44,9 @@ LogicalResult OpenFhePkeHeaderEmitter::translate(Operation &op) {
 }
 
 LogicalResult OpenFhePkeHeaderEmitter::printOperation(ModuleOp moduleOp) {
-  os << getModulePrelude(scheme_, importType_) << "\n";
+  // FIXME: grab the scheme from moduleOP directly
```
j2kun, line 47: address FIXME

j2kun (changes requested):
Looks great!

lib/Dialect/Secret/Conversions/SecretToBGV/SecretToBGV.cpp
AlexanderViand-Intel: Will do @putting them somewhere! Unfortunately, the module op verifier will complain if it's not `<somedialect>.<whatever>`, so reversing it might not work.

ZenithalHourlyRate: Would it be better to put the string in `lib/Dialect/BGV/BGVDialect.td` as its prefix is the dialect. We have done similar things for `secret.secret` / `mgmt.mgmt`.

AlexanderViand-Intel: > Would it be better to put the string in `lib/Dialect/BGV/BGVDialect.td` as its prefix is the dialect. We have done similar things for `secret.secret` / `mgmt.mgmt`.

 I was able to flip the names, so for now I put it where @j2kun suggested.

Since I'm fairly sure this attribute will be made redundant by better parameter/keygen handling farily soon (e.g., if we want to be able to emit openfhe code that uses multiple schemes in the same function), I'm not too worried about where exactly it goes for now.

lib/Target/OpenFhePke/OpenFhePkeHeaderEmitter.cpp
AlexanderViand-Intel: Oops 🙈 I'm surprised how this go through the tests, though - do we not have any end-to-end tests with OpenFHE and CKKS? I'd expect _something_ to break if you specify BGV but actually do CKKS....

j2kun (approved):

PR #1307, opened 2025-01-24
frontend: support simple loops structs in python frontend

asraa:
Support single loops in python frontend.

Limitations:
* Must be static start, stop, steps in the range call
* Will raise an error for nested loops now (see linked issue, shouldn't be too hard to support in a follow-up)
* Supports only a sinlge iter_arg for now (also will not be hard, just requires creating new temp vars for the iter_args)

### DIFF
--- a/heir_py/mlir_emitter.py
+++ b/heir_py/mlir_emitter.py
@@ -2,8 +2,120 @@
 
 import operator
 import textwrap
+from dataclasses import dataclass
+from collections import deque
 
 from numba.core import ir
+
+
+class HeaderInfo:
+  """
+  Header info provides info from a loop's header block
+
+  Attributes:
+    phi_var: An ir.Var representing the loop iterator
+    body_id: The block id of the loop body
+    next_id: The block id of the block after the loop
+  """
+
+  def __init__(self, header_block):
+    body = header_block.body
+    assert len(body) == 5
+    self.phi_var = body[3].target
+    self.body_id = body[4].truebr
+    self.next_id = body[4].falsebr
+
+
+class RangeArgs:
+  """
+  Range args provides a ranges start, stop, step args.
+
+  Attributes:
+    start: An ir.Var or int providing the lower bound
+    stop: An ir.Var or int providing the upper bound
+    step: An ir.Var or int providing the step
+  """
+
+  def __init__(self, range_call):
+    args = range_call.value.args
+    self.start = 0
+    self.step = 1
+    match len(args):
+      case 1:
+        self.stop = args
+      case 2:
+        self.start, self.stop = args
+      case 3:
+        self.start, self.stop, self.step = args
+
+
+@dataclass
+class Loop:
+  """
+  A dataclass containing Loop info.
+
+  Attributes:
+    header_id: The block id of the loop's header
+    header: A HeaderInfo object with the iterator and loop block ids
+    rage: A RangeArg object with the loop's range info
+    inits: A list of ir.Vars that initialize any loop-carried vars
+  """
+
+  header_id: int
+  header: HeaderInfo
+  range: RangeArgs
+  inits: list[ir.Var]
+
+
+def build_loop_from_call(index, body, blocks):
+  # Build a loop from a range call starting at index
+  header_id = body[index + 3].target
+  header = HeaderInfo(blocks[header_id])
+  range_args = RangeArgs(body[index])
+
+  # Loop body must start with assigning the local iter var
+  loop_body = blocks[header.body_id].body
+  assert loop_body[0].value == header.phi_var
+
+  inits = []
+  for instr in loop_body[1:]:
+    if type(instr) == ir.Assign and not instr.target.is_temp:
+      inits.append(instr.target)
+  if len(inits) > 1:
+    raise NotImplementedError("Multiple iter_args not supported")
+
+  return Loop(
+      header_id,
+      header,
+      range_args,
+      inits,
+  )
+
+
+def is_start_of_loop(index, body, ssa_ir):
+  instr = body[index]
+  # True if instr is a range call that begins a loop
+  match instr:
+    case ir.Assign(value=ir.Expr(op="call")):
+      func = instr.value.func
+    case _:
+      return False
+
+  assert func.name in ssa_ir._definitions
+  func_def_list = ssa_ir._definitions[func.name]
+  assert len(func_def_list) == 1
+  func_def = func_def_list[0]
+
+  match func_def:
+    case ir.Global(name="range"):
+      # The next instrs should be getiter, iternext, and jump
+      next_instrs = map(type, body[index + 1 : index + 4])
+      if list(next_instrs) != [ir.Assign, ir.Assign, ir.Jump]:
+        return False
+    case _:
+      return False
+
+  return True
 
 
 class TextualMlirEmitter:
@@ -13,6 +125,7 @@ class TextualMlirEmitter:
     self.temp_var_id = 0
     self.numba_names_to_ssa_var_names = {}
     self.globals_map = {}
+    self.loops = {}
 
   def emit(self):
     func_name = self.ssa_ir.func_id.func_name
@@ -25,7 +138,7 @@ class TextualMlirEmitter:
     # TODO(#1162): get inferred or explicit return types
     return_types_str = "i64"
 
-    body = self.emit_body()
+    body = self.emit_blocks()
 
     mlir_func = f"""func.func @{func_name}({args_str}) -> ({return_types_str}) {{
 {textwrap.indent(body, '  ')}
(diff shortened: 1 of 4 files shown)


j2kun:

WoutLegiest (approved):
Partially followed writing this PR during pair programming. 
Cool work, again step closer to our E2E goal!

heir_py/mlir_emitter.py
```diff
@@ -25,38 +116,60 @@ def emit(self):
...
+          loop = build_loop_from_call(i, block.body, blocks, self.ssa_ir)
+          self.loops[instr.target] = loop
+          self.omit_block_header[loop.header.next_id] = True
 
-      if first:
-        first = False
+    # print blocks
+    str_blocks = []
+    for block_id, block in sorted(blocks.items()):
+      if block_id in self.printed_blocks:
```
j2kun, line 144: Could you add a comment here explaining why the program might revisit a block multiple times? (b/c emit_loop prints multiple blocks)

Another option is to start by putting the blocks in a `deque`, change the for loop to a `while blocks_deque`, and then pass the `deque` to `emit_loop` and `emit_block` so it can `popleft` from the deque as it prints stuff out.

The reason I think this might be cleaner is because it must be true that the loop blocks are contiguous until the loop ends. The invariant that would be preserved is: the `blocks_deque` contains as its first element the next block that hasn't yet been emitted.

heir_py/mlir_emitter.py
```diff
@@ -125,6 +242,10 @@ def emit_assign(self, assign):
       case ir.Global():
         self.globals_map[assign.target.name] = assign.value.name
         return ""
+      case ir.Var():
+        # FIXME: keep track of loop var, and assign
```
j2kun, line 246: Convert FIXME to TODO

heir_py/mlir_emitter.py
```diff
@@ -191,6 +312,49 @@ def emit_branch(self, branch):
...
+    for_str = (
+        f"affine.for {itvar} = {loop.range.start} to {loop.range.stop} step"
+        f" {loop.range.step}"
+    )
+
+    if len(loop.inits) == 1:
+      # Note: we must generalize to inits > 1
+      init_val = self.get_name(loop.inits[0])
+      # Within the loop, forward the name for the init val to a new temp var
+      iter_arg = "itarg"
```
j2kun, line 329: This name will need to be uniquified, no? Maybe give it a name that includes the block id.

heir_py/mlir_emitter.py
```diff
@@ -191,6 +312,49 @@ def emit_branch(self, branch):
...
+        raise NotImplementedError("Nested loops are not supported")
+
+    body_str = self.emit_block(loop_block)
+    if len(loop.inits) == 1:
+      # Yield the iter arg
+      yield_var = self.get_name(loop.inits[0])
+      yield_str = f"affine.yield {yield_var} : i64"
+      body_str += "\n" + yield_str
+
+    # After we emit the body, we need to update the printed block map and replace all uses of the iterarg after the block with the result of the for loop.
```
j2kun, line 350: nit: format comment across multiple lines

heir_py/mlir_emitter.py
```diff
@@ -25,38 +116,60 @@ def emit(self):
...
     for block_id, block in sorted(blocks.items()):
-      instructions = []
-      for instr in block.body:
-        result = self.emit_instruction(instr)
-        if result:
-          instructions.append(result)
+      for i in range(len(block.body)):
+        # Detect a range call
+        instr = block.body[i]
+        if is_range_call(instr, self.ssa_ir):
```
j2kun, line 136: This assumes that the only place `range` is used is within a for loop. A decent simplifying assumption for now, but we might want to tighten that a little bit to something like "is_start_of_static_loop(block)" that checks for the range and the subsequent instructions in the block to match it as a loop header. Otherwise the failure mode here is an assertion failure while attempting to build a loop.

j2kun (changes requested):

heir_py/mlir_emitter.py
```diff
@@ -2,17 +2,108 @@
...
+    self.phi_var = body[3].target
+    self.body_id = body[4].truebr
+    self.next_id = body[4].falsebr
+
+
+class RangeArgs:
+
+  def __init__(self, range_call, ssa_ir):
+    args = range_call.value.args
+    self.stop = get_constant(args[0], ssa_ir)
```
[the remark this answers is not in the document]
asraa: Ah yes, we can defer this to MLIR. The get_constant restriction was only on my end it seems, since affine.for can take SSA values for the lower/upper bound and step.

That should be a pretty easy fix, so fixing and adding a test!

heir_py/loop.py
```diff
@@ -0,0 +1,24 @@
...
+_heir_foo = pipeline.run_compiler(foo).module
+
+cc = _heir_foo.foo__generate_crypto_context()
+kp = cc.KeyGen()
+_heir_foo.foo__configure_crypto_context(cc, kp.secretKey)
+arg0_enc = _heir_foo.foo__encrypt__arg0(cc, 2, kp.publicKey)
+res_enc = _heir_foo.foo(cc, arg0_enc)
+res = _heir_foo.foo__decrypt__result0(cc, res_enc, kp.secretKey)
+
+print(res)  # should be 8
```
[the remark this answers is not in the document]
asraa: I'll rewrite the file as a test - it seems like adding more than one test to e2e_test.py right now results in a failure...

```
==================== Test output for //heir_py:e2e_test:
Running tests under Python 3.10.9: /usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/sandbox/linux-sandbox/6134/execroot/heir/bazel-out/k8-opt/bin/heir_py/e2e_test.runfiles/python3_10_x86_64-unknown-linux-gnu/bin/python3
[ RUN      ] EndToEndTest.test_loop
[       OK ] EndToEndTest.test_loop
[ RUN      ] EndToEndTest.test_simple_arithmetic
[  FAILED  ] EndToEndTest.test_simple_arithmetic
======================================================================
ERROR: test_simple_arithmetic (__main__.EndToEndTest)
EndToEndTest.test_simple_arithmetic
----------------------------------------------------------------------
  File "/usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/sandbox/linux-sandbox/6134/execroot/heir/bazel-out/k8-opt/bin/heir_py/e2e_test.runfiles/heir/heir_py/e2e_test.py", line 10, in test_simple_arithmetic
    def foo(a, b):
  File "/usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/sandbox/linux-sandbox/6134/execroot/heir/bazel-out/k8-opt/bin/heir_py/e2e_test.runfiles/heir/heir_py/decorator.py", line 135, in decorator
    compilation_result = run_compiler(
  File "/usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/sandbox/linux-sandbox/6134/execroot/heir/bazel-out/k8-opt/bin/heir_py/e2e_test.runfiles/heir/heir_py/pipeline.py", line 165, in run_compiler
    bound_module = importlib.import_module(module_name)
  File "/usr/local/google/home/asraa/.cache/bazel/_bazel_asraa/dd6a35b586d0a2d65a5a8939e7a649cf/execroot/heir/external/python3_10_x86_64-unknown-linux-gnu/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 674, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1176, in create_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
ImportError: generic_type: type "PublicKey" is already registered!
```

heir_py/mlir_emitter.py
asraa: Ah yes, right now since there's no nested loop support it doesn't since its scoped to that region, but you're right that I need to change this to future-proof the code.

heir_py/loop.py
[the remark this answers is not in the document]
asraa: So i'm gathering that the openfhe types are being repeatedly registered, and i would want to follow https://stackoverflow.com/questions/76072215/pybind11-generic-type-type-is-already-registered-when-importing-two-modul to see if i can make sure that theyre module local, but I'm not sure at all how I can do that. If you do know, would love to track your fix.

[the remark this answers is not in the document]
j2kun: > type "PublicKey" is already registered!

Yeah file a bug and put the test in a separate file, I can look into that later.

heir_py/mlir_emitter.py
asraa: Ah, that's handled now in the loop emission with self.forward_name. Old note.

heir_py/mlir_emitter.py
```diff
@@ -191,6 +345,77 @@ def emit_branch(self, branch):
...
+        raise NotImplementedError("Nested loops are not supported")
+
+    body_str = self.emit_block(loop_block, blocks_to_print)
+    if len(loop.inits) == 1:
+      # Yield the iter arg
+      yield_var = self.get_name(loop.inits[0])
+      yield_str = f"affine.yield {yield_var} : i64"
+      body_str += "\n" + yield_str
+
+    # After we emit the body, we need to update the printed block map and
```
j2kun, line 410:
```suggestion
    # After we emit the body, we need to
```

j2kun (approved):

PR #1308, opened 2025-01-24
decouple openfhe pipeline

AlexanderViand-Intel:
Another openfhe pipeline cleanup motivated by making things nicer for the frontend.

This splits `--mlir-to-openfhe-bgv` into `--mlir-to-bgv --to-openfhe` (and ditto for CKKS).

I'd have liked to do Lattigo, too, but since it doesn't yet support CKKS it'd have been harder to remove the scheme option dependency, and it also does some overriding of the add-client-interface options that the OpenFHE backend doesn't.

On that topic: why *do* we have the option for either per-arg encrypt functions or a single encrypt function? All the OpenFHE E2E tests (and the Frontend) use the former.... If we want to keep the latter, we should (a) make the frontend aware of it so that it generates the right Python helpers and (b) make sure all backends can generate code for either option.

### DIFF
--- a/docs/content/en/docs/Design/simd.md
+++ b/docs/content/en/docs/Design/simd.md
@@ -62,10 +62,10 @@ and index-based accesses into tensors (e.g., `tensor.extract` and
 entire tensors. While its implementation does not depend on any FHE-specific
 details or even the Secret dialect, this transformation is likely only useful
 when lowering a high-level program to an arithmetic-circuit-based FHE scheme
-(e.g., B/FV, BGV, or CKKS). The `-mlir-to-openfhe-bgv` pipeline demonstrates the
-intended flow: augmenting a high-level program with `secret` annotations, then
-applying the SIMD optimization (and any other high-level optimizations) before
-lowering to BGV operations and then exiting to OpenFHE.
+(e.g., B/FV, BGV, or CKKS). The `--mlir-to-bgv --scheme-to-openfhe` pipeline
+demonstrates the intended flow: augmenting a high-level program with `secret`
+annotations, then applying the SIMD optimization (and any other high-level
+optimizations) before lowering to BGV operations and then exiting to OpenFHE.
 
 > **Warning** The current SIMD vectorizer pipeline supports only one-dimensional
 > tensors. As a workaround, one could reshape all multi-dimensional tensors into
--- a/tools/heir-opt.cpp
+++ b/tools/heir-opt.cpp
@@ -366,18 +366,15 @@ int main(int argc, char **argv) {
       "BGV.",
       mlirToRLWEPipelineBuilder(mlir::heir::RLWEScheme::bgvScheme));
 
-  PassPipelineRegistration<mlir::heir::MlirToRLWEPipelineOptions>(
-      "mlir-to-openfhe-bgv",
-      "Convert a func using standard MLIR dialects to FHE using BGV and "
-      "export "
-      "to OpenFHE C++ code.",
-      mlirToOpenFheRLWEPipelineBuilder(mlir::heir::RLWEScheme::bgvScheme));
+  PassPipelineRegistration<mlir::heir::OpenfheOptions>(
+      "scheme-to-openfhe",
+      "Convert code expressed at FHE scheme level to OpenFHE C++ code.",
+      toOpenFhePipelineBuilder());
 
-  PassPipelineRegistration<mlir::heir::MlirToRLWEPipelineOptions>(
+  PassPipelineRegistration<mlir::heir::LattigoOptions>(
       "mlir-to-lattigo-bgv",
       "Convert a func using standard MLIR dialects to FHE using BGV and "
-      "export "
-      "to Lattigo GO code.",
+      "export to Lattigo GO code.",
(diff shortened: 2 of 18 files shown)


j2kun:
> why do we have the option for either per-arg encrypt functions or a single encrypt function

I'd be happy to remove it. I think I added it because, in theory, MLIR supports multiple return values and so you could do it all in one function if you wanted (or in C++ we could return a std::tuple). But I implemented the "do it all in one function" version first and only later decided to split it up for the C++ API.

asraa:
Maybe another thing is that the mlir JIT runner that comes with the MLIR tools doesn't support multiple return args, but I suspect when we have a full poly to llvm lowering maybe we just write our own runner.

tools/heir-opt.cpp
```diff
@@ -366,18 +366,15 @@ int main(int argc, char **argv) {
...
       mlirToRLWEPipelineBuilder(mlir::heir::RLWEScheme::bgvScheme));
 
-  PassPipelineRegistration<mlir::heir::MlirToRLWEPipelineOptions>(
-      "mlir-to-openfhe-bgv",
-      "Convert a func using standard MLIR dialects to FHE using BGV and "
-      "export "
-      "to OpenFHE C++ code.",
-      mlirToOpenFheRLWEPipelineBuilder(mlir::heir::RLWEScheme::bgvScheme));
+  PassPipelineRegistration<mlir::heir::EntryFunctionOptions>(
+      "to-openfhe",
```
j2kun, line 370:
```suggestion
      "scheme-to-openfhe",
```

I'd prefer a more specific name. At least, using `to-openfhe` makes it harder to find it via grep due to being a subset of other strings (`*-to-openfhe`).

j2kun (changes requested):

j2kun:
> maybe we just write our own runner.

I can't wait for that day because removing the `mlir-cpu-runner` will dramatically speed up build times.

AlexanderViand-Intel:
> > maybe we just write our own runner.
> 
> I can't wait for that day because removing the `mlir-cpu-runner` will dramatically speed up build times.

Do we need it to be a *runner*? We're already relying on clang for the E2E OpenFHE stuff, so could we lower to LLVM, emit LLVM IR and then compile instead?

AlexanderViand-Intel:
> > why do we have the option for either per-arg encrypt functions or a single encrypt function
> 
> I'd be happy to remove it. I think I added it because, in theory, MLIR supports multiple return values and so you could do it all in one function if you wanted (or in C++ we could return a std::tuple). But I implemented the "do it all in one function" version first and only later decided to split it up for the C++ API.

I think getting rid of the single-function-for-all approach might indeed be best then! 
@ZenithalHourlyRate Is there a specific reason that the Lattigo backend uses the single-function approach or could you change it over to also use per-arg encryption functions?

ZenithalHourlyRate:
> I think getting rid of the single-function-for-all approach might indeed be best then!
> @ZenithalHourlyRate Is there a specific reason that the Lattigo backend uses the single-function approach or could you change it over to also use per-arg encryption functions?

I used the single-function approach just because it is handy when writing that pass. Will change it to use the per-arg version.

j2kun:
> > > maybe we just write our own runner.
> > 
> > 
> > I can't wait for that day because removing the `mlir-cpu-runner` will dramatically speed up build times.
> 
> Do we need it to be a _runner_? We're already relying on clang for the E2E OpenFHE stuff, so could we lower to LLVM, emit LLVM IR and then compile instead?

Yes, this is what I mean when I say "our own runner": just lower to LLVM and run it for real.

docs/content/en/docs/Design/simd.md
```diff
@@ -62,10 +62,10 @@ and index-based accesses into tensors (e.g., `tensor.extract` and
 entire tensors. While its implementation does not depend on any FHE-specific
 details or even the Secret dialect, this transformation is likely only useful
 when lowering a high-level program to an arithmetic-circuit-based FHE scheme
-(e.g., B/FV, BGV, or CKKS). The `-mlir-to-openfhe-bgv` pipeline demonstrates the
-intended flow: augmenting a high-level program with `secret` annotations, then
-applying the SIMD optimization (and any other high-level optimizations) before
-lowering to BGV operations and then exiting to OpenFHE.
+(e.g., B/FV, BGV, or CKKS). The `--mlir-to-bgv --to-openfhe` pipeline
```
j2kun, line 65:
```suggestion
(e.g., B/FV, BGV, or CKKS). The `--mlir-to-bgv --scheme-to-openfhe` pipeline
```

j2kun (approved):

docs/content/en/docs/Design/simd.md
AlexanderViand-Intel: Fixed!

j2kun:
@AlexanderViand-Intel can you rebase over main?

AlexanderViand-Intel:
> @AlexanderViand-Intel can you rebase over main?

Done! Also renamed the "EntryFunctionOptions" to "OpenfheOptions" as they now include the new "debug" option

PR #1309, opened 2025-01-25
pass analysis state through external func.call ops

j2kun:
Hongren was doing some debug work involving external func calls (`func.func private ...`), and Asra is also working on a prototype that involves external calls, and we identified a few places where `func.call` is not supported in the various passes and lowerings.

This PR adds support for `func.call` to the three analysis passes involved in `annotate-mgmt`. The core assumption here is that external calls imply a join of the lattices of its operands. For secretness this makes a lot of sense (return type is secret if any operand is secret, since we can't know what else happens in the function). For Level/Dimension it's a little bit less clear, but at least for functions with a single operand this implicitly causes the lattice to pass through the call op, where it would otherwise be an uninitialized lattice state.

### DIFF
--- a/lib/Analysis/Utils.h
+++ b/lib/Analysis/Utils.h
@@ -1,0 +1,37 @@
+#ifndef LIB_ANALYSIS_UTILS_H_
+#define LIB_ANALYSIS_UTILS_H_
+
+#include "mlir/include/mlir/Analysis/DataFlowFramework.h"  // from @llvm-project
+#include "mlir/include/mlir/Interfaces/CallInterfaces.h"   // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"                // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+// A generalized version of visitExternalCall that joins all arguments to the
+// func.call op, and then propagates this joined value to all results. This is
+// useful for debug functions where there is a single operand that is not
+// changed by the external call, but can also be useful for some analyses like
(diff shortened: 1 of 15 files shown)


lib/Analysis/Utils.h
```diff
@@ -0,0 +1,40 @@
...
+                           &propagateIfChanged) {
+  StateT resultState = StateT();
+
+  for (const LatticeT *operand : argumentLattices) {
+    const StateT operandState = operand->getValue();
+    resultState = StateT::join(resultState, operandState);
+    if (resultState.isInitialized() && resultState.get()) {
+      break;
+    }
+  }
```
ZenithalHourlyRate, line 30: I am OK with this now, the implicit conversion of `resultState.get()` to `bool` is a little bit unclear though.

In the future we might let the analysis define a default transfer function and reuse it in both `visitOperation` and `visitExternalCall`. For now we do not have the specific need for it.

ZenithalHourlyRate (approved):
MulResultAnalysis should also be updated.

lib/Analysis/Utils.h
j2kun: It's a good idea.

For this change, I think actually I will remove the break statement. Just blindly join all operands. This was originally a short-circuit optimization for secretness, knowing that once secretness is `true` every join will continue to produce `true`. For level/dimension this is not necessarily true.

j2kun: Removed the break

j2kun:
Added to MulResultAnalysis as well

Issue #1209, opened 2024-12-17
Enable a faster backend for OpenFHE


j2kun:
Cf. https://github.com/google/heir/pull/1195 for discussion of the impact. We will need to add OpenMP as a dependency.

j2kun:
We patch in the openfhe config that would be generated by Cmake here: https://github.com/google/heir/blob/main/bazel/openfhe/add_config_core.patch

First step is figuring out what the right generated header file is to replace that patch, then work through build issues.

@ZenithalHourlyRate do you have a copy of the generated src/core/include/config_core.h that came with your faster build from #1195 ?

ZenithalHourlyRate:
I was just using `openfhe-development`, not even `openfhe-hexl`.
```cpp
$ cat build/src/core/config_core.h
//==================================================
// DO NOT TOUCH THIS FILE! It is generated by cmake
//==================================================
#ifndef __CMAKE_GENERATED_CONFIG_CORE_H__
#define __CMAKE_GENERATED_CONFIG_CORE_H__

/* #undef WITH_BE2 */
#define WITH_BE4
/* #undef WITH_NOISE_DEBUG */
/* #undef WITH_NTL */
/* #undef WITH_TCM */

#define CKKS_M_FACTOR 1
#define HAVE_INT128 TRUE
#define HAVE_INT64 TRUE
#define MATHBACKEND 4
#define NATIVEINT 64
```

I think the real problem is these lines

https://github.com/google/heir/blob/9031113340c97fb7e02568566a4aa7038c7f9e89/bazel/openfhe/openfhe.BUILD#L28-L30

I tried to delete the `"OMP_NUM_THREADS=1"` line but the resulting program is still single threaded. Do not know why there is such limitation, maybe it is related to bazel job control?

asraa:
I did a little more exploring - just deleting the OMP_NUM_THREADS=1 is not enough because we need to pass the openmp copts and linkopts to to the bazel rules. I added these flags (rest of the diff omitted but the rules are updated with these new opts)

```
diff --git a/bazel/openfhe/openfhe.BUILD b/bazel/openfhe/openfhe.BUILD
index a4b0ba96..eb5c8245 100644
--- a/bazel/openfhe/openfhe.BUILD
+++ b/bazel/openfhe/openfhe.BUILD
@@ -25,10 +25,22 @@ OPENFHE_COPTS = [
     "-fexceptions",
 ]
 
+LINKOPTS = [
+    "-fopenmp",
[...]

PR #1314, opened 2025-01-26
Enable multithread on OpenFHE with OpenMP support

eymay:
Included suggested flags from discussion #1209. In addition,
compile and link flags added to `tests/Examples/openfhe` to support
OpenMP. Multithreaded execution is enabled by default with this setup.
Clang based toolchain is supported currently.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -68,11 +68,11 @@ jobs:
       # requires the `PATH` to find the `cargo` binary.
       - name: "Run `bazel build`"
         run: |
-          bazel build --noincompatible_strict_action_env  -c fastbuild //...
+          bazel build --noincompatible_strict_action_env --//:enable_openmp=0 -c fastbuild //...
 
       - name: "Run `bazel test`"
         run: |
-          bazel test --noincompatible_strict_action_env -c fastbuild //...
+          bazel test --noincompatible_strict_action_env --//:enable_openmp=0 -c fastbuild //...
 
       # Tests specifically for the tfhe-rs codegen
       - name: rustup toolchain install
--- a/BUILD
+++ b/BUILD
@@ -21,8 +21,33 @@ package_group(
     packages = [],
 )
 
-# Disables Yosys deps for CLI tools and tests.
-# use by passing `--//:enable_yosys=0` to `bazel build` or `bazel test`
+# Disables deps for CI tools and tests.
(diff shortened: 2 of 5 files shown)


j2kun:
This looks great, but the next minor problem is that OpenMP is not guaranteed to be installed on the host system. We should probably gate this by a config option (like [the yosys one](https://github.com/google/heir/blob/dd91785625a729e29f3b20b88ae5d08668c11d2a/BUILD#L24-L43)) so that it can build and run when OpenMP is not installed, and then enable it in our CI runners.

eymay:
> Wonderful, thanks so much for pushing forward on that thread! I'll repro it, I guess I was missing putting the copts / linkopts into the top-level cc_test.
> 
> If we could, it might be nice to add some docs into the readme suggesting that if a user wants to link generated openfhe code with a library that they include those copts. Maybe somewhere here... or tips for working with Bazel. https://heir.dev/docs/development/#openfhe

The thing is the suggested [CMakeLists.txt](https://github.com/openfheorg/openfhe-development/blob/main/CMakeLists.User.txt) of OpenFHE includes the OpenMP support by default so I guess most of the users will already have a working setup with flags. 

While tips at [https://heir.dev/docs/development/#openfhe](https://heir.dev/docs/development/#openfhe) are useful, are they related to the OpenFHE build happening with Bazel? It seems to be targeting an external configuration.

A place I can think of similar to Yosys config is the [Getting Started](https://github.com/google/heir/blob/main/docs/content/en/docs/getting_started.md#clone-and-build-the-project). But feel free to document as I don't want to clutter unnecessarily :)

eymay:
> We should probably gate this by a config option (like [the yosys one](https://github.com/google/heir/blob/dd91785625a729e29f3b20b88ae5d08668c11d2a/BUILD#L24-L43)) so that it can build and run when OpenMP is not installed, and then enable it in our CI runners.

I added an option in Bazel and disabled OpenMP for CI in `build_and_test.yml`. Or should we be installing OpenMP instead in the runners?

eymay:
Currently, the default behaviour is assuming an installed OpenMP with the config being enabled. We can also opt for making it disabled by default and enabling with only the config.

j2kun:
This looks great. Assuming CI passes, I'm happy to merge this in and we can enable OpenMP in the CI runner later if desired (it's probably not going to have a big impact on our test runtimes).

j2kun (approved):

PR #1321, opened 2025-01-27
Add a graph coloring algorithm

j2kun:
This is the main workhorse of https://github.com/google/heir/issues/919 to compute a shift network. This PR was extracted from the in-progress https://github.com/google/heir/pull/1286, as this part is not affected by the remaining issues blocking that PR from being ready for review.

CC @jellevos

### DIFF
--- a/lib/Utils/Graph/Graph.h
+++ b/lib/Utils/Graph/Graph.h
@@ -4,7 +4,10 @@
 #include <algorithm>
 #include <cstdint>
 #include <map>
+#include <queue>
 #include <set>
+#include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include "mlir/include/mlir/Support/LogicalResult.h"  // from @llvm-project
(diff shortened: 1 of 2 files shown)


lib/Utils/Graph/Graph.h
```diff
@@ -154,6 +157,160 @@ class Graph {
...
+      updateSaturationDegree(graph, vertex);
+    }
+
+    while (!queue.empty()) {
+      auto current = queue.top();
+      queue.pop();
+
+      // Skip if vertex is already colored or info is outdated. Could avoid
+      // having "stale" info by using a std::set instead of a priority queue,
+      // but then that would incur log(n) lookups and log(n) updates. Probably
```
asraa, line 229: Thanks for this comment! I was exactly wondering whether pushing the same vertices on the stack when it's neighbor is colored (from `updateNeighborSaturation` pushing onto the queue) was an issue. 

I think you could also erase it from the queue before you update the neighborColors map on line 293, but maybe only if it's performance reason related, since it would be nice then to move the queue update out of the helper in `updateSaturationDegree`.

asraa (approved):
Nice!

j2kun:
> Cool. Any specific reason to implement a graph class yourself, instead of using Boost (or something else)?

Boost is disallowed in Google, though boost::graph is allowed, I feel nervous adding a large dependency for simple stuff.

lib/Utils/Graph/Graph.h
j2kun: Since erasing has log-time lookups, I feel it would be faster overall for sparse graphs to wait until they are popped naturally and skip them. You would do this operation a constant number of times (degree(v) times is constant for sparse graphs) while it could be log(num_vertices) deletion even for sparse graphs.

PR #1334, opened 2025-01-29
relax constraints on ctxt dimension in add/sub/mul

AlexanderViand-Intel:
Removes constraints that ctxt dimension must match from 
`bgv.add`,`bgv.sub`,`bgv.mul`,
`ckks.add`,`ckks.sub`,`ckks.mul`,
`lwe.radd`,`lwe.rsub`,`lwe.rmul`,
`openfhe.add`,`openfhe.sub`,`openfhe.mul_no_relin`.

This became necessary now that the relinearization ILP no longer enforces this constraint.

Fixes #1329

PR #1339, opened 2025-01-29
Inplace binary operation class for the OpenFHE IR + sum and sub inplace operations

ShirChoiEX:
This is my attempt of adding the inplace operations for the Openfhe IR. I added inplace addition/subtraction definitions to check if it would build, which it did. 

I would like to have some comments on the implementation, if it is correct, and how can it be improved.


lib/Dialect/Openfhe/IR/OpenfheOps.td
```diff
@@ -57,6 +57,21 @@ class Openfhe_BinaryOp<string mnemonic, list<Trait> traits = []>
   let results = (outs NewLWECiphertext:$output);
 }
 
+class Openfhe_BinaryInPlaceOp<string mnemonic, list<Trait> traits = []>
+  : Openfhe_Op<mnemonic, traits # [
+    AllTypesMatch<["lhs", "rhs"]>,
```
ZenithalHourlyRate, line 62: There is a recent PR (#1334) that relaxes the `AllTypesMatch` requirement to `    [SameOperandsAndResultRings, InferTypeOpAdaptor]`

Discussion for maintainers: We might also need to register a `MemoryEffects<[MemWrite]>` to make liveness analysis happy, though I found `--cse --canonicalize` does not eliminate these inplace ops.

ZenithalHourlyRate (commented):
Thanks for the contribution!

Please install a pre-commit hook (check https://heir.dev/docs/contributing/) to make PR format clean.

Others are great!

lib/Dialect/Openfhe/IR/OpenfheOps.td
ShirChoiEX: is there a need to add `[SameOperandsAndResultRings]` trait for the inplace operations? It seems to be analyzing both inputs and the output, but there is no output for an inplace function. There seem to be no alternative trait like `[SameOperandsRings]`, or does the `[SameOperandsAndResultRings]` ignore the result check if it is not created? Also, regarding the [#1334](https://github.com/google/heir/pull/1334) PR, he uses `[SameOperandsAndResultRings]` in [AddOp](https://github.com/AlexanderViand-Intel/heir/blob/06b8dfba9e45da795d90f772cb45fefc7d261236/lib/Dialect/Openfhe/IR/OpenfheOps.td#L158) and [SubOp](https://github.com/AlexanderViand-Intel/heir/blob/06b8dfba9e45da795d90f772cb45fefc7d261236/lib/Dialect/Openfhe/IR/OpenfheOps.td#L163) but not in the [MulOp](https://github.com/AlexanderViand-Intel/heir/blob/06b8dfba9e45da795d90f772cb45fefc7d261236/lib/Dialect/Openfhe/IR/OpenfheOps.td#L194) why is it so?

AlexanderViand-Intel: > Also, regarding the [#1334](https://github.com/google/heir/pull/1334) PR, he uses `[SameOperandsAndResultRings]` in [AddOp](https://github.com/AlexanderViand-Intel/heir/blob/06b8dfba9e45da795d90f772cb45fefc7d261236/lib/Dialect/Openfhe/IR/OpenfheOps.td#L158) and [SubOp](https://github.com/AlexanderViand-Intel/heir/blob/06b8dfba9e45da795d90f772cb45fefc7d261236/lib/Dialect/Openfhe/IR/OpenfheOps.td#L163) but not in the [MulOp](https://github.com/AlexanderViand-Intel/heir/blob/06b8dfba9e45da795d90f772cb45fefc7d261236/lib/Dialect/Openfhe/IR/OpenfheOps.td#L194) why is it so?

If the question is why no `SameOperandsAndResultRings` is needed for `openfhe.mul` it's because 
`AllTypesMatch<["lhs", "rhs", "output"]>` implies `SameOperandsAndResultRings`, but if the question is instead why I didn't "relax" the constraints on `openfhe.mul` to "only" `SameOperandsAndResultRings`, that's just because the op isn't used in our default lowering pipeline so there was no need to.

ZenithalHourlyRate: > or does the `[SameOperandsAndResultRings]` ignore the result check if it is not created?

From the following line we can know that if there is no result, the trait only checks the equivalence among operands.

https://github.com/google/heir/blob/ad59e8fcdc110b8e9a59df561afe559695d9fede/lib/Dialect/LWE/IR/LWETraits.h#L19-L50

ShirChoiEX:
I rebased the branch + fixed merge conflicts. All tests should be passing now

lib/Dialect/Openfhe/IR/OpenfheOps.td
```diff
@@ -58,6 +58,21 @@ class Openfhe_BinaryOp<string mnemonic, list<Trait> traits = []>
...
+    ]> {
+
+  let summary = "In-place binary operation for OpenFHE";
+
+  let arguments = (ins
```
j2kun, line 68:
```suggestion

  let arguments = (ins
```

The summary is irrelevant in the base class because it is overridden by all instantiated defs. I can be removed with no impact.

j2kun (approved):
Sorry for the last minute comment. Otherwise, LGTM!

j2kun:
I'll get this in an patch the minor nitpick internally :)

PR #1340, opened 2025-01-29
Adding Loop support to arith-to-cggi pass

WoutLegiest:
Lower the MNIST Tosa model to CGGI and keeping parallel loops
Redefinition of the memref conversion: 
- Memref can be kept in the arith world if it only stores constants
- Now if the same allocOp is used to store both constants and lwe elements: bring to LWE world and add createTrivalOps
- If AllocOp changes: add attribute to the affected operations. This attribute is used to convert the store and load ops


lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -220,15 +295,20 @@ struct ArithToCGGI : public impl::ArithToCGGIBase<ArithToCGGI> {
...
+      // accepts Check if there is at least one Store op that is a constants
       return (llvm::all_of(op->getUses(),
                            [&](OpOperand &op) {
                              auto defOp =
                                  dyn_cast<memref::StoreOp>(op.getOwner());
                              if (defOp) {
-                               return isa<mlir::arith::ConstantOp>(
-                                   defOp.getValue().getDefiningOp());
+                               return allowedRemainArith(defOp);
+                               return (isa<mlir::arith::ConstantOp>(
```
j2kun, line 306: This return is unreachable.

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -220,15 +295,20 @@ struct ArithToCGGI : public impl::ArithToCGGIBase<ArithToCGGI> {
...
+      // Check if all Store ops are constants or GetGlobals, if not store op,
+      // accepts Check if there is at least one Store op that is a constants
       return (llvm::all_of(op->getUses(),
                            [&](OpOperand &op) {
                              auto defOp =
                                  dyn_cast<memref::StoreOp>(op.getOwner());
                              if (defOp) {
-                               return isa<mlir::arith::ConstantOp>(
-                                   defOp.getValue().getDefiningOp());
+                               return allowedRemainArith(defOp);
```
j2kun, line 305: `allowedRemainArith` always returns false for a `StoreOp`. Perhaps this logic is intended to be different? Why not just check `allowedRemainInArith` for all op users?

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -248,14 +332,24 @@ struct ArithToCGGI : public impl::ArithToCGGIBase<ArithToCGGI> {
...
+        return true;
+      }
+
+      if (auto lweAttr =
+              op->getAttrOfType<mlir::StringAttr>("lwe_annotation")) {
+        return false;
+      }
+
+      if (auto *defOp =
+              dyn_cast<memref::StoreOp>(op).getValue().getDefiningOp()) {
```
j2kun, line 346:
```suggestion
              cast<memref::StoreOp>(op).getValue().getDefiningOp()) {
```

Since the operation is guaranteed to be a StoreOp, you can use `cast` instead of `dyn_cast`

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -264,17 +358,40 @@ struct ArithToCGGI : public impl::ArithToCGGIBase<ArithToCGGI> {
...
+
+      if (auto lweAttr =
+              op->getAttrOfType<mlir::StringAttr>("lwe_annotation")) {
+        return false;
+      }
+
+      return true;
+    });
+
+    // Convert LoadOp if memref comes from an argument
```
j2kun, line 374: Fix comment copied from above

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -220,15 +295,20 @@ struct ArithToCGGI : public impl::ArithToCGGIBase<ArithToCGGI> {
         });
 
     target.addDynamicallyLegalOp<memref::AllocOp>([&](Operation *op) {
-      // Check if all Store ops are constants, if not store op, accepts
-      // Check if there is at least one Store op that is a constants
+      // Check if all Store ops are constants or GetGlobals, if not store op,
+      // accepts Check if there is at least one Store op that is a constants
       return (llvm::all_of(op->getUses(),
```
j2kun, line 300: nit: can we split this condition up into multiple lines to make it more readable?

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -191,6 +242,30 @@ struct ConvertArithBinOp : public OpConversionPattern<SourceArithOp> {
...
+
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      mlir::memref::AllocOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+    for (auto *userOp : op->getUsers()) {
+      userOp->setAttr("lwe_annotation",
```
j2kun, line 257: There is some circular logic here that I'm not able to understand easily, so it at least needs a comment explaining it. When processing an alloc, the code marks all downstream users with this annotation, but then when evaluating if an alloc should be converted, the code (in part) checks some downstream users to see if they have this annotation.

Perhaps it would be worth separating this logic into three steps: the first step walks the IR (without any pattern application) and decides which ops to mark with the annotation, the second step reads the annotations to determine legality, and the third step is the pattern application, which only modifies the op in question.

j2kun (changes requested):

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
WoutLegiest: It is more, we only care to check the store Ops, the rest of all the operands, we can just ignore

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
WoutLegiest: Oww yeah! Thanks for catching this circular depedency

I now removed the checking of the attribute from the `allowedRemainArith` function and implemented it in a separate function. One thing I don't want to loose about the allowedRemainArith function is the recursive calling to figure out legality of extSiOps (if they come from Get global, they are actually allowed).

j2kun (approved):

PR #1341, opened 2025-01-30
Configure pyink formatter in pre-commit and apply to all files

j2kun:
Note https://github.com/google/pyink is a fork of `black` with some patches to support Google's internal style guide. The configuration I set in `pyproject.toml` corresponds to the options making pyink consistent with Google's internal style guide.

Eventually they hope to upstream all their patches to black, and they have been making some progress on this. But in the end we may be stuck with `pyink` for a while.

Their repo has instructions for getting it set up with VSCode. I haven't yet had the time to get it set up with vim/NeoVim, but if there are any other ~~dinosaurs~~ wizards like me who want my config let me know and I can share it.


AlexanderViand-Intel (approved):
Nice! Could you also add a mention to the "Developer" documentation, something like the sentence from the PR description about setting up vscode?

j2kun:
Added a note and squashed

PR #1343, opened 2025-01-30
Introduce NoiseAnalysis Framework

ZenithalHourlyRate:
Before incorporating parameter selection procedures, it is vital to support various forms of noise analyse in the HE circuit, and the analysis itself worths design discussion.

Although there are multiple papers discussing noise analysis (i.e. worst-case v.s. average-case, coefficient embedding v.s. canonical embedding, implementation-agnostic v.s. implementation-specific), I think we can model them using MLIR Dataflow framework with different `LatticeState` and re-use the `Analysis`. The focus of this PR is not the detailed function in `Noise.cpp/NoiseAnalysis.cpp` but the design of `NoiseAnalysis.h/Noise.h/Params.h`.

The tricky part of parameter selection and noise analysis is that, noise analysis requires concrete parameter, yet parameter should be selected according to the analysis result, forming a circular dependency.

To break this dependency, we can use an iterative approach by first giving a conservative param and later passes can give optimized one. Previous papers would use optimizer to solve it at once, but for us that would be left for future PR.

The `BGV/Noise.cpp` uses formulas from KPZ21, and there are various assumptions made to the circuit, but that should not be critical and we could revisit it when we incorporate the param selection procedure and have tested various programs.

I could not think of a test for such analysis. The test should be considered when param selection is in.

### Example

Analysed result

```shell
$ --mlir-to-secret-arithmetic --secret-insert-mgmt-bgv $PWD/tests/Examples/openfhe/dot_product_8.mlir --debug-only=secret-insert-mgmt-bgv --debug-only=NoiseAnalysis --debug-only=BGVNoise

ringDim: 16384
plaintextModulus: 4295294977
level: 2
logqi: 60 60 60 
dnum: 2
logpi: 60 
Propagating 45.27 to <block argument> of type 'tensor<8xi16>' at index: 0
Propagating 45.27 to <block argument> of type 'tensor<8xi16>' at index: 1
Propagating 98.53 to %1 = arith.muli %input0, %input1 {mgmt.mgmt = #mgmt.mgmt<level = 2, dimension = 3>} : tensor<8xi16>
Propagating 98.53 to %2 = mgmt.relinearize %1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
Propagating 98.53 to %3 = tensor_ext.rotate %2, %c4 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>, index
Propagating 99.53 to %4 = arith.addi %2, %3 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
Propagating 99.53 to %5 = tensor_ext.rotate %4, %c2 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>, index
Propagating 100.53 to %6 = arith.addi %4, %5 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
Propagating 100.53 to %7 = tensor_ext.rotate %6, %c1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>, index
Propagating 101.53 to %8 = arith.addi %6, %7 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16>
Propagating 41.76 to %9 = mgmt.modreduce %8 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<8xi16>
Propagating 113.76 to %extracted = tensor.extract %9[%c7] {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<8xi16>
Propagating 53.76 to %10 = mgmt.modreduce %extracted {mgmt.mgmt = #mgmt.mgmt<level = 0>} : i16
```

Real noise dumped

```
$ ./bazel-out/k8-dbg/bin/tests/Examples/openfhe/dot_product_8_debug_test

( 1 2 3 4 5 6 7 8 ... )
cv 2 Ql 4 logQ: 167 logqi: [ 47 52 52 16 ] budget 123.579 noise: 42.4211
( 2 3 4 5 6 7 8 9 ... )
cv 2 Ql 4 logQ: 167 logqi: [ 47 52 52 16 ] budget 123.69 noise: 42.3096
( 2 6 12 20 30 42 56 72 ... )
cv 3 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 66.4859 noise: 83.5141
( 2 6 12 20 30 42 56 72 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 66.4859 noise: 83.5141
( 30 42 56 72 2 6 12 20 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 66.4859 noise: 83.5141
( 32 48 68 92 32 48 68 92 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 66.096 noise: 83.904
( 68 92 32 48 68 92 32 48 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 66.096 noise: 83.904
( 100 140 100 140 100 140 100 140 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 65.6047 noise: 84.3953
( 140 100 140 100 140 100 140 100 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 65.6047 noise: 84.3953
( 240 240 240 240 240 240 240 240 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 64.8355 noise: 85.1645
( 240 240 240 240 240 240 240 240 ... )
cv 2 Ql 3 logQ: 151 logqi: [ 47 52 52 ] budget 64.8355 noise: 85.1645
( 0 0 0 0 0 0 0 240 ... )
cv 2 Ql 2 logQ: 99 logqi: [ 47 52 ] budget 26.5106 noise: 71.4894
( 240 ... )
cv 2 Ql 2 logQ: 99 logqi: [ 47 52 ] budget 26.5106 noise: 71.4894
( 240 ... )
cv 2 Ql 2 logQ: 99 logqi: [ 47 52 ] budget 26.5106 noise: 71.4894
( 240 ... )
cv 2 Ql 2 logQ: 99 logqi: [ 47 52 ] budget 26.5106 noise: 71.4894
```

### DIFF
--- a/lib/Analysis/NoiseAnalysis/BGV/Noise.cpp
+++ b/lib/Analysis/NoiseAnalysis/BGV/Noise.cpp
@@ -1,4 +1,4 @@
-#include "lib/Analysis/NoiseAnalysis/Noise.h"
+#include "lib/Analysis/NoiseAnalysis/BGV/Noise.h"
 
 #include <cmath>
 
@@ -6,12 +6,14 @@
 #include "llvm/include/llvm/Support/Debug.h"        // from @llvm-project
 #include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
 
-#define DEBUG_TYPE "Noise"
+#define DEBUG_TYPE "BGVNoise"
 
 namespace mlir {
 namespace heir {
+namespace bgv {
 
-std::string Noise::toString() const {
+template <bool W>
+std::string Noise<W>::toString() const {
   switch (noiseType) {
     case (NoiseType::UNINITIALIZED):
       return "Noise(uninitialized)";
@@ -20,7 +22,8 @@ std::string Noise::toString() const {
   }
 }
 
-std::string Noise::toBound(const LocalParam &param) const {
+template <bool W>
+std::string Noise<W>::toBound(const LocalParam &param) const {
   auto t = param.getSchemeParam()->getPlaintextModulus();
   auto bound = log(t * getValue()) / log(2);
   std::stringstream stream;
@@ -28,41 +31,50 @@ std::string Noise::toBound(const LocalParam &param) const {
   return stream.str();
 }
 
-llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Noise &variance) {
+template <bool W>
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Noise<W> &variance) {
   return os << variance.toString();
 }
 
-Diagnostic &operator<<(Diagnostic &diagnostic, const Noise &variance) {
+template <bool W>
+Diagnostic &operator<<(Diagnostic &diagnostic, const Noise<W> &variance) {
   return diagnostic << variance.toString();
 }
 
-double Noise::getExpansionFactor(const LocalParam &param) {
+template <bool W>
+double Noise<W>::getExpansionFactor(const LocalParam &param) {
   auto n = param.getSchemeParam()->getRingDim();
-  // openfhe average case
-  auto expansionFactor = 2.0 * sqrt(n);
-  // worst-case
-  // auto expansionFactor = n;
-  return expansionFactor;
+  if constexpr (W) {
+    // worst-case
+    return n;
+  } else {
+    // openfhe average case
+    return 2.0 * sqrt(n);
+  }
 }
 
-double Noise::getBoundErr(const LocalParam &param) {
+template <bool W>
+double Noise<W>::getBoundErr(const LocalParam &param) {
   auto std0 = param.getSchemeParam()->getStd0();
   auto assurance = 6;
   auto boundErr = std0 * assurance;
   return boundErr;
 }
 
-double Noise::getBoundKey(const LocalParam &param) {
+template <bool W>
+double Noise<W>::getBoundKey(const LocalParam &param) {
   auto boundKey = 1.0;
   return boundKey;
 }
 
-Noise Noise::evalConstant(const LocalParam &param) {
+template <bool W>
+Noise<W> Noise<W>::evalConstant(const LocalParam &param) {
   auto t = param.getSchemeParam()->getPlaintextModulus();
   return Noise::of(t);
 }
 
-Noise Noise::evalEncryptPk(const LocalParam &param) {
+template <bool W>
+Noise<W> Noise<W>::evalEncryptPk(const LocalParam &param) {
   auto boundErr = getBoundErr(param);
   auto boundKey = getBoundKey(param);
   auto expansionFactor = getExpansionFactor(param);
@@ -71,11 +83,14 @@ Noise Noise::evalEncryptPk(const LocalParam &param) {
   return Noise::of(fresh);
 }
 
-Noise Noise::evalAdd(const Noise &lhs, const Noise &rhs) {
+template <bool W>
+Noise<W> Noise<W>::evalAdd(const Noise &lhs, const Noise &rhs) {
   return Noise::of(lhs.getValue() + rhs.getValue() + 1);
 }
-Noise Noise::evalMultNoRelin(const LocalParam &resultParam, const Noise &lhs,
-                             const Noise &rhs) {
+
+template <bool W>
+Noise<W> Noise<W>::evalMultNoRelin(const LocalParam &resultParam,
+                                   const Noise<W> &lhs, const Noise<W> &rhs) {
   auto t = resultParam.getSchemeParam()->getPlaintextModulus();
   auto expansionFactor = getExpansionFactor(resultParam);
 
@@ -84,7 +99,9 @@ Noise Noise::evalMultNoRelin(const LocalParam &resultParam, const Noise &lhs,
                     rhs.getValue() + 1));
 }
 
-Noise Noise::evalModReduce(const LocalParam &inputParam, const Noise &input) {
+template <bool W>
+Noise<W> Noise<W>::evalModReduce(const LocalParam &inputParam,
+                                 const Noise<W> &input) {
   auto cv = inputParam.getDimension();
   assert(cv == 2);
   double modulus = 1L << inputParam.getSchemeParam()
@@ -98,13 +115,23 @@ Noise Noise::evalModReduce(const LocalParam &inputParam, const Noise &input) {
   return Noise::of(scaled + added);
 }
 
-Noise Noise::evalRelinearize(const LocalParam &inputParam, const Noise &input) {
+// assume relinearize does not introduce error larger than mult
+template <bool W>
+Noise<W> Noise<W>::evalRelinearize(const LocalParam &inputParam,
+                                   const Noise &input) {
   return input;
 }
 
-Noise Noise::evalRotate(const LocalParam &inputParam, const Noise &input) {
+template <bool W>
+Noise<W> Noise<W>::evalRotate(const LocalParam &inputParam,
+                              const Noise &input) {
   return Noise::evalRelinearize(inputParam, input);
 }
 
+// instantiation
+template class Noise<true>;
+template class Noise<false>;
+
+}  // namespace bgv
 }  // namespace heir
 }  // namespace mlir
--- a/lib/Analysis/NoiseAnalysis/BGV/Noise.cpp
+++ b/lib/Analysis/NoiseAnalysis/BGV/Noise.cpp
@@ -32,13 +32,13 @@ std::string Noise<W>::toBound(const LocalParam &param) const {
 }
 
 template <bool W>
-llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Noise<W> &variance) {
-  return os << variance.toString();
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Noise<W> &noise) {
+  return os << noise.toString();
 }
 
 template <bool W>
-Diagnostic &operator<<(Diagnostic &diagnostic, const Noise<W> &variance) {
-  return diagnostic << variance.toString();
+Diagnostic &operator<<(Diagnostic &diagnostic, const Noise<W> &noise) {
+  return diagnostic << noise.toString();
 }
 
 template <bool W>
@@ -104,8 +104,11 @@ Noise<W> Noise<W>::evalModReduce(const LocalParam &inputParam,
                                  const Noise<W> &input) {
   auto cv = inputParam.getDimension();
   assert(cv == 2);
-  double modulus = 1L << inputParam.getSchemeParam()
-                             ->getLogqi()[inputParam.getCurrentLevel()];
+
+  auto currentLogQi =
+      inputParam.getSchemeParam()->getLogqi()[inputParam.getCurrentLevel()];
+
+  double modulus = pow(2.0, currentLogQi);
 
   auto expansionFactor = getExpansionFactor(inputParam);
   auto boundKey = getBoundKey(inputParam);
@@ -122,6 +125,8 @@ Noise<W> Noise<W>::evalRelinearize(const LocalParam &inputParam,
   return input;
 }
 
+// assume rotation after mult...
+// should be fixed later.
 template <bool W>
 Noise<W> Noise<W>::evalRotate(const LocalParam &inputParam,
                               const Noise &input) {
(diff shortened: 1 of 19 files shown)


lib/Transforms/SecretInsertMgmt/SecretInsertMgmtBGV.cpp
```diff
@@ -87,6 +92,37 @@ struct SecretInsertMgmtBGV
     pipeline.addPass(createCSEPass());
     pipeline.addPass(mgmt::createAnnotateMgmt());
     (void)runPipeline(pipeline, getOperation());
+
+    // The following part is experimental noise analysis
+    // for BGV. Should observe the result using --debug-only=NoiseAnalysis
+    // and --debug-only=secret-insert-mgmt-bgv
```
j2kun, line 98:
```suggestion
    // for BGV. Should observe the result using --debug-only=NoiseAnalysis,secret-insert-mgmt-bgv
```

FYI these flags accept a comma-separated list of debug-type ids.

lib/Analysis/NoiseAnalysis/Params.h
```diff
@@ -0,0 +1,84 @@
...
+        level(level),
+        logqi(logqi),
+        dnum(dnum),
+        logpi(logpi) {}
+
+ private:
+  // the N in Z[X]/(X^N+1)
+  int ringDim;
+
+  // the plaintext modulud for BGV
```
j2kun, line 28:
```suggestion
  // the plaintext modulus for BGV
```

lib/Analysis/NoiseAnalysis/Params.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include <cmath>
+#include <vector>
+
+#include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+class SchemeParam {
```
j2kun, line 12: I think these classes are fine, but I would prefer if we made two small tweaks:

1. Name them specific to the scheme they are operating on (if not the scheme + variant).
2. Put them in a new directory, maybe `lib/Parameters`.

The reason for (2) is that I suspect we will wrap these classes in IR attributes with custom parser/printers so we can attach them to the IR later. So having them in an isolated place allows us to avoid cyclic dependencies more easily.

lib/Analysis/NoiseAnalysis/NoiseAnalysis.h
```diff
@@ -0,0 +1,49 @@
...
+
+  void setToEntryState(NoiseLattice<Noise> *lattice) override {
+    // At an entry point, we have no information about the noise.
+    this->propagateIfChanged(lattice, lattice->join(Noise::uninitialized()));
+  }
+
+  LogicalResult visitOperation(
+      Operation *op, ArrayRef<const NoiseLattice<Noise> *> operands,
+      ArrayRef<NoiseLattice<Noise> *> results) override;
+
```
j2kun, line 41: You may also want to override visitExternalCall if you are passing things to a debug function, otherwise the default behavior is to reset the lattice to uninitialized.

lib/Analysis/NoiseAnalysis/BGV/Noise.h
```diff
@@ -0,0 +1,111 @@
...
+#include "lib/Analysis/NoiseAnalysis/Params.h"
+#include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Diagnostics.h"       // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace bgv {
+
+// is worst case or not
+template <bool W = false>
```
j2kun, line 20: I think a template with a boolean option is not the right modeling construct here. First, most of the formulas above don't depend on it, while the one that does switches on a constexpr (presumably because you can't add this to the `Noise` constructor without mucking up the lattice). Second, more advanced noise models would likely have completely different APIs; e.g., they might track a lower and upper bound for a given probability. We don't want to leak that throughout the rest of the program, an in this case it's leaking through the template instantiation.

I can think of a few possible improvements. 

(1) Add a base class to represent the common parts of the two noise models given here, with virtual methods for the parts that differ. Since the parts that differ are only used internally, you can further simplify the analysis pass below to accept a `NoiseModel<NoiseBase>` and avoid the templating.

(2) Separate the noise model from the noise data. Since the lattice itself only needs "join", the `visit` helpers could instead live in a `NoiseModel` class with a more well-defined interface (e.g., `propagateThroughOp` as the front-facing function, which takes as input the operand noises and produces result noises), and then the `NoiseAnalysis` can avoid hard-coding anything about how the noise model works internally (in particular, which ops need type switches) while two noise models that both use "single scalar with a max join" can reuse the same `Noise` struct. In this case you might still need a `Noise` base class or template, if two different model classes use two different kinds of noise, but after this refactoring that would be relatively straightforward and could be deferred until the point where we need it.

I like (2) better because I can see how it will be more future-proof: if we have a noise model that uses an upper bound on error probability, that upper bound would be data given to the noise model, and not live in the noise struct, and would only be visible during the noise model construction time.

Open to any further ideas.

lib/Analysis/NoiseAnalysis/BGV/Noise.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "llvm/include/llvm/Support/Debug.h"        // from @llvm-project
+#include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+
+#define DEBUG_TYPE "BGVNoise"
+
+namespace mlir {
+namespace heir {
+namespace bgv {
+
+template <bool W>
```
j2kun, line 15: Please add a code comment linking to the source for these formulas. Either a paper or code reference.

j2kun (changes requested):
Added some comments on the structure/organization.

Having a noise analysis that assumes known scheme parameters is the right way to start. Even if we have a sophisticated parameter selection optimization, we will want to regularly verify noise bounds are still respected after various program transformations.

To that end, let's test this by creating a utility function `validateNoise` that accepts as input a top-level op, known parameters, and a noise model, and returns a LogicalResult for success/failure if the noise stays within correctness bounds.

Then the right way to test this PR would be to wrap the `validateNoise` in a trivial pass `validate-noise` and add test IRs that pass and fail the noise bounds.

lib/Analysis/NoiseAnalysis/BGV/Noise.h
ZenithalHourlyRate: I have refactored the `NoiseAnalysis.h` with the following idea in mind

* `NoiseAnalysis` should take only one user-friendly template argument
* The concrete `Model` and `State` should be decoupled from `NoiseAnalysis` and `NoiseLattice`
* All the thing should be designed in a template way. No inheritance.

For inheritance I think it becomes nasty because there is so much templating here, so I did not go for (1).

For (2), I agree that "NoiseModel class with a more well-defined interface" could be helpful. I took one step further that the template class `NoiseAnalysis` can have [Explicit specialization](https://en.cppreference.com/w/cpp/language/template_specialization) given the template argument, so the architecture becomes 

* `NoiseAnalysis` tooks a `NoiseModel` template argument which has all `SchemeType` / `StateType` defined.
* `NoiseAnalysis` is a header-only library and does not have default implementation.
* For each `NoiseModel`, it defines the transfer function for `StateType`, and it also specialize `NoiseAnalysis<NoiseModel>` in its cpp file.
* The dependency becomes
`End-user -> Some NoiseModel (with Analysis specialization) -> NoiseAnalysis`.

I think this is quite extensible now, either for the promised `bgv-noise-by-variance-coeff-average-case`/`bgv-noise-by-bound-canon-worst-case` impl in near-future PR, or for CKKS/CGGI noise analysis.

lib/Transforms/ValidateNoise/ValidateNoise.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    return success();
+  }
+
+  template <typename NoiseAnalysis>
+  LogicalResult validate(
+      DataFlowSolver *solver,
+      const typename NoiseAnalysis::SchemeParamType &schemeParam) {
+    auto res = getOperation()->walk([&](func::FuncOp funcOp) {
+      return funcOp->walk([&](secret::GenericOp genericOp) {
```
j2kun, line 96: It seems like this could be collapsed into a single walk on Generics. The Func is not used anywhere in the body.

lib/Transforms/ValidateNoise/ValidateNoise.cpp
```diff
@@ -0,0 +1,176 @@
...
+    auto localParam = getLocalParam(value);
+
+    int total = 0;
+    auto logqi = localParam.getSchemeParam()->getLogqi();
+    for (auto i = 0; i <= localParam.getCurrentLevel(); ++i) {
+      total += logqi[i];
+    }
+
+    auto bound = NoiseModel::toLogBound(localParam, noiseState);
+    auto boundString = NoiseModel::toLogBoundString(localParam, noiseState);
```
j2kun, line 76:
```suggestion
    [[maybe_unused]] auto boundString = NoiseModel::toLogBoundString(localParam, noiseState);
```

You could also put this inside the debug below.

lib/Transforms/ValidateNoise/ValidateNoise.cpp
```diff
@@ -0,0 +1,176 @@
...
+    // logQ - logNoise - 1
+    int budget = total - bound - 1;
```
j2kun, line 78: Why isn't this `- 1` part of the NoiseModel itself? It seems somewhat arbitrary here.

lib/Parameters/BGV/Params.h
```diff
@@ -0,0 +1,86 @@
...
+  int64_t plaintextModulus;
+
+  // RNS level, from 0 to L
+  int level;
+
+  // logarithm of the modulus of each level
+  // logqi.size() == level + 1
+  std::vector<int> logqi;
+
+  // number of digits
```
j2kun, line 39: Clarify in the comment what it's the number of digits of.

lib/Parameters/BGV/Params.h
```diff
@@ -0,0 +1,86 @@
...
+  const std::vector<int> &getLogpi() const { return logpi; }
+  double getStd0() const { return std0; }
+
+  void print(llvm::raw_ostream &os) const;
+
+  static SchemeParam getConservativeSchemeParam(int level,
+                                                int64_t plaintextModulus);
+};
+
+class LocalParam {
```
j2kun, line 63: Please add a comment above the class explaining what "local" means. I think it means: the set of parameters specific to an SSA value at a particular point in the program execution.

lib/Parameters/BGV/Params.cpp
```diff
@@ -0,0 +1,73 @@
...
+namespace mlir {
+namespace heir {
+namespace bgv {
+
+struct RLWEParam {
+  int ringDim;
+  int maxQ;
+};
+
+// uniform tenary
```
j2kun, line 14:
```suggestion
// parameters for uniform ternary secret distribution
```

lib/Parameters/BGV/Params.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include <cassert>
+
+namespace mlir {
+namespace heir {
+namespace bgv {
+
+struct RLWEParam {
+  int ringDim;
+  int maxQ;
```
j2kun, line 11: Maybe clarify this is the log2 of Q, if not in the name then in a comment.

lib/Parameters/BGV/Params.cpp
j2kun: I don't recall if BGV has a sparsity condition for its secret distribution, but if so please add the hamming weight. Or better yet, if you can cite an external doc that would be helpful.

j2kun (changes requested):

lib/Transforms/ValidateNoise/ValidateNoise.cpp
ZenithalHourlyRate: Moved the budget and total calculation to the noise model.

lib/Parameters/BGV/Params.cpp
ZenithalHourlyRate: In the homormorphic encryption standard there is no sparse setting. Added link to HES and Openfhe.

ZenithalHourlyRate:
Comments addressed.

PR #1367, opened 2025-02-06
Add Chebyshev interpolation

j2kun:
This PR adds a Chebyshev interpolation routine, which is a subroutine needed for Caratheodory-Fejer approximation.

Adds a new (very lightweight) dependency: `pocketfft`, the fft implementation used by numpy.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -218,6 +218,7 @@ prospectively choose to deem waived or otherwise exclude such Section(s) of
 the License, but only in their entirety and only with respect to the Combined
 Software.
 
+# Yosys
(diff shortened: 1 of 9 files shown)


WoutLegiest (approved):
Thanks, cool code! 
Just wondering if we could link a reference document that explain the math behind it? Or are we putting it into the paper

j2kun:
> Thanks, cool code! Just wondering if we could link a reference document that explain the math behind it? Or are we putting it into the paper

Added a citation. I think the method in this PR is not novel (it's a very old and well known technique) and the main routine (coming next) is the Caratheodory-Fejer approximation which has proper citations.

PR #1374, opened 2025-02-07
Lattigo: Add debug func impl for noise

ZenithalHourlyRate:
Related to #1343, in which the debug func impl for noise for Openfhe is provided.

Lattigo is honest with regard to compiler instructions, whereas Openfhe does some AUTOmation so not all modreduce are respected.

So we can see a good noise trace that corresponds to the result predicted by the noise analysis.

```
[1 2 3 4 5 6 7 8]
Noise: 38.78
[2 3 4 5 6 7 8 9]
Noise: 38.88
[2 6 12 20 30 42 56 72]
Noise: 83.28
[2 6 12 20 30 42 56 72]
Noise: 83.28
[30 42 56 72 2 6 12 20]
Noise: 83.28
[32 48 68 92 32 48 68 92]
Noise: 83.83
[68 92 32 48 68 92 32 48]
Noise: 83.83
[100 140 100 140 100 140 100 140]
Noise: 84.37
[140 100 140 100 140 100 140 100]
Noise: 84.37
[240 240 240 240 240 240 240 240]
Noise: 84.99
[240 240 240 240 240 240 240 240]
Noise: 54.14
[0 0 0 0 0 0 0 240]
Noise: 87.11
[240 0 0 0 0 0 0 0]
Noise: 87.11
[240 0 0 0 0 0 0 0]
Noise: 87.11
[240 0 0 0 0 0 0 0]
Noise: 53.77
```

(I changed the param provided to the below, which coresponds to the NoiseAnalysis input)
```go
	param, err15 := bgv.NewParametersFromLiteral(bgv.ParametersLiteral{
		LogN:             14,
		LogQ:             []int{60, 60, 60},
		LogP:             []int{60},
		PlaintextModulus: 4295294977,
	})
```

I observe some abnormality with the trace according to my past experience of observing noise growth though (i.e. the mod reduce noise should be less than freshly encrypted noise). I believe this difference is implementation-specific (yes different lib, like HElib/Openfhe, has different function impl so their analysis is different, not everything is textbook impl). I will investigate more closely when dealing with the correctness of noise model later. For now I think this PR is OK.

This PR is partially inspired by https://github.com/tuneinsight/lattigo/discussions/414

### DIFF
--- a/tests/Examples/lattigo/dot_product_8_debug.go
+++ b/tests/Examples/lattigo/dot_product_8_debug.go
@@ -13,4 +13,26 @@ func __heir_debug(evaluator *bgv.Evaluator, param bgv.Parameters, encoder *bgv.E
 	pt := decryptor.DecryptNew(ct)
 	encoder.Decode(pt, value)
 	fmt.Println(value)
+
+	// get a new pt with no noise
+	// in Lattigo, Decrypt won't mod T
+	// Decode will mod T so we Decode then Encode
+	valueFull := make([]int64, ct.N())
+	encoder.Decode(pt, valueFull)
+	// set the level and scale of the plaintext
+	ptNoNoise := bgv.NewPlaintext(param, ct.Level())
+	ptNoNoise.Scale = ct.Scale
+	encoder.Encode(valueFull, ptNoNoise)
+
+	// subtract the message from the ciphertext
+	vec, _ := evaluator.SubNew(ct, ptNoNoise)
+	// get infty norm of the noise
+	_, _, max := rlwe.Norm(vec, decryptor)
+	// get logQi for current level
+	total := 0
+	for i := 0; i <= ct.LevelQ(); i++ {
+		total += param.LogQi()[i]
+	}
+	// t * e for BGV
+	fmt.Printf("Noise: %.2f Total: %d\n", max+param.LogT(), total)
 }


ZenithalHourlyRate:
See trace from #1379 

So the problem is that when doing ct-pt op, the pt has to set the scale. i.e. Lattigo use GHS variant of BGV and this is related to #1169.

For OpenFHE we have MakePackedPlaintext that hides the detail, but for Lattigo we have to manually set the scale

```
	pt := bgv.NewPlaintext(param, ct.Level())
	// set scale otherwise the calculation is incorrect
	pt.Scale = ct.Scale
``` 

Trace now for #1379 (plaintext modulus 65537)
```
[1 2 3 4 5 6 7 8]
Noise: 22.36 Budget 112
[2 3 4 5 6 7 8 9]
Noise: 22.54 Budget 112
[2 6 12 20 30 42 56 72]
Noise: 49.83 Budget 112
[2 6 12 20 30 42 56 72]
Noise: 49.83 Budget 112
[30 42 56 72 2 6 12 20]
Noise: 49.83 Budget 112
[32 48 68 92 32 48 68 92]
Noise: 50.39 Budget 112
[68 92 32 48 68 92 32 48]
Noise: 50.39 Budget 112
[100 140 100 140 100 140 100 140]
Noise: 51.00 Budget 112
[140 100 140 100 140 100 140 100]
Noise: 51.00 Budget 112
[240 240 240 240 240 240 240 240]
Noise: 51.48 Budget 112
[240 240 240 240 240 240 240 240]
Noise: 22.41 Budget 65
[0 0 0 0 0 0 0 240]
Noise: 40.01 Budget 65
[240 0 0 0 0 0 0 0]
Noise: 40.01 Budget 65
[240 0 0 0 0 0 0 0]
Noise: 40.01 Budget 65
[240 0 0 0 0 0 0 0]
Noise: 22.36 Budget 25
```

ZenithalHourlyRate:
Also, for publickey encryption, Lattigo gives the error of magnititude 22 (close to mod reduce error); where Openfhe gives the error of 26 (close to what noise model predicts).

This is because Lattigo does the public key encryption in QP then mod reduce it to Q, so the error looks like being mod reduced.

Openfhe does the public key encryption in Q, but it has a FLEXIBLEAUTOEXT mode to achieve similar effect (mod reduce before the first multiplication).

So for lattigo, the noise model could be refined and the option `include-first-mul` for `secret-insert-mgmt-bgv` is not useful.

Lattigo also supports secret key encryption which gives even smaller error (magnititude 19)

Ref https://pkg.go.dev/github.com/tuneinsight/lattigo/v6@v6.1.0/core/rlwe#Encryptor.EncryptZero

j2kun (approved):

PR #1382, opened 2025-02-09
Lattigo: Support Leveled SHE part of CKKS

ZenithalHourlyRate:
This is a mirror of supporting BGV in Lattigo, of course, the LHE part.

GO does not have f8/f16 so I just use float32 for them when emitting GO code.

This yet does not support full feature set of CKKS in HEIR, like bootstrapping, 2dim vector, affine for loop.


lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
```diff
@@ -0,0 +1,166 @@
...
+}
+
+def Lattigo_CKKSEncodeOp : Lattigo_CKKSOp<"encode"> {
+  let summary = "Encode a plaintext value in the Lattigo CKKS dialect";
+  let description = [{
+    This operation encodes a plaintext value using the specified encoder in the Lattigo CKKS dialect.
+  }];
+  let arguments = (ins
+    Lattigo_CKKSEncoder:$encoder,
+    AnyType:$value,
```
asraa, line 52: why not use RankedTensorType here instead of using the custom verifier? (Here and for the decodeop result)

lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
```diff
@@ -0,0 +1,166 @@
...
+}
+
+def Lattigo_CKKSRotateOp : Lattigo_CKKSOp<"rotate"> {
+  let summary = "Rotate slots of a ciphertext in the Lattigo CKKS dialect";
+  let description = [{
+    This operation rotates slots of a ciphertext value in the Lattigo CKKS dialect.
+
+    For CKKS, the maximum number of slots is N/2.
+
+    offset is valid in (-N/4, N/4).
```
asraa, line 156: worth putting a verifier for this condition?

lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
ZenithalHourlyRate: Changed the AnyType to AnyRankedTensor. Verifier still needed for element type though.

j2kun: You should be able to simplify this further by using `RankedTensorOf<[AnyInteger]>`, and for the float/complex cases you can give `RankedTensorOf<[AnyFloat, Complex<AnyFloat>]>` I think this is enough to remove the verifiers in many of these ops.

ZenithalHourlyRate: Used `RankedTensorOf<[AnyInteger]>` and `AllTypesMatch` to remove the verifiers.

ZenithalHourlyRate:
Rebased to resolve merge conflict.

j2kun (approved):

PR #1383, opened 2025-02-10
LWE: call debug func for arguments

ZenithalHourlyRate:
Separated from #1343. Added test.

### DIFF
--- a/lib/Dialect/LWE/Transforms/AddDebugPort.cpp
+++ b/lib/Dialect/LWE/Transforms/AddDebugPort.cpp
@@ -73,22 +73,34 @@ LogicalResult insertExternalCall(func::FuncOp op, Type lwePrivateKeyType) {
   // implicit assumption the first argument is private key
   auto privateKey = op.getArgument(0);
 
-  ImplicitLocOpBuilder b =
-      ImplicitLocOpBuilder::atBlockBegin(module.getLoc(), module.getBody());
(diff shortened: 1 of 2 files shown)


lib/Dialect/LWE/Transforms/AddDebugPort.cpp
```diff
@@ -73,22 +73,33 @@ LogicalResult insertExternalCall(func::FuncOp op, Type lwePrivateKeyType) {
...
   auto privateKey = op.getArgument(0);
 
-  ImplicitLocOpBuilder b =
-      ImplicitLocOpBuilder::atBlockBegin(module.getLoc(), module.getBody());
+  ImplicitLocOpBuilder b = ImplicitLocOpBuilder::atBlockBegin(
+      op.getLoc(), &op.getBody().getBlocks().front());
+
+  auto insertCall = [&](Value value) {
+    Type valueType = value.getType();
+    if (auto lweCiphertextType = dyn_cast<NewLWECiphertextType>(valueType)) {
```
asraa, line 81: this won't work for arguments that have some shaped type of a ciphertext, right? maybe add a quick comment here, but i assume we will eventually have tensors of ciphertexts after we figure out layouts properly

asraa (approved):

lib/Dialect/LWE/Transforms/AddDebugPort.cpp
ZenithalHourlyRate: Added the comment.

Issue #266, opened 2023-11-17
Implement a polynomial approximation solver


j2kun:
Reading through https://arxiv.org/abs/2311.08610 and I wanted to highlight some of the polynomial approximation techniques that we could incorporate into HEIR when the time is right:

> 2.3 Polynomial Approximation
> Polynomial networks are commonly obtained by approximating the non-polynomial functions of pretrained networks, e.g., (Lee et al., 2022d; Takabi et al., 2019; Mohassel and Zhang, 2017; Hesamifard et al., 2017; Lee et al., 2021), or by substituting ReLU during or after a dedicated training process (Baruch et al., 2022, 2023). The Remez  algorithm (Remez, 1934; Pachón and Trefethen,  2009; Egidi et al., 2020) is commonly used for finding the optimal polynomial approximation of a function f(x) in a certain degree within a predefined range [a, b], assuming a uniform distribution of x. Alternatively, iterative methods such as  the Newton–Raphson method (Raphson, 1702) offer another polynomial approximation approach.  Specifically, (Panda, 2022) focused on approximating √(1/x)  in the interval [a, b], by dividing the interval  into sub-intervals and approximating over each via Newton–Raphson method, before aggregating the results with another polynomial. However, the input range to the non-polynomial layers [a, b] can be extremely large, which results in poor and no practical approximations. This paper employs polynomials for ReLU, as defined in (Lee et al., 2021), for layer normalization (inverse square root) from (Panda, 2022) and for GELU using polynomials derived from the Remez algorithm.

j2kun:
Looking to the remez algorithm, if we want a port/implementation we will need four components as dependencies:

- An arbitrary precision number implementation (APInt or MPInt)
- A linear system solver, perhaps with eigen since we want to incorporate OR-tools into the build anyway http://eigen.tuxfamily.org/index.php?title=Lapack
- An extrema-finding algorithm; (successive parabolic interpolation or Brent’s method), not sure if there is a good dep we could use to get this for free
- A root-finding algorithm (bisection method should be simple to implement)

j2kun:
Reference implementation: https://github.com/samhocevar/lolremez

j2kun:
Maks also pointed me to an LLVM-internal ConstraintSystem solver https://github.com/llvm/llvm-project/blob/2c875719c841ff13b9b250e6ea97fc3e0aca2070/llvm/include/llvm/Analysis/ConstraintSystem.h#L4

This isn't quite a linear system solver, but may come in handy.

[...]

PR #1396, opened 2025-02-11
Add Caratheodory-Fejer approximation

j2kun:
Part of https://github.com/google/heir/issues/266

This PR adds the Caratheodory-Fejer polynomial approximation routine for [-1, 1].

Next: Support non-unit intervals by rescaling.

This PR also fixes a bug in which I was using Chebyshev polynomials of the second kind instead of chebyshev polynomials of the first kind, which was revealed once I tried to actually convert the output of the CF approximation (which is in the Chebyshev basis) to the monomial basis.

### DIFF
--- a/lib/Utils/Approximation/CaratheodoryFejer.cpp
+++ b/lib/Utils/Approximation/CaratheodoryFejer.cpp
@@ -1,0 +1,122 @@
+#include "lib/Utils/Approximation/CaratheodoryFejer.h"
+
+#include <cstdint>
+#include <functional>
+#include <optional>
+
+#include "Eigen/Dense"  // from @eigen
+#include "lib/Utils/Approximation/Chebyshev.h"
+#include "lib/Utils/Polynomial/Polynomial.h"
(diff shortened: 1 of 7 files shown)


lib/Utils/Approximation/CaratheodoryFejer.cpp
```diff
@@ -0,0 +1,122 @@
...
+  for (int32_t i = 1; i < eigenvalues.size(); ++i) {
+    if (std::abs(eigenvalues(i)) > maxEigenvalue) {
+      maxEigenvalue = std::abs(eigenvalues(i));
+      maxIndex = i;
+    }
+  }
+  VectorXd maxEigenvector = eigenvectors.col(maxIndex);
+
+  // A debug for comparing the eigenvalue solver with the reference
+  // implementation.
```
WoutLegiest, line 76: An option to put it into an LLVM_DEBUG env?

WoutLegiest (approved):
Cool and efficient implementation

asraa (approved):

PR #1379, opened 2025-02-08
BGV: Populate Moduli Chain from Noise Analysis

ZenithalHourlyRate:
Depends on #1343. Implements part of #1168.

For `dot_product.mlir`, it will generate the following parameter:

```
ringDim: 8192
plaintextModulus: 65537
level: 2
logqi: 25 40 47 
qi: 33832961 1099511922689 140737488486401 
dnum: 2
logpi: 47 47 
pi: 140737488928769 140737489256449 
```

### Idea

The idea behind the parameter generation is simple: For each level, the mod reduce should be able to decrease the noise by that amount (e.g. from 69.5 to 23), so the modulus for that level should be at least 47 bit.

For special modulus, the selection is conservative in that we only make sure `P > Q / dnum` and make no other optimization. We default to the HYBRID key switching technique for now.

I understand there are implementation using optimizers like KPZ21 and MML+23. This PR is meant for making the pipeline aware of concrete parameters, like `LWE` dialect and backend. Further PR could optimize the selection process.

### Current status / Discussion

* The parameter selection procedure lives in `--validate-noise` pass for simplicity. It should be separated into another pass use some attribute to communicate with other passes (e.g. `bgv.schemeParam` attribute). Need design discussion.
* `LWE` dialect (`secret-to-bgv`) has not been touched. I now directly pass the generated parameter to attr, then backend will pick it up somewhere; further design should pass the information in `LWE` Types and backend will get information there.
* For the prime generation part, I directly use utility functions from `@openfhe//nbtheory.h`. It is quite crude for now.
* We need a way for user to specify the plaintext modulus. The current default modulus 4295294977 is too large, though when set it to smaller one the box_blur test (iirc) will fail.

### Example

The noise analysis for the generated param

```
Noise Bound: 29.27 Budget: 81 Total: 112 for value: <block argument> of type 'tensor<8xi16>' at index: 0 
Noise Bound: 29.27 Budget: 81 Total: 112 for value: <block argument> of type 'tensor<8xi16>' at index: 1 
Noise Bound: 66.53 Budget: 44 Total: 112 for value: %1 = arith.muli %input0, %input1 {mgmt.mgmt = #mgmt.mgmt<level = 2, dimension = 3>} : tensor<8xi16> 
Noise Bound: 66.53 Budget: 44 Total: 112 for value: %2 = mgmt.relinearize %1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<8xi16> 
[...]

PR #1400, opened 2025-02-12
CKKS: Generate Scheme Param

ZenithalHourlyRate:
In a similar style with #1379.

CKKS noise analysis is not as suggestive as BGV, which only gives the precision info, while the later can be used for determining overflow or not (correctness). So we may only support user providing the the default scale and the first mod size and we generate the parameter for them.

There are many scale-related techniques like https://github.com/google/heir/issues/1169#issuecomment-2616357068 and https://github.com/google/heir/issues/1169#issuecomment-2617786886. This PR only serves as a baseline for making the pipeline work, as it just generates the primes in an increasing order without caring the scaling factor. These management passes should be implemented later.

I put the parameter generation process in `secret-to-ckks`. As we are not using noise model (so not `validate-noise`), we do not have other pass to put it.

The CKKS SchemeParamAttr is for later integration like the Lattigo CKKS backend #1382.

### DIFF
--- a/lib/Dialect/CKKS/IR/CKKSAttributes.td
+++ b/lib/Dialect/CKKS/IR/CKKSAttributes.td
@@ -1,21 +1,21 @@
-#ifndef LIB_DIALECT_BGV_IR_BGVATTRIBUTES_TD_
-#define LIB_DIALECT_BGV_IR_BGVATTRIBUTES_TD_
+#ifndef LIB_DIALECT_CKKS_IR_CKKSATTRIBUTES_TD_
+#define LIB_DIALECT_CKKS_IR_CKKSATTRIBUTES_TD_
 
-include "BGVDialect.td"
+include "lib/Dialect/CKKS/IR/CKKSDialect.td"
 
 include "mlir/IR/AttrTypeBase.td"
 include "mlir/IR/DialectBase.td"
 include "mlir/IR/OpBase.td"
 
-class BGV_Attribute<string attrName, string attrMnemonic>
-    : AttrDef<BGV_Dialect, attrName> {
+class CKKS_Attribute<string attrName, string attrMnemonic>
+    : AttrDef<CKKS_Dialect, attrName> {
     let mnemonic = attrMnemonic;
     let assemblyFormat = "`<` struct(params) `>`";
 }
 
-def BGV_SchemeParam
-    : BGV_Attribute<"SchemeParam", "scheme_param"> {
-    let summary = "BGV Scheme Parameters";
+def CKKS_SchemeParam
+    : CKKS_Attribute<"SchemeParam", "scheme_param"> {
+    let summary = "CKKS Scheme Parameters";
     let description = [{
       This attribute is used for recording the scheme parameters for CKKS.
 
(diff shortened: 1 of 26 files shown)


lib/Dialect/CKKS/IR/CKKSAttributes.td
```diff
@@ -0,0 +1,40 @@
...
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
+
+class CKKS_Attribute<string attrName, string attrMnemonic>
+    : AttrDef<CKKS_Dialect, attrName> {
+    let summary = "Attribute for CKKS";
+    let description = [{
+      This attribute represents the values for CKKS.
+    }];
```
j2kun, line 15:
```suggestion
```

Since it's a base class, the def should always override these fields. They can be removed

lib/Dialect/Secret/Conversions/SecretToCKKS/SecretToCKKS.cpp
```diff
@@ -272,16 +276,46 @@ class SecretGenericTensorInsertConversion
...
-        getRlweRNSRing(context, maxLevel, coefficientModBits, polyModDegree);
+    // generate scheme parameters
+    auto maxLevel = getMaxLevel();
+    std::vector<double> logPrimes;
+    logPrimes.push_back(firstModBits);
+    for (int i = 0; i < maxLevel; i++) {
+      logPrimes.push_back(scalingModBits);
+    }
+
+    auto schemeParam = ckks::SchemeParam::getConcreteSchemeParam(45, logPrimes);
```
j2kun, line 312: Should this 45 use the pass flag?

lib/Parameters/CKKS/Params.cpp
j2kun, line 1: A lot of stuff in this file looks copied from BGV/Params.cpp. Maybe extract the prime generation to a new library in `lib/Utils/NumberTheory.{h,cpp}`? Or maybe a new `Utils` lib in `lib/Parameters` that includes the prime generation and the other common functions like `computeDnum` and `computeRingDim` and the `HEStd_128_classic`

Then it seems a large part of `getConcreteSchemeParams` is also copied. Because that function is so large, it's hard for me to see what differs between the BGV/CKKS versions of it. It seems like the only difference is `logDefaultScale` replaced `plaintextModulus`, and `logDefaultScale` is passed through while `plaintextModulus` is used as the first prime. It might be worth abstracting the common parts so they can be reused to make it easier to keep the two implementations in sync and also to show the differences between them more clearly. It's not clear to me what's the right way to do this, though. Is it possible to have a baseline RLWE parameter generation that is agnostic of BGV/CKKS, and then modify/pass particular options to do the CKKS/BGV specific parts?

j2kun (changes requested):

lib/Parameters/CKKS/Params.cpp
ZenithalHourlyRate: Now I separate them into

* `RLWESecurityParams`, including the HEStd_128_classic (updated to the recent "Security guidelines for implementing homomorphic encryption") and computeRingDim
* `RLWEParams`, for common fields and the prime generation
* `{BGV/CKKS}/Params`, for special field for BGV/CKKS

For now I think it is OK as we are only making the pipeline work. In the future for CKKS there might be a different prime generation process living in `CKKS/Params.cpp` because Openfhe has a level specific scaling factor and to maintain that Openfhe uses a carefully designed algorithm in "Approximate Homomorphic Encryption with Reduced Approximation Error"

PR #1403, opened 2025-02-13
Adding support of the Hello World Small example to Arith-to-CGGI

WoutLegiest:
Extending the CGGI Ops with the required operations to support the Hello World Small Example.

Updating the Arith to CGGI pass to work with these new operations, including 'deeper' checking for when an operation needs to go to CGGI and when to stay in arith. 
In the Hello World Small example, some operations can stay in the arith domain. Since these constants are stored in a memref, the compiler cannot fold them away.


WoutLegiest:
Goal is to add signed operations to the CGGI pipeline.

WoutLegiest:
And small update to the boolean vectorizer test

lib/Dialect/CGGI/IR/CGGIOps.td
```diff
@@ -26,6 +27,10 @@ class CGGI_Op<string mnemonic, list<Trait> traits = []> :
 
 // --- Operations for a gate-bootstrapping API of a CGGI library ---
 
+/*******************
+ * BOOL OPERATIONS *
```
j2kun, line 31: If you feel it's too messy in one file, one thing you can do is split this into multiple files, a `CGGIOpBase.td` for the base class, and then separate `BoolOps.td`, etc. for each class of ops. This is how it's done in, say, TOSA upstream: https://github.com/llvm/llvm-project/tree/main/mlir/include/mlir/Dialect/Tosa/IR

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -51,15 +51,53 @@ static bool allowedRemainArith(Operation *op) {
...
         // Default case for operations that don't match any of the types
         return false;
       });
 }
 
 static bool hasLWEAnnotation(Operation *op) {
-  return static_cast<bool>(
-      op->getAttrOfType<mlir::StringAttr>("lwe_annotation"));
+  auto check =
+      static_cast<bool>(op->getAttrOfType<mlir::StringAttr>("lwe_annotation"));
```
j2kun, line 74: I think this returns nullptr if it's not found, so you could just store it as a StringAttr and then compare it (implicitly) to nullptr. Same as below

lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
```diff
@@ -51,15 +51,53 @@ static bool allowedRemainArith(Operation *op) {
...
+
+  // Check recursively if a defining op has a LWE annotation
+  return llvm::TypeSwitch<Operation *, bool>(op)
+      .Case<mlir::arith::ExtUIOp, mlir::arith::ExtSIOp, mlir::arith::TruncIOp>(
+          [](auto op) {
+            if (auto *defOp = op.getIn().getDefiningOp()) {
+              return hasLWEAnnotation(defOp);
+            }
+            return static_cast<bool>(
+                op->template getAttrOfType<mlir::StringAttr>("lwe_annotation"));
```
j2kun, line 86: This check seems duplicated with the one on line 73.

j2kun (approved):

lib/Dialect/CGGI/IR/CGGIOps.td
WoutLegiest: Thanks! I would indeed be better to split it out into different files

j2kun (approved):

PR #1411, opened 2025-02-13
Adding support of the Hello World Small example to CGGI-to-tfhe-rs

WoutLegiest:
Further upgrade of the lowering of the Hello World Example to CGGI-to-tfhe-rs by extending the operations and introducing a new type: the tfhe_rust.bool, which can be using using the tfhe-rs parameters and not the tfhe-rs-bool implementation


WoutLegiest:
+ Update to the bool vectorizer again

j2kun (approved):

PR #1421, opened 2025-02-18
Detect entry-function using heuristics

### LINKED ISSUE #1418: Detect `entry-function` for `configure-crypto-context` and `add-debug-port`
Currently the user has to specify the `entry-function` to properly get the passes in title working. When the function name is missing or incorrect there is no warning and some wanted feature may silently fail or result in cryptic error, e.g. #1414.

We can safely detect the entry function using the following heuristic

* no double underline `__` (often generated by heir)
* has lwe type in its argument
* the first function satisfying the above requirement

For multi-function support we may need to use a `UnitAttr` to determine which one is the entry, but this situation is infrequent for now.

ZenithalHourlyRate:
Fixes #1418 

The detection code could be put into a separate compilation unit, looking for where to put it.

There are also entry-function option for `tosa-to-boolean-jaxite` but as I am not so familiar there I did not touch them.

### DIFF
--- a/lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
@@ -11,6 +11,7 @@
 #include "lib/Dialect/Lattigo/IR/LattigoOps.h"
 #include "lib/Dialect/Lattigo/IR/LattigoTypes.h"
 #include "lib/Dialect/ModuleAttributes.h"
+#include "lib/Utils/TransformUtils.h"
 #include "llvm/include/llvm/Support/raw_ostream.h"      // from @llvm-project
 #include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
 #include "mlir/include/mlir/IR/BuiltinAttributes.h"     // from @llvm-project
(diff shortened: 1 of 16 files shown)


lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -282,17 +282,51 @@ struct ConfigureCryptoContext
...
-  void runOnOperation() override {
-    auto result =
-        getOperation()->walk<WalkOrder::PreOrder>([&](func::FuncOp op) {
-          auto funcName = op.getSymName();
-          if ((funcName == entryFunction) && failed(convertFunc(op))) {
-            op->emitError("Failed to configure the crypto context for func");
-            return WalkResult::interrupt();
-          }
+  func::FuncOp getEntryFunction() {
+    ModuleOp moduleOp = cast<ModuleOp>(getOperation());
```
asraa, line 286: could you please factor some of this logic into a utility that can be used in the other pass pipelines please?

ZenithalHourlyRate: Put it in `lib/Utils/TransformUtils.h` now

asraa (approved):
thanks!

PR #1422, opened 2025-02-18
Pass Attribute to debug function call

### LINKED ISSUE #1415: Support passing attribute to debug function call
In `lwe-add-debug-port`, for each value there is a `__heir_debug` function call corresponding to it, like

```mlir
%0 = bgv.add ...
call @__heir_debug(%sk, %0)
```

For passes like `--validate-noise`, they would calculate some property for `%0` like `bound = 26.5`, and such analysis result could be passed to the debug function for further uses

```mlir
%0 = bgv.add ... {bound = "26.5"}
call @__heir_debug(%sk, %0) {bound = "26.5", op = "bgv.add"} 
```

For translater, the signature of the debug function would need to change to

```c++
// c++ style
void __heir_debug(PrivateKeyT, CiphertextT, std::map<std::string, std::string>);
```

and the translater is responsible for translating the attribute to the `map`.

ZenithalHourlyRate:
Fixes #1415

Because heir-translate does not have the `func::FuncOp OpAsmOpInterface` hack some `%arg0` is present instead of `%evaluator` or `%cc`. Should be fixed later when I get `OpAsmTypeInterface` integrated upstream in MLIR (https://github.com/llvm/llvm-project/pull/124700).

<del>For Openfhe, the debug output is quite verbose because of LWE type. For printing op in generic form by AsmPrinter we can not have type alias.</del> resolved now with post-processing in debug function

#### Lattigo

```
<block argument> 
  [1 2 3 4 5 6 7 8]
  Noise: 22.30 Total: 96
<block argument> 
  [2 3 4 5 6 7 8 9]
  Noise: 22.49 Total: 96
%ct = lattigo.bgv.mul %arg0, %arg4, %arg5 
  [2 6 12 20 30 42 56 72]
  Noise: 49.79 Total: 96
%ct_0 = lattigo.bgv.relinearize %arg0, %ct 
  [2 6 12 20 30 42 56 72]
  Noise: 49.79 Total: 96
```

#### Openfhe

```
<block argument> 
  ( 1 2 3 4 5 6 7 8 ... )
  cv 2 Ql 4 logQ: 122.323 logqi: [ 31.0008 37 37 17.3219 ] budget 94.7219 noise: 26.6009
<block argument> 
  ( 2 3 4 5 6 7 8 9 ... )
  cv 2 Ql 4 logQ: 122.323 logqi: [ 31.0008 37 37 17.3219 ] budget 94.843 noise: 26.4798
%ct = openfhe.mul_no_relin %arg0, %arg2, %arg3 
  ( 2 6 12 20 30 42 56 72 ... )
  cv 3 Ql 3 logQ: 105.001 logqi: [ 31.0008 37 37 ] budget 52.7384 noise: 51.2625
%ct_0 = openfhe.relin %arg0, %ct 
  ( 2 6 12 20 30 42 56 72 ... )
  cv 2 Ql 3 logQ: 105.001 logqi: [ 31.0008 37 37 ] budget 52.7384 noise: 51.2625
```

### DIFF
--- a/tests/Examples/openfhe/dot_product_8_debug_test.cpp
+++ b/tests/Examples/openfhe/dot_product_8_debug_test.cpp
@@ -44,15 +44,26 @@ DCRTPoly DecryptCore(const std::vector<DCRTPoly>& cv,
   return b;
 }
 
+#define OP
 #define DECRYPT
 #define NOISE
 
-void __heir_debug(CryptoContextT cc, PrivateKeyT sk, CiphertextT ct) {
+void __heir_debug(CryptoContextT cc, PrivateKeyT sk, CiphertextT ct,
+                  const std::map<std::string, std::string>& debugAttrMap) {
+#ifdef OP
+  auto isBlockArgument = debugAttrMap.at("asm.is_block_arg");
+  if (isBlockArgument == "1") {
+    std::cout << "Input" << std::endl;
+  } else {
+    std::cout << debugAttrMap.at("asm.op_name") << std::endl;
+  }
+#endif
+
 #ifdef DECRYPT
   PlaintextT ptxt;
   cc->Decrypt(sk, ct, &ptxt);
   ptxt->SetLength(8);
-  std::cout << ptxt << std::endl;
+  std::cout << "  " << ptxt << std::endl;
 #endif
 
 #ifdef NOISE
@@ -73,7 +84,7 @@ void __heir_debug(CryptoContextT cc, PrivateKeyT sk, CiphertextT ct) {
     logQ += logqi;
   }
 
-  std::cout << "cv " << cv.size() << " Ql " << sizeQl << " logQ: " << logQ
+  std::cout << "  cv " << cv.size() << " Ql " << sizeQl << " logQ: " << logQ
             << " logqi: " << logqi_v << " budget " << logQ - noise - 1
             << " noise: " << noise << std::endl;
 #endif
--- a/lib/Target/Lattigo/LattigoEmitter.cpp
+++ b/lib/Target/Lattigo/LattigoEmitter.cpp
@@ -164,12 +164,46 @@ LogicalResult LattigoEmitter::printOperation(func::ReturnOp op) {
 }
 
 LogicalResult LattigoEmitter::printOperation(func::CallOp op) {
+  // build debug attribute map for debug call
+  auto debugAttrMapName = getDebugAttrMapName();
+  if (isDebugPort(op.getCallee())) {
+    os << debugAttrMapName << " := make(map[string]string)\n";
+    for (auto attr : op->getAttrs()) {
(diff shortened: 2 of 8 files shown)


tests/Examples/openfhe/dot_product_8_debug_test.cpp
```diff
@@ -44,15 +44,33 @@ DCRTPoly DecryptCore(const std::vector<DCRTPoly>& cv,
...
+#define OP
 #define DECRYPT
 #define NOISE
 
-void __heir_debug(CryptoContextT cc, PrivateKeyT sk, CiphertextT ct) {
+void __heir_debug(CryptoContextT cc, PrivateKeyT sk, CiphertextT ct,
+                  const std::map<std::string, std::string>& debugAttrMap) {
+#ifdef OP
+  auto op = debugAttrMap.at("op");
+  auto sepBlockArgument = op.find("of type");
```
j2kun, line 55: What kind of of attribute generates text like this? It looks like it's a printed block argument (`block argument at index #0 of type...`). Maybe there's an additional special case you can add to the map builder to avoid the string munging here and pass exactly what you want?

j2kun (approved):

tests/Examples/openfhe/dot_product_8_debug_test.cpp
ZenithalHourlyRate: > What kind of of attribute generates text like this?

```cpp
    // Use AsmPrinter to print Value to print the defining op
    auto ciphertext = op->getOperand(op->getNumOperands() - 1);
    os << debugAttrMapName << R"(["op"] = ")" << ciphertext << "\"\n";
```

Here I called `operator<<(raw_osstream&, Value)` to print the Value, which will be either `<block argument> of type (!type) at index` or `%0 = op %1...`, so it is MLIR AsmPrinter's behavior.

> Maybe there's an additional special case you can add to the map builder to avoid the string munging here and pass exactly what you want?

Well I do can process this in Emitter by copying buffer from raw_ostream, but then that is too reliant on MLIR's behavior. Once the upstream changed the text string it has to adapt to that.

Another point I want to make is that, we should provide full op information to the debug function and let end user decide whether they want to see the verbose type (so the change to the debug function is just a template here), instead of stripping it earlier.

ZenithalHourlyRate: Updated to have more information at construction time.

Though instead of seeing detailed asm like op with operand name and result name.

```
%ct = openfhe.mul_no_relin %arg0, %arg2, %arg3 
   ...
%ct_0 = openfhe.relin %arg0, %ct 
   ...
```

we only get the op name for convenience now.

```
Input
  ...
Input
  ...
openfhe.mul_no_relin
  ...
openfhe.relin
  ...
```

j2kun (approved):

PR #1430, opened 2025-02-19
Preserve attribute during secret.generic lowering

### LINKED ISSUE #1426: Support preserving attributes in various lowering
For debugging, it is especially useful to preserve some attribute during lowering, like comparing the noise analysis result with real bound that can only be obtained in runtime in debug function

```mlir
(validate-noise annotate the result)
arith.add {bound = 50}
(lowered to)
bgv.add {bound = 50}
(lowered to)
lwe.radd {bound = 50}
(then passed to the lwe debug function)
```

Technically, we need to copy the attribute in all the lowering code.

This generally falls in the category of Attribute (lifetime) management. Related issues are #1414, #1415

There is also an issue related to refactoring secret.generic op lowering #1413. I might make a quick change to these lowerings if these refactoring is not on the way now.

ZenithalHourlyRate:
Fixes #1426 

* `secret-distribute-generic` preserves the dialect attrs from secret.generic op argument to the func.func op argument
* `secret-to-<scheme>` preserves the dialect attrs from secret.generic's inner op to converted op
* `<scheme>-to-lwe` preserves the attribute originally
* `lwe-add-debug-port` will transplant the attr to the CallOp, for debugging purpose, see #1422 
* `lwe-to-<backend>` will preserve the attr in CallOp
  + It will discard attributes for backend ops, so `lwe.add {attr}` will be converted to `backend.add` instead of `backend.add {attr}`
  + Due to MLIR issue, these attributes are lost during type conversion, so there is no test now, should be fixed after https://github.com/llvm/llvm-project/pull/127772 is in
  
Conceptually, this PR suggests that dialect attribute should be preserved and managed by the programmer (of corresponding dialect) while other attribute has no guarantee on how long it will live.

### DIFF
--- a/lib/Utils/ConversionUtils.h
+++ b/lib/Utils/ConversionUtils.h
@@ -248,7 +248,23 @@ class SecretGenericOpConversion
       if (failed(result)) return failure();
     }
 
-    return matchAndRewriteInner(op, resultTypes, inputs, innerOp.getAttrs(),
+    // only preserve dialect attrs
+    // we do not want op attrs like overflowFlags from arith.add
+    SmallVector<NamedAttribute> dialectAttrs;
+    for (auto &namedAttr : innerOp.getDialectAttrs()) {
+      dialectAttrs.push_back(namedAttr);
+    }
+
+    // special treatment for memref.alloc, which has alignment attr
+    // that should be preserved
+    // TODO: secret-to-cggi should handle this instead of here
+    for (auto &namedAttr : innerOp.getAttrs()) {
+      if (namedAttr.getName().getValue() == "alignment") {
+        dialectAttrs.push_back(namedAttr);
+      }
+    }
+
+    return matchAndRewriteInner(op, resultTypes, inputs, dialectAttrs,
                                 rewriter);
   }
 
@@ -278,7 +294,8 @@ class SecretGenericOpCipherConversion : public SecretGenericOpConversion<T, Y> {
         })) {
       return failure();
     }
-    rewriter.replaceOpWithNewOp<Y>(op, outputTypes, inputs);
+    rewriter.replaceOpWithNewOp<Y>(op, outputTypes, inputs)
+        ->setDialectAttrs(attributes);
     return success();
   }
 };
@@ -324,7 +341,8 @@ class SecretGenericOpCipherPlainConversion
         ciphertextTy.getPlaintextSpace().getEncoding(),
         ciphertextTy.getPlaintextSpace().getRing());
 
-    rewriter.replaceOpWithNewOp<Y>(op, ciphertext, plaintext);
+    rewriter.replaceOpWithNewOp<Y>(op, ciphertext, plaintext)
+        ->setDialectAttrs(attributes);
     return success();
   }
 };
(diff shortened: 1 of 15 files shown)


lib/Utils/ConversionUtils.h
```diff
@@ -248,7 +248,23 @@ class SecretGenericOpConversion
...
+    // we do not want op attrs like overflowFlags from arith.add
+    SmallVector<NamedAttribute> dialectAttrs;
+    for (auto &namedAttr : innerOp.getDialectAttrs()) {
+      dialectAttrs.push_back(namedAttr);
+    }
+
+    // special treatment for memref.alloc, which has alignment attr
+    // that should be preserved
+    // TODO: secret-to-cggi should handle this instead of here
+    for (auto &namedAttr : innerOp.getAttrs()) {
```
j2kun, line 261: Don't we actually want to preserve the attributes on the `secret.generic` op? That's where the `mgmt.mgmt` attributes are set during the lowering, no?

I bring this up because I was working on a refactor of the context-aware type conversion in https://github.com/google/heir/pull/1428 and I noticed this was happening, and the patterns were not able to find the mgmt attribute when only looking at the inner op.

ZenithalHourlyRate: Well I did a special treatment for mgmt.mgmt

* First it is annotated on the inner op: `%0 = arith.add {mgmt}`
* in `secret-distribute-generic`, the mgmt attr is moved to secret.generic, so `%0 = secret.generic {mgmt} { arith.add }`
* context aware type converter for `%0` then will look up the mgmt attr in `secret.generic` as now it becomes `%0`'s defining op so there is no inner op involved

So it is `secret-distribute-generic`'s responsibility to maintain `secret.generic`'s attribute. Generally there are no other attribute in `secret.generic` as only `secret-distribute-generic` will create these ops (well for the wrap-generic created secret.generic its attribute should be cleaned, represented by `moveDialectAttrsToFuncArgument` here). All meaningful other attribute are annotated in the inner op.

j2kun: Sorry, I think I misspoke. I was running into a situation like this:

```
%0 = secret.generic {mgmt1}
  ...
%1 = secret.generic (%0) {mgmt2}
```

When the lowering converts the first generic to a `ckks` op, it doesn't currently preserve any attributes. So when the second generic is processed, it looks for the `mgmt` attr of the replacement for `%0`, which had no attribute at all.

In my PR I originally tried copying over just the inner op's attrs to the new `ckks` op, but the `mgmt` attr was not present on the inner op, just the enclosing generic.

I'm actually still confused about how the current code was able to find the mgmt attr in this situation, given that it should not have had access to the original op that defined `%0` after the replacement was made.

ZenithalHourlyRate: > When the lowering converts the first generic to a ckks op, it doesn't currently preserve any attributes.

I intentionally made so. In old version of my PR I did preserve the mgmt attribute from first secret.generic to a ckks op and expect the second secret.generic to find the mgmt attribute there, but after inspecting OpAdaptor during conversion I find the following fun fact about DialectConversion:

During DialectConversion the old op and new op will both be present and the uses of the old op, while asked `rewriter` to `replaceUses`, is not _committed_ until the last step (which is called unlink internally in DialectConversion). That is why all op modificaiton in DialectConversion has to happen through the rewriter. So for the second op, it will find the following IR

```mlir
%0 = secret.generic {mgmt1}
%0_new = ckks.op /* no_attr */
  ...
// %0 has not been replaced by %0_new in %1's view!
%1 = secret.generic (%0) {mgmt2}
```

This is not undefined behavior but intended behavior of DialectConversion in MLIR, because ideally op conversion should not affect each other and the view of each `ConvertOpRewritePattern` should be the original IR instead of the intermediate IR, so it is guaranteed for the second op to find the original first op.

ZenithalHourlyRate:
From the discussion (Cc @AlexanderViand-Intel ) on debug function being unaware of LWE underlying type, after #1422 is in we can now pass the information to the debug function

1. `lwe-add-debug-port` will now add an attribute `message.size` to the debug call

```mlir
    %ct_11 = bgv.rotate %ct_10 {offset = 7 : index} : !ct_L1_1
    call @__heir_debug_2(%sk, %ct_11) {message.size = "8"} : (!skey_L2_, !ct_L1_1) -> ()
    %ct_12 = lwe.reinterpret_underlying_type %ct_11 : !ct_L1_1 to !ct_L1_
    call @__heir_debug_3(%sk, %ct_12) {message.size = "1"} : (!skey_L2_, !ct_L1_) -> ()
```

2. In `BackendEmitter`, such attribute is passed to a `debugFuncAttrMap` debug func arg (of type `std::map<string, string>` in C++) and debug function body can use it at will

```mlir
  debugAttrMap11["message.size"] = "8"
  __heir_debug(cc, sk, ct11, debugAttrMap11);
```

Should add test in this PR after the main branch is fixed.

ZenithalHourlyRate:
> Should add test in this PR after the main branch is fixed.

Rebased and added test on message.size attr

j2kun (approved):

PR #1435, opened 2025-02-20
Migrate TypeAsmInterface to OpAsmTypeInterface

ZenithalHourlyRate:
After the upstreaming of https://discourse.llvm.org/t/rfc-introduce-opasm-type-attr-interface-for-pretty-print-in-asmprinter/83792, we now have `OpAsmTypeInterface` and `OpAsmAttrInterface` that is directly hooked into the MLIR AsmPrinter, so there is a series of update

* Migrate the `TypeAsmInterface` to `OpAsmTypeInterface getAsmName` method (this PR)
* Mirgate `OpAsmDialectInterface getAlias(Type)` to `OpAsmTypeInterface getAlias`
* Mirgate `OpAsmDialectInterface getAlias(Attr)` to `OpAsmAttrInterface getAlias`

The good thing about using `OpAsmTypeInterface` is that we now do not need to hack into `func::FuncOp OpAsmOpInterface` to make it also work for function argument; also all our op and op from MLIR (like `affine.for`) would respect the TypeInterface instead of manually hook into their OpInterface.

One behavior change is the grouping behavior of multi-result op. Previously it would be `%0:4` now it becomes four separate SSA.


j2kun (approved):
Thank you so much for upstreaming this!

Issue #1186, opened 2024-12-13
Lattigo: Correctly lowering 1-dim bgv.rotate to RotateColumns/RotateRows for full packing


ZenithalHourlyRate:
Lattigo natively exposes the `Z_t/(X^N+1)` plaintext space as a SIMD packing of a matrix of size `N/2 * 2`, which element in `Z_t`, which is due to the nature of automorphism.

> The Galois automorphisms form a group (under composition), which is isomorphic to `Zn/2 × Z2`. The first factor is generated by `gal_elt = 3`, and the second factor is generated by `gal_elt = m − 1`

> https://www.microsoft.com/en-us/research/uploads/prod/2017/11/sealmanual-2-3-1.pdf , section 5.6

Their API allows RotateColumns by k (either to the left or right, size N/2) or RotateRows (swapping rows).

In the meantime, OpenFHE exposes it just as a vector of `N`, and provides an API generating indexes of rotation of size N.

If we just pack half of the plaintext, it is OK to just use RotateColumns. But if we do full packing for BGV for Lattigo, we need to correctly combine the Lattigo API to provide a 1-dim view of the data.

ZenithalHourlyRate:
OpenFHE also shows a `n/2 by 2` feature, and when rotate right by 1, these two rows does not mix, so we does not have a generic 1-dim bgv.rotate for index > n / 2.

For full packing, `tensor_ext.rotate` has distinct semantic as `bgv.rotate`. The workaround is to always half-full-packing for BGV for now. (Note that in cyclic repetition code [I did a half packing](https://github.com/google/heir/blob/c9dd3afd5f01f682586c4c57762cd5b8fda92958/lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp#L531-L533))

For example, the following code get the full batching for BGV, but when rotated right by 1, we get

```c++
auto batchSize = cryptoContext->GetCryptoParameters()->GetEncodingParams()->GetBatchSize();
    std::vector<int64_t> vectorFull;
    for (size_t i = 0; i < batchSize; i++) {
        vectorFull.push_back(i);
    }
    Plaintext pt = cryptoContext->MakePackedPlaintext(vectorFull);
    auto ct      = cryptoContext->Encrypt(keyPair.publicKey, pt);
    auto ct_rot  = cryptoContext->EvalRotate(ct, -1);
    Plaintext dec;
    cryptoContext->Decrypt(keyPair.secretKey, ct_rot, &dec);
    std::cout << "[";
[...]

PR #1439, opened 2025-02-20
Lattigo: Add more e2e tests

ZenithalHourlyRate:
For more coverage.

Now I come to understand why roberts_cross and box_blur has to select big plaintext modulus: if it is too small, the parameter selection (either by Openfhe or by --validate-noise) will select logN < 13 and the 4096 input vector wont fit in the ciphertext...

You might question: why logN = 12 (namely N = 4096) wont work. This is related to #1186, so in brief when N = 4096 we have the structure Z2048xZ2 so there is only 2048 slots in one row and the heir-vectorizer is not aware of this.

This should also be handled by the layout conversion pass. 

These issues should be handled temporarily in another PR by enabling the parameter selection process to have a minimal logN (specified by ciphertext-degree, the naming issue also related to #1402)

Also fix a bug when `pow(5, index)` becomes so large like `5 ** 16 = 152587890625`

### DIFF
--- a/lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Lattigo/Transforms/ConfigureCryptoContext.cpp
@@ -241,7 +241,11 @@ LogicalResult convertFuncForScheme(func::FuncOp op) {
   // generate Galois Keys on demand
   auto rotIndices = findAllRotIndices(op);
   for (auto rotIndex : rotIndices) {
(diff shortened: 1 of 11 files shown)


j2kun (approved):

j2kun:
FYI, internally I am forced to replace the commented out go_test build rule with `t.Skip` inside the actual test. So this PR will land with that minor patch applied.

Issue #1370, opened 2025-02-06
Which versions of Python should the frontend support?


j2kun:
Since the correctness of the Python will soon (if it doesn't already) depend on the Python version. In particular, we read python bytecode as part of emitting mlir (nontrivially in https://github.com/google/heir/pull/1307) and the bytecode spec changes version to version. Since Python is now on version 3.13, I think supporting as far back as 3.9/3.10 might be overkill.

Also, our dependency numba is [gradually dropping support](https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information) of earlier versions of Python. The version we have currently pinned (numba 0.60) is the last one to support Python3.9 and the last _not_ to support Python3.13. So for now at least we are stuck with Python 3.9-3.12, and if we want to upgrade numba to get support for Python3.13, we would need to drop Python3.9.

AlexanderViand-Intel:
The way I'm expressing the tensor type hints (i.e., `Tensor[16,16, I32]` for `tensor<16x16xi32>` ) in the frontend will require Python 3.11 because that's when variadic generics became available: https://peps.python.org/pep-0646/ , so I think even Python 3.10 support might be difficult.

j2kun:
Sounds like Python3.11 is a minimum version, and the remaining question is, should we aim to additionally support 3.12 & 3.13 as a public-facing guarantee (i.e., add it to a CI version matrix, add the requisite special cases of bytecode, and bump the numba version) or let it remain best-effort?

j2kun:
For now I'm going to say "let's only officially support Python3.11" and call it a day until someone asks for broader support

PR #1445, opened 2025-02-20
Migrate to bazel 8, start migrating to bazelmod

j2kun:
This PR sets up a hybrid Bazelmod + WORKSPACE configuration, where the "hard" dependencies remain in a workspace file while the supported dependencies are migrated to `MODULE.bazel`

This allows us to use the latest bazel version. I also went ahead and bumped versions of various dependencies, including rules_python and rules_hdl, which have material impacts on the project.

Summary of impact:

- Bazel 8 uses `_main` instead of `heir` in many bazel runfiles paths, which broke yosys techmap file finding, lit configuration, and our heir-opt bazel rules.
- Globs that have empty matches are now hard errors
- Go dependencies now must come in a go.mod file
- rules_python now has a different build target naming scheme (`@heir_py_pip_deps_numba//:pkg` -> `@heir_py_pip_deps//numba`)
- rules_python now has stricter hermeticity conditions, breaking our use of LLVM's in-tree `lit` tool, so I moved `lit` to a pip dependency.
- I updated our `rules_python` to use Pyhton3.11 by default (Cf. https://github.com/google/heir/issues/1370)
- Some deps we were building manually are on the bazel registry, which had minor changes to the import paths (e.g. `eigen`)
- The new rules_go decides on bazel project names for you, so instead of `@lattigo` we have to do `@com_github_tuneinsight_lattigo_v6`


j2kun:
> Bazel 8 uses _main instead of heir in many bazel runfiles paths, which broke yosys techmap file finding, lit configuration, and our heir-opt bazel rules.

I worked with @raghav198 on a bug related to this: apparently on MacOS platforms bazel starts inside `_main`, while on my linux system it was outside `_main`. Need to investigate this a bit closer.

j2kun:
@asraa @AlexanderViand-Intel do you think we can get this PR in before the Python frontend? I think getting this in with the python version bump would help with the internal patch for Alex's python frontend PR

MODULE.bazel
```diff
@@ -0,0 +1,85 @@
...
+# compile_commands extracts the relevant compile data from bazel into
+# `compile_commands.json` so that clangd, clang-tidy, etc., can use it.
+# Whenever a build file changes, you must re-run
+#
+#   bazel run @hedron_compile_commands//:refresh_all
+#
+# to ingest new data into these tools.
+#
+# See the project repo for more details and configuration options
+# https://github.com/hedronvision/bazel-compile-commands-extractor
```
asraa, line 79: we should use renovatebot to manage the sha update - should we file an issue? eventually we should have some dependency management automated, and bzlmod at least takes care of it for the

asraa (approved):
Thanks! I also remember getting a survey from the bazel folks at google to list out the blocking deps for bzlmod migration and at the time I didn't know exactly which were blocking. I'll send them over the remaining ones

MODULE.bazel
j2kun: File an issue :)

j2kun:
I think I can fix the remaining frontend issues, almost got it.

AlexanderViand-Intel:
> I think I can fix the remaining frontend issues, almost got it.

I just had a look and `is_mlir_type_annotation(..)` looks...interesting 😉 What made this necessary/broke the simple "is sublass" check?

j2kun:
OK the internal patch is now done and tested, just waiting on a Googler to click the button to merge it. Sorry for the delay all!

PR #1449, opened 2025-02-22
Polynomial approximation patterns

j2kun:
Part of https://github.com/google/heir/issues/1217

This PR adds polynomial approximation patterns for all unary `math` dialect ops that operate on floats, e.g., `math.exp`. These should all be elementwise mappable, so extent to tensors<f32> and similar for use in the secret+tensor_ext_layer.

The pass supports extracting the desired degree/interval of approximation from the op attributes, and otherwise uses a default [-1, 1] degree 5 approximation.

What's missing:

- Smarter error handling (e.g., someone asks for an approximation of an op outside its domain)
- Binary/ternary math ops, when all but one argument is statically known (filed a TODO) 
- Patterns that match on higher level math ops, e.g., stablehlo.relu lowers to something like `linalg.map { arith.maximumf }`

### DIFF
--- a/lib/Dialect/Polynomial/IR/PolynomialOps.td
+++ b/lib/Dialect/Polynomial/IR/PolynomialOps.td
@@ -340,35 +340,35 @@ def Polynomial_INTTOp : Polynomial_Op<"intt", [Pure]> {
   let hasVerifier = 1;
 }
 
-def Polynomial_EvalOp : Polynomial_Op<"eval", [AllTypesMatch<["value", "output"]>]> {
-    let summary = "Evaluate a static polynomial attribute at a given SSA value.";
-    let description = [{
-        Evaluates the result of a polynomial specified as a static attribute at
-        a given SSA value. The result represents the evaluation of the
-        polynomial at the input value and produces a corresponding scalar
-        value.
+def Polynomial_EvalOp : Polynomial_Op<"eval", [AllTypesMatch<["value", "output"]>, ElementwiseMappable]> {
+  let summary = "Evaluate a static polynomial attribute at a given SSA value.";
+  let description = [{
+      Evaluates the result of a polynomial specified as a static attribute at
+      a given SSA value. The result represents the evaluation of the
+      polynomial at the input value and produces a corresponding scalar
+      value.
 
-        Example:
+      Example:
 
-        ```mlir
-        !poly_ty = !polynomial.polynomial<ring=<coefficientType=i32>>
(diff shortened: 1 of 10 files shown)


lib/Transforms/PolynomialApproximation/PolynomialApproximation.td
```diff
@@ -0,0 +1,55 @@
...
+
+include "mlir/Pass/PassBase.td"
+
+def PolynomialApproximation : Pass<"polynomial-approximation"> {
+  let summary = "Approximate ops by polynomials";
+  let description = [{
+    This pass replaces certain operations that are incompatible
+    with the FHE computational model with polynomial approximations.
+
+    The pass applies to all `math` dialect ops except:
```
AlexanderViand-Intel, line 12: nit: I think it might be safer to liste the ops we support, since upstream might add `math` ops without us noticing

AlexanderViand-Intel (changes requested):
Amazing! Just a nit and one real request: could you also add this pass to the `--mlir-to-secret-arithmetic` pipeline, and add at least one `math.<something>` op in an e2e test? Thanks!

lib/Dialect/Polynomial/IR/PolynomialOps.td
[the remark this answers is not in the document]
j2kun: This was a manual change to be consistent with the rest of the file's indentation level. I'm not aware of any tablegen formatter or LSP, sadly...

j2kun:
> could you also add this pass to the `--mlir-to-secret-arithmetic` pipeline, and add at least one `math.<something>` op in an e2e test? Thanks!

I actually cannot, because we don't have a lowering for `polynomial.eval` yet.

lib/Transforms/PolynomialApproximation/PolynomialApproximation.td
j2kun: Fixed

AlexanderViand-Intel:
> I actually cannot, because we don't have a lowering for `polynomial.eval` yet.

:(

AlexanderViand-Intel (approved):
> > could you also add this pass to the `--mlir-to-secret-arithmetic` pipeline, and add at least one `math.<something>` op in an e2e test? Thanks!
> 
> I actually cannot, because we don't have a lowering for `polynomial.eval` yet.

:( That's a good reason, though xD

asraa (approved):

PR #1451, opened 2025-02-23
lattigo-alloc-to-inplace: greedily use availible storage as inplace output

ZenithalHourlyRate:
See https://github.com/google/heir/pull/1407#discussion_r1962686905

This optimization extends the alloc-to-inplace pass to use available storage (dead value after this operation) instead of judging only whether one of its operand is dead. This further reduces the memory usage.

Detailed explanation of the optimization could be found in the code comments.

Marked as draft as waiting for #1407 to merge.

### DIFF
--- a/lib/Dialect/Lattigo/Transforms/AllocToInplace.cpp
+++ b/lib/Dialect/Lattigo/Transforms/AllocToInplace.cpp
@@ -3,98 +3,225 @@
 #include <utility>
 
 #include "lib/Dialect/Lattigo/IR/LattigoOps.h"
-#include "mlir/include/mlir/Analysis/Liveness.h"  // from @llvm-project
-#include "mlir/include/mlir/IR/MLIRContext.h"     // from @llvm-project
-#include "mlir/include/mlir/IR/PatternMatch.h"    // from @llvm-project
-#include "mlir/include/mlir/Support/LLVM.h"       // from @llvm-project
-#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
+#include "lib/Utils/Tablegen/InplaceOpInterface.h"
+#include "mlir/include/mlir/Analysis/Liveness.h"        // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/MLIRContext.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"          // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"             // from @llvm-project
+#include "mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h"  // from @llvm-project
 
 namespace mlir {
 namespace heir {
 namespace lattigo {
 
+// There are two types of Value in the IR:
+// 1. Storage: the actual memory allocated for the value
+// 2. Referring Value: the value that refers to the storage (e.g., the
+// returned SSA of an inplace operation)
+//
+// This class is similar to Disjoint-set data structure.
+// Each Storage is the root, and all Referring Values are in its set.
+//
+// At the beginning StorageInfo should be initialized based on current relation
+// in program.
+//
+// During rewriting, when we find available Storage for an AllocOp, we replace
+// it with an InplaceOp and update the StorageInfo by merging the Storage of the
+// AllocOp to the available Storage.
+//
+// This allows a mix of AllocOp and InplaceOp in input IR for the pass.
+class StorageInfo {
+ public:
+  StorageInfo() = default;
+
+  void addStorage(Value value) {
+    if (mlir::isa<RLWECiphertextType>(value.getType())) {
+      storageToReferringValues[value] = {};
+    }
+  }
+
+  void addReferringValue(Value storage, Value value) {
+    storageToReferringValues[storage].push_back(value);
+  }
+
+ private:
+  // maintenance should be called internally
+  void removeStorage(Value value) { storageToReferringValues.erase(value); }
+
+  void mergeStorage(Value from, Value to) {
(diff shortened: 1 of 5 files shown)


ZenithalHourlyRate:
Ready for review after #1407 is in

lib/Dialect/Lattigo/Transforms/AllocToInplace.cpp
```diff
@@ -3,16 +3,92 @@
...
+      }
+    }
+    return Value();
+  };
+
+  auto *block = op->getBlock();
+  for (auto arg : block->getArguments()) {
+    updateStorage(arg);
+  }
+  block->template walk<WalkOrder::PreOrder>([&](Operation *prevOp) {
```
j2kun, line 46: This walk implies a quadratic time algorithm for the whole pass. I suspect that is not necessary, since the OpRewritePattern materializes the changes to the IR instantly, but it may also be more convenient to walk the IR and modify it directly rather than use the pattern application.

In that version, I imagine you'd keep track of the mapping from references to storage and mutate it as you go, as well as keep track of a set of storage values seen so far and their liveness, and only update those values that are impacted by the currently visited op.

Is there anything preventing this here? Maybe the Liveness would become stale and must be re-generated, or else no changes materialized until the end of the walk?

We can merge it in as-is, but in that case let's put a TODO and file an issue to convert this to a linear-time algorithm.

ZenithalHourlyRate: Now it is linear time.

The main logic is contained in a `StorageInfo` class that maintains the Storage/ReferringValue map like a disjoint set. After each rewrite, the map is updated with small cost.

In my understaning `Liveness` is initialized once and later query will only work for the input IR. However, as the Storage value is not mutated, the information for it is always valid given our rewrite pattern.

j2kun (approved):
Looks great! Thanks for that improvement.

j2kun:
I'm getting internal failures like

```
exit status 2
WARNING: Logging before InitGoogle() is written to STDERR
F0000 00:00:1740696474.398074    8408 logging.cc:62] assert.h assertion failed at third_party/llvm/llvm-project/llvm/include/llvm/ADT/SmallVector.h:884 in iterator llvm::SmallVectorImpl<mlir::Value>::insert(iterator, ItTy, ItTy) [T = mlir::Value, ItTy = mlir::Value *]: this->isReferenceToStorage(I) && "Insertion iterator is out of bounds."
*** Check failure stack trace: ***
    @     0x564888d71269  absl::log_internal::LogMessage::SendToLog()
    @     0x564888d711de  absl::log_internal::LogMessage::Flush()
    @     0x564888d451a4  __assert_fail
    @     0x564883509510  llvm::SmallVectorImpl<>::insert<>()
    @     0x564883509101  mlir::heir::lattigo::StorageInfo::mergeStorage()
    @     0x564883508c89  mlir::heir::lattigo::StorageInfo::replaceAllocWithInplace()
    @     0x56488350a21d  mlir::heir::lattigo::ConvertBinOp<>::matchAndRewrite()
    @     0x5648853e044e  llvm::function_ref<>::callback_fn<>()
```

Looks like mergeStorage needs some improvement. But this seems like it's not just ASAN, the assertion is actually failing, which should happen if you run blaze with `-c dbg`

lib/Dialect/Lattigo/Transforms/AllocToInplace.cpp
```diff
@@ -3,98 +3,223 @@
...
+  void addReferringValue(Value storage, Value value) {
+    storageToReferringValues[storage].push_back(value);
+  }
+
+ private:
+  // maintenance should be called internally
+  void removeStorage(Value value) { storageToReferringValues.erase(value); }
+
+  void mergeStorage(Value from, Value to) {
+    storageToReferringValues[to].insert(storageToReferringValues[to].end(),
```
j2kun, line 53: I'm guessing you need a reserve/resize here so you're not inserting past the bounds of the array?

ZenithalHourlyRate:
I could not reproduce that with `-c dbg`

lib/Dialect/Lattigo/Transforms/AllocToInplace.cpp
ZenithalHourlyRate: Added a reserve here

j2kun: ok that was my guess. I will be able to dig in a bit deeper internally tomorrow or later tonight.

lib/Dialect/Lattigo/Transforms/AllocToInplace.cpp
```diff
@@ -3,98 +3,225 @@
...
-    // InplaceOp has the form: output = InplaceOp(evaluator, lhs, rhs, inplace)
-    // where inplace is the actual output but for SSA form we need to return a
-    // new value
-    rewriter.replaceOpWithNewOp<InplaceOp>(op, op.getOperand(1).getType(),
-                                           op.getOperand(0), op.getOperand(1),
-                                           op.getOperand(2), op.getOperand(1));
+    // InplaceOp has the form: output = InplaceOp(evaluator, lhs, rhs,
+    // inplace) where inplace is the actual output but for SSA form we need to
+    // return a new value
+    auto inplaceOp = rewriter.replaceOpWithNewOp<InplaceOp>(
```
j2kun, line 143: I patched the problem internally, so no changes are needed on this PR, but here was the problem: you were replacing the op before calling `replaceAllocWithInplace`, which then queries the (now deleted) `oldOp`'s results.

For whatever reason, externally that deletion was not yet overwritten so it works, but internally the memory was cleared and it led to `oldOp->getNumResults()` being like 28376422345, and hitting a timeout while looping over the "results" (or hitting a memory error when trying to access the iterator used for `reserve`).

So this PR will land with a patch that creates the op, then calls `replaceAllocWithInplace`, then does the replacement (in each of the three patterns).

Issue #1441, opened 2025-02-20
Openfhe: Make the backend aware of the minimal slot requirement


ZenithalHourlyRate:
See #1439, where too small plaintext modulus will make the openfhe parameter selection process generate logN < 13.

This can be controled by `SetBatchSize`

ZenithalHourlyRate:
See https://github.com/google/heir/pull/1905#issuecomment-2981184052

We could not tell OpenFHE the minimum slot requirement via SetBatchSize, as it does not participate in the param gen process. This fact can be established by finding `GetBatchSize` in https://github.com/openfheorg/openfhe-development/blob/main/src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp where `GetBatchSize` only participates in the assertion phase, not the generation phase.

PR #1452, opened 2025-02-23
Parameters: Allow specifying minimum slot number

### LINKED ISSUE #1440: BGV: Make parameter generation process aware of the minimal slot requirement
See #1439, where box blur/roberts cross has input of size 4096 yet the BGV parameter generation process generate a logN of 12.

ZenithalHourlyRate:
Fixes #1440

Now for lattigo, boxblur uses the ptm of 786433 and roberts cross uses the ptm of 536903681, which is the same as those used in openfhe tests.

Seems that during computation the number in slots do can exceed 65537 so these ptm are required. Have not inspected the intermediate values during computation though. Not sure if #1441 is necessary or not.

Indeed we need a plaintext execution result from the secret-arithmetic IR (both in Z and Zp) to determine whether the execution is correct given a plaintext modulus, or the incorrectness is from the ciphertext side (i.e. scheme parameter). Ref to #1266


ZenithalHourlyRate:
Rebased on top of the bazelmod migration.

j2kun (approved):

j2kun:
Needs a rebase, then LGTM!

ZenithalHourlyRate:
Rebased

Issue #1419, opened 2025-02-17
B/FV support


ZenithalHourlyRate:
As I am also experimenting with B/FV noise estimation method, I would prefer also integrating the B/FV support to HEIR.

Technically, the hacky way is quite simple as I only need to get the translater work (emitting B/FV code instead of BGV code), as for Lattigo, BFV/BGV is only a flag difference in `bgv.Evaluator.ScaleInvariant`, and for Openfhe they only differs in the name `CryptoContextBFVRNS`. Though they have some mgmt difference like there is no `mgmt.modreduce` now.

However, for full support there would be much duplication as then we need to introduce a `bfv` dialect and all the ops/attributes need to be copied.

I wonder which way do we prefer, the way I prefer is

* Add support in translater
* Modify mgmt passes to support bfv style
* Use `bgv` dialect to represent bfv ops.

PR #1456, opened 2025-02-23
Initial support of B/FV via reusing BGV

ZenithalHourlyRate:
Addresses #1419

As B/FV operation is highly similar to BGV, to make the pipeline work with minimum effort we can reuse BGV to represent B/FV.

One exception is that `bgv.modulus_switch` is not available for B/FV (most libraries does not implement the leveled version). Happily the two backends we have, when configured in B/FV mode, just ignores the compiler-issued modulus switching instruction.

  * Lattigo implements it as no-op
  * Openfhe, when in FIXEDMANUAL mode, throws exception. In other cases just no-op.

The side effect is that in `LWE` IR, we will see the level decreasing but in the backend it is not decreasing. This should be resolved by a future `secret-insert-mgmt-bfv` pass.
  
I think the time we want to separate B/FV from BGV is when we want to model/implement the multiplication technique of B/FV.

### DIFF
--- a/lib/Dialect/BGV/IR/BGVDialect.td
+++ b/lib/Dialect/BGV/IR/BGVDialect.td
@@ -10,7 +10,14 @@ def BGV_Dialect : Dialect {
   let name = "bgv";
 
   let description = [{
-    The BGV dialect defines the types and operations of the BGV cryptosystem.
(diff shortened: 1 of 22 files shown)


ZenithalHourlyRate:
Rebased on top of the bazelmod migration.

j2kun (approved):
Looks great!

Issue #1420, opened 2025-02-17
Support different flavor of lowering in `--mlir-to-<scheme>`


ZenithalHourlyRate:
Passes in `mlir-to-<scheme>` have multiple places that are tailored for different backends, like

* `modulus-switch-before-first-mul` in `secret-insert-mgmt`, which is only for Openfhe FLEXIBLEAUTOEXT mode
* `validate-noise` noise models for Lattigo as Lattigo has a different public key encryption technique, see https://github.com/google/heir/pull/1374#issuecomment-2646153253
* `AddAndKeySwitchCountAnalysis`, which is only for Openfhe
* `optimize-relinearization` which has an option on whether supporting mixed degree op, see also https://github.com/google/heir/pull/1295#issuecomment-2616340672 and #1169 

A `--mlir-to-<scheme>=flavor=lattigo --scheme-to-lattigo` can make full potential of these passes, yet this is quite redundant.

Another option is to mark the backend in module, but when we want to switch to another backend we have to modify the string in IR, which is inconvenient

```mlir
module attributes {backend.lattigo} {...}
```

Yet another solution is to add a pass to select the backend in the beginning like

```bash
--openfhe --mlir-to-scheme --to-backend
```
where `--openfhe` just modify the module attribute to `backend.openfhe`.

ZenithalHourlyRate:
In the office hour, with loop support (#1569) and the pending scale management PR (#1459) as the background, we discussed that the current status in main is not perfect for the scale management to land, however we should not wait until the loop support shifted to that direction, so we may need a partial solution here.

The solution is to introduce two kinds of pipeline that one is fully managed (taking care of level, cross level, dimension, noise/parameter generation, scale) and the other is managed the least (only relinearize, and generate a fallback parameter). For example, `--mlir-to-bgv=mgmt=false` (which is quite ugly) or something like that.

The observation is that, openfhe does not respect our management anyway, and the loop support is currently focused on the openfhe pipeline, this least managed pipeline is just for openfhe, so we may merge that idea into this issue about different flavors of lowering.

My opinion now favors the module attribute solution as it is early in the pipeline and all other pipeline can just read it from IR instead of from pass options.

[...]

PR #1459, opened 2025-02-24
BGV/CKKS: support scale management

### LINKED ISSUE #785: Encoding "Taxonomy"

### LINKED ISSUE #1169: Track scaling factor in encoding attr for BGV (GHS variant) and handle cross-level op

### LINKED ISSUE #1364: AnnotateMgmt for programs with loops fails - LevelAnalysis yields infinite loop
Root cause: Analyses do not understand region bearing ops (for loops) properly, we need an implementation of `visitNonControlFlowArguments` most likely.

Reproducer:

```
$ heir-opt --secret-insert-mgmt-ckks
  func.func @loop(%arg0: !secret.secret<tensor<1x1024xf32>>) -> !secret.secret<tensor<1x1024xf32>> {
    %0 = secret.generic ins(%arg0 : !secret.secret<tensor<1x1024xf32>>) {
    ^body(%input0: tensor<1x1024xf32>):
      %1 = affine.for %arg1 = 0 to 3 iter_args(%arg3 = %input0) -> (tensor<1x1024xf32>) {
        %3 = arith.mulf %arg3, %arg3 : tensor<1x1024xf32>
        affine.yield %3 : tensor<1x1024xf32>
      }
      secret.yield %1 : tensor<1x1024xf32>
    } -> !secret.secret<tensor<1x1024xf32>>
    return %0 : !secret.secret<tensor<1x1024xf32>>
  }
```

This program has a single ctxt-ctxt multiplication inside, and the result should have a relinearization inserted after the mul so that the block arg and yielded value always have dimension 2. The pass also inserts modreduce before the multiplication even when first mul option is off because it seems that the args are from a mulresult too.

So I think LevelAnalysis, DimensionAnalysis, and MulResultAnalysis all need to be aware of the region bearing operation.

When you run the reproducer, it will run into an infinite loop when it re-runs the solver after the relinearization insertion. I added some debugs to the level analysis and saw it continuining to increment the level of multiplication over and over in the loop.

```
LevelAnalysis: Visiting operation: %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
propogate: %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>is changed 1 to state 2769
LevelAnalysis: Visiting operation: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
propogate: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>is changed 1 to state 2769
LevelAnalysis: Visiting operation: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
propogate: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>is changed 0 to state 2769
LevelAnalysis: Visiting operation: %4 = mgmt.relinearize %3 : tensor<1x1024xf32>
propogate: %4 = mgmt.relinearize %3 : tensor<1x1024xf32>is changed 1 to state 2769
LevelAnalysis: visitNonControlFlowArguments: %1 = affine.for %arg1 = 0 to 3 iter_args(%arg2 = %input0) -> (tensor<1x1024xf32>) {
  %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
  %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
  %4 = mgmt.relinearize %3 : tensor<1x1024xf32>
  affine.yield %4 : tensor<1x1024xf32>
}
LevelAnalysis: visitNonControlFlowArguments: %1 = affine.for %arg1 = 0 to 3 iter_args(%arg2 = %input0) -> (tensor<1x1024xf32>) {
  %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
  %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
  %4 = mgmt.relinearize %3 : tensor<1x1024xf32>
  affine.yield %4 : tensor<1x1024xf32>
}
LevelAnalysis: Visiting operation: %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
propogate: %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>is changed 1 to state 2770
LevelAnalysis: Visiting operation: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
propogate: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>is changed 1 to state 2770
LevelAnalysis: Visiting operation: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
propogate: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>is changed 0 to state 2770
LevelAnalysis: Visiting operation: %4 = mgmt.relinearize %3 : tensor<1x1024xf32>
propogate: %4 = mgmt.relinearize %3 : tensor<1x1024xf32>is changed 1 to state 2770
LevelAnalysis: visitNonControlFlowArguments: %1 = affine.for %arg1 = 0 to 3 iter_args(%arg2 = %input0) -> (tensor<1x1024xf32>) {
  %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
  %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
  %4 = mgmt.relinearize %3 : tensor<1x1024xf32>
  affine.yield %4 : tensor<1x1024xf32>
}
LevelAnalysis: visitNonControlFlowArguments: %1 = affine.for %arg1 = 0 to 3 iter_args(%arg2 = %input0) -> (tensor<1x1024xf32>) {
  %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
  %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
  %4 = mgmt.relinearize %3 : tensor<1x1024xf32>
  affine.yield %4 : tensor<1x1024xf32>
}
LevelAnalysis: Visiting operation: %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>
propogate: %2 = mgmt.modreduce %arg2 : tensor<1x1024xf32>is changed 1 to state 2771
LevelAnalysis: Visiting operation: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
propogate: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>is changed 1 to state 2771
LevelAnalysis: Visiting operation: %3 = arith.mulf %2, %2 : tensor<1x1024xf32>
```

cc @ZenithalHourlyRate

ZenithalHourlyRate:
See #1169

Fixes #1169 Fixes #1364 Fixes #785

I am afraid we can only do this for Lattigo backend, as Openfhe does not have explicit port for setting scale. Although the policy implemented is in line with Openfhe's implementation, and Openfhe does that automatically.

The detailed rational/impl of the scale management should be put in design doc within this PR.

There are a few changes to support scale

  * `mgmt.level_reduce`, `mgmt.adjust_scale` op to support corresponding operation
  * Modify `secret-insert-mgmt-bgv` to use these ops to handle cross level op, where `adjust_scale` is a place holder
  * `--validate-noise` will generate parameters aware of these management op
  * `--populate-scale` (better name wanted) to concretely fill the scale based on the parameter
  
### TODO
  * [x] Deal with the complicacy caused by the `include-first-mul` option.
  * [x] `LWE` dialect scale support 
  * [x] Lowering to Lattigo backend.
  * [x] Option to just ignore these mgmt op for Openfhe, as it has automation. See also #1420
  * [x] Handle ct-pt op, need to set pt scale as ct scale

Cc @AlexanderViand-Intel: A comment on https://github.com/google/heir/pull/1295#issuecomment-2616340672 is that, the two backends we have can safely `Add(ct0, ct1)` with ciphertexts of different scale as internally when they find scale mismatching they would just adjust scale themselves. So the mixed-degree option for `optimize-relinearization` can be on without affecting correctness, though the noise is different. The merging of this PR does not fix the scale mismatching problem possibly induced by optimize-relinearization for our current two backends, but it indeed pave the way for our own `poly` backend which must be scale aware. 

### Example

the input mlir

```mlir
func.func @cross_level_add(%base: tensor<4xi16> {secret.secret}, %add: tensor<4xi16> {secret.secret}) -> tensor<4xi16> {
  // same level
  %base0 = arith.addi %base, %add : tensor<4xi16>
  // increase one level
  %mul1 = arith.muli %base0, %base0 : tensor<4xi16>
  // cross level add
  %base1 = arith.addi %mul1, %add : tensor<4xi16>
  // increase one level
  %mul2 = arith.muli %base1, %base1 : tensor<4xi16>
  // cross level add
  %base2 = arith.addi %mul2, %add : tensor<4xi16>
  // increase one level
  %mul3 = arith.muli %base2, %base2 : tensor<4xi16>
  // cross level add
  %base3 = arith.addi %mul3, %add : tensor<4xi16>
  return %base3 : tensor<4xi16>
}
```

After `secret-insert-mgmt-bgv`, we get

```mlir
      %1 = arith.addi %input0, %input1 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
      %2 = arith.muli %1, %1 {mgmt.mgmt = #mgmt.mgmt<level = 3, dimension = 3>} : tensor<4xi16>
      %3 = mgmt.relinearize %2 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
      %4 = arith.addi %3, %input1 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
      %5 = mgmt.modreduce %4 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
      %6 = arith.muli %5, %5 {mgmt.mgmt = #mgmt.mgmt<level = 2, dimension = 3>} : tensor<4xi16>
      %7 = mgmt.relinearize %6 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
      %8 = mgmt.adjust_scale %input1 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
      %9 = mgmt.modreduce %8 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
      %10 = arith.addi %7, %9 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
      %11 = mgmt.modreduce %10 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
      %12 = arith.muli %11, %11 {mgmt.mgmt = #mgmt.mgmt<level = 1, dimension = 3>} : tensor<4xi16>
      %13 = mgmt.relinearize %12 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
      %14 = mgmt.level_reduce %input1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
      %15 = mgmt.adjust_scale %14 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
      %16 = mgmt.modreduce %15 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
      %17 = arith.addi %13, %16 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
      %18 = mgmt.modreduce %17 {mgmt.mgmt = #mgmt.mgmt<level = 0>} : tensor<4xi16>
```
where adjust_scale has no concrete scale parameter

After --validate-noise and `--populate-scale`, we will get the per-level scale, and the value to fill for each `adjust_scale`

```mlir
PopulateScale: scale = [57802, 46604, 21845, 1, ]
PopulateScale: scaleBig = [60481, 36636, 29128, 1, ]
PopulateScale: adjustScale = [1, 2528, 13431, 21845, ]
Propagate ScaleState(1) to <block argument> of type 'tensor<4xi16>' at index: 0
Propagate ScaleState(1) to <block argument> of type 'tensor<4xi16>' at index: 1
Propagate ScaleState(1) to %1 = arith.addi %input0, %input1 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
Propagate ScaleState(1) to %2 = arith.muli %1, %1 {mgmt.mgmt = #mgmt.mgmt<level = 3, dimension = 3>} : tensor<4xi16>
Propagate ScaleState(1) to %3 = mgmt.relinearize %2 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
Propagate ScaleState(1) to %4 = arith.addi %3, %input1 {mgmt.mgmt = #mgmt.mgmt<level = 3>} : tensor<4xi16>
Propagate ScaleState(21845) to %5 = mgmt.modreduce %4 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
Propagate ScaleState(29128) to %6 = arith.muli %5, %5 {mgmt.mgmt = #mgmt.mgmt<level = 2, dimension = 3>} : tensor<4xi16>
Propagate ScaleState(29128) to %7 = mgmt.relinearize %6 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
Propagate ScaleState(21845) to %8 = mgmt.adjust_scale %input1 {mgmt.mgmt = #mgmt.mgmt<level = 3>, scale = 21845 : i64} : tensor<4xi16>
Propagate ScaleState(29128) to %9 = mgmt.modreduce %8 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
Propagate ScaleState(29128) to %10 = arith.addi %7, %9 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
Propagate ScaleState(46604) to %11 = mgmt.modreduce %10 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
Propagate ScaleState(36636) to %12 = arith.muli %11, %11 {mgmt.mgmt = #mgmt.mgmt<level = 1, dimension = 3>} : tensor<4xi16>
Propagate ScaleState(36636) to %13 = mgmt.relinearize %12 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
Propagate ScaleState(1) to %14 = mgmt.level_reduce %input1 {mgmt.mgmt = #mgmt.mgmt<level = 2>} : tensor<4xi16>
Propagate ScaleState(13431) to %15 = mgmt.adjust_scale %14 {mgmt.mgmt = #mgmt.mgmt<level = 2>, scale = 13431 : i64} : tensor<4xi16>
Propagate ScaleState(36636) to %16 = mgmt.modreduce %15 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
Propagate ScaleState(36636) to %17 = arith.addi %13, %16 {mgmt.mgmt = #mgmt.mgmt<level = 1>} : tensor<4xi16>
Propagate ScaleState(57802) to %18 = mgmt.modreduce %17 {mgmt.mgmt = #mgmt.mgmt<level = 0>} : tensor<4xi16>
```

Where the first three lines are purely calculated from bgv scheme parameter and the later is the analysis to validate whether the scale matches.

The initial scaling factor is chosen to be 1 for both `include-first-mul={true,false}`, as for `include-first-mul=false`, the scaling factor of the last level must be the same, so we have `1 * 1 = 1`.


ZenithalHourlyRate:
It has been quite messy supporting scale, as we have to change these things below

* mgmt insertion policy for both BGV and CKKS
  + insert rescale after mult, before mult, and before mult with first mult
  + note that the original cross-level policy for CKKS is wrong
* populate scale with regard to all three rescale insertion policy
* LWE type, which if we add a new field quite a bunch of test file need to change
* Two backends

My idea is to skip the LWE type support and use attribute to pass information temporarily, and skip openfhe as it does support that anyway.

The pipeline works now for Lattigo, see example

```mlir
func.func @cross_level_add(%base: tensor<4xi16> {secret.secret}, %add: tensor<4xi16> {secret.secret}) -> tensor<4xi16> {
  // increase one level
  %mul1 = arith.muli %base, %add : tensor<4xi16>
  // cross level add
  %base1 = arith.addi %mul1, %add : tensor<4xi16>
  return %base1 : tensor<4xi16>
}
```

After properly managed and calculation of scale we get
```mlir
      %1 = mgmt.modreduce %input0 {mgmt.mgmt = #mgmt.mgmt<level = 1, scale = 4>} : tensor<4xi16>
      %2 = mgmt.modreduce %input1 {mgmt.mgmt = #mgmt.mgmt<level = 1, scale = 4>} : tensor<4xi16>
      %3 = arith.muli %1, %2 {mgmt.mgmt = #mgmt.mgmt<level = 1, dimension = 3, scale = 16>} : tensor<4xi16>
      %4 = mgmt.relinearize %3 {mgmt.mgmt = #mgmt.mgmt<level = 1, scale = 16>} : tensor<4xi16>
      // need to adjust the scale by mul_const delta_scale
      %5 = mgmt.adjust_scale %input1 {delta_scale = 4 : i64, mgmt.mgmt = #mgmt.mgmt<level = 2, scale = 4>, scale = 4 : i64} : tensor<4xi16>
      %6 = mgmt.modreduce %5 {mgmt.mgmt = #mgmt.mgmt<level = 1, scale = 16>} : tensor<4xi16>
      %7 = arith.addi %4, %6 {mgmt.mgmt = #mgmt.mgmt<level = 1, scale = 16>} : tensor<4xi16>
      %8 = mgmt.modreduce %7 {mgmt.mgmt = #mgmt.mgmt<level = 0, scale = 65505>} : tensor<4xi16>
```

`adjust_scale` is materialized as the following

```mlir
    %cst = arith.constant dense<1> : tensor<4xi16>
    %pt = lwe.rlwe_encode %cst {encoding = #full_crt_packing_encoding, lwe.scale = 4 : i64, ring = #ring_Z65537_i64_1_x4_} : tensor<4xi16> -> !pt
    %ct_5 = bgv.mul_plain %ct_0, %pt : (!ct_L2_, !pt) -> !ct_L2_
```

When emitted to lattigo with debug handler, we can observe the scale change exactly the same

```
Input
  Scale:  1
Input
  Scale:  1
lattigo.bgv.rescale_new
  Scale:  4
lattigo.bgv.rescale_new
  Scale:  4
lattigo.bgv.mul_new
  Scale:  16
lattigo.bgv.relinearize_new
  Scale:  16
// this is adjust_scale
lattigo.bgv.mul_new
  Scale:  4
lattigo.bgv.rescale_new
  Scale:  16
lattigo.bgv.add_new
  Scale:  16
lattigo.bgv.rescale_new
  Scale:  65505
Result [4 9 16 25]
```

j2kun:
Talking about this in office hours. Some ideas:

- Have one integer scaling factor attribute that specifies its own bitwidth, to support large bitwidth scaling factors (e.g., for CKKS). The the lowerings to a particular backend (e.g., C++/OpenFHE) would need to pick the appropriate type (perhaps long double) to represent that scaling factor in the target language.
- Make the scaling factor attribute optional on the LWE type to avoid having to update the entire codebase, and raise errors or handle the default cause when the scaling factor is not present. We could also have some backlog work to go update the rest of the codebase so that there are scaling factors and we can later remove optionality.

ZenithalHourlyRate:
> Until now 99 files changed...would be insane if more changes are introduced.

> The hard part of supporting scale management is that, making scale match everywhere. The current state of the PR will break the loop support.

I think I am going to pend this PR as the loop part is important. I intend to open another PR that possibly add a pass taking a loop and analyse if it is expressible by FHE world, for at least LevelAnalysis, then we can make ScaleAnalysis happy.

In the meantime, some non-critical part of this is split out into like #1540 and further PRs.

j2kun:
I am still digesting many of these details, but this one stuck out for me:

> We can not even deal with unrolled version because we need some kind of back-propagation

We had the same problem with layout management, a `tensor.empty` has no intrinsic layout when we visit it. We added the `assign_layout` op to give it a default placeholder layout, and then a back-propagation pass would support merging downstream layout changes into this op to "override" it with a better layout discovered later. Could you use the same idea for scale? I.e., set some default `assign_scale` in a forward-propagation pass, then in a backward propagation pass you could hoist `adjust_scale` ops backward through the IR, and if you encounter `assign_scale` you can replace it with the desired adjusted scale and then use that to set the scale of the initial empty tensor.

ZenithalHourlyRate:
After dependent PRs merged #1540 #1586 #1611 this PR now is much cleaner, still it is big enough (85 files changed with 3k additional lines)

Notable changes

* Now Loop support is back-alive, as we can skip management passes entirely by specifying `backend=openfhe`, and additionally providing a scheme parameter could skip all the analysis so infinite loop like #1364 won't happen.
* Only the lattigo pipeline has scale management. And all the messy code (back-propagation mentioned above) is now handled by Backward Analysis provided by MLIR framework, i.e. `ScaleAnalysisBackward`. A similar `LevelAnalysisBackward` is also introduced. The backward analyses are responsible for handling plaintext and adjust_scale, which naturally needs back-prop from ciphertext.
* For plaintext operand, `mgmt.no_op` is inserted for them as a holder for mgmt attribute, in case canonicalizer merges `arith.constant` despite they have different annotation.
* `adjust_scale` is materialized to `mul_plain %ct, %c1` with `%c1` encoded at specific scaling factor in `populate-scale-<scheme>`
* different modulus switching policy and scale management policy test cases and e2e test cases are added.

Other parts are described in the `management.md` design doc.

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+
+1. After multiplication: `[ab / qi]_t`.
+
+1. Before multiplication: `[a / qi * b / qi]_t` = `[ab / (qi^2)]_t`.
+
+This is messy enough. To ease the burden, we can impose additional requirement:
+Mandate a constant scale `Delta_i` for each level. This is called
+_level-specific scaling factor_. With this in mind, addition within one level
+can happen without caring about the scale, and the scale of the next level is
+also the same.
```
j2kun, line 115:
```suggestion
This is messy enough. To ease the burden, we can impose additional requirement:
mandate a constant scale `Delta_i` for all ciphertexts at level `i`. 
This is called the _level-specific scaling factor_.
With this in mind, addition within one level
can happen without caring about the scale.
```

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+1. After multiplication: `Delta_{i-1} = [Delta_i^2 / qi]_t`
+
+1. Before multiplication: `Delta_{i-1} = [Delta_i^2 / (qi^2)]_t`
+
+### BGV: Cross Level Operation
+
+With the level-specific scaling factor, one may wonder how could addition and
+multiplication of ciphertext of different levels happen as they have different
+scaling factor. This can be solved by adjusting the level and scale of the
+ciphertext of higher level.
```
j2kun, line 126:
```suggestion
With the level-specific scaling factor, one may wonder how to perform addition and
multiplication of ciphertexts on different levels. This can be done by adjusting
the level and scale of the ciphertext at the higher level.
```

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+1. Assume the level and scale of two ciphertexts are `l1` and `l2`, `s1` and
+   `s2` respectively. WLOG assume `l1 > l2`.
+
+1. Drop `l1 - l2 - 1` limbs for the first ciphertext to make it at level
+   `l2 + 1`, if those extra limbs exist.
+
+1. Adjust scale from `s1` to `s2 * q_{l2 + 1}` by multiplying
+   `[s2 * q_{l2 + 1} / s1]_t` for the first ciphertext.
+
+1. Modulusing switching from `l2 + 1` to `l2`, the scale becomes `s2` for the
```
j2kun, line 141:
```suggestion
1. Modulus switch from `l2 + 1` to `l2`, producing scale `s2` for the
```

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+### BGV: Implementation in HEIR
+
+In HEIR the different modulus switching policy is controlled by the pass option
+for `--secret-insert-mgmt-bgv`. The pass defaults to the "Before Multiplication"
+policy. If user wants other policy, the `after-mul` or
+`before-mul-include-first-mul` option may be used. The `mlir-to-bgv` pipeline
+option `modulus-switch-before-first-mul` corresponds to the latter option.
+
+The `secret-insert-mgmt` pass is also responsible for managing cross-level
+operations. However, as the scheme parameters has not been generated, the
```
j2kun, line 153:
```suggestion
operations. However, as the scheme parameters are not generated at this point, the
```

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+concrete scale could not be instantiated so some placeholder operations are
+inserted.
+
+After the modulus switching policy applied, user could use `generate-param-bgv`
+pass to generate scheme parameters. Optionally, user could skip this pass by
+manually providing scheme parameter as an attribute at module level.
```
j2kun, line 159:
```suggestion
After the modulus switching policy is applied, the `generate-param-bgv`
pass generates scheme parameters. Optionally, user could skip this pass by
manually providing scheme parameter as an attribute at module level.
```

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+
+Then `populate-scale-bgv` comes into play by using the scheme parameters to
+instantiate concrete scale, and turn those placeholder operations into concrete
+multiplication operation.
+
+## CKKS
+
+CKKS is a leveled scheme where each level has a modulus `qi`. The level is
+numbered from `0` to `L` where `L` is the input level and `0` is the output
+level. CKKS ciphertext contains a scaled message $\\Delta m$ where $\\Delta$
```
j2kun, line 169: Though I am not an expert here, I had believed that the scale in CKKS was also partly used for message precision, in the sense of a fixed-point numeric scale (and given the infinite input domain for CKKS encoding). But for BGV it feels mainly like an artifact of the GHS optimization. Are there any meaningful differences in how these two kinds of "scale" are handled? I'm guessing based on this document that the answer is not for the purposes of `mgmt`, but maybe there is something about choosing an input encoding scale that is left unsaid.

lib/Analysis/LevelAnalysis/LevelAnalysis.cpp
```diff
@@ -50,6 +57,16 @@ LogicalResult LevelAnalysis::visitOperation(
...
       })
+      .Case<mgmt::LevelReduceOp>([&](auto levelReduceOp) {
+        // implicitly ensure that the operand is secret
+        const auto *operandLattice = operands[0];
+        if (!operandLattice->getValue().isInitialized()) {
+          return;
+        }
+        auto level = operandLattice->getValue().getLevel();
+        propagate(levelReduceOp.getResult(),
+                  LevelState(level + levelReduceOp.getLevelToDrop()));
```
j2kun, line 68: I would expect this to be a subtraction, if the input starts at level `L` and goes down to 0. Though some of your tests have this increasing when a level reduce op is applied.

Similar below where you lattice join with a max instead of a min (though maybe that makes sense because you're backpropagating?)

lib/Analysis/LevelAnalysis/LevelAnalysis.h
```diff
@@ -94,8 +94,29 @@ class LevelAnalysis
   }
 };
 
+class LevelAnalysisBackward
```
j2kun, line 97: Maybe a little code comment to state why you need a separate backwards analysis.

lib/Dialect/Mgmt/IR/MgmtOps.td
```diff
@@ -86,4 +113,56 @@ def Mgmt_BootstrapOp : Mgmt_Op<"bootstrap"> {
...
+    AnyType:$input,
+    DefaultValuedAttr<I64Attr, "0">:$scale
+  );
+  let results = (outs AnyType:$output);
+  let assemblyFormat = "operands attr-dict `:` type($output)";
+
+  let hasCanonicalizer = 1;
+}
+
+def Mgmt_NoOp : Mgmt_Op<"no_op"> {
```
j2kun, line 149: Not a huge fan of the name here. Some alternative ideas:

- `mgmt.assume` or `mgmt.require`, if it is intended to be used to mark/override mgmt attributes mid-IR
- `mgmt.init` or `mgmt.seed`, if it is intended to set up initial mgmt attributes for static constants int he IR.

lib/Analysis/ScaleAnalysis/ScaleAnalysis.h
```diff
@@ -0,0 +1,180 @@
...
+    propagateIfChanged(state, changed);
+  }
+
+ private:
+  const SchemeParamType schemeParam;
+  int64_t inputScale;
+};
+
+template <typename ScaleModelT>
+class ScaleAnalysisBackward
```
j2kun, line 137: Maybe the same comment here to explain why there are two analyses

lib/Analysis/ScaleAnalysis/ScaleAnalysis.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+int64_t CKKSScaleModel::evalMulScaleBackward(const ckks::LocalParam &param,
+                                             int64_t result, int64_t lhs) {
+  return result - lhs;
+}
+
+int64_t CKKSScaleModel::evalModReduceScale(const ckks::LocalParam &inputParam,
+                                           int64_t scale) {
+  const auto *schemeParam = inputParam.getSchemeParam();
+  // TODO: rescale using logqi instead of logDefaultScale
```
j2kun, line 79: Track this with an issue or remove, same below.

lib/Analysis/ScaleAnalysis/ScaleAnalysis.cpp
```diff
@@ -0,0 +1,435 @@
...
+    assert(false && "ScaleLattice not initialized");
+    return 0;
+  }
+  return lattice->getValue().getScale();
+}
+
+int64_t getScaleFromMgmtAttr(Value value) {
+  auto mgmtAttr = mgmt::findMgmtAttrAssociatedWith(value);
+  if (!mgmtAttr) {
+    assert(false && "MgmtAttr not found");
```
j2kun, line 400:
```suggestion
    assert(false && "MgmtAttr not found");
    return 0;
```

j2kun (changes requested):
I still have to review lib/Transforms, but the Scale analysis looks good so far and I think the direction is good. I am very excited about the documentation writeup here. I apologize for all the suggested edits, but I hope the result will be well polished as a result.

lib/Transforms/PopulateScale/PopulateScaleBGV.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+      auto mgmtAttr = op->getAttrOfType<mgmt::MgmtAttr>(
+          mgmt::MgmtDialect::kArgMgmtAttrName);
+
+      auto inputType = op.getInput().getType();
+
+      APInt one(getElementTypeOrSelf(inputType).getIntOrFloatBitWidth(), 1);
+      TypedAttr constantAttr;
+      if (auto inputTensorType = dyn_cast<RankedTensorType>(inputType)) {
+        constantAttr = DenseElementsAttr::get(inputTensorType, one);
```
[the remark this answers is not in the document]
j2kun: I think this could also be used for the CKKS side, but I'm not sure if this helper handles all these float semantics (probably it does?)

lib/Transforms/SecretInsertMgmt/SecretInsertMgmtCKKS.cpp
```diff
@@ -184,28 +209,51 @@ struct SecretInsertMgmtCKKS
...
+      patternsMulDepth.add<MatchCrossMulDepth<arith::AddIOp>>(
+          &getContext(), &scaleCounter, getOperation(), &solver);
+      patternsMulDepth.add<MatchCrossMulDepth<arith::SubIOp>>(
+          &getContext(), &scaleCounter, getOperation(), &solver);
+      patternsMulDepth.add<MatchCrossMulDepth<arith::MulFOp>>(
+          &getContext(), &scaleCounter, getOperation(), &solver);
+      patternsMulDepth.add<MatchCrossMulDepth<arith::AddFOp>>(
+          &getContext(), &scaleCounter, getOperation(), &solver);
+      patternsMulDepth.add<MatchCrossMulDepth<arith::SubFOp>>(
+          &getContext(), &scaleCounter, getOperation(), &solver);
```
j2kun, line 245:
```suggestion
      patternsMulDepth.add<MatchCrossMulDepth<arith::MulIOp>, MatchCrossMulDepth<arith::AddIOp>, MatchCrossMulDepth<arith::SubIOp>, MatchCrossMulDepth<arith::MulFOp>, MatchCrossMulDepth<arith::AddFOp>, MatchCrossMulDepth<arith::SubFOp>>(
          &getContext(), &scaleCounter, getOperation(), &solver);
```

Since they all take the same constructor arguments, I suspect you can use the variable template args version of `RewritePatternSet::add`, here and elsewhere

lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.h
```diff
@@ -43,12 +60,69 @@ struct ModReduceBefore : public OpRewritePattern<Op> {
...
+                                PatternRewriter &rewriter) const override;
+
+ private:
+  int *scaleCounter;
+  Operation *top;
+  DataFlowSolver *solver;
+};
+
+template <typename Op>
+struct MatchCrossMulDepth : public OpRewritePattern<Op> {
```
j2kun, line 89: A brief docstring comment for each of these patterns would be helpful.

lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
```diff
@@ -76,66 +104,176 @@ LogicalResult ModReduceBefore<Op>::matchAndRewrite(
...
+    auto mulDepthState =
+        solver->lookupState<MulDepthLattice>(operand->get())->getValue();
+    if (!mulDepthState.isInitialized()) {
+      return failure();
+    }
+    auto mulDepth = mulDepthState.getMulDepth();
+    mulDepths.push_back(mulDepth);
+  }
+
+  // only the input level can have mul depth mismatch.
```
j2kun, line 236: why is that?

lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.h
```diff
@@ -43,12 +60,69 @@ struct ModReduceBefore : public OpRewritePattern<Op> {
...
+      : OpRewritePattern<Op>(context, /*benefit=*/1),
+        scaleCounter(scaleCounter),
+        top(top),
+        solver(solver) {}
+
+  LogicalResult matchAndRewrite(Op op,
+                                PatternRewriter &rewriter) const override;
+
+ private:
+  int *scaleCounter;
```
j2kun, line 83: Related to my earlier comment, a code comment here may be helpful to explain the role of scaleCounter.

lib/Utils/ContextAwareConversionUtils.h
```diff
@@ -253,16 +254,40 @@ class SecretGenericOpCipherPlainConversion
...
+    Attribute ciphertextEncoding =
+        ciphertextTy.getPlaintextSpace().getEncoding();
+    Attribute plaintextEncoding = lwe::getEncodingAttrWithNewScalingFactor(
+        ciphertextEncoding, cleartextMgmtAttr.getScale());
+
+    if (!plaintextEncoding) {
+      return failure();
+    }
+
+    // TODO: inherit level/dimension from no-op mgmt attr
```
j2kun, line 278: Add a tracking issue

tests/Examples/lattigo/bgv/cross_level/cross_level.mlir
```diff
@@ -0,0 +1,50 @@
...
+    // increase one level
+    %mul3 = arith.muli %base2, %base2 : tensor<4xi16>
+    // cross level add
+    %base3 = arith.addi %mul3, %add : tensor<4xi16>
+    return %base3 : tensor<4xi16>
+  }
+}
+
+// module attributes {backend.lattigo, bgv.schemeParam = #bgv.scheme_param<logN = 13, Q = [134250497, 17179967489, 17180262401, 68720050177], P = [68720066561, 68720295937], plaintextModulus = 65537, encryptionTechnique = extended>, scheme.bgv} {
+//   func.func @cross_level(%arg0: !secret.secret<tensor<4xi16>>, %arg1: !secret.secret<tensor<4xi16>>) -> !secret.secret<tensor<4xi16>> {
```
j2kun, line 22: This comment is intended to show the sample output? Not sure we should check that in since it will become stale.

tests/Examples/lattigo/ckks/cross_level/cross_level.mlir
```diff
@@ -0,0 +1,57 @@
...
+    %base2 = arith.addf %mul2, %add : tensor<4xf32>
+    // increase one level
+    %mul3 = arith.mulf %base2, %base2 : tensor<4xf32>
+    // cross level add
+    %base3 = arith.addf %mul3, %add : tensor<4xf32>
+    return %base3 : tensor<4xf32>
+  }
+}
+
+// "--annotate-module=backend=lattigo" --mlir-to-secret-arithmetic --secret-insert-mgmt-ckks='slot-number=4 before-mul-include-first-mul' --generate-param-ckks --populate-scale-ckks='before-mul-include-first-mul'
```
j2kun, line 21: ditto

j2kun (changes requested):

docs/content/en/docs/Design/management.md
```diff
@@ -0,0 +1,176 @@
...
+
+Note that, as multiplication has two operands, the actual circuit for the latter
+two policies is `mult(ms(ct0), ms(ct1))`, whereas in the first policy the
+circuit is `ms(mult(ct0, ct1))`.
+
+The third policy has one more modulus switching than others, so it will need one
+more modulus.
+
+There are also other insertion strategy like inserting it dynamically based on
+current noise (see HElib) or lazy rescaling. Those are not implemented.
```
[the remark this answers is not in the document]
ZenithalHourlyRate: It becomes `lazy modulus switching` now. It is an optimization.

lib/Analysis/LevelAnalysis/LevelAnalysis.cpp
ZenithalHourlyRate: See comment in `LevelAnalysis` docstring. Inside level analysis we count from 0 to L instead of L to 0 as we does not know L in the first place.

lib/Dialect/Mgmt/IR/MgmtCanonicalization.td
```diff
@@ -0,0 +1,48 @@
+#ifndef LIB_DIALECT_MGMT_IR_MGMTCANONICALIZATION_TD_
+#define LIB_DIALECT_MGMT_IR_MGMTCANONICALIZATION_TD_
+
+include "MgmtOps.td"
+include "lib/Utils/DRR/Utils.td"
+include "mlir/IR/PatternBase.td"
+
+// mod reduce should be put after adjust scale
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Put a link to `SecretInsertMgmtPatterns`. I still want to use the canonicalizer instead of a separate pass as it is handy. Some transformation is generic like merging LevelReduce, though some transformation listed here do is not equivalent. I'll leave it here for simplicity until one day we have more required pattern by other passes.

lib/Dialect/Mgmt/IR/MgmtCanonicalization.td
```diff
@@ -0,0 +1,48 @@
...
+  (Mgmt_LevelReduceOp:$levelReduceOp2 (Mgmt_LevelReduceOp:$levelReduceOp1 $input, $levelToDrop1), $levelToDrop2),
+  (Mgmt_LevelReduceOp $input, (AddI64Attr $levelToDrop1, $levelToDrop2)),
+  [(HasOneUse:$levelReduceOp1), (HasOneUse:$levelReduceOp2)]
+>;
+
+def LevelOne : NativeCodeCall<"$_builder.getI64IntegerAttr(1)">;
+
+// Technically this transformation is not equivalent, as two mod reduce ops will
+// change the scaling factor twice. This assumes there is an AdjustScaleOp somewhere
+// in the line to correct the scaling factor.
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Commented above.

lib/Parameters/CKKS/Params.h
```diff
@@ -33,6 +33,18 @@ class SchemeParam : public RLWESchemeParam {
   static SchemeParam getSchemeParamFromAttr(SchemeParamAttr attr);
 };
 
+// Parameter for each SSA ciphertext SSA value.
+class LocalParam : public RLWELocalParam {
+ public:
+  LocalParam(const SchemeParam *schemeParam, int currentLevel, int dimension)
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Well only after `populate-scale` the local param of scale is meaningful. For now I do not add it for simplicity. When we actually need it (CKKS noise model on the way!), that new PR will add the scale.

By then we will have something like `secret-insert-mgmt-ckks -> generate-param-ckks -> populate-scale-ckks (logScale) -> generate-param-ckks-reduced-error (high-precision-scale) -> populate-scale-ckks (again)`.

lib/Transforms/PopulateScale/PopulateScaleBGV.cpp
```diff
@@ -0,0 +1,155 @@
...
+        : OpRewritePattern<mgmt::AdjustScaleOp>(context, /*benefit=*/1),
+          plaintextModulus(plaintextModulus) {}
+
+    LogicalResult matchAndRewrite(mgmt::AdjustScaleOp op,
+                                  PatternRewriter &rewriter) const override {
+      auto inputScale =
+          mgmt::findMgmtAttrAssociatedWith(op.getInput()).getScale();
+      auto scale = mgmt::findMgmtAttrAssociatedWith(op).getScale();
+      // no need to adjust scale
+      if (scale == inputScale) {
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Now it becomes `adjust_scale {id = counter}` so on its own it does not know how to fold.

lib/Transforms/PopulateScale/PopulateScaleCKKS.cpp
```diff
@@ -0,0 +1,169 @@
...
+    OpPassManager annotateMgmt("builtin.module");
+    annotateMgmt.addPass(mgmt::createAnnotateMgmt());
+    (void)runPipeline(annotateMgmt, getOperation());
+
+    // convert adjust scale to mul plain
+    RewritePatternSet patterns(&getContext());
+    patterns.add<ConvertAdjustScaleToMulPlain>(&getContext());
+    walkAndApplyPatterns(getOperation(), std::move(patterns));
+
+    // run canonicalizer and CSE to clean up arith.constant and move no-op out
```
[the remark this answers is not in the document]
ZenithalHourlyRate: I insert `mgmt.init` right before the HE operation because I used to encounter a situation where plaintext is in the loop

```mlir
for 
  pt = tensor.extract weight
  0 = mgmt.init pt
  1 = add ct, 0 
```

though currently we totally skip this pass for loops, I still maintain it in secret.generic and ask canoicalizer to move it only when possible.

lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
ZenithalHourlyRate: See `MatchCrossMulDepth` docstring.

ZenithalHourlyRate (commented):
Sorry for not replying every review comments and directly marking them as resolved as people will receive too many emails for them. These are handled in code accordingly.

(one annoying part of github review is that once I push a new version and comments become outdated, I could not submit reply comments in batch)

As for those questions, I put explanation in the code comment as I believe every other person will have similar questions.

ZenithalHourlyRate:
Fixes #1169 #1364 #785

lib/Dialect/Mgmt/IR/MgmtOps.td
```diff
@@ -86,4 +113,56 @@ def Mgmt_BootstrapOp : Mgmt_Op<"bootstrap"> {
...
+
+  let description = [{
+    This is scheme-agonistic operation that adjust the scale of the input
+    ciphertext to the specified scale.
+
+    The scale is specified by the `scale` attribute.
+
+    At the time of `secret-insert-mgmt-<scheme>`, the concrete scale
+    is not known as the scheme parameter is not generated. After that pass
+    the scale defaults to _distinct_ negative numbers so it could not be lowered.
```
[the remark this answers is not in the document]
j2kun: Understood. I think mostly the wording was confusing me.

j2kun (approved):
One minor conflict to resolve and then we are good to go!

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -85,7 +85,8 @@ def CKKS_MulOp : CKKS_Op<"mul", [Commutative, SameOperandsAndResultRings, InferT
   let hasVerifier = 1;
 }
 
-def CKKS_MulPlainOp : CKKS_CiphertextPlaintextOp<"mul_plain", [InferTypeOpAdaptor, AllCiphertextTypesMatch, Commutative]> {
+// MulPlain op result ciphertext type could be different from the input
+def CKKS_MulPlainOp : CKKS_CiphertextPlaintextOp<"mul_plain", [InferTypeOpAdaptor, Commutative]> {
```
ZenithalHourlyRate, line 89: Cc @asraa on #1620: BGV/CKKS MulPlainOp does not require result and input ciphertext types to be the same in that they may have different scale. The verification of result scale = ct scale + pt scale is done by InferTypeOpAdaptor in https://github.com/google/heir/blob/55d883daa628be169db3f8c0e05b1076e9b3328d/lib/Dialect/LWE/IR/LWEOps.h#L69-L78

ZenithalHourlyRate (commented):
Rebased to resolve merge conflict

ZenithalHourlyRate:
Rebased to resolve merge conflict

ZenithalHourlyRate:
Merged by copybara in c7af5286429668dcdc9b3f35be5373e6cc9cedbd

Issue #1311, opened 2025-01-25
Support `func.call` / Remove single-function assumption


AlexanderViand-Intel:
Currently, various places in the compiler assume that the entire program is contained inside a single function.
However, we clearly want to support the ability to call other functions, though probably with the limitation that the call graph must be trivially resolvable at compile time and must have no cycles (i.e., no recursion).

Related PRs:
 
* #1309
* #1310

PR #1466, opened 2025-02-25
Support cleartext execution mode

ZenithalHourlyRate:
With the discussion on _ciphertext semantic cleartext execution_, it might be preferable to lower the secret arithmetic IR to LLVM and execute them in mlir-runner to inspect the execution result.

There should be two versions of cleartext execution, one with the original type (i16, f16) and one with the FHE underlying type (e.g. `mod_arith<65537>`, f64?)

This PR implements the lowering for tensor_ext.rotate to tensor dialect, and with that we can lower to LLVM using `heir-polynoimal-to-llvm`.

```bash
--mlir-to-secret-arithmetic --tensor-ext-to-tensor --secret-forget-secrets --cse
# on dot_product_8.mlir
``` 

The trouble is that, for the test file `dot_product_8_test.mlir`, we can not run

```bash
--mlir-to-secret-arithmetic --tensor-ext-to-tensor --secret-forget-secrets --cse --heir-polynomial-to-llvm
```

As we are facing these two problems

* printMemrefI32 is external function, which runs into segfault somewhere
* in `test` body we have `call @dot_product`. However, for `--wrap-generic` the signature of `dot_product` will change and func.call will fail validation.

### DIFF
--- a/lib/Dialect/TensorExt/Conversions/TensorExtToTensor/TensorExtToTensor.cpp
+++ b/lib/Dialect/TensorExt/Conversions/TensorExtToTensor/TensorExtToTensor.cpp
@@ -1,0 +1,100 @@
+#include "lib/Dialect/TensorExt/Conversions/TensorExtToTensor/TensorExtToTensor.h"
+
+#include "lib/Dialect/TensorExt/IR/TensorExtDialect.h"
+#include "lib/Utils/ConversionUtils.h"
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"           // from @llvm-project
+#include "mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h"  // from @llvm-project
+
+namespace mlir::heir::tensor_ext {
+
+#define GEN_PASS_DEF_TENSOREXTTOTENSOR
+#include "lib/Dialect/TensorExt/Conversions/TensorExtToTensor/TensorExtToTensor.h.inc"
+
+struct ConvertRotateOp : public OpRewritePattern<RotateOp> {
+  ConvertRotateOp(mlir::MLIRContext *context)
+      : OpRewritePattern<RotateOp>(context) {}
+
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(RotateOp op,
+                                PatternRewriter &rewriter) const override {
+    auto shift = op.getShift();
+    auto constantOp =
+        dyn_cast_or_null<arith::ConstantOp>(shift.getDefiningOp());
+    if (!constantOp) {
+      return failure();
+    }
+
+    auto shiftValue = cast<IntegerAttr>(constantOp.getValue()).getInt();
+    auto tensorShape =
+        cast<RankedTensorType>(op.getTensor().getType()).getShape();
+
+    // only support 1D tensors
+    // TODO(#924): Currently RotateOp only supports rotating a 1-D vector, or a
+    // vector with only one non-unit dimension that is treated as the major
+    // dimension.
+    if (tensorShape.size() != 1) {
+      return failure();
+    }
+
+    auto tensorSize = tensorShape[0];
+    if (shiftValue < 0) {
+      shiftValue += tensorSize;
+    }
+
+    auto tensorElementType = getElementTypeOrSelf(op.getTensor().getType());
+    auto leftTensorType =
+        RankedTensorType::get({shiftValue}, tensorElementType);
+    auto rightTensorType =
+        RankedTensorType::get({tensorSize - shiftValue}, tensorElementType);
+
+    auto left = rewriter.create<tensor::ExtractSliceOp>(
+        op.getLoc(), leftTensorType, op.getTensor(), ArrayRef<Value>{},
+        ArrayRef<Value>{}, ArrayRef<Value>{},
+        /*offsets=*/ArrayRef<int64_t>{0},
+        /*sizes=*/ArrayRef{shiftValue}, /*strides=*/ArrayRef<int64_t>{1});
+    auto right = rewriter.create<tensor::ExtractSliceOp>(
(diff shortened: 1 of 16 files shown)


ZenithalHourlyRate:
> As we are facing these two problems
> 
> * printMemrefI32 is external function, which runs into segfault somewhere

Fixed that

> * in `test` body we have `call @dot_product`. However, for `--wrap-generic` the signature of `dot_product` will change and func.call will fail validation.

I just put the following signature with no `secret.secret` attached. Seems that `heir-vectorizer` works well for ops that is not in `secret.generic` (is that expected? that may cause unwanted effect like transforming other test codes)
```
func.func @dot_product(%arg0: tensor<8xi16>, %arg1: tensor<8xi16>) -> i16
```

With that said, although the e2e test looks good now. We should be cautious that `mlir-to-secret-arithmetic` for non-wraped version and wraped version may be different, (e.g. operation-balancer only transforms op in `secret.generic`)

One solution is to put `{secret.secret}` in test file, and when wrap generic, it transforms both the main function and the call the the main function (using `secret.conceal`) and when forget-secretness these transforms are dropped.

The full pipeline for the cleartext backend should be

```bash
{--secretize/warn if no {secret.secret} found) --mlir-to-secret-arithmetic --tensor-ext-to-tensor --secret-forget-secrets --heir-polynomial-to-llvm
```

ZenithalHourlyRate:
> With that said, although the e2e test looks good now. We should be cautious that `mlir-to-secret-arithmetic` for non-wraped version and wraped version may be different, (e.g. operation-balancer only transforms op in `secret.generic`)
> 
> One solution is to put `{secret.secret}` in test file, and when wrap generic, it transforms both the main function and the call the the main function (using `secret.conceal`) and when forget-secretness these transforms are dropped.

Changed `wrap-generic` to support func.call. Related to #1311.

There are two use cases for this early pipeline, one is for testing, another is for multi-function

```mlir
func.func @main({secret.secret})

func.func @test_main() {
  call @main
}
```

```mlir
func.func @func1({secret.secret})

func.func @func2({secret.secret})

func.func @main({secret.secret}) {
  call @func1
  call @func2
}
```

For the first case the current change works well, yet for the second case we will get the following IR

```mlir
func.func @main(!secret.secret) {
  secret.generic {
    secret.conceal
    call @func1
    secret.reveal
    secret.conceal
    call @func2
    secret.reveal
  }
}
```

To support that we might change the secret canonicalization rule to split the secret.generic, yet this will make the "single-secret.generic" assumption from later passes invalid.

lib/Dialect/TensorExt/Conversions/TensorExtToTensor/TensorExtToTensor.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace mlir::heir::tensor_ext {
+
+#define GEN_PASS_DEF_TENSOREXTTOTENSOR
+#include "lib/Dialect/TensorExt/Conversions/TensorExtToTensor/TensorExtToTensor.h.inc"
+
+class TensorExtToTensorTypeConverter : public TypeConverter {
+ public:
+  TensorExtToTensorTypeConverter(MLIRContext *ctx) {
+    addConversion([](Type type) { return type; });
```
j2kun, line 17: If there is no type conversion then we can skip the whole DialectConversion framework and just implement this with the walk-based rewriter.

tests/Examples/cleartext/dot_product_8_test.mlir
```diff
@@ -0,0 +1,39 @@
+// RUN: heir-opt %s --mlir-to-secret-arithmetic | FileCheck %s --check-prefix=CHECK_MLIR_TO_SECRET_ARITHMETIC
+
+// RUN: heir-opt %s --mlir-to-secret-arithmetic --tensor-ext-to-tensor --secret-forget-secrets --heir-polynomial-to-llvm \
+// RUN:   | mlir-runner -e test_dot_product -entry-point-result=void \
```
j2kun, line 4: Some of the complexity is in the use of mlir-runner. We could instead lower all the way to machine code and use a C++ test harness. See for example https://github.com/google/heir/blob/9c2848bae24aac36dd8d449c1eacfc0dbe965038/tests/Dialect/Arith/Transforms/runner/lower_split.mlir#L2

The bonus of not using mlir-runner is that it alone contributes to more than half the build time of the entire project (because it pulls in all dialects and transforms from upstream). So let's try avoiding to use it.

lib/Transforms/Secretize/WrapGeneric.cpp
```diff
@@ -103,15 +106,81 @@ struct WrapWithGeneric : public OpRewritePattern<func::FuncOp> {
...
+                                PatternRewriter &rewriter) const override {
+    auto module = mlir::cast<ModuleOp>(top);
+    auto callee = module.lookupSymbol<func::FuncOp>(op.getCallee());
+    if (callee.isDeclaration()) {
+      return success();
+    }
+
+    SmallVector<Value> newOperands;
+    auto funcResultTypes = llvm::to_vector(callee.getResultTypes());
+    llvm::dbgs() << funcResultTypes.size();
```
j2kun, line 123:
```suggestion
```

j2kun (changes requested):
I like the wrapping in conceal/reveal, I think that will be nice and future proof.

tests/Examples/cleartext/dot_product_8_test.mlir
ZenithalHourlyRate: Using c test harness means we have to deal with the memref descriptor for _tensor type_, which is quite implicit.

I had some experience dealing with it when doing the previous version of plaintext execution at https://github.com/ZenithalHourlyRate/heir/commit/617e0d163690710c5ed44840553e9193d936ea33 so I am now able to know which signature I would use, for either raw signature or the `llvm.emit_c_interface` signature, but I would admit that is painful.

Also this introduces clang and mlir-translate dependency, and if we enable it by default (instead of a "manual" tag in BUILD), we will have more build time.

j2kun: I think this is much better, at least in part because `mlir-translate` doesn't have all the LLVM passes in it, and `clang` is a dependency we can get from the host machine if need be, while `mlir-runner` pulls in every dialect and transform from MLIR upstream, plus all the LLVM infrastructure they rely on, which is a LOT.

j2kun (approved):

Issue #1216, opened 2024-12-20
Add Canonicalization Patterns for ModArith


AlexanderViand-Intel:
Upstream MLIR defines [a whole range of peephole optimizations](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/IR/ArithCanonicalization.td) for operations in the `arith` dialect in the form of [canonicalization ](https://mlir.llvm.org/docs/Canonicalization/) patterns. 

In HEIR, our `mod_arith` Dialect, which models arithmetic modulo some (frequently non-power-of-two) modulus, could benefit from similar arithmetization patterns.

j2kun:
@AlexanderViand-Intel had a potential masters student lined up for this. Is that still a potential collaboration, or is it open now?

j2kun:
@Maokami are there any more ideas for canonicalizations for ModArith? Or should we close this issue for now?

Maokami:
I think the basic operations like `add`, `sub`, ... are mostly covered, but there’s still room to add more canonicalizations for operations I haven’t addressed yet — for example, something like `mod_arith.subifge(x, 0) -> x`.

Alternatively, we could close this issue and continue extending the canonicalization pass for ModArith in a new one. I’m happy to go either way — feel free to decide what works best!

j2kun:
Let's add any ideas we have for specific canonicalizations to a new issue.

PR #1470, opened 2025-02-26
heir-vectorizer: support mod arith input

ZenithalHourlyRate:
In an effort to make the _ciphertext semantic cleartext backend_, which should support execution in Zt, the plaintext modulus ring, we may transform the input from `i16` to `mod_arith<ptm>`, which could be done by the existing `--arith-to-mod-arith` with small modification.

Note that `--arith-to-mod-arith` should happen before `--wrap-generic` as it seems designed to be so.

Meanwhile, the heir-vectorizer is unaware of mod_arith type so no simd optimization will happen if we apply `--mlir-to-secret-arithmetic`. This PR is meant for making various passes in heir-vectorizer understand mod arith type and produce the same circuit for `i16` input and `mod_arith` input.

### Current status

* `insert-rotation` works after adding DRR rules
* `cse` following `insert-rotation` seems to ignore some optimization chance (i.e. no clean up as expected). Might be resolved by #1216?
* `rotate-and-reduce` does not work
* `arith-to-mod-arith` should support custom modulus in pass options.

### Side note

We may also just cheat through the heir-vectorizer by the following pipeline. However, `forget-secrets` then `secretize` will lose information on argument secretness. E.g. plaintext func argument will be secret now.

```bash
# first make i16 input transformed to the wanted circuit
# then use arith-to-mod-arith to convert type
--mlir-to-secret-arithmetic --secret-forget-secrets \
--secretize --arith-to-mod-arith --wrap-generic
```

### DIFF
--- a/lib/Analysis/RotationAnalysis/BUILD
+++ b/lib/Analysis/RotationAnalysis/BUILD
@@ -9,6 +9,7 @@ cc_library(
     srcs = ["RotationAnalysis.cpp"],
     hdrs = ["RotationAnalysis.h"],
     deps = [
+        "@heir//lib/Dialect/ModArith/IR:Dialect",
         "@heir//lib/Dialect/TensorExt/IR:Dialect",
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:Analysis",
--- a/lib/Analysis/RotationAnalysis/RotationAnalysis.cpp
+++ b/lib/Analysis/RotationAnalysis/RotationAnalysis.cpp
@@ -1,5 +1,6 @@
 #include "lib/Analysis/RotationAnalysis/RotationAnalysis.h"
 
+#include "lib/Dialect/ModArith/IR/ModArithOps.h"
 #include "lib/Dialect/TensorExt/IR/TensorExtOps.h"
 #include "llvm/include/llvm/ADT/STLExtras.h"   // from @llvm-project
 #include "llvm/include/llvm/ADT/TypeSwitch.h"  // from @llvm-project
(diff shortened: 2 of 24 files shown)


ZenithalHourlyRate:
Ready for review. Now the heir-vectorizer is fully aware of mod_arith type and existing heir vectorizer test also applies for mod arith.

Though I did not add the test for box_blur_64x64 and roberts_cross_64x64 as they already take a long time to run.

arith-to-mod-arith now also supports a custom modulus. So once #1483 is implemented, we can support _ciphertext semantic cleartext execution in Zp_ similar to #1466

j2kun:
I'm not sure if I understand the rationale here fully, and I am noticing this by being confused why we're running `arith-to-mod-arith` at all. Maybe we should take a step back and outline how the overall pipeline should work.

I would expect that the main work would be to add a new `secret-to-mod-arith` pass (or `secret-to-plaintext` if we'd lower to `mod_arith` polynomials?) that starts from the same IR as the `secret-to-<scheme>` and lowers to plaintext types and ops.

If we start making special changes for `mod_arith` to the first part of the pipeline to support this, then we will inevitably introduce bugs where the plaintext evaluation is not equivalent to the scheme evaluation because of some differing behavior in early passes. The only "acceptable" surface area for bugs here should be where the pipelines actually differ, which would be after lowering to a scheme.

@AlexanderViand-Intel thoughts?

AlexanderViand-Intel:
> @AlexanderViand-Intel thoughts?

I think Jeremy's description is also what I had in mind: basically, adding a "scheme" lowering that, where the BGV lowering might lower `arith.addi %a, %b : i16` to `bgv.add %cta, %ctb` , this would lower it to `mod_arith.add %a, %b` with the modulus set to the ptxt mod the scheme would have had.

ZenithalHourlyRate:
> I would expect that the main work would be to add a new secret-to-plaintext pass, which lowers secret types to mod_arith polynomials, and which starts from the same IR as the secret-to-<scheme> and lowers to plaintext types and ops.

So it is `--mlir-secret-to-arithmetic --arith-to-mod-arith`. But as you can see the current `arith-to-mod-arith` pass does not support converting `secret.generic` as it also has its uses in CGGI pipeline. If we do only want to convert things in `secret.generic`, we could add another pass `--secret-arith-to-mod-arith`.

However, that pass would duplicate a lot with `--arith-to-mod-arith`, so I am trying to make `--arith-to-mod-arith --mlir-secret-to-arithmetic` equivalent to `--mlir-secret-to-arithmetic --arith-to-mod-arith`.

> which starts from the same IR as the `secret-to-<scheme>` and lowers to plaintext types and ops.

Then we will have a `bgv-to-arith`, which is just the inverse of `secret-to-bgv`, then we could start from the beginning of `--mlir-to-secret-arithmetic --secret-insert-mgmt-<scheme> --optimize-relinearization --secret-distribute-generics`, then as in the lowering we just discard mgmt ops, we can start from `--mlir-to-secret-arithmetic`, then if `--mlir-to-secret-arithmetic --lowering` works, we could also make `--lowering --mlir-to-secret-arithmetic` work.

On the other hand, I think for the integer arithmetic pipeline, shouldn't we convert arith to mod arith in the first pass? As we never have `i16` semantic in the whole FHE pipeline, ideally the whole pipeline should only see `mod_arith`.

I would propose for all bgv/bfv lowerings `--arith-to-mod-arith` becomes the first pass as it is the actual underlying math structure for user input instead of `i16`. Of course we might in the future add a pass to say the user input `i16` and we decide smartly when fitting it into `mod_arith`, but for all subsequent pipeline they should see `mod-arith` instead of `i16`.

AlexanderViand-Intel:
> I would propose for all bgv/bfv lowerings `--arith-to-mod-arith` becomes the first pass as it is the actual underlying math structure for user input instead of `i16`. Of course we might in the future add a pass to say the user input `i16` and we decide smartly when fitting it into `mod_arith`, but for all subsequent pipeline they should see `mod-arith` instead of `i16`.

I see the reasoning behind this, but I don't think it's a good idea to do this, as it means we'd never be able to use any `arith`-using upstream passes in the arithmetization phase of HEIR. I think our semantics (if it doesn't overflow in `i16` it won't overflow in `ptxt-mod`) are easy enough to achieve that we don't need to convert to `mod_arith` that early

ZenithalHourlyRate:
Then for cleartext backend with Zp we may work around by `--mlir-to-secret-arithmetic --secret-forget-secrets --arith-to-mod-arith=ptm --to-llvm-lowering` where

1. `arith-to-mod-arith` works on a transformed IR with no secret
2. no second `wrap-generic`

j2kun:
IMO the question of how to ensure that user inputs fit into scheme plaintexts is an independent question. I think we will probably need a pass at some point that does things like convert a 64-bit integer program into appropriately-chosen RNS types that are small enough to fit in scheme plaintexts. Such a pass would probably end up emitting mod-arith, and for the reasons Alex says I'd expect it to not be the very first thing we do, but I'm not sure exactly where it would live.

If we can get everything to work as expected by forgetting secrets and then doing arith-to-mod-arith, then that might suffice. Maybe we can wrap it in a single pass called `-secret-to-plaintext` so that (1) we can later augment it to support CGGI and (2) we can hide the implementation details of the pass in case we change it later and (3) to make it easier to tell what the purpose of the pass is, give it nice docs, etc.

ZenithalHourlyRate:
Will split useful changes into other PR

PR #1484, opened 2025-02-27
`mod_arith.constant` parser fixes / improvements

### LINKED ISSUE #1482: `mod_arith.constant` silently fails for tensor types
Using a tensor type as return type for `mod_arith.constant` *should* be impossible (it's defined with `let results = (outs ModArith_ModArithType:$output);`) but it somehow does not trigger a verifier error. Instead, it causes `heir-opt` to produce no output on stdout or stderr, but it does not crash.  See example reproducer below:

```llvm
// RUN: heir-opt %s
!t = tensor<2x!mod_arith.int< 17 : i32>>

func.func @foo() -> !t {
    %c = mod_arith.constant 42 : !t
    func.return %c : !t
}
```

I'm pretty stumped by this, not even sure where to start debugging on this. Of course, we probably want to enable `mod_arith.constant dense<42>` similar to how `arith` handles this anyway, and that should make this bug moot, but it's so odd that I thought it might be worth investigating first.

AlexanderViand-Intel:
Fixes #1482 and improves the somewhat complicated handling around `mod_arith.constant` in general:

* removes redundant `ModArithAttr` attribute that was only used by `mod_arith.constant`
* removes return type inference for `mod_arith.constant` based on type redundancy caused by `ModArithAttr`
* changes type of `mod_arith.constant` value from `ModArithAttr` to `IntegerAttr`
* adds verifier for `mod_arith.constant` to ensure that the constant value bitwidth is smaller than the underlying type bitwidth
* ensures that all paths in the custom parser will emit an error (no more silent failures in parsing)

It also does the last one for `polynomial.constant`'s parser.


lib/Dialect/ModArith/IR/ModArithDialect.cpp
```diff
@@ -135,52 +127,38 @@ LogicalResult BarrettReduceOp::verify() {
...
-  if (parsedValue.getActiveBits() > outputBitWidth)
-    return parser.emitError(parser.getCurrentLocation(),
-                            "constant value is too large for the modulus");
+LogicalResult ConstantOp::verify() {
+  auto valueBW = getValue().getBitWidth();
+  auto modBW = getType().getModulus().getValue().getBitWidth();
+  // Note: comparison will crash if the bitwidths aren't the same
+  if (valueBW > modBW ||
+      (valueBW == modBW && !getValue().ult(getType().getModulus().getValue())))
+    return emitOpError("Constant value must be smaller than modulus.");
```
j2kun, line 150: I recall discussing this at length a while back, and I don't think it makes sense to enforce this at parse time. In particular, negative attributes would bypass this check, and we have cases where `-1` is the most convenient value to provide to a `constant` op.

Instead, I think it would be easier to allow any integer as input, and let canonicalization or other ops like `mod_arith.reduce` handle converting a value to its canonical representative in the ring.

j2kun (changes requested):

AlexanderViand-Intel:
> The purpose of `ModArithAttr` is also for constant folding, but I suppose we can make it work with just `IntegerAttr`

Oh, good point! I think it should indeed be fine with `IntegerAttr` (and `DenseIntElementsAttr` in the future) though, as I don't think we'd ever need the modulus information that the `ModArithAttr` wrapper would give us (since any folding/etc would only happen in cases where the moduli agree).

AlexanderViand-Intel:
I just found a bug/issue with the verifier: it the modulus has `spare` bitwidth, we can get cases like 
`%const123 = mod_arith.constant 19 : !mod_arith.int<17 : i32>` which passes, because 19 gets bitwidth `8`, and the actual number-to-number comparison can only be run when the bitwidths are equal (or APInt.ult will throw an errror).

So I need to look at activeBits instead, I think!

j2kun:
> I just found a bug/issue with the verifier: it the modulus has `spare` bitwidth, we can get cases like `%const123 = mod_arith.constant 19 : !mod_arith.int<17 : i32>` which passes, because 19 gets bitwidth `8`, and the actual number-to-number comparison can only be run when the bitwidths are equal (or APInt.ult will throw an errror).
> 
> So I need to look at activeBits instead, I think!

Cf. my review comment, I don't think we should do any verification on the actual value.

AlexanderViand-Intel:
> > I just found a bug/issue with the verifier: it the modulus has `spare` bitwidth, we can get cases like `%const123 = mod_arith.constant 19 : !mod_arith.int<17 : i32>` which passes, because 19 gets bitwidth `8`, and the actual number-to-number comparison can only be run when the bitwidths are equal (or APInt.ult will throw an errror).
> > So I need to look at activeBits instead, I think!
> 
> Cf. my review comment, I don't think we should do any verification on the actual value.

Yeah, I think that makes sense, as I just remembered `mod_arith` explicitly allows values that go over the modulus (but of course not the underlying Integer Type width) at some points. I'll simplify the verifier to only check the value fits into the underlying type!

j2kun (approved):

ZenithalHourlyRate:
The following IR will fail parsing

```
// Constant Value's bitwidth must be smaller than underlying type
%c0 = mod_arith.constant 0 : !mod_arith.int<255 : i32>
```

while, interestingly, the following IR is ok

```
// generic form: "mod_arith.constant"() <{value = 1 : i4}> : () -> !mod_arith.int<255 : i32>
mod_arith.constant 1 : !mod_arith.int<255 : i32>
```

Write it in IR and heir-lsp would complain these when hovering on that op.

AlexanderViand-Intel:
> The following IR will fail parsing
> 
> ```
> // Constant Value's bitwidth must be smaller than underlying type
> %c0 = mod_arith.constant 0 : !mod_arith.int<255 : i32>
> ```
> 
> while, interestingly, the following IR is ok
> 
> ```
> // generic form: "mod_arith.constant"() <{value = 1 : i4}> : () -> !mod_arith.int<255 : i32>
> mod_arith.constant 1 : !mod_arith.int<255 : i32>
> ```


:(  It seems like the Integer parser will give non-zero types the next power-of-two bitwidth that fits, but will give zero an `i64` size by default...

Issue #1488, opened 2025-02-27
TODO-Backlinks broken


AlexanderViand-Intel:
I'm assuming @j2kun has seen this already, but posting here for completeness:

As of the last few commits to main, the TODO-Backlinks workflow fails with errors like
```
Updating #1174
    update.comment.edit(update.new_comment_body)
  File "/usr/local/lib/python3.12/site-packages/github/IssueComment.py", line 121, in edit
    headers, data = self._requester.requestJsonAndCheck("PATCH", self.url, input=post_parameters)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/github/Requester.py", line 494, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, self.__customConnection(url)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/github/Requester.py", line 525, in __check
    raise self.createException(status, responseHeaders, data)
github.GithubException.GithubException: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/issues/comments#update-an-issue-comment", "status": "403"}
```

PR #1489, opened 2025-02-27
set permissions on todo backlinks workflow

j2kun:
I think something changed with org-wide default permissions for runners, so this should override that with explicit permissions for this action.

Maybe fixes https://github.com/google/heir/issues/1488 ? Hard to tell


AlexanderViand-Intel (approved):
Let's see if this does the trick🤞

PR #1493, opened 2025-02-28
ModArith: support some folding method for add

ZenithalHourlyRate:
Splitted from #1470 

Folders copied from ArithOps.cpp of MLIR

Should be rebased after #1484 is in.


ZenithalHourlyRate:
Closed in favor of #1506

PR #1494, opened 2025-02-28
Split parameter generation into standalone pass

ZenithalHourlyRate:
Parameter generation process in `validate-noise` is no longer adaquate and deserves a standalone pass.

This also adds a default parameter generation process for BFV that generates all 60bit moduli chain, whose length is just the multiplicative depth. Further noise model for BFV could refine such process.

Now `secret-to-<scheme>` requires a explicit scheme param attr for proper lowering to LWE type.


j2kun (approved):
Needs a rebase, then LGTM!

ZenithalHourlyRate:
Rebased

PR #1497, opened 2025-02-28
enable (1-dim) dense `mod_arith.constant`

### LINKED ISSUE #1483: mod_arith.constant should support creating constant tensor
In user program, `arith.constant : tensor` often appears, so during `arith-to-mod-arith` they should be converted.

Also TODO holder for #1470

AlexanderViand-Intel:
* (Partially) fixes #1483 
* In addition to adding dense, this PR also fixes the `zero becomes i64` issue @ZenithalHourlyRate  spotted (See #1484)

I'm not very happy with or attached to the current implementation, so happy to hear improvement suggestions!

~~For some reason, I was set on avoiding a `dyn_cast` of the parsedType to get at the underlying type/bitwidth of `mod_arith.int` but I think that's the only reasonable way to do this. Right now, the attribute on the op might have any IntegerType, as long as its smaller than the underlying type of `mod_arith.int`, but that just seems like it'd cause annoying crashes down the line in folders/lowerings/etc, due to bitwidth-equality asserts being all over the place in MLIR.~~ 
Update: I tried this, and it actually causes MORE headaches to have them (need to) match, as we can't just grab an attribute of `arith.constant` and hand it off to `mod_arith.constant`, which is a pretty common use case.

~~Also, there's still the open question on how to enable multi-dim dense values: more re-inventing parsers, or accepting a 64-bit limit on dense values? (I know `polynomial`'s attributes already have a 64-bit limit, probably for similar reasons?)~~
While the `ArrayAttr` approach does parse mult-dim arrays, it doesn't let you access them in a helpful way, so both are probably equal amounts of effort. I think I'll leave adding multi-dim support for another time, as this "quick fix" is already spiraling out of control anyway. The "correct" solution is probably to expose more parsing stuff upstream anyway.

### DIFF
--- a/lib/Dialect/ModArith/IR/ModArithOps.td
+++ b/lib/Dialect/ModArith/IR/ModArithOps.td
@@ -91,8 +91,8 @@ def ModArith_ConstantOp : Op<ModArith_Dialect, "constant", [Pure]> {
     %0 = mod_arith.constant 123 : !mod_arith.int<65537:i32>
     ```
   }];
-  let arguments = (ins APIntAttr:$value);
-  let results = (outs ModArith_ModArithType:$output);
+  let arguments = (ins TypedAttrInterface:$value);
+  let results = (outs ModArithLike:$output);
   let hasCustomAssemblyFormat = 1;
   let hasVerifier = 1;
 }
--- a/lib/Dialect/ModArith/IR/ModArithDialect.cpp
+++ b/lib/Dialect/ModArith/IR/ModArithDialect.cpp
@@ -127,33 +127,101 @@ LogicalResult BarrettReduceOp::verify() {
 }
 
 ParseResult ConstantOp::parse(OpAsmParser &parser, OperationState &result) {
-  APInt parsedInt;
+  unsigned minBitwidth = 4;  // bitwidth assigned by parser to integer `1`
   Type parsedType;
+  if (parser.parseOptionalKeyword("dense").succeeded()) {
+    // Dense case
+    // We parse the integers as a list, rather than an ArrayAttr, so we can more
+    // easily convert them to the correct bitwidth (ArrayAttr forces I64)
+    std::vector<APInt> parsedInts;
+    if (parser.parseLess() ||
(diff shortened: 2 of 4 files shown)


AlexanderViand-Intel:
Reminder to myself: add test cases for both scalar and dense versions with `0`, since it _is_ somewhat of a special case. Also probably a good idea to add some test cases with negative numbers.

AlexanderViand-Intel:
I realized I'd forgotten to update the printer, but the `syntax.mlir` tests didn't trigger because they only test parsing - they dont' have "real" `//CHECK` lines (just test for presence) and therefore don't care if the printer produces some garbage that's different than then expected syntax. I think the lesson here is that it might be good practice to actually have "real" `//CHECK` statements even in syntax test files, especially where custom parsers are involved (types, attributes, Ops with custom parsers, etc).

lib/Dialect/ModArith/IR/ModArithOps.td
```diff
@@ -91,8 +91,8 @@ def ModArith_ConstantOp : Op<ModArith_Dialect, "constant", [Pure]> {
     %0 = mod_arith.constant 123 : !mod_arith.int<65537:i32>
     ```
```
ZenithalHourlyRate, line 92: Maybe add a dense example here?

ZenithalHourlyRate (approved):
LGTM!

Some future suggestion is to support syntax like `dense<0> : tensor<1024xmod_arith>`, lets make that an issue.

I indeed encounter multi-dim arith.constant in HEIR in gx_kernel_8x8.mlir, which after `--canonicalize` will emit a `%cst = arith.constant dense<[[1, -1, 2], [-2, 1, -1], [0, 0, 0]]> : tensor<3x3xi16>`, though after `--heir-vectorier` such thing would disappear.

j2kun:
FYI, we have an intermediate workaround for stuff we don't support in the constant syntax: define it as an arith constant and then use the mod_arith.encapsulate op to convert to mod arith. That should work for all tensors.

j2kun (approved):

PR #1499, opened 2025-02-28
Remove cfg=exec from bazel executable targets

j2kun:
And update module lock file

Cf. https://github.com/google/heir/issues/847#issuecomment-2691438353

This reduces build time by ~40% in `-c fastbuild` mode, and 38% in `-c opt`.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -937,7 +937,7 @@
     },
     "@@rules_python+//python/extensions:pip.bzl%pip": {
       "general": {
-        "bzlTransitiveDigest": "b5FkZYymLJqXrpFALT3yo4vmvkIJlCXRa+rRlrNfsi8=",
+        "bzlTransitiveDigest": "npnHl1jGdnSt5hglsMHC97vRzHsub5IeUUtFGwSXIyw=",
         "usagesDigest": "ogzJv1UVhgnb/xpM7ytUxAUUpEs+C8eyACas/3Xgl0A=",
         "recordedFileInputs": {
           "@@//frontend/requirements.txt": "af2cd992a762dabd7d9b4dc32667c9aecc82550f90aab96d637bed7496e14963",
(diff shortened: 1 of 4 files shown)


ZenithalHourlyRate (approved):
MLIR would not be built twice if I call `test //tests/...` now. Glad to see this!

I have the following result

```bash
# heir-opt built, only rerun all tests
$ bazel test --cache_test_results=no //tests/...
INFO: Elapsed time: 228.330s, Critical Path: 225.86s
```

Those 3 box_blur (from heir_vectorizer, openfhe/lattigo e2e test)  are the critical path.

For `-c fastbuild` it is the same time and same critical path.

For `test -c opt`, indeed MLIR will take a longer time to compile, especially for really big file like `LLVMDialect.cpp` which takes around 200s.

For testing files, box_blur now complete quickly, but
`naive_matmul_test_lib.inc.cc` will take above 200s to compile as its loop is unrolled (3.5k loc).

```bash
# heir-opt built
$ bazel test -c opt --cache_test_results=no //tests/...
INFO: Elapsed time: 258.526s, Critical Path: 256.51s
```

j2kun:
@ZenithalHourlyRate I imagine you already know this, but for most of the time during development, you can limit the scope of the targets you test using, e.g., `bazel test //tests/path/to/subdir/...:all` and avoid running the long end to end tests frequently, as well as rebuilding stuff not used by the targets you're testing. So that may also help.

We could also probably reduce the dimension of box blur to 32, since the HECO port is long since done.

asraa:
> For testing files, box_blur now complete quickly, but
naive_matmul_test_lib.inc.cc will take above 200s to compile as its loop is unrolled (3.5k loc).

We also will have for loop support pretty soon so we can reduce the # of lines of code here too. I'll test it out and update it in https://github.com/google/heir/pull/1469

asraa (approved):

Issue #1412, opened 2025-02-14
python frontend: support nested loops


asraa:
https://github.com/google/heir/pull/1307 adds support for single loops with at most one iter_arg, so add support for nested loops (propogate the inner body's iter_arg through the loop nest) and generalize to an arbitrary number of iter_args.

asraa:
Blocked on https://github.com/llvm/llvm-project/issues/129901

PR #1502, opened 2025-03-01
frontend: add support for multiple iter_args and complex control flow

asraa:
* adds loops with > 1 init_arg
* adds support for loops with complex control flow (including nested loops - which are just now blocked on an upstream MLIR verification issue #1412)
* adds a few more binary ops
* fixes a bug in type inference when return type is specified

### DIFF
--- a/frontend/heir/mlir/types.py
+++ b/frontend/heir/mlir/types.py
@@ -121,7 +121,7 @@ def to_numba_str(type) -> str:
   if issubclass(type, MLIRTypeAnnotation):
     return type.numba_str()
 
-  raise TypeError(f"Unsupported type annotation: {type}, {type.__origin__}")
+  raise TypeError(f"Unsupported type annotation: {type}, {get_origin(type)}")
 
 
 def parse_annotations(annotations):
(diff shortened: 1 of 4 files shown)


frontend/heir/mlir/types.py
```diff
@@ -121,15 +121,17 @@ def to_numba_str(type) -> str:
...
 
 def parse_annotations(annotations):
   if not annotations:
     raise TypeError("Function is missing type annotations.")
   signature = ""
   secret_args = []
-  for idx, (_, arg_type) in enumerate(annotations.items()):
+  for idx, (name, arg_type) in enumerate(annotations.items()):
+    if name == 'return':
+      continue
```
AlexanderViand-Intel, line 134: Wouldn't this mean one could specify an "incorrect" return type (i.e., something other than what the type inference determines) and the frontend wouldn't complain? That sounds like it could be pretty misleading/confusing.

How about we save the return type and later check it against the inferred return type?

asraa: saved! i added a debug warning if they were incorrect for now, assuming we can fix the I8/I64 stuff issues later

frontend/heir/mlir/types.py
```diff
@@ -121,19 +121,28 @@ def to_numba_str(type) -> str:
...
 def parse_annotations(annotations):
   if not annotations:
     raise TypeError("Function is missing type annotations.")
   signature = ""
   secret_args = []
-  for idx, (_, arg_type) in enumerate(annotations.items()):
+  rettype = None
+  for idx, (name, arg_type) in enumerate(annotations.items()):
+    if name == "return":
+      # A user should not annotate the return type as secret
```
AlexanderViand-Intel, line 135: Not directly related to this PR, but why should they not annotate the return type as secret? 
I understand why we don't bother with it most of the time, but it might give a nice "sense of security" when the return type is explicitly marked secret. Assuming, of course, we correctly throw an error if the return type gets inferred to something non-secret instead

AlexanderViand-Intel (approved):
LGTM! I rebased my mixed-bitwidth PR on top of this and things seem to work there, too!

frontend/heir/mlir/types.py
asraa: I wasn't really sure! In the MLIR "frontend" users only expect to annotate args as secret, and then we infer secretness from there. Actually, I handle users annotating it as plaintext and secret in this code, so maybe I should not make the comment sound so scary.

asraa:
I'll pull it in and get @j2kun's internal approval for it to be merged!

PR #1503, opened 2025-03-01
Frontend. Fix cryptic TypeError on non-configured repo

mr0re1:
```sh
$ python frontend/example.py

# Before
...
  File "/Users/orlov/wp/heir/frontend/heir/heir_cli/heir_cli_config.py", line 33, in <module>
    heir_opt_path=repo_root / "tools" / "heir-opt",
                  ~~~~~~~~~~^~~~~~~~~
TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'

# After
...
RuntimeError: Could not build development config. Did you run bazel?
```

### DIFF
--- a/frontend/heir/heir_cli/heir_cli_config.py
+++ b/frontend/heir/heir_cli/heir_cli_config.py
@@ -15,24 +15,28 @@ class HEIRConfig:
   heir_translate_path: str | Path
 
 
-def find_above(dirname: str):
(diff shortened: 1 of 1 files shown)


j2kun (approved):
Thanks! On the agenda is to publish this on PyPI, and I feel like the right dev approach at that stage will be to always use bazel.

PR #1505, opened 2025-03-03
Add two different lowerings with Horner and Paterson-Stockmeyer Methods for poly.eval along with lower_eval test

zmbekdemir:
Lowerings are based on a threshold condition. This implementation is only for integer values for now. #1217

### DIFF
--- a/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
+++ b/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
@@ -1219,6 +1219,200 @@ struct ConvertINTT : public OpConversionPattern<INTTOp> {
   }
 };
 
+struct ConvertEvaltoHorner : public OpConversionPattern<EvalOp> {
+  ConvertEvaltoHorner(mlir::MLIRContext *context)
+      : OpConversionPattern<EvalOp>(context) {}
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      EvalOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+    auto attr = dyn_cast<TypedIntPolynomialAttr>(op.getPolynomialAttr());
+    if (!attr) return failure();
+
+    const IntPolynomial &poly = attr.getValue().getPolynomial();
+    auto x = op.getOperand();
+    auto terms = poly.getTerms();
+
+    auto type = dyn_cast<PolynomialType>(attr.getType());
+    if (!type) return failure();
+    IntegerType coeffType =
+        dyn_cast<IntegerType>(type.getRing().getCoefficientType());
+    if (!coeffType) return failure();
+
+    int bitWidth = coeffType.getWidth();
+
+    if (terms.empty()) {
+      // Handle empty polynomial case
+      auto zeroConst = b.create<arith::ConstantIntOp>(0, bitWidth);
+      rewriter.replaceOp(op, zeroConst);
+      return success();
+    }
+
+    // Get the highest degree
+    int64_t maxDegree = terms.back().getExponent().getSExtValue();
+    const int DEGREE_THRESHOLD = 5;
+    if (maxDegree >= DEGREE_THRESHOLD) return failure();
+
+    // Create a map from exponent to coefficient for easy lookup
+    std::unordered_map<int64_t, APInt> coeffMap;
+    for (auto term : terms) {
+      coeffMap[term.getExponent().getSExtValue()] = term.getCoefficient();
+    }
+
+    // Start with the coefficient of the highest degree term
+    Value result = b.create<arith::ConstantIntOp>(
+                        coeffMap[maxDegree].getSExtValue(), bitWidth)
+                       .getResult();
+
+    // Apply Horner's method, accounting for possible missing terms
+    for (int64_t i = maxDegree - 1; i >= 0; i--) {
+      // Multiply by x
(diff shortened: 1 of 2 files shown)


lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -1268,6 +1268,190 @@ struct ConvertINTT : public OpConversionPattern<INTTOp> {
...
+            chunkValue = b.create<arith::ConstantOp>(
+                b.getI64Type(), b.getIntegerAttr(b.getI64Type(), coeffMap[j]));
+            hasCoeff = true;
+          } else {
+            // Apply Horner's step: multiply by x and add next coefficient
+            chunkValue = b.create<arith::MulIOp>(chunkValue, x);
+            auto coeff = b.create<arith::ConstantOp>(
+                b.getI64Type(), b.getIntegerAttr(b.getI64Type(), coeffMap[j]));
+            chunkValue = b.create<arith::AddIOp>(chunkValue, coeff);
+          }
```
j2kun, line 1411: Why not use the precomputed xPowers values here? In the case of FHE, line 1407 would correspond to a ciphertext-ciphertext multiplication, while using the precomputed xPowers you would get a much cheaper plaintext-ciphertext multiplication. Plus you have already paid for the ciphertext-ciphertext multiplications during the precompute step.

(Of course this is implemented in the context of modular arithmetic and integer polynomials, but if we intend to generalize it to FHE ciphertexts, that would become a problem)

Maybe you could link in a code comment at the top to your source for the algorithm. The first place I thought to look was Wikipedia: https://en.wikipedia.org/wiki/Polynomial_evaluation#Matrix_polynomials, but its description differs from what is here. In the Wikipedia description, each chunk uses the precomputed powers (with x^{tk} factored out for each t)

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/lower_eval.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: heir-opt --polynomial-to-mod-arith %s | FileCheck %s
+
+!eval_poly_ty = !polynomial.polynomial<ring=<coefficientType=i64>>
+#eval_poly = #polynomial.typed_int_polynomial<4+x+x**2> : !eval_poly_ty
+
+func.func @test_eval() -> i64 {
+    %c6 = arith.constant 6 : i64
+    %0 = polynomial.eval #eval_poly, %c6 : i64
```
j2kun, line 8: Please add `// CHECK` directives here for the expected output ops. See https://github.com/google/heir/blob/c90c3e26fcb909a0620261157f21bb4f34f7741e/tests/Dialect/Polynomial/IR/canonicalization.mlir as a good example of the FileCheck syntax using regular expressions to capture SSA variable names.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -1268,6 +1268,190 @@ struct ConvertINTT : public OpConversionPattern<INTTOp> {
...
+          chunkValue = b.create<arith::MulIOp>(chunkValue, x);
+        }
+      }
+
+      if (hasCoeff) {
+        chunkValues[i] = chunkValue;
+      }
+    }
+
+    // Combine chunks using Horner's method with x^k
```
j2kun, line 1423: This is a nice idea! Vs precomputing powers x^{tk} it saves k=sqrt(n) plaintext-ciphertext ops.

j2kun (changes requested):
This is looking very good! With a few small improvements and added tests I think we can get this checked in.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
zmbekdemir: Okay, I realized it is not necessary. I think I kinda forgot we already computed the powers of x. It must be fixed now.

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/lower_eval.mlir
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// CHECK-LABEL : @test_eval_for_paterson
+func.func @test_eval_for_paterson() -> i64 {
+    // CHECK-NOT: polynomial.eval
+    %c7 = arith.constant 6 : i64
+    %0 = polynomial.eval #eval_poly_for_paterson, %c7 : i64
+    // CHECK: %[[C6:.*]] = arith.constant 6 : i64
+    // CHECK: %[[C1:.*]] = arith.constant 1 : i64
+    // CHECK: arith.muli
```
j2kun, line 32: I think this PR is basically ready to be merged, but these tests should be a bit more precise. These checks should assert what is being multiplied/added at each step. For example, I think these first two muls are the precomputed powers, and so it should be something like

```
// CHECK: %[[xpow2:.*]] = arith.muli [[C6]], [[C6]]
// CHECK: %[[xpow3:.*]] = arith.muli [[xpow2]], [[C6]]
```

And similar for the rest of the assertions.

j2kun (changes requested):

tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/lower_eval.mlir
zmbekdemir: @j2kun I've pushed the new test file, can u check?

j2kun (approved):
LGTM!

PR #1506, opened 2025-03-03
Add canonicalization patterns and constant foldings for ModArith binary ops

Maokami:
This implements parts of #1216, focusing on binary arithmetic operations.
- Added constant folding
  - Assumes that all inputs and outputs of operations follow the canonical representation.
- Added canonicalization patterns
  - Most of these patterns are ported from [arith canonicalization patterns](https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Arith/canonicalize.mlir).
- Added corresponding tests
  - TODO? : I’m exploring the possibility of using the lean-mlir project for translation validation of these manually written tests. (related to #817)

### DIFF
--- a/lib/Dialect/ModArith/IR/ModArithCanonicalization.td
+++ b/lib/Dialect/ModArith/IR/ModArithCanonicalization.td
@@ -1,0 +1,280 @@
+#ifndef LIB_DIALECT_MODARITH_IR_MODARITHCANONICALIZATION_TD_
+#define LIB_DIALECT_MODARITH_IR_MODARITHCANONICALIZATION_TD_
+
+include "lib/Dialect/ModArith/IR/ModArithOps.td"
+include "mlir/IR/OpBase.td"
+include "mlir/IR/PatternBase.td"
+
+def Equal : Constraint<CPred<"$0 == $1">>;
+
+def IsZero : Constraint<CPred<"dyn_cast<IntegerAttr>($0).getValue().isZero()">>;
+def IsOne : Constraint<CPred<"dyn_cast<IntegerAttr>($0).getValue() == APInt(dyn_cast<IntegerAttr>($0).getValue().getBitWidth(), 1, true)">>;
+
+def IsMinusOne : Constraint<CPred<"dyn_cast<IntegerAttr>($1).getValue().zextOrTrunc(dyn_cast<ModArithType>($0.getType()).getModulus().getValue().getBitWidth()) == dyn_cast<ModArithType>($0.getType()).getModulus().getValue() - APInt(dyn_cast<ModArithType>($0.getType()).getModulus().getValue().getBitWidth(), 1)">>;
+
+def CreateZeroConstant : NativeCodeCall<
+   "$_builder.create<ConstantOp>($0.getLoc(), dyn_cast<ModArithType>($0.getType()), IntegerAttr::get(dyn_cast<ModArithType>($0.getType()).getModulus().getType(), 0));">
+  ;
+
+// add x, 0 -> x
+def AddZero : Pat<
+  (ModArith_AddOp $lhs, (ModArith_ConstantOp $cst)),
+  (replaceWithValue $lhs),
+  [
+    (IsZero $cst),
+  ]
+>;
+
+// sub x, 0 -> x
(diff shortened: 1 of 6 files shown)


lib/Dialect/ModArith/IR/ModArithDialect.cpp
```diff
@@ -158,6 +163,154 @@ void ConstantOp::print(OpAsmPrinter &p) {
...
 
+OpFoldResult ConstantOp::fold(FoldAdaptor adaptor) {
+  // constant(c0) -> c0 mod q
+  APInt cst = adaptor.getValue();
+
+  auto modType = dyn_cast_if_present<ModArithType>(getType());
+  if (!modType) return {};
+
+  // Retrieve the modulus value
+  int64_t modulus = modType.getModulus().getValue().getSExtValue();
```
j2kun, line 174: If the input is larger than fits in an int64_t, this will likely raise an error.

Is there anything stopping you from using `APInt` methods directly (E.g., `APInt::urem` and `APInt::operator+`)? Then this will be more generally useful.

I think the same is true for the binary op folders below.

lib/Dialect/ModArith/IR/ModArithCanonicalization.td
```diff
@@ -0,0 +1,248 @@
...
+    (ModArith_ConstantOp:$c1 $b)),
+  (ModArith_SubOp
+    (ModArith_AddOp
+      $c0,
+      $c1),
+    $x),
+  []
+>;
+
+// add(x, mul(y, (modulus - 1))) -> sub(x, y)
```
j2kun, line 100: Would it be worth adding `sub`-rooted versions of this and the next pattern?

j2kun (changes requested):
Looks great! The APInt suggestion I think we should do, and the other is optional.

lib/Dialect/ModArith/IR/ModArithDialect.cpp
Maokami: That’s a good point!

lib/Dialect/ModArith/IR/ModArithCanonicalization.td
Maokami: I think so!

lib/Dialect/ModArith/IR/ModArithDialect.cpp
```diff
@@ -158,6 +163,168 @@ void ConstantOp::print(OpAsmPrinter &p) {
...
+  if (!modType) return {};
+
+  // Retrieve the modulus value and its bit width
+  APInt modulus = modType.getModulus().getValue();
+  unsigned modBitWidth = modulus.getBitWidth();
+
+  // Adjust cst's bit width to match modulus if necessary
+  if (cst.getBitWidth() != modBitWidth) {
+    cst = cst.zext(modBitWidth);
+  }
```
j2kun, line 180:
```suggestion
  cst = cst.zextOrTrunc(modBitWidth);
```

I think this is the right method

j2kun: Similar elsewhere

j2kun:
Looks like a little build error and then let's get it in

lib/Dialect/ModArith/IR/ModArithDialect.cpp
Maokami: lol

Maokami:
Yes, it seems like there have been some modifications to `mod_arith::ConstantOp` in the latest commits.
I'll take care of it soon

Maokami:
Hello @AlexanderViand-Intel  ! Could you review this PR?
There were recent changes related to `mod_arith::ConstantOP`,
and I thought it would be good if you could check whether this PR properly reflects those updates.
Thanks!

j2kun (approved):

j2kun:
> Hello @AlexanderViand-Intel ! Could you review this PR? There were recent changes related to `mod_arith::ConstantOP`, and I thought it would be good if you could check whether this PR properly reflects those updates. Thanks!

If the tests pass, then I'd say it properly reflects those updates :)

PR #1492, opened 2025-02-28
arith-to-mod-arith: support custom modulus

ZenithalHourlyRate:
Splitted from #1470.

Should be rebased after #1484 is in.

PR #1513, opened 2025-03-04
Plaintext backend: support mod arith semantic

### LINKED ISSUE #1509: tests: should find clang from env
After #1466 is in, clang now becomes a dependency when running `bazel test //tests/Examples/cleartext/...`.

We should make lit access `clang` from environment.

ZenithalHourlyRate:
After #1492 is in, we can now support arbitrary plaintext modulus execution in the plaintext backend, so we can know what happens when we using `i16` and `Zt`.

The notable example these two RUNs in `roberts_cross.mlir`, where bigger plaintext modulus will pass and smaller one will fail, indicating to the user/compiler that if the user/compiler chooses small plaintext modulus the failure is not from ciphertext side but from the Zt arithmetic.

Also this fixes #1509 by using `cc` from environment.


ZenithalHourlyRate:
The lesson I learnt from playing with the plaintext modulus is that, even though the user input may be i16 and the output does not exceed that, after early pipeline like vectorizer/layout (which often assumes things happen in Z instead of Zt or Z_i16) the user is totally unaware of what the circuit becomes, so the user might need the help of plaintext backend to select an appropriate plaintext modulus with typical inputs. We might write a guide on selecting plaintext modulus. Note that other libraries take the same approach by either suggesting a large ptm or let the user try each one.

In such situation compiler also could not help. As one may expect, integer range analysis just gives worst-case estimate and it could be too loose. What the range analysis can do is some partial effort. The same thing should hold for the float point pipeline where range analysis is also wanted.

tests/Examples/plaintext/dot_product/dot_product_8.mlir
```diff
@@ -0,0 +1,34 @@
+// Secret arithmetic plaintext backend with original integer type semantic
+//
+// RUN: heir-opt %s --mlir-to-plaintext-backend \
+// RUN:   | mlir-translate --mlir-to-llvmir | llc -filetype=obj > %t
+// RUN: cc %S/libdot_product_8_test.a %t -o a.out
```
j2kun, line 5: I see what you mean now with `cc` just being used for linking. This is fine. I think if it's easy to change this to `$CC` with an extra line in lit.cfg.py, something like `os.environ["CC"] = os.environ.get("CC", "/usr/bin/cc")`, that would be slightly preferable, but I am also happy to merge this as-is.

tests/Examples/plaintext/dot_product/dot_product_8_mod_test.c
```diff
@@ -0,0 +1,30 @@
...
+    int64_t *allocated, int64_t *aligned, int64_t offset, int64_t size,
+    int64_t stride,
+    /* arg 1*/
+    int64_t *allocated_2, int64_t *aligned_2, int64_t offset_2, int64_t size_2,
+    int64_t stride_2);
+
+int main() {
+  int64_t arg0[8] = {1, 2, 3, 4, 5, 6, 7, 8};
+  int64_t arg1[8] = {2, 3, 4, 5, 6, 7, 8, 9};
+  int64_t expected = 240;
```
j2kun, line 17: I think it would be worth choosing a plaintext modulus that is small enough in one of the tests so that the expected value is different from the test in which no modulus is set. How about a modulus of `17`, so that the expected value is `2`?

Another thing to make this easier is to have this `main` print out `res`, and then have the test assertion happen in FileCheck using `--check-prefix=CHECK_MY_UNIQUE_PREFIX` as you do in other lit tests in this file.

[edit]: I noticed you did this for roberts cross, but having it for a smaller test might be better so regressions could be caught by smaller, faster running tests.

j2kun (approved):
LGTM!

j2kun:
I'll wait until the morning to mark this as pull_ready so you have a chance to respond to the optional comments.

tests/Examples/plaintext/dot_product/dot_product_8.mlir
ZenithalHourlyRate: Seems that natively lit does not support bash-style command line in RUN

```
# .---command stderr------------
# | '$CC': command not found
# `-----------------------------
```

To fix it we have to use something like 
```
sh -c '$CC %S/libdot_product_8_test.a %t -o a.out'
```

I think that would be quite inelegant. Lets make cc here and whenever someone runs into issue we adopt this way (if they run into issue they already need to set $CC anyway so some changes are ok, afterall, they should not pay special care to some test cases that infrequently changes and could be convered in CI)

tests/Examples/plaintext/dot_product/dot_product_8_mod_test.c
ZenithalHourlyRate: Added the modulus=17 test case

Issue #1522, opened 2025-03-04
Noise analysis fails non-graciously with assertion on "unsupported" IR


AlexanderViand-Intel:
I wanted to quickly write a failing test ahead of looking at adding `arith` truncation/extension support and ran into an unexpected assertion crash rather than the "could not lower op ..." Error message I was expecting. 

```llvm
// RUN: heir-opt --mlir-to-bgv %s

func.func @foo(%a : i16 {secret.secret})  -> i32 {
    %aa = arith.extui %a : i16 to i32
    return %aa : i32
}
```
crashes with the following assertion:

```
heir-opt: ./lib/Analysis/NoiseAnalysis/BGV/Noise.h:48: const double &mlir::heir::bgv::NoiseState::getValue() const: 
Assertion `isKnown()' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
```

`--mlir-to-bfv`/`--mlir-to-ckks` fail with the expected `error: failed to legalize operation 'secret.generic' that was explicitly marked illegal` (Not that that's a great error message or anything.... but it beats a crash 😉)

AlexanderViand-Intel:
Interestingly, this fails even *after* I added support for `arith.extui` (working fine for ckks/bfv) which seems to indicate that I'd need to update the noise analysis to know about the new op in some way? @ZenithalHourlyRate

ZenithalHourlyRate:
I think this is the same type of issue as #1397. I think we need to introduce a `--detect-legal-secret-arithmetic-ops-<scheme>` before all the mgmt passes to error out to the user that certain op has not been lowered then it could not be managed

AlexanderViand-Intel:
> I think there might be no FHE op corresponding to arith.extui so updating the noise model can be wierd

Yeah, the whole point is that it's a no-op from the FHE perspective

Issue #1520, opened 2025-03-04
frontend: type inference for return type is "incorrect" when constants are in the program


asraa:
Reproducer:
```
## Simple Example
@compile(debug=True)  # defaults to scheme="bgv", OpenFHE backend, and debug=False
def func(x: Secret[I16], y: Secret[I16]):
  sum = x + y + 255
  diff = x - y
  mul = x * y
  expression = sum * diff + mul
  deadcode = expression * mul
  return expression
```

Numba will automatically make 255 an i64, which means the result type will be inferred to be an i64. Right now this example fails to compile due to a ValueError complaining about the type of the constant differing from the i16s that are otherwise annotated in the program.

The numba TypeInferer seems to have a lot of parameters. I'll investigate whether there's something I can tell to it to make that estimate more inline.

@AlexanderViand-Intel

AlexanderViand-Intel:
I have a fix for the numba side of this already, but it "only" makes the constants "right sized" for their value, the more tedious part is supporting truncation/extension since we'll still get things like i8 + i32 which arith doesn't like. However, we need to support trunc/ext in HEIR anyway for things like arith.select-to-MUX

AlexanderViand-Intel:
> I should have a PR for this tomorrow :)

@asraa  Done, see #1526 and the PRs it depends on

PR #1523, opened 2025-03-05
add support for `arith.extui`/`arith.extsi`

AlexanderViand-Intel:
* Necessary step towards both #1520 and supporting ops such as `arith.select` which require one `i1` and one `i16`/`i32`/etc operand. 

Approach:
* Since the ptxt mod must already be selected large enough for the largest `i??` type appearing in the program to fit, the extension becomes a no-op under FHE.
* Whenever we see an extension operation in a `SecretTo<Scheme>` lowering, we emit a `lwe.reintrepret_application_data` op as it's "FHE op". 
* The OpenFHE emitter already had code to turn this op into a trivial assignment (e.g., `auto ct2 = ct;`), the lattigo backend has it's own non-parametrized ctxt type, so the op becomes a true no-op and is automagically removed.

marked as draft for now because:
* Depends on #1525
*  #1522 needs to be fixed (or at least worked-around for this specific case) before it'll work for `--mlir-to-bgv`.

### DIFF
--- a/lib/Dialect/Secret/Conversions/SecretToBGV/SecretToBGV.cpp
+++ b/lib/Dialect/Secret/Conversions/SecretToBGV/SecretToBGV.cpp
@@ -12,6 +12,7 @@
 #include "lib/Dialect/BGV/IR/BGVOps.h"
 #include "lib/Dialect/LWE/IR/LWEAttributes.h"
 #include "lib/Dialect/LWE/IR/LWEDialect.h"
+#include "lib/Dialect/LWE/IR/LWEOps.h"
 #include "lib/Dialect/LWE/IR/LWETypes.h"
 #include "lib/Dialect/Mgmt/IR/MgmtAttributes.h"
 #include "lib/Dialect/Mgmt/IR/MgmtDialect.h"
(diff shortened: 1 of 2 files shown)


ZenithalHourlyRate:
Maybe we can convert `arith.extsi` to `lwe.reinterpret_underlying_type`?

AlexanderViand-Intel:
> Maybe we can convert `arith.extsi` to `lwe.reinterpret_underlying_type`?

Ohh, that'd certainly look nicer than the builtin casts! Will change tomorrow :) 
Slightly off-topic, but shouldn't that op be renamed now that we went from underlying_type in the old LWE types to message_type?

ZenithalHourlyRate:
> Slightly off-topic, but shouldn't that op be renamed now that we went from underlying_type in the old LWE types to message_type?

I agree.

AlexanderViand-Intel:
> Maybe we can convert `arith.extsi` to `lwe.reinterpret_underlying_type`?

Done! This was doubly nice, not only is it more clear, but the OpenFHE emitter actually already had code to handle this op, so the diff for this PR just got a lot smaller (well, once the rename PR is merged)

asraa (approved):
Approved modulo rebase once the PR is merged (it's in the merge queue)

AlexanderViand-Intel:
> Approved modulo rebase once the PR is merged (it's in the merge queue)

Rebased :)

PR #1524, opened 2025-03-05
NoiseAnalysis: pass through unknown op

### LINKED ISSUE #1522: Noise analysis fails non-graciously with assertion on "unsupported" IR

ZenithalHourlyRate:
Fixes #1522

For unknown op the analysis now short circuit through the op and emit a remark that the op is unknown.


lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
```diff
@@ -106,9 +106,45 @@ LogicalResult CountAnalysis::visitOperation(
...
+              }
+              first = countState;
+              break;
+            }
+
+            if (!mlir::isa<arith::ExtSIOp, arith::ExtUIOp, arith::ExtFOp>(op)) {
+              op.emitRemark()
+                  << "Unsupported operation for count analysis encountered: "
+                  << op;
+            }
```
AlexanderViand-Intel, line 142: Two sugestions here:

1. I think this should still be a hard error, rather than just a remark, since I'd rather have HEIR fail to lower some code than lower something to incorrect code.

2. Can you re-order this, so that it's a bit easier to follow? I.e., f`if( isa<no-op op like ext>) { // inherit count.... ,  } else { // throw error}`?

lib/Analysis/NoiseAnalysis/BGV/NoiseCoeffModelAnalysis.cpp
AlexanderViand-Intel, line 1: same as above

AlexanderViand-Intel (changes requested):
Thanks for the quick turnaround! Just two small requests:

AlexanderViand-Intel (approved):
Thanks!

j2kun (approved):

PR #1525, opened 2025-03-05
rename `lwe.reinterpret_underlying_type` to `lwe.reinterpret_application_data`

AlexanderViand-Intel:
Inspired by a discussion in #1523: with the new LWE types, the ctxt type no longer has an `underlying_type` and instead has `application_data` (I originally assumed `message_type` and suggested that rename, but that's actually "inside" application data). This simply renames the op to match.


ZenithalHourlyRate (approved):
LGTM!

j2kun (approved):

PR #1526, opened 2025-03-05
Frontend: support mixed bitwidths in `arith` ops via `arith.extui`

### LINKED ISSUE #1520: frontend: type inference for return type is "incorrect" when constants are in the program

AlexanderViand-Intel:
* Fixes #1520 

Marked as draft as it has a few dependencies:
*  #1523 which in turn depends on:
   * #1525
   * #1524
* #1502


AlexanderViand-Intel:
Oh, looks like I might need some insights from @asraa  as to how to get this to work with her loop code, which currently hardcodes integers to `i64` in two places, and in both, the thing that's being emitted isn't even present in the typemap as far as I can tell?

frontend/heir/mlir_emitter.py
```diff
@@ -379,6 +380,9 @@ def emit_assign(self, assign):
         return ""
     raise NotImplementedError(f"Unsupported IR Element: {assign}")
 
+  # TODO (#1162): this function seems to be unused and should probably be removed.
```
asraa, line 383: Oh yeah, it looks like the different ir.Expr types are handled in the switch statement for emit_assign - @j2kun maybe this should be removed?

frontend/heir/mlir_emitter.py
```diff
@@ -423,37 +427,74 @@ def emit_expr(self, expr):
...
+    if lhs_type == rhs_type:
+      return self.get_name(lhs), self.get_name(rhs), ""
+
+    # types aren't integer types
+    if not isinstance(lhs_type, types.Integer) or not isinstance(
+        rhs_type, types.Integer
+    ):
+      raise NotImplementedError(
+          "Extension handling for non-integer (e.g., tensor) types not yet"
+          " supported."
```
asraa, line 444: floating points too right? although the support for that would be pretty easy to add later (just extf instead)

asraa (approved):
Thank you!!

frontend/heir/mlir_emitter.py
AlexanderViand-Intel: kind-of: Numba also turns all python floats into doubles/`F64`, but it seems a lot less obvious to me what the "inherent" type of a literal should be for floats than for integers. For example, what precision/bit width should we assign `0.1`?  I think for floats, a solution would also have to add support for in-function type hints (using `heir.mlir` type hints) so that we can get an "intended" type for literals (or, more specifically, the variables they're assigned to)

frontend/mixed_bitwidth_test.py
```diff
@@ -0,0 +1,20 @@
+from heir import compile
+from heir.mlir import I32, Secret
+
+from absl.testing import absltest  # fmt: skip
+
+
+class EndToEndTest(absltest.TestCase):
+
+  def test_loop(self):
```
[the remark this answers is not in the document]
AlexanderViand-Intel: Oops 🙈 Fixed!

frontend/heir/mlir_emitter.py
AlexanderViand-Intel: I've made the error message a bit more explicit and added todos for the float case and tensor case

j2kun:
Asra's out today so I'll pick up the remaining review

j2kun (approved):

PR #1529, opened 2025-03-06
Migrate to new AsmPrinter alias behavior

ZenithalHourlyRate:
In the past when an alias is generated with a trailing digit, it is sanitized with a trailing underscore to avoid potential conflict with other auto-generated ids.

After https://github.com/llvm/llvm-project/pull/127993, such behavior is no longer present as AsmPrinter will record what has been generated and decide which one should append `_N` suffix.

Hope Integrate LLVM people find this PR as they may encounter test failure when bumping llvm commit.


j2kun:
The llvm integrate change has been a headache for the last week, but they're ready to merge and because the asm printer confused the on call, I'm going to mark this as pull_ready so it can be merged once they merge the LLVM integrate.

PR #1530, opened 2025-03-06
CountAnalysis: do not emit error on arith.constant

ZenithalHourlyRate:
After #1524 when running test I would observe

```
<unknown>:0: error: Unsupported operation for count analysis encountered: %5 = "arith.constant"() <{value = 31 : index}> : () -> index
<unknown>:0: note: see current operation: %5 = "arith.constant"() <{value = 31 : index}> : () -> index
```

This fixes such emitting.

### DIFF
--- a/lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
+++ b/lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
@@ -111,10 +111,10 @@ LogicalResult CountAnalysis::visitOperation(
       })
       .Default(
           [&](auto &op) {
-            if (!mlir::isa<arith::ExtSIOp, arith::ExtUIOp, arith::ExtFOp>(op)) {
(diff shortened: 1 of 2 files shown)


asraa (approved):

lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
```diff
@@ -111,7 +111,8 @@ LogicalResult CountAnalysis::visitOperation(
       })
       .Default(
           [&](auto &op) {
-            if (!mlir::isa<arith::ExtSIOp, arith::ExtUIOp, arith::ExtFOp>(op)) {
+            if (!mlir::isa<arith::ConstantOp, arith::ExtSIOp, arith::ExtUIOp,
+                           arith::ExtFOp>(op)) {
               op.emitError()
```
asraa, line 116: just a random side comment, since you already use `op.emitError()` I don't think you need to print the op again in the error string

ZenithalHourlyRate: Changed that.

PR #1536, opened 2025-03-06
update hedron compile commands

j2kun:


asraa (approved):
Thanks for fixing this! I noticed the errors in the codespaces but didn't know the issue

PR #1562, opened 2025-03-13
NoiseAnalysis: refactor to use options from SchemeParam

ZenithalHourlyRate:
As the number of noise model becomes larger, it might be favorable to reduce possible template parameters like -pk/-sk. Instead these options could be read from SchemeParam.

This PR made the following changes

* Added `usePublicKey` option to `SchemeParam` for BGV/BFV/CKKS
* Added `encryptionTechniqueExtended` option to `SchemeParam` for BGV/BFV
  + In short, messages are first encrypted at Qp then mod reduced to Q, this will have less initial noise
  + This is useful for Lattigo BGV/BFV as it defaults to the extended technique
  + Openfhe also support this option, only for BFV
  + I am still studying the noise model using Lattigo so this mode available would be convenient otherwise the model will be inaccurate
* Added these option in `SchemeParamAttr` accordingly
* Changed noise model to respect these option from `SchemeParam` and remove numerous template instantiation
* Changed `lwe-add-client-interface` to recognize `usePublicKey` from `SchemeParam` instead of from pass options
* Changed Openfhe `ConfigureCryptoContext`/`Emitter` to recognize the encryption technique.

### DIFF
--- a/tests/Examples/lattigo/bgv/dot_product_8_sk/BUILD
+++ b/tests/Examples/lattigo/bgv/dot_product_8_sk/BUILD
@@ -9,7 +9,7 @@ heir_lattigo_lib(
     name = "dot_product_8_sk",
     go_library_name = "dotproduct8sk",
     heir_opt_flags = [
-        "--mlir-to-bgv=ciphertext-degree=8 use-public-key=false",
+        "--mlir-to-bgv=ciphertext-degree=8 use-public-key=false encryption-technique-extended=true",
         "--scheme-to-lattigo",
     ],
     mlir_src = "dot_product_8.mlir",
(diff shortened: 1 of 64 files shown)


tests/Examples/lattigo/bgv/dot_product_8_sk/BUILD
```diff
@@ -9,7 +9,7 @@ heir_lattigo_lib(
     name = "dot_product_8_sk",
     go_library_name = "dotproduct8sk",
     heir_opt_flags = [
-        "--mlir-to-bgv=ciphertext-degree=8 use-public-key=false",
+        "--mlir-to-bgv=ciphertext-degree=8 use-public-key=false encryption-technique-extended=true",
```
j2kun, line 12: Should this be true by default, even for OpenFHE (maybe with an error when it's not properly set for OpenFHE+{BGV, CKKS})? Or maybe have a wrapper pipeline that sets it to true by default for lattigo...

I just worry that lattigo users will not know to set this option if it's not the default.

j2kun (approved):
LGTM! One optional suggestion about lattigo default options.

tests/Examples/lattigo/bgv/dot_product_8_sk/BUILD
ZenithalHourlyRate: I think this should be addressed by #1420, but did not get a immediate plan on how to implement that

PR #1564, opened 2025-03-14
B/FV: use msb encryption type in LWE type

ZenithalHourlyRate:
B/FV should has encryption type specified as `msb` instead of the BGV `lsb` in LWE type. For CKKS it should be `mixed`.


asraa (approved):
thanks!

PR #1570, opened 2025-03-17
B/FV: add variance based model

ZenithalHourlyRate:
This adds a noise model for B/FV based on variance analysis in coefficient embedding from BMCM23 (https://ia.cr/2023/600)

BMCM23 gives a tight bound *within* 1-bit (experimentally verified) when all inputs are independently generated ciphertext, and it won't underestimate as the depth goes on, unlike the BGV model.

This is achieved by considering the dependence factor on `s`. Note that `Var(s^2) != Var(s) * Var(s)`, so higher degree of `s` will introduce more error. For each ciphertext, its error could be written as `sum k_i * s^i`, so the max degree `T` of it will affect how error grows. They give a correction factor for multiplication `f(T + 1)` as in each multiplicaiton its degree is increased by 1.

However, for dependent ciphertext like consequtive multiplication of one ciphertext, _this implementation_ will give underestimation. Although in their paper they give partial result on dependent cases, they say it could not be generalized to all circuit so the problem remains open.

### Experiment setup

We give 5 cases, `mult_indep_{8,16,32}` and `mult_dep_{8,16}` where the independent cases are the multiplication of independently generated `k` ciphertext, and the dependent case is the multiplication of one ciphertext `k` times. We rely on `operation-balancer` to balance the input MLIR into a tree.

As expected, all independent cases survive the real noise, while all depedent cases fail. The log for `mult_indep_32` is digested below

```
Input
  [1]
  Noise: 6.83 Total: 160
  Noise Bound: 7.35 Gap: 0.52
// Depth 1
lattigo.bgv.mul
  [1]
  Noise: 35.78 Total: 160
  Noise Bound: 36.27 Gap: 0.49
// Depth 2
lattigo.bgv.mul
  [1]
  Noise: 65.27 Total: 160
  Noise Bound: 65.48 Gap: 0.21
// Depth 3
lattigo.bgv.mul
  [1]
  Noise: 94.21 Total: 160
  Noise Bound: 94.89 Gap: 0.68
// Depth 4
lattigo.bgv.mul
  [1]
  Noise: 124.23 Total: 160
  Noise Bound: 124.47 Gap: 0.24
// Depth 5
lattigo.bgv.mul
  [1]
  Noise: 153.88 Total: 160
  Noise Bound: 154.17 Gap: 0.29
```

### DIFF
--- a/lib/Analysis/NoiseAnalysis/BFV/NoiseByVarianceCoeffModel.cpp
+++ b/lib/Analysis/NoiseAnalysis/BFV/NoiseByVarianceCoeffModel.cpp
@@ -1,4 +1,4 @@
-#include "lib/Analysis/NoiseAnalysis/BGV/NoiseByVarianceCoeffModel.h"
+#include "lib/Analysis/NoiseAnalysis/BFV/NoiseByVarianceCoeffModel.h"
 
 #include <cassert>
 #include <cmath>
@@ -12,14 +12,15 @@
 
 namespace mlir {
 namespace heir {
-namespace bgv {
+namespace bfv {
 
-// the formulae below are mainly taken from MP24
-// "A Central Limit Framework for Ring-LWE Noise Analysis"
-// https://ia.cr/2019/452
-// and CLP23
-// "Optimisations and tradeoffs for HElib"
-// https://ia.cr/2023/104
+// the formulae below are mainly taken from BMCM23
+// "Improving and Automating BFV Parameters Selection: An Average-Case Approach"
+// https://ia.cr/2023/600
+// with modification that it works on Var(e) instead of invariant noise
+// Reader may also refer to KPZ21
+// "Revisiting Homomorphic Encryption Schemes for Finite Fields"
+// to see how the formulae there are adapted to this model
 
 using Model = NoiseByVarianceCoeffModel;
 
@@ -42,10 +43,11 @@ double Model::toLogBudget(const LocalParamType &param, const StateType &noise) {
 double Model::toLogTotal(const LocalParamType &param) {
   double total = 0;
   auto logqi = param.getSchemeParam()->getLogqi();
-  for (auto i = 0; i <= param.getCurrentLevel(); ++i) {
+  for (auto i = 0; i <= param.getSchemeParam()->getLevel(); ++i) {
     total += logqi[i];
   }
-  return total - 1.0;
+  double logT = log2(param.getSchemeParam()->getPlaintextModulus());
+  return total - logT - 1.0;
 }
 
 std::string Model::toLogBoundString(const LocalParamType &param,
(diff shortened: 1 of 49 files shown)


lib/Analysis/NoiseAnalysis/NoiseWithDegree.h
```diff
@@ -0,0 +1,98 @@
...
+#include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Diagnostics.h"       // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+// This class could be shared among all noise models that tracks the noise by a
+// single value. Noise model could have different interpretation of the value.
+// In BGV/BFV world, most noise model just use a single value, either as bound
+// or as variance.
```
j2kun, line 18:
```suggestion

```

lib/Analysis/NoiseAnalysis/NoiseWithDegree.h
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  static NoiseWithDegree uninitialized() {
+    return NoiseWithDegree(NoiseType::UNINITIALIZED, std::nullopt);
+  }
+  static NoiseWithDegree of(double value, int degree) {
+    return NoiseWithDegree(NoiseType::SET, value, degree);
+  }
+
+  /// Create an integer value range lattice value.
```
j2kun, line 35:
```suggestion
```

lib/Transforms/GenerateParam/GenerateParam.td
```diff
@@ -103,6 +104,11 @@ def GenerateParamBFV : Pass<"generate-param-bfv"> {
     of the coefficients, the first being $2 \sqrt{N}$ and the second
     being $N$.
 
+    The third model is taken from BMCM23. It works by tracking the variance
+    of the coefficient embedding of the ciphertexts. This gives a much tighter
+    noise estimate for independent ciphertext input, but may give underestimation
+    for dependent ciphertext input. See the paper for more details.
```
j2kun, line 110:
```suggestion
    for dependent ciphertext input. See [the paper](https://ia.cr/2023/600) for more details.
```

lib/Analysis/NoiseAnalysis/BFV/NoiseByVarianceCoeffModel.h
```diff
@@ -0,0 +1,76 @@
...
+  // the error 'e', assuming coefficients are IID.
+  // Note that BMCM23 tracks the invariant noise t/q(c0 + c1 * s).
+  // We adapt it to track the 'e' in [(q/t)m + e]_q for convenience
+  // and consistency with the other models.
+  //
+  // The main contribution of BMCM23 is that it considers the dependence between
+  // random variables caused by 's', so the max degree T in 's^T' of the
+  // invariant noise needs to be tracked.
+  //
+  // MP24/CCH+23 states that for two polynomial multipication, the variance of
```
j2kun, line 32: Link to these two papers. I see them referred to in the BGV models as well. Maybe we should have a README under NoiseAnalysis that gives citations for all the papers and then we can refer to the acronyms in code comments (with the understanding that `grep MP24` will find the README)

lib/Analysis/NoiseAnalysis/BFV/NoiseByVarianceCoeffModel.cpp
```diff
@@ -0,0 +1,301 @@
...
+// https://ia.cr/2023/600
+// with modification that it works on Var(e) instead of invariant noise
+// Reader may also refer to KPZ21
+// "Revisiting Homomorphic Encryption Schemes for Finite Fields"
+// to see how the formulae there are adapted to this model
+
+using Model = NoiseByVarianceCoeffModel;
+
+// https://stackoverflow.com/questions/27229371/inverse-error-function-in-c
+static double erfinv(double a) {
```
j2kun, line 26: Share this impl with BGV. Maybe `lib/Utils/MathUtils.h`?

j2kun (changes requested):
LGTM, but maybe @kragall would be open to review the implementation of the noise model itself?

ZenithalHourlyRate:
Comments addressed.

Should be merged after #1586 as the heir-opt flags in lattigo tests should be changed accordingly.

ZenithalHourlyRate:
Rebased after #1586

j2kun (approved):

Issue #920, opened 2024-08-19
Determine what parts of Vos-Contri-Erkin 24 can be ported to HEIR


AlexanderViand-Intel:
Jelle Vos, Mauro Conti, Zekeriya Erkin
_Depth-Aware Arithmetization of Common Primitives in Prime Fields._ 
https://eprint.iacr.org/2024/1200

They consider the "arithmetization" problem (i.e., converting high-level operations to add/mul for arithmetic FHE) from a very general viewpoint. They consider a more complex cost model than usual, with mul depth and total number of muls (with a little twist: they allow for tracking that squaring is usually cheaper than generic mul). However, afaik, they do not consider that multiplications get cheaper the "further down" they are (i.e., multiplying two fresh ctxts is often more expensive than multiplying a depth-l ctxt, after it's shed around l of its RNS limbs due to rescale/modswitch).

They give arithmetizations for logical operations (i.e., emulating Z_2 in Z_t for t > 2), exponentiations, arbitrary multiplication trees, etc and then show how that applies to optimizing things such as polynomial evaluation, equality checks, etc.

As an aisde, they also provide a rebalancing algorithm which they claim improves upon the one in EVA Improved (which @lawrencekhlim 
 implemented in #878 as part of #836):
![image](https://github.com/user-attachments/assets/1811f16d-fb48-4cd4-b823-cfaf2f19f3d5)

In addition to showing a few nice instances like the ones pictured above, they define/refine an optimization problem (something something changing from MILP to MaxSAT - this is more @j2kun's area of expertise so I must admit I skipped most of that) and produce a pareto front of circuits. 
Finally, they briefly talk about how their approach is really only optimal for the operation in isolation, but not if, e.g., a comparison appears halfway "down" an FHE computation, though they provide some heuristics on how to handle this situation.

j2kun:
This paper also seems to have a lot to say about polynomial evaluations of low multiplicative depth. Pinging https://github.com/google/heir/issues/1565 to link the issue here.

j2kun:
I read this paper on the flight to HES, I think the maxsat approach will be too slow to run at compile time, but I did talk a lot with Jelle at FHE.org/RWC and we discussed possibly using the best-discovered arithmetizations for common operations and common moduli and check them in as hard-coded kernels. If we can do that in a compact form (PDLL?) that might make for a nice, manageable set of cheap rules.

Issue #827, opened 2024-07-25
Add `arith.cmpi` lowering from Iliashenko & Zucca to `--secret-to-bgv` 


AlexanderViand-Intel:
Implement techniques from ["Faster homomorphic comparison operations for BGV and BFV"](https://petsymposium.org/popets/2021/popets-2021-0046.php) [^1]

Caveat: this approach is only really practical for small plaintext modulus (e.g.,$p = 257$), but that requires advanced CRT splitting into $R_{p^d}$ for $d > 1$, which is not supported by OpenFHE.

[^1]: Iliashenko, I. and Zucca, V. 2021. Faster homomorphic comparison operations for BGV and BFV. 
Proceedings on Privacy Enhancing Technologies. 2021, 3 (Jul. 2021), 246–264.

PR #1571, opened 2025-03-17
frontend: support conditionals as scf.if statements

asraa:
Adds support for conditional.

Really simple cases will now work, but there's a few follow-ups that I will go:
* You will hit https://github.com/google/heir/issues/1621 if you have a complex branch situation that might involve multiple levels of plaintext
* This will error out if you have an if/else that doesn't modify the same variables (for e.g. an if statement alone with an empty else statement). I will make a follow-up to union over all the vars that the branches modify so that the MLIR blocks yield the same variables.

### DIFF
--- a/frontend/conditional_test.py
+++ b/frontend/conditional_test.py
@@ -1,0 +1,25 @@
+from heir import compile
+from heir.mlir import I1, I64, Secret
+
+from absl.testing import absltest  # fmt: skip
+
+
+class EndToEndTest(absltest.TestCase):
+
+  def test_cond(self):
+
+    @compile(debug=True)
+    def cond(cond1: Secret[I1], a: I64, b: I64):
+      result = 0
+      if cond1:
+        result = a
+      else:
+        result = b
+      return result
+
+    self.assertEqual(2, cond(1, 2, 4))
+    self.assertEqual(4, cond(0, 2, 4))
+
+
+if __name__ == "__main__":
+  absltest.main()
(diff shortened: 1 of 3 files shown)


AlexanderViand-Intel:
> TODO: the rlwe pipeline doesn't allow for arith.cmpi so I need some other way of lowering it

Now that we have polynomial evaluation, a lowering for CMP (see #827) should be feasible, but we won't be able to run such code in OpenFHE, as it's only feasible for small plaintext moduli.

j2kun:
> > TODO: the rlwe pipeline doesn't allow for arith.cmpi so I need some other way of lowering it
> 
> Now that we have polynomial evaluation, a lowering for CMP (see #827) should be doable, but we won't be able to run such code in OpenFHE, as it's only feasible for small plaintext moduli which OpenFHE does not (afaik) do.

Why not use the comparison work from Jelle Vos's paper https://github.com/google/heir/issues/920 ? I finally got a chance to read in on the flight to HES and it gives comparison kernels in terms of polynomial evaluation.

j2kun:
> polynomial evaluation

Also we are still missing

AlexanderViand-Intel:
> Why not use the comparison work from Jelle Vos's paper #920 ? I finally got a chance to read in on the flight to HES and it gives comparison kernels in terms of polynomial evaluation.

Good point! The comparison issue predates Jelle's paper by just a few days but I guess it should be updated

asraa:
OK! I rebased this and fixed a few issues! Please take a look

j2kun (approved):

PR #1574, opened 2025-03-17
add `-select-rewrite` pass that converts `arith.select` to HE-friendly add/mul

### LINKED ISSUE #1572: rlwe: support arith::selectop operation in secret-to-<rlwe scheme> lowerings
Can we implement select as a cmux style operation `(cond) * true_br + (1 - cond) * false_br`?

AlexanderViand-Intel:
Fixes #1572

This is a rebase/revive of https://github.com/google/heir/commit/60ee9c66301a852a97c625fe618da304f33fada which wasn't viable back then because we didn't have `arith.extui` support (#1526) at the time.

### DIFF
--- a/lib/Transforms/SelectRewrite/BUILD
+++ b/lib/Transforms/SelectRewrite/BUILD
@@ -1,0 +1,52 @@
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library")
+
+package(
+    default_applicable_licenses = ["@heir//:license"],
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "SelectRewrite",
+    srcs = ["SelectRewrite.cpp"],
+    hdrs = ["SelectRewrite.h"],
+    deps = [
+        ":pass_inc_gen",
+        "@heir//lib/Utils:ConversionUtils",
+        "@llvm-project//mlir:ArithDialect",
+        "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:Support",
+        "@llvm-project//mlir:TensorDialect",
+        "@llvm-project//mlir:TransformUtils",
+        "@llvm-project//mlir:Transforms",
+    ],
+)
+
+gentbl_cc_library(
+    name = "pass_inc_gen",
+    tbl_outs = [
+        (
+            [
+                "-gen-pass-decls",
+                "-name=SelectRewrite",
+            ],
+            "SelectRewrite.h.inc",
+        ),
+        (
+            ["-gen-pass-doc"],
+            "SelectRewritePasses.md",
+        ),
+        (
+            ["-gen-rewriters"],
+            "SelectRewrite.cpp.inc",
+        ),
+    ],
+    tblgen = "@llvm-project//mlir:mlir-tblgen",
+    td_file = "SelectRewrite.td",
+    deps = [
+        "@heir//lib/Utils/DRR",
+        "@llvm-project//mlir:ArithOpsTdFiles",
+        "@llvm-project//mlir:OpBaseTdFiles",
+        "@llvm-project//mlir:PassBaseTdFiles",
+    ],
+)
--- a/lib/Transforms/SelectRewrite/SelectRewrite.cpp
+++ b/lib/Transforms/SelectRewrite/SelectRewrite.cpp
@@ -1,0 +1,55 @@
+#include "lib/Transforms/SelectRewrite/SelectRewrite.h"
+
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/TypeUtilities.h"          // from @llvm-project
+#include "mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h"  // from @llvm-project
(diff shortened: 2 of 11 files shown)


lib/Transforms/SelectRewrite/BUILD
```diff
@@ -0,0 +1,52 @@
...
+        "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:Support",
+        "@llvm-project//mlir:TensorDialect",
+        "@llvm-project//mlir:TransformUtils",
+        "@llvm-project//mlir:Transforms",
+    ],
+)
+
+gentbl_cc_library(
+    name = "pass_inc_gen",
```
ZenithalHourlyRate, line 26: Maybe we can modify `add_heir_transforms` to support also generating rewriters

lib/Transforms/SelectRewrite/SelectRewrite.cpp
```diff
@@ -0,0 +1,55 @@
...
+  return op;
+}
+
+TypedAttr getMatchingOne(Value op) {
+  auto intType = getElementTypeOrSelf(op);
+  if (auto st = mlir::dyn_cast<ShapedType>(op.getType())) {
+    auto containerType = st.cloneWith(st.getShape(), intType);
+    return DenseElementsAttr::get(containerType, true);
+  }
+  return IntegerAttr::get(intType, true);
```
ZenithalHourlyRate, line 32: Got a little confused by the implicit conversion from true to 1.

lib/Transforms/SelectRewrite/SelectRewrite.h
```diff
@@ -0,0 +1,18 @@
+#ifndef LIB_TRANSFORMS_SelectRewrite_SelectRewrite_H_
+#define LIB_TRANSFORMS_SelectRewrite_SelectRewrite_H_
```
ZenithalHourlyRate, line 2: Upper case here. Also for the td file.

ZenithalHourlyRate (approved):
Just a few nitpicking comments.

lib/Transforms/SelectRewrite/BUILD
AlexanderViand-Intel: 👍 Let's file an issue for that, I don't think I want to tackle this as part of this PR 😅

lib/Transforms/SelectRewrite/SelectRewrite.cpp
AlexanderViand-Intel: The `DenseElementsAttr::get` actually *has* to be `true` (it's a different overload than the `int` version of `get`), as it won't accept `1` for a shaped type with element type `i1` 🤷. I added an assert to check that we only call this helper on `i1` types, as it's not really obvious from the C++ context that that's what's going on here.

The `IntegerAttr`, on the other hand was indeed an implicit conversion and will happily take `1` even for `i1`, so I changed that.

lib/Transforms/SelectRewrite/SelectRewrite.cpp
```diff
@@ -0,0 +1,55 @@
...
+#define GEN_PASS_DEF_SELECTREWRITE
+#include "lib/Transforms/SelectRewrite/SelectRewrite.h.inc"
+
+Value matchShapedType(OpBuilder builder, Value op, Value target) {
+  auto targetShapedType = mlir::dyn_cast<ShapedType>(target.getType());
+  auto opShapedType = mlir::dyn_cast<ShapedType>(op.getType());
+  if (targetShapedType && !opShapedType) {
+    auto newShapedType =
+        targetShapedType.cloneWith(targetShapedType.getShape(), op.getType());
+    return builder.create<tensor::SplatOp>(op.getLoc(), op, newShapedType);
```
asraa, line 21: just a small nit: i think ShapedType can also include vectors and memrefs, so maybe for now restrict ShapedType to TensorType since at this point in the pipeline i think we want to keep tensor semantics.

lib/Transforms/SelectRewrite/SelectRewrite.cpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+  return op;
+}
+
+TypedAttr getMatchingOne(Value op) {
```
asraa, line 26: if this is a floating point program intended for CKKS lowering, then a select might be `(i1, f32, f32) -> f32` and I think your code will handle that since it will get the matching one for the f32 type. can you add a quick test for that?

asraa (approved):
thanks for being so quick!

lib/Transforms/SelectRewrite/SelectRewrite.cpp
AlexanderViand-Intel: Ohh, I think it'll actually fail for that case! Not just because the `new_cond` still keeps the element type of `cond`, but also because you can't `arith.extui` an `i1` into an `f32`! 

Since this *is* an important use case, I'll add some logic to handle this case later today :)

lib/Transforms/SelectRewrite/SelectRewrite.cpp
AlexanderViand-Intel: The helper (and the pattern) should indeed work for any ShapedType, so I figured it's best to make it as generic as possible.

lib/Transforms/SelectRewrite/SelectRewrite.cpp
AlexanderViand-Intel: @asraa done, sorry for the delay on this!

asraa (approved):
Thanks!!

PR #1552, opened 2025-03-12
Halevi-Shoup matvec kernel

j2kun:
Rebased over https://github.com/google/heir/pull/1428

PR #1580, opened 2025-03-19
Add setAttributeAssociatedWith, getUndistinguishedResultAttr

j2kun:
setAttributeAssociatedWith is the converse of findAttributeAssociatedWith. getUndistinguishedResultAttr is a generalized helper to allow an op to specify attributes for multiple results via an ArrayAttr. This also makes it so that one can have both a single-result op with a single attr and a single-result op with an ArrayAttr of size 1, though the former is preferred.

Extracted from https://github.com/google/heir/pull/1552

### DIFF
--- a/lib/Utils/AttributeUtils.cpp
+++ b/lib/Utils/AttributeUtils.cpp
@@ -55,6 +55,21 @@ Attribute findAttributeForBlockArgument(BlockArgument blockArg,
       .Default([](Operation *op) { return nullptr; });
 }
 
+Attribute getUndistinguishedResultAttr(Operation *op, int resultNumber,
+                                       StringRef attrName) {
+  Attribute attr = op->getAttr(attrName);
+  // Some ops store the relevant attribute as an array attr with a name
+  // (e.g., "tensor_ext.layout" being an array of layout attrs for a
+  // multi-result op. In this case, we assert the array entries match
+  // with operation results, and return the relevant entry.
+  if (auto arrayAttr = dyn_cast_or_null<ArrayAttr>(attr)) {
+    assert(arrayAttr.size() == op->getNumResults() &&
+           "Array attr does not match number of results");
+    return arrayAttr[resultNumber];
+  }
+  return attr;
+}
+
 Attribute findAttributeForResult(Value result, StringRef attrName) {
   auto *parentOp = result.getDefiningOp();
   assert(parentOp &&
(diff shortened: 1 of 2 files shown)


lib/Utils/AttributeUtils.cpp
```diff
@@ -97,6 +102,63 @@ FailureOr<Attribute> findAttributeAssociatedWith(Value value,
...
+  int resultNumber = cast<OpResult>(result).getResultNumber();
+
+  llvm::TypeSwitch<Operation *>(parentOp)
+      .Case<OperandAndResultAttrInterface>(
+          [&](auto op) { op.setResultAttr(resultNumber, attrName, attr); })
+      .Default([&](Operation *op) {
+        if (auto arrayAttr = dyn_cast_or_null<ArrayAttr>(attr)) {
+          assert(arrayAttr.size() == op->getNumResults() &&
+                 "Array attr does not match number of results");
+          arrayAttr[resultNumber] = attr;
```
ZenithalHourlyRate, line 145: Why the arrayAttr dyn_casted from attr has to set its i-th content to attr, should not it be `arrayAttr = op->getAttr` then if it is array then set it?

j2kun: I don't know what I was thinking. Updated it to be more precise.

lib/Utils/AttributeUtils.cpp
```diff
@@ -97,6 +102,70 @@ FailureOr<Attribute> findAttributeAssociatedWith(Value value,
...
+
+  llvm::TypeSwitch<Operation *>(parentOp)
+      .Case<OperandAndResultAttrInterface>(
+          [&](auto op) { op.setResultAttr(resultNumber, attrName, attr); })
+      .Default([&](Operation *op) {
+        // If the op has an existing array attr, set the attribute in the
+        // existing array attr's index.
+        auto existingAttr = op->getAttr(attrName);
+        if (auto existingArrayAttr =
+                dyn_cast_or_null<ArrayAttr>(existingAttr)) {
```
ZenithalHourlyRate, line 146: Still ArrayAttr here is quite ambiguous. For example `secret.execution_result` is an ArrayAttr and it does not have the semantic of each element corresponding to each element. If we setAttributeForResult for it twice, the second one will run into assertion.

I would suggest determining the shape (ArrayAttr of ArrayAttr, or ArrayAttr of Attr, or Attr) by the number of result.

Anyway, lets merge it and when we run into issue other PR can change it.

ZenithalHourlyRate (approved):

lib/Utils/AttributeUtils.cpp
j2kun: I don't think every attribute necessarily needs to use these utils, it's mainly meant for a common case.

PR #1582, opened 2025-03-19
Update bazelmod deps

j2kun:
Add gitattributes for bazel-provided merge conflict resolution in the lockfile.

Added a patch to re2 (transitive dep via or-tools) due to https://github.com/google/re2/issues/525

Extracted from https://github.com/google/heir/pull/1552


ZenithalHourlyRate (approved):
Document somewhere about the command devs need to run?

Those `jq` and `git config` stuff.

j2kun:
I think the failure mode is: "bazel-lockfile-merge not found" then grep repo for `bazel-lockfile-merge` and find the comment in `.gitattributes`, or else google it and find the page with instructions.

PR #1584, opened 2025-03-19
Feat: integrate belfort FPGA support

Tuditi:
This PR adds support for the Belfort FPGA's that are integrated with AWS F2 instances.


tests/Examples/tfhe_rust_hl/fpga/README.md
```diff
@@ -0,0 +1,47 @@
...
+set_server_key(fpga_key.clone());
+```
+
+## Running the example
+
+To transpile the .mlir file to rust code, run the following command:
+
+```bash
+cd heir/tests/Examples/tfhe_rust_hl/fpga
+~/heir-translate ./arith.mlir --emit-tfhe-rust-hl > ./src/fn_under_test.rs
```
j2kun, line 40: We could also provide a bazel command here instead of using the nightly:

```
bazel run //tools:heir-translate -- --emit-tfhe-rust-hl $PWD/arith.mlir > $PWD/src/fn_under_test.rs
```

j2kun (approved):

tests/Examples/tfhe_rust_hl/fpga/README.md
Tuditi: Using bazel requires an extra tool as well as a correct clang version. For me downloading the binary was more straightforward. I'm happy to include the bazel command if you want, but following the KISS principle, I would exclude it.

asraa: Maybe we can make the pre-reqs generic and link to https://heir.dev/docs/getting_started/#getting-heir to get `heir-translate` either from nightly / built from source.

Then we can just use `heir-translate` generically below to indicate use.

asraa:
One last thing - could you please squash the commits when you're ready? We require that before merging to make history clean with the automatic migration tooling

tests/Examples/tfhe_rust_hl/fpga/README.md
j2kun: I agree the nightly is simpler setup. This command would be needed to test local changes to HEIR. Though I suppose someone who is already making local changes to HEIR would not have trouble constructing the right bazel command. It's probably fine.

Tuditi:
> One last thing - could you please squash the commits when you're ready? We require that before merging to make history clean with the automatic migration tooling

Sure thing! ☺️

PR #1594, opened 2025-03-20
Enable tensors in Frontend

AlexanderViand-Intel:
From python, we convert
```Python
@compile()
def func(x: Secret[Tensor[2, 3, 4, I16]], y: Tensor[2, 3, 4, I16]):
  ....
 ```
 
 into 
 ```llvm
func.func @func(%x: tensor<?x?x?xi16> {secret.secret, shape.shape=[2,3,4]} loc("example.py":10:0),
                 %y: tensor<?x?x?xi16> {shape.shape=[2,3,4]} loc("example.py":10:0)) -> (tensor<?x?x?xi16>) {
 ```

Saving the shape into a function argument attribute on the func is easier than actually emitting a shaped tensor type, as it avoids the need to either cast the argument back to `tensor<?x?x?xi16>` later or try to do shape inference in Python/Numba.

I've also added a `--shape-inference` pass which will (a) update the actual argument types based on the annotations (b) walk the function body (using `inferOpTypeInterface`'s `inferResultTypes` to update types) , and (c) update the return type(s) of the function based on the inferred type(s) of the return op.

Sorry for the delay on this one, I thought the MLIR side would be a quick one and, with hindsight, it's actually not that bad, but I just didn't get around to finishing it before HACS.

### DIFF
--- a/frontend/heir/mlir/types.py
+++ b/frontend/heir/mlir/types.py
@@ -2,6 +2,8 @@
 
 from abc import ABC, abstractmethod
 from typing import Generic, Self, TypeVar, TypeVarTuple, get_args, get_origin
+from numba.core.types import Type as NumbaType
+from numba.core.types import boolean, int8, int16, int32, int64, float32, float64
 
 T = TypeVar("T")
 Ts = TypeVarTuple("Ts")
@@ -16,7 +18,7 @@ class MLIRTypeAnnotation(ABC):
 
   @staticmethod
   @abstractmethod
-  def numba_str():
+  def numba_type() -> NumbaType:
     raise NotImplementedError(
         "No numba type exists for a generic MLIRTypeAnnotation"
     )
@@ -34,77 +36,70 @@ class MLIRTypeAnnotation(ABC):
 class Secret(Generic[T], MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
+  def numba_type() -> NumbaType:
     raise NotImplementedError("No numba type exists for a generic Secret")
 
 
 class Tensor(Generic[*Ts], MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
+  def numba_type() -> NumbaType:
     raise NotImplementedError("No numba type exists for a generic Tensor")
 
 
 class F32(MLIRTypeAnnotation):
   # TODO (#1162): For CKKS/Float: allow specifying actual intended precision/scale and warn/error if not achievable  @staticmethod
   @staticmethod
-  def numba_str():
-    return "float32"
+  def numba_type() -> NumbaType:
+    return float32
 
 
 class F64(MLIRTypeAnnotation):
   # TODO (#1162): For CKKS/Float: allow specifying actual intended precision/scale and warn/error if not achievable  @staticmethod
   @staticmethod
-  def numba_str():
-    return "float64"
+  def numba_type() -> NumbaType:
+    return float64
 
 
 class I1(MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
-    return "boolean"
-
-
-class I4(MLIRTypeAnnotation):
-
-  @staticmethod
-  def numba_str():
-    return "int4"
+  def numba_type() -> NumbaType:
+    return boolean
 
 
 class I8(MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
-    return "int8"
+  def numba_type() -> NumbaType:
+    return int8
 
 
 class I16(MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
-    return "int16"
+  def numba_type() -> NumbaType:
+    return int16
 
 
 class I32(MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
-    return "int32"
+  def numba_type() -> NumbaType:
+    return int32
 
 
 class I64(MLIRTypeAnnotation):
 
   @staticmethod
-  def numba_str():
-    return "int64"
+  def numba_type() -> NumbaType:
+    return int64
 
 
 # Helper functions
 
 
-def to_numba_str(type) -> str:
+def to_numba_type(type: type) -> NumbaType:
   if get_origin(type) == Secret:
     raise TypeError(
         "Secret type should not appear inside another type annotation."
@@ -112,37 +107,58 @@ def to_numba_str(type) -> str:
 
   if get_origin(type) == Tensor:
     args = get_args(type)
-    inner_type = args[-1]
+    if len(args) != 2:
+      raise TypeError(
+          "Tensor should contain exactly two elements: a shape list and a"
+          f" type, but found {type}"
+      )
+    shape = args[0]
+    inner_type = args[1]
     if get_origin(inner_type) == Tensor:
       raise TypeError("Nested Tensors are not yet supported.")
+    # This is slightly cursed, as numba constructs array types via slice syntax
     # Cf. https://numba.pydata.org/numba-doc/dev/reference/types.html#arrays
-    return f"{to_numba_str(inner_type)}[{','.join([':'] * (len(args) - 1))}]"
+    ty = to_numba_type(inner_type)[(slice(None),) * len(shape)]
+    # We augment the type object with `shape` for the actual sizes
+    ty.shape = shape  # type: ignore
+    return ty
 
   if issubclass(type, MLIRTypeAnnotation):
-    return type.numba_str()
+    return type.numba_type()
 
   raise TypeError(f"Unsupported type annotation: {type}, {get_origin(type)}")
 
 
-def parse_annotations(annotations):
+def parse_annotations(
+    annotations,
+) -> tuple[list[NumbaType], list[int], NumbaType | None]:
+  """Converts a python type annotation to a list of numba types.
+  Args:
+    annotations: A dictionary of type annotations, e.g. func.__annotations__
+  Returns:
+    A tuple of (args, secret_args, rettype) where:
+    - args: a list of numba types for the function arguments
+    - secret_args: a list of indices of secret arguments
+    - rettype: the numba type of the return value
+  """
   if not annotations:
     raise TypeError("Function is missing type annotations.")
-  signature = ""
-  secret_args = []
+  args: list[NumbaType] = []
+  secret_args: list[int] = []
   rettype = None
   for idx, (name, arg_type) in enumerate(annotations.items()):
     if name == "return":
       # A user may not annotate the return type as secret
-      rettype = to_numba_str(arg_type)
+      rettype = to_numba_type(arg_type)
       continue
     if get_origin(arg_type) == Secret:
       assert len(get_args(arg_type)) == 1
-      numba_arg = to_numba_str(get_args(arg_type)[0])
+      numba_arg = to_numba_type(get_args(arg_type)[0])
       if name == "return":
         rettype = numba_arg
         continue
       secret_args.append(idx)
-      signature += f"{numba_arg},"
+      args.append(numba_arg)
     else:
-      signature += f"{to_numba_str(arg_type)},"
-  return signature, secret_args, rettype
+      args.append(to_numba_type(arg_type))
+  return args, secret_args, rettype
--- a/frontend/tensor_loop_test.py
+++ b/frontend/tensor_loop_test.py
@@ -1,0 +1,31 @@
+from heir import compile
+from heir.mlir import I16, Secret, Tensor
+import numpy as np
+from numpy.testing import assert_array_equal
+
+
+from absl.testing import absltest  # fmt: skip
+class TensorLoopTest(absltest.TestCase):
+
+  def test_tensor_loop(self):
+
+    @compile()  # defaults to scheme="bgv", OpenFHE backend, and debug=False
+    def func(x: Secret[Tensor[[1024], I16]], y: Secret[Tensor[[1024], I16]]):
+      result = x
+      for i in range(10):
+        result = result + y
+      return result * result
+
+    a = np.array([7] * 1024, dtype=np.int16)
+    b = np.array([8] * 1024, dtype=np.int16)
+    expected = (a + 10 * b) ** 2
+
+    # Test cleartext functionality
+    assert_array_equal(expected, func.original(a, b))
+
+    # Test FHE functionality
+    assert_array_equal(expected, func(a, b))
+
+
+if __name__ == "__main__":
+  absltest.main()
--- a/frontend/tensor_test.py
+++ b/frontend/tensor_test.py
@@ -1,0 +1,28 @@
+from heir import compile
+from heir.mlir import I16, Secret, Tensor
+import numpy as np
+from numpy.testing import assert_array_equal
+
+
+from absl.testing import absltest  # fmt: skip
+class TensorTest(absltest.TestCase):
+
+  def test_basic_tensors(self):
+
+    @compile()  # defaults to BGV and OpenFHE
+    def foo(a: Secret[Tensor[[1024], I16]], b: Secret[Tensor[[1024], I16]]):
+      return a * a - b * b
+
+    a = np.array([7] * 1024, dtype=np.int16)
+    b = np.array([8] * 1024, dtype=np.int16)
+    expected = a * a - b * b
+
+    # Test cleartext functionality
+    assert_array_equal(expected, foo.original(a, b))
+
+    # Test FHE functionality
+    assert_array_equal(expected, foo(a, b))
+
+
+if __name__ == "__main__":
+  absltest.main()
--- a/lib/Transforms/ShapeInference/ShapeInference.cpp
+++ b/lib/Transforms/ShapeInference/ShapeInference.cpp
@@ -1,0 +1,425 @@
+#include "lib/Transforms/ShapeInference/ShapeInference.h"
+
+#include <llvm/Support/Debug.h>
+
+#include "llvm/include/llvm/Support/Debug.h"            // from @llvm-project
+#include "llvm/include/llvm/Support/FormatVariadic.h"   // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"          // from @llvm-project
+#include "mlir/include/mlir/IR/Verifier.h"              // from @llvm-project
+#include "mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h"  // from @llvm-project
+
+#define DEBUG_TYPE "shape-inference"
+
+namespace mlir {
+namespace heir {
+
+#define GEN_PASS_DEF_SHAPEINFERENCE
+#include "lib/Transforms/ShapeInference/ShapeInference.h.inc"
+
+namespace {
+FailureOr<SmallVector<int64_t>> getShapeFromArrayAttr(ArrayAttr shapeAttr,
+                                                      ShapedType shapedType,
+                                                      Location loc) {
+  SmallVector<int64_t> shape;
+  for (auto attr : shapeAttr) {
+    if (auto intAttr = dyn_cast<IntegerAttr>(attr))
+      shape.emplace_back(intAttr.getInt());
+    else
+      return emitError(loc,
+                       "Invalid shape attribute (expected ArrayAttr of "
+                       "only IntegerAttr)");
+  }
+
+  if (shape.size() != shapedType.getRank())
+    return emitError(
+        loc, llvm::formatv("Shape annotation has wrong number of elements "
+                           "(expected {0}, got {1}).",
+                           shapedType.getRank(), shape.size()));
+  return shape;
+}
+
+/// Helper function to handle RegionBranchOpInterface operations
+// (e.g., scf.while, scf.if, affine.for)
+void handleInterface(RegionBranchOpInterface regionBranchOp) {
+  LLVM_DEBUG(llvm::dbgs() << "ShapeInference:\t\t\tcurrent op has "
+                             "RegionBranchOpInterface.\n");
+
+  // Get all successors that can be reached from the current op
+  // (starting from current op is indicated by passing
+  // RegionBranchPoint::parent() as the branch point. Successors can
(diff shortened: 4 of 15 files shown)

### CODE REVIEW

File: frontend/heir/mlir/types.py [hunk 1]
```diff
@@ -34,115 +36,128 @@ def __mul__(self, other) -> Self:
...
+      )
     inner_type = args[-1]
     if get_origin(inner_type) == Tensor:
       raise TypeError("Nested Tensors are not yet supported.")
+    # This is slightly cursed, as numba constructs array types via slice syntax
     # Cf. https://numba.pydata.org/numba-doc/dev/reference/types.html#arrays
-    return f"{to_numba_str(inner_type)}[{','.join([':'] * (len(args) - 1))}]"
+    ty = to_numba_type(inner_type)[(slice(None),) * (len(args) - 1)]
+    # We augment the type object with `shape` for the actual sizes
+    ty.shape = args[:-1]  # type: ignore
```
j2kun, line 122: Possible to have the shape here be its own tuple/list?

`Tensor[(2,3,4), I16]` instead of mixing the member type and the shape? I think that would be slightly more Pythonic.

frontend/tensor_test.py
```diff
@@ -0,0 +1,22 @@
...
+class TensorTest(absltest.TestCase):
+
+  def test_basic_tensors(self):
+
+    @compile()  # defaults to BGV and OpenFHE
+    def foo(a: Secret[Tensor[1024, I16]], b: Secret[Tensor[1024, I16]]):
+      return a * a - b * b
+
+    # Test cleartext functionality
+    self.assertEqual(-15, foo.original(7, 8))
```
j2kun, line 15: I get why this can't be a type error, but maybe it would make more sense to use a numpy array here (and what happens if you try to use a numpy array in the FHE version below?)

lib/Transforms/ShapeInference/ShapeInference.td
```diff
@@ -0,0 +1,14 @@
+#ifndef LIB_TRANSFORMS_SHAPEINFERENCE_SHAPEINFERENCE_TD_
+#define LIB_TRANSFORMS_SHAPEINFERENCE_SHAPEINFERENCE_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def ShapeInference : Pass<"shape-inference"> {
+  let summary = "Infer shapes for shaped types";
+
+  let dependentDialects = [
```
j2kun, line 9: Maybe add a note that this is used primarily for the python frontend.

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+      if (shape.size() != type.getRank())
+        return emitError(
+            funcOp.getLoc(),
+            llvm::formatv("Shape annotation has wrong number of elements "
+                          "(expected {0}, got {1}).",
+                          type.getRank(), shape.size()));
+
+      // save the updated argument type
+      type = RankedTensorType::get(shape, type.getElementType());
```
j2kun, line 83: This actually mutates the container being looped over?? :fearful:

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,187 @@
...
+    // update the type of the "real" arguments (block arguments),
+    // and we also remove the attribute argument from the signature
+    // (primarily for better readability of the IR after shape inference).
+    //
+    // We could delay updating the function type until the end,
+    // when we also know the inferred return type(s),
+    // but that would leave the func op in a less consistent state longer
+    auto newInputTypes = llvm::to_vector(funcOp.getFunctionType().getInputs());
+    for (auto [idx, type] : dynTensorArgs) {
+      funcOp.getArgument(idx).setType(type);
```
j2kun, line 104: I'm assuming you didn't access the actual block arguments above to support function declarations with this pass as well (otherwise maybe you would want to use `AttributeUtils::findAttributeAssociatedWith` to simplify the lookups above). But if my assumption is correct, then this will fail if the function is just a declaration.

```suggestion
      if (!funcOp.isDeclaration()) funcOp.getArgument(idx).setType(type);
```

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,187 @@
...
+      funcOp.removeArgAttr(idx, "shape.shape");
+      newInputTypes[idx] = type;
+    }
+    auto newFunctionType = rewriter.getFunctionType(
+        newInputTypes, funcOp.getFunctionType().getResults());
+    funcOp.setFunctionType(newFunctionType);
```
j2kun, line 110: Any reason to do this here instead of merging it with the equivalent call on line 165-166? Since the block argument types are already changed it shouldn't affect the type inference step 3.

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    /////////////////////////////////////////////////////////////////////
+    //// Step 3: Walk the body, inferring and updating new types     ////
+    /////////////////////////////////////////////////////////////////////
+
+    auto walkResult = funcOp.getBody().walk([&](Operation *op) {
```
j2kun, line 116: Also guard by `funcOp.isDeclaration()`

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+struct ShapeInference : impl::ShapeInferenceBase<ShapeInference> {
+  using ShapeInferenceBase::ShapeInferenceBase;
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    RewritePatternSet patterns(context);
+    // FIXME: This probably shouldn't even be a pattern anymore...
+    patterns.add<InferShape>(context);
+    (void)applyPatternsGreedily(getOperation(), std::move(patterns));
```
j2kun, line 182:
```suggestion
    patterns.add<InferShape>(context);
    walkAndApplyPatterns(getOperation(), std::move(patterns));
```

IMO it's fine to be a pattern, but no need for greedy rewrite engine

j2kun (changes requested):
Generally looks good! A few nits.

j2kun:
Testing this out locally, I am getting the weirdest behavior here, which seems like what I was seeing before but couldn't reproduce.

```
> arg_type
frontend.heir.mlir.types.Secret[frontend.heir.mlir.types.Tensor[1024, frontend.heir.mlir.types.I16]]
> get_origin(arg_type)
<class 'frontend.heir.mlir.types.Secret'>
> get_origin(arg_type) == Secret
False
```

j2kun:
I tried with this example and got this error message, which suggests the shape inference isn't quite handling for loops properly

```
from heir import compile
from heir.mlir import I16, Secret, Tensor


@compile()  # defaults to scheme="bgv", OpenFHE backend, and debug=False
def func(x: Secret[Tensor[1024, I16]], y: Secret[Tensor[1024, I16]]):
    result = x
    for i in range(10):
        result = result + y
    return result * result


x = [v for v in range(1024)]
y = [2 * v for v in range(1024)]
func(x, y)
```

Error (loc's stripped):

```
<stdin>:4:12: error: 'affine.for' op 0-th init and 0-th region iter_arg have different type: 'tensor<1024xi16>' != 'tensor<?xi16>'
    %1:1 = affine.for %3 = 0 to %2 iter_args(%4 = %x) -> (tensor<?xi16>) {
           ^
<stdin>:4:12: note: see current operation: 
%0 = "affine.for"(%arg0) <{lowerBoundMap = affine_map<() -> (0)>, operandSegmentSizes = array<i32: 0, 0, 1>, step = 1 : index, upperBoundMap = affine_map<() -> (10)>}> ({
^bb0(%arg2: index, %arg3: tensor<?xi16>):
  %2 = "arith.addi"(%arg3, %arg1) <{overflowFlags = #arith.overflow<none>}> : (tensor<?xi16>, tensor<1024xi16>) -> tensor<?xi16>
  "affine.yield"(%2) : (tensor<?xi16>) -> ()
}) : (tensor<1024xi16>) -> tensor<?xi16>


input was:

func.func @func(%x: tensor<?xi16> {secret.secret, shape.shape=[1024]}, %y: tensor<?xi16> {secret.secret, shape.shape=[1024]}) -> (tensor<?xi16>) {
    %0 = arith.constant 10 : i32 
    %2 = arith.index_cast %0 : i32 to index
    %1:1 = affine.for %3 = 0 to %2 iter_args(%4 = %x) -> (tensor<?xi16>) {
      %5 = arith.addi %4, %y : tensor<?xi16> 
      affine.yield %5 : tensor<?xi16>
    }
    %6 = arith.muli %1#0, %1#0 : tensor<?xi16> 
    func.return %6 : tensor<?xi16> 
}

options were:

--shape-inference --view-op-graph

```

AlexanderViand-Intel:
Good catches on both 😅  I'll have a look tomorrow/over the weekend!

frontend/tensor_test.py
AlexanderViand-Intel: Good point! The FHE version works fine with numpy arrays, so I changed that, too

AlexanderViand-Intel:
> I tried with this example and got this error message, which suggests the shape inference isn't quite handling for loops properly

I think it might generally not account for ops with regions and block arguments. I wonder to what extent I can handle this generically. E.g., a `linalg.generic` would need to be handled, too, I guess, but wouldn't be captured by any of the loop/control-flow traits the pass could match on?

AlexanderViand-Intel:
> We can get what you have merged in, modulo the minor comments above, and support the more complex loop case as a followup. Sound good?

:+1:  I'll fix the other issues and see if there's a way to fail more gracefully than throwing a verifier error

lib/Transforms/ShapeInference/ShapeInference.cpp
AlexanderViand: Actually, I don't think I can support declarations of functions, because how would I determine what the static shape of a return type of, e.g., `tensor<?x?xi32>` should be, so I'll just guard the whole pattern against `isDeclaration()`. I *will* switch to `findAttributeAssociatedWith`, just didn't remember that it existed.

lib/Transforms/ShapeInference/ShapeInference.cpp
AlexanderViand: It does, but not in a way that would invalidate the iterators (afaik). I'm not sure there's a much nicer way to express this except the trivial approach of taking a copy of the container beforehand?

lib/Transforms/ShapeInference/ShapeInference.cpp
AlexanderViand: (I guess we could extend the shape annotation to also work for return arguments (does `func.func` have attributes for its returns?) but I think that's out of scope for now)

lib/Transforms/ShapeInference/ShapeInference.cpp
j2kun: It's not the end of the world, but IMO a better design would be to split this step off into a helper function that returns or populates a new vector containing the materialized types, or else just merge the loop from 31-36 with this loop so that the emplace_back on line 35 can be done here instead (the check on func attrs can be moved to the beginning, and the empty check on line 38 is effectively a duplicate of like 87, paired with the continue on line 34 that would be merged into this loop)

lib/Transforms/ShapeInference/ShapeInference.cpp
j2kun: > That seems to be looking at attributes on the actual block args, though, where the ArgAttrs for a funcOp are just in a special dictionary inside the main op attribute dictionary for the func op itself?

That is handled as a special case for different op types, because block arguments can't actually have attributes: https://github.com/google/heir/blob/1b11fcf16add488f84cd57b7f107e4e2a046a551/lib/Utils/AttributeUtils.cpp#L34-L36

> does func.func have attributes for its returns?

Yes, `getResultAttr` works similarly to `getOperandAttr`. So function declarations should be supportable this way. Let's do it in a followup.

j2kun:
if the loop test is out of scope for this PR, let's remove/disable it and get this one merged.

frontend/tensor_loop_test.py
j2kun, line 1: nit: This one should be in a proper test, since it's invoked in a `_test` target in the BUILD file.

tests/Transforms/shape_inference/shape_inference.mlir
```diff
@@ -0,0 +1,50 @@
...
+        //CHECK-NEXT: affine.yield %[[NEW_SUM]] : tensor<42xi16>
+        affine.yield %new_sum : tensor<?xi16>
+    }
+    //CHECK: return %[[FOR]] : tensor<42xi16>
+    func.return %0 : tensor<?xi16>
+}
+
+// FIXME: Add tests for other control flow ops (e.g., scf.if)
+
+// FIXME: Add tests for linalg.generic and similar region carrying non-control-flow ops
```
j2kun, line 50: Remove or migrate to TODOs

tests/Transforms/shape_inference/shape_inference.mlir
```diff
@@ -0,0 +1,50 @@
+// RUN: heir-opt --shape-inference  %s | FileCheck %s
+
+//CHECK-LABEL: func @no_inference
```
j2kun, line 3: the name says "no inference" but it appears to be doing inference. Maybe remove the attributes or change the name?

tests/Transforms/shape_inference/shape_inference.mlir
```diff
@@ -0,0 +1,50 @@
...
+//CHECK-LABEL: func @one_dimensional
+//CHECK-SAME:(%[[X:.*]]: tensor<42xi16>, %[[Y:.*]]: tensor<42xi16>) -> tensor<42xi16>
+func.func @one_dimensional(%x: tensor<?xi16> {shape.shape=[42]} , %y: tensor<?xi16> {shape.shape=[42]}) -> (tensor<?xi16>) {
+    //CHECK-NEXT: %[[ADD:.*]] = arith.addi %[[X]], %[[Y]] : tensor<42xi16>
+    %0 = arith.addi %x, %y : tensor<?xi16>
+    //CHECK-NEXT: return %[[ADD]] : tensor<42xi16>
+    func.return %0 : tensor<?xi16>
+}
+
+//CHECK-LABEL: func @two_dimensional
```
j2kun, line 21: here and elsewhere, just use `// CHECK:`. The regex in the pre-commit should have caught this, but there's no space before `C` after `//`, so I'm not even sure these assertions are being triggered by `FileCheck` (I thought the space was mandatory, but maybe it's not?)

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        // Shape annotation on non-ShapedType type
+        if (shapeAttr && !shapedType) {
+          emitWarning(
+              funcOp.getLoc(),
+              "Ignoring shape annotation on argument with non-ShapedType type");
+        }
+
+        inferredInputTypes.emplace_back(type);
```
j2kun, line 93: nit

```suggestion
        changed = true;
        continue;
      }
      
      // Dynamic ShapedType but no {shape.shape} annotation
      if (isDynamicShapedType && !shapeAttr) {
        emitWarning(funcOp.getLoc(),
                    "Found argument with dynamic shaped type but no shape "
                    "annotation.");
      }

      // Shape annotation on non-ShapedType type
      if (shapeAttr && !shapedType) {
        emitWarning(
            funcOp.getLoc(),
            "Ignoring shape annotation on argument with non-ShapedType type");
      }

      inferredInputTypes.emplace_back(type);
```

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,300 @@
...
+    /////////////////////////////////////////////////////////////////////
+    //// Step 2: Walk the body, inferring and updating new types     ////
+    /////////////////////////////////////////////////////////////////////
+
+    // Only do this if there's been an annotated type and there's a body
+    if (changed && !funcOp.isDeclaration()) {
+      rewriter.startOpModification(funcOp);
+
+      //// Update Block Argument types based on Annotations
+      for (auto [idx, type] : llvm::enumerate(inferredInputTypes)) {
```
j2kun, line 106: could be simplified with `llvm::zip(inferredInputTypes, funcOp.getArguments())`

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,300 @@
...
+        funcOp.getArgument(idx).setType(type);
+      }
+
+      // WalkOrder::PreOrder is essential so that we can handle mapping
+      // operands to block arguments for region-bearing ops before
+      // the region itself is walked.
+      auto walkResult =
+          funcOp.getBody().walk<WalkOrder::PreOrder>([&](Operation *op) {
+            llvm::dbgs() << "walking op: " << op->getName() << "\n";
```
j2kun, line 115:
```suggestion
            LLVM_DEBUG(llvm::dbgs() << "walking op: " << op->getName() << "\n");
```

Here and elsewhere

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,300 @@
...
+              regionBranchOp.getEntrySuccessorRegions(operandAttributes,
+                                                      successors);
+              for (auto &successor : successors) {
+                llvm::dbgs() << "found successor: \n";
+
+                auto succesorInputs = successor.getSuccessorInputs();
+                for (auto [idx, input] : llvm::enumerate(succesorInputs)) {
+                  llvm::dbgs() << "found successor input " << input
+                               << " with index " << idx << "\n";
+                  input.setType(op->getOperandTypes()[idx]);
```
j2kun, line 133: this I don't know: does RegionBranchOpInterface guarantee these types must agree?

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,300 @@
...
+                    << "updating parent op result types: " << parent->getName()
+                    << "\n";
+                // FIXME: is this always safe to do?
+                for (int i = 0; i < parent->getNumResults(); ++i) {
+                  parent->getResult(i).setType(op->getOperand(i).getType());
+                }
+              }
+            }
+
+            op->dump();
```
j2kun, line 191: nit: remove

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,300 @@
...
+                                                   ->getOperandTypes());
+
+      if (terminatorOperandTypes.size() !=
+          funcOp.getFunctionType().getResults().size())
+        return emitError(funcOp.getLoc(),
+                         "Mismatch between inferred number of returns from "
+                         "function body and function signature.");
+    }
+
+    SmallVector<Type> inferredReturnTypes{};
```
j2kun, line 221:
```suggestion
    SmallVector<Type> inferredReturnTypes;
```

AlexanderViand:
Thanks for the comments! I think some have already beeen resolved as I continued to hack away at this, but I'll make sure to incorporate the other ones when I clean this up.

 I think this is pretty much good to go, except for how I deal with terminators (there's also a n interrface-based way to get the "correct" mapping rather than just hoping the orders match, as my current code does)

tests/Transforms/shape_inference/shape_inference.mlir
AlexanderViand: Done @CHECK-LABEL, the space isn't mandatory as far as I can tell (had plenty of such CHECKs trigger during debugging 😉) and we've already got a bunch of other tests without spaces, so the pre-commit regex might need updating :)

lib/Transforms/ShapeInference/ShapeInference.cpp
AlexanderViand: this should be op.getEntrySuccessorOperands(successor) instead of just op.getOperands/op.getOperandsTypes, then it is indeed guaranteed by the interface verifier.

AlexanderViand:
@j2kun I've cleaned this up, including adding a bunch of comments on what my best current understanding of the RegionBranchOp Interface(s) is and why things should work (though I'd not be shocked if someone finds a way to break some of it almost immediately 😉).

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,406 @@
...
+                funcOp.getLoc(),
+                "Mismatch between inferred return type and annotated "
+                "return type.");
+          }
+          newType = terminatorType;
+        }
+        changed = true;
+        inferredReturnTypes.emplace_back(newType);
+      } else {
+        // FIXME: emit warnings here, too
```
j2kun, line 358:
```suggestion
```

Fix or remove

lib/Transforms/ShapeInference/ShapeInference.cpp
```diff
@@ -0,0 +1,406 @@
...
+                          "return type(s) after its operands' types were "
+                          "updated during shape inference.");
+                return WalkResult::interrupt();
+              }
+            }
+
+            // If an operation is a (RegionBranch) terminator, we need to update
+            // the return types of the results of the parent operation to match
+            // the types of the terminators operands.
+            if (auto terminator =
```
j2kun, line 229: My only concern now is readability: this function is massive and some parts, like the handlers for each interface type, seem like they can be split into independent functions without too much hassle.

j2kun (approved):

j2kun:
Also looks like it's not picking up your CLA?

AlexanderViand:
> Also looks like it's not picking up your CLA?

I think that's because I created the PR itself with my old account which is no longer associated with the email the CLA system knows about. Once I have a minute during the conference, I'll fix the FIXMEs and create a new PR from this account :)

AlexanderViand:
Closing this to re-open from right account:

> > Also looks like it's not picking up your CLA?
> 
> I think that's because I created the PR itself with my old account which is no longer associated with the email the CLA system knows about. Once I have a minute during the conference, I'll fix the FIXMEs and create a new PR from this account :)

Issue #1604, opened 2025-03-22
LWE: distinguish ct-pt mul and ct-scalar mul


ZenithalHourlyRate:
In LWE dialect, we have ct-ct multiplication already, ct-pt multiplication modeled in bgv/ckks dialect. I propose we should also support ct-scalar mul (some literature also call it ct-constant mul)

### All those multiplication

For a **ct** of form (-**as** + **m** + **te**, **a**), where bold letter means it is a ring element (or vector), we have the following possible way to multiply it

* Tensor product: tensor product by another ciphertext **ct**'
* Multiply by a ring element in `Zt[X]/(X^N + 1)`, like **m**': the result becomes (-**asm**' + **mm**' + t**em**', **am**')
* Multiply by a scalar in Zt, like k: the result becomes (-k**as** + k**m** + tk**e**, **a**k)

### Different lowering

For `arith.mul %ct, %c2` where `%c2 = arith.constant 2`, to correctly lowering, we have the following ways

* encrypt 2 to get **ct**' and do ct-ct multiplication
* splat 2 into a vector `[2, 2, ...]` and properly encode (full crt encoding or inverse canonical encoding) it into **m**', then do ct-pt multiplication
* directly do ct-scalar

### Different noise behavior

One benefit of the modelling of ct-scalar mul is that, it has much smaller noise growth. For the multiplying by 2 example, we have the noise term of t**em** and tk**e** respectively. Of course the latter is much friendly.

Note that the overflow of [**mm'**]_t into noise also counts.

### Cases where ct-scalar mul is wanted

In polynomial approximation output, there would be a lot of ct-scalar multiplication, if we instead just use ct-pt version, the noise growth would be quite unexpected.

Specificially, the sine function approximation in CKKS bootstrapping needs ct-scalar multiplication, and if we do not control the noise/precision carefully, the result could be unexpected.

For scale management, the adjust scale operation is also a ct-scalar multiplication.

### Backend supports such difference

For Openfhe, they offer three sets of API

```
EvalMul(ct, ct)
EvalMul(ct, pt)
EvalMul(ct, NativeInteger constant) or double constant for CKKS
```

For Lattigo, it also supports three sets of API, accepting either rlwe ct, vector of input, and a scalar.

AlexanderViand-Intel:
[...]

PR #1626, opened 2025-03-30
Lattigo: use DropLevel API for level reduce

ZenithalHourlyRate:
DropLevel in Lattigo has the implementation:

```GO
// DropLevel reduces the level of op0 by levels and returns the result in op0.
// No rescaling is applied during this procedure.
func (eval Evaluator) DropLevel(op0 *rlwe.Ciphertext, levels int) {
	op0.Resize(op0.Degree(), op0.Level()-levels)
}
```

which is the same as what LattigoEmitter implements.

In the meantime, CKKS has DropLevelNew but BGV does not have one, so the emitter does not use the DropLevelNew API. Will PR Lattigo to try to add this API.

### DIFF
--- a/lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
@@ -306,18 +306,24 @@ struct ConvertRlweRotateOp : public OpConversionPattern<RlweRotateOp> {
   }
 };
 
-template <typename LevelReduceOp, typename LattigoLevelReduceOp>
+template <typename EvaluatorType, typename LevelReduceOp,
+          typename LattigoLevelReduceOp>
 struct ConvertRlweLevelReduceOp : public OpConversionPattern<LevelReduceOp> {
   using OpConversionPattern<LevelReduceOp>::OpConversionPattern;
 
   LogicalResult matchAndRewrite(
       LevelReduceOp op, typename LevelReduceOp::Adaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
+    FailureOr<Value> result =
+        getContextualEvaluator<EvaluatorType>(op.getOperation());
+    if (failed(result)) return result;
+
+    Value evaluator = result.value();
     rewriter.replaceOp(
         op, rewriter.create<LattigoLevelReduceOp>(
                 op.getLoc(),
                 this->typeConverter->convertType(op.getOutput().getType()),
-                adaptor.getInput(), op.getLevelToDrop()));
+                evaluator, adaptor.getInput(), op.getLevelToDrop()));
     return success();
   }
 };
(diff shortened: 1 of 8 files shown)


Pro7ech:
> DropLevel in Lattigo has the implementation:
> 
> ```go
> // DropLevel reduces the level of op0 by levels and returns the result in op0.
> // No rescaling is applied during this procedure.
> func (eval Evaluator) DropLevel(op0 *rlwe.Ciphertext, levels int) {
> 	op0.Resize(op0.Degree(), op0.Level()-levels)
> }
> ```
> 
> which is the same as what LattigoEmitter implements.
> 
> In the meantime, CKKS has DropLevelNew but BGV does not have one, so the emitter does not use the DropLevelNew API. Will PR Lattigo to try to add this API.

@ZenithalHourlyRate I'm curious about which scenario this method would be called. This more an artifact of the early stages of Lattigo than anything else because the evaluator automatically manages the levels of the ciphertexts, so this method never has to be called during circuit evaluation. The one case I can think of is when marshalling the result of a computation when the output ciphertext is not at the minimum level (still the circuit should ensure that), as calling this method can reduce the data sent back to the client. However in this circtumstance it can also be done in place.

ZenithalHourlyRate:
The MLIR SSA form (we are pinning to value-semantic) could not express inplace operation, so a drop level operation in HEIR would be `%new = lattigo.rlwe.drop_level_new %old`, which is for the ease of our dialect lowering.

We have an optimization pass called `alloc-to-inplace` that replaces the allocation version to inplace version (that Lattigo prefers), but still two versions of them are needed in case such replacement does not succeed.

As for why DropLevel is needed by the compiler. The philosophy is that the compiler should manage all the metadata on its own so corresponding instructions should be issued. Specifically, this instruction is used by the scale management pass (#1459 in-progress) where reducing to certain level is needed by _cross-level_ operations.

Pro7ech:
I see. But drop level isn't even used for scale management in Lattigo. Contrary to Open FHE, Lattigo doesn't use a different scale for each level. Instead the scale is maintained to the default scale by tweaking the scaling factor of the inputs to ensure the scaling factor, before a rescaling, is the product between the default scale and the prime to be consumed. Cross level operations are thus just taking the minimum level of the inputs, without having to do intermediate rescaling to march scales. 

Open FHE's approach can be emulated, but additional care must be given, such as manually specifying the encoding factors, else cross level operations will result in precision loss.

ZenithalHourlyRate:
> Open FHE's approach can be emulated, but additional care must be given, such as manually specifying the encoding factors, else cross level operations will result in precision loss.

The intended solution of the in-progress PR for emulating Openfhe's cross-level behavior in Lattigo is indeed manually specifying the encoding factors by multiplying by 1 with scaling factor (unfortunately HEIR does not suport multiply with constant now #1604)

```
pt = bgv.NewPlaintext
pt.Scale = [compiler specified]
encoder.Encode([1, 1, ...], pt)
```

The emulation behavior is one baseline approach for enabling the basic scaling factor management functionality of HEIR (as HEIR does not have fine control over other backends). I do not think we will pin to that solution and further techniques (like the one Lattigo does) could be ported to HEIR.

> Just making sure it is not mistaken for something it is not, the point being that, in Lattigo, wether this method is called or not before cross level operations will not have any effect on the output ciphertext.

In my understanding Lattigo on its own management does not need such API. HEIR should be able to manage the ciphertext such that `matchScaleThenEvaluateInPlace` will not be called, which 
 means Lattigo's management is bypassed.

> Will PR Lattigo to try to add this API.

As for the API itself, after the discussion I think I can close that request as Lattigo itself does not need it, and HEIR can emulate it anyway.

Pro7ech:
Thanks for the clarification :)

j2kun (approved):

PR #1633, opened 2025-03-31
Halevi Shoup e2e test

j2kun:
This is a properly rebased version of https://github.com/google/heir/pull/1600, but I left https://github.com/google/heir/pull/1600 intact in case I messed up the rebase, as it's been a week or so since I have worked on this and I forgot what I was doing.


j2kun:
So far the e2e pipeline is generating good openfhe code, with the last step being to lower the `assign_layout` kernel (a `linalg.generic`) to something `heir-translate` can handle.

j2kun:
OK this last step is going to be more annoying than I thought, because the linalg.generic can't be lowered to loops without bufferizing the op, but you can't bufferize one op you have to bufferize the whole func, and the operands are still in tensor semantics.

Looks like today is the day I port `convert-linalg-to-affine-loops` to HEIR for tensor semantics.

j2kun:
The e2e example seems to be generating the right code, but CKKS decryption fails and I can't yet figure out why. The generated mlir and openfhe: https://gist.github.com/j2kun/7a4f82484388915f5b871e266e5730f2

j2kun:
Ok found the bug, it's in my kernel

j2kun:
Note to self: remaining failing tests

```
//tests/Transforms/convert_to_ciphertext_semantics:dot_product.mlir.test FAILED in 0.7s
//tests/Transforms/mlir_to_openfhe_bgv:simple_sum.mlir.test              FAILED in 1.1s
//tests/Transforms/mlir_to_openfhe_ckks:bootstrap_waterline.mlir.test    FAILED in 0.6s
//tests/Transforms/mlir_to_openfhe_ckks:dot_product_float.mlir.test      FAILED in 0.7s
//tests/Transforms/mlir_to_openfhe_ckks:matmul.mlir.test                 FAILED in 0.7s
//tests/Transforms/mlir_to_openfhe_ckks:naive_matmul.mlir.test           FAILED in 0.7s
//tests/Transforms/mlir_to_openfhe_ckks:simple_sum.mlir.test             FAILED in 1.2s
//tests/Transforms/optimize_relinearization:force_equal_args.mlir.test   FAILED in 0.7s
//tests/Transforms/optimize_relinearization:multiple_return_values.mlir.test FAILED in 0.7s
//tests/Transforms/optimize_relinearization:optimize_relinearization.mlir.test FAILED in 0.8s
//tests/Transforms/secret_insert_mgmt:bootstrap_waterline.mlir.test      FAILED in 0.7s
```

j2kun:
This is getting out of hand. Starting to extract bits from this PR into different sub-PRs

j2kun:
Next step (note to self): migrate lwe-add-client-interface to use the `tensor_ext.original_type` attribute to identify the right encryption/decryption target types.

lib/Dialect/ModuleAttributes.cpp
```diff
@@ -31,6 +34,27 @@ bool moduleIsCGGI(Operation *moduleOp) {
...
+  }
+
+  for (auto schemeAttrName : schemeAttrNames) {
+    if (auto schemeAttr = moduleOp->getAttr(schemeAttrName)) {
+      return schemeAttr;
+    }
+  }
+
+  // A default case to avoid nullptr errors
+  return UnitAttr::get(op->getContext());
```
ZenithalHourlyRate, line 55: I actually always use `nullptr` for empty Attribute and assume attribute could be nullptr everywhere and add checks there. I think this is rather standard? https://github.com/llvm/llvm-project/blob/65cede26a6b06ba02c08284fada06c46c0289704/mlir/include/mlir/IR/Attributes.h#L38-L48

j2kun: I had considered this, and then I was uncertain what would happen if you pass the result to a type switch. I will give it a try though.

j2kun: Using nullptr causes TypeSwitch to stack trace when the module does not have an attribute attached. This is because TypeSwitch internally uses dyn_cast instead of dyn_cast_or_null.

j2kun: I ended up continuing to use UnitAttr, but having the TypeSwitches that use it return a `FailureOr<T>` for more graceful failures.

j2kun:
Some remaining failing tests

```
heir-opt: external/llvm-project/llvm/include/llvm/Support/Casting.h:566: decltype(auto) llvm::cast(const From &) [To = mlir::IntegerAttr, From = mlir::TypedAttr]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
Stack dump:
0.      Program arguments: bazel-out/k8-dbg/bin/tools/heir-opt --mlir-to-plaintext-backend=plaintext-modulus=65537 -o bazel-out/k8-dbg/bin/tests/Examples/plaintext/dot_product/dot_product_8_mod_65537_heir_opt.mlir tests/Examples/common/dot_product_8.mlir

-------------

ERROR: /home/j2kun/fhe/heir/tests/Examples/plaintext/roberts_cross/BUILD:30:15: Action tests/Examples/plaintext/roberts_cross/roberts_cross_heir_opt.mlir failed: (Exit 1): heir-opt failed: error executing Action command (from target //tests/Examples/plaintext/roberts_cross:roberts_cross_heir_opt) bazel-out/k8-dbg/bin/tools/heir-opt --mlir-to-plaintext-backend -o bazel-out/k8-dbg/bin/tests/Examples/plaintext/roberts_cross/roberts_cross_heir_opt.mlir tests/Examples/common/roberts_cross_64x64.mlir

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
tests/Examples/common/roberts_cross_64x64.mlir:28:13: error: 'tensor_ext.rotate' op requires a 1-D input tensor or tensor with single non-unit dimension, but found 'tensor<4x1024xi16>'
      %14 = arith.subi %9, %13 : i16
            ^
tests/Examples/common/roberts_cross_64x64.mlir:28:13: note: see current operation: %4 = "tensor_ext.rotate"(%arg1, %2) : (tensor<4x1024xi16>, index) -> tensor<4x1024xi16>

----------------


FAIL: //tests/Examples/plaintext/dot_product:dot_product_8_cc_test (Segmentation fault) (see /home/j2kun/.cache/bazel/_bazel_j2kun/b07a48cf9da8c73a462d3608e5c508dd/execroot/_main/bazel-out/k8-dbg/testlogs/tests/Examples/plaintext/dot_product/dot_product_8_cc_test/test.log)
INFO: From Testing //tests/Examples/plaintext/dot_product:dot_product_8_cc_test:
==================== Test output for //tests/Examples/plaintext/dot_product:dot_product_8_cc_test:
```

j2kun:
Figured out one problem: the plaintext backend is failing because mod_arith.constant doesn't support fold when the attribute it contains is a dense elements attr.

j2kun:
OK, aside from the two cleanup PRs https://github.com/google/heir/pull/1857 and https://github.com/google/heir/pull/1858, this is ready for review! @asraa I am at your disposal for walking you through the changes. The majority of the complication is in tests.

j2kun:
SIGH ok there's another problem. Now that the lowering includes non-secret ops, the plaintext lowering fails when combining secret-forget-secrets with arith-to-mod-arith (the cleartext packing operations get converted to mod arith ops).

So now instead of forgetting secrets, I need a secret-to-mod-arith pass that only does type conversion on the secret types.

j2kun:
OK so I resolved the issue as per our discussions, and ended up disabling some of the debug tests involving floats. Let's get this in and I can fix that up next.

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
```diff
@@ -94,7 +94,16 @@ LogicalResult NoiseAnalysis<NoiseModel>::visitOperation(
 
   auto res =
       llvm::TypeSwitch<Operation &, LogicalResult>(*op)
-          .Case<secret::GenericOp>([&](auto genericOp) {
+          .template Case<secret::RevealOp, secret::ConcealOp>([&](auto op) {
+            // Reveal outputs are not secret, so no noise. Conceal outputs are
+            // a fresh encryption. Both are handled properly by setToEntryState
```
asraa, line 99: Are conceal outputs a fresh encryption or no noise? If they are concealed server-side, then I would expect these are plaintexts that have no noise, rather than a public key encryption.

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
```diff
@@ -145,28 +154,6 @@ LogicalResult NoiseAnalysis<NoiseModel>::visitOperation(
             propagate(rotateOp.getResult(), rotate);
             return success();
           })
-          // NOTE: special case for ExtractOp... it is a mulconst+rotate
```
asraa, line 148: for later: why get rid of extract?

lib/Dialect/ModuleAttributes.cpp
```diff
@@ -31,6 +34,26 @@ bool moduleIsCGGI(Operation *moduleOp) {
          nullptr;
 }
 
+Attribute getSchemeParamAttr(Operation *op) {
+  SmallVector<StringLiteral> schemeAttrNames = {
+      bgv::BGVDialect::kSchemeParamAttrName,
+      ckks::CKKSDialect::kSchemeParamAttrName,
```
asraa, line 40: Why do we have both `kBGVSchemeAttrName` and  `bgv::BGVDialect::kSchemeParamAttrName`? (They're different, scheme.bgv and bgv.schemeParam). The rest of the code in this file is checking for kBGVSchemeAttrName while this code checks for the dialect attr.

lib/Dialect/Secret/Conversions/SecretToModArith/SecretToModArith.h
```diff
@@ -0,0 +1,16 @@
+#ifndef LIB_DIALECT_SECRET_CONVERSIONS_SECRETTOModArith_SECRETTOModArith_H_
+#define LIB_DIALECT_SECRET_CONVERSIONS_SECRETTOModArith_SECRETTOModArith_H_
```
asraa:
```suggestion
#ifndef LIB_DIALECT_SECRET_CONVERSIONS_SECRETTOMODARITH_SECRETTOMODARITH_H_
#define LIB_DIALECT_SECRET_CONVERSIONS_SECRETTOMODARITH_SECRETTOMODARITH_H_
```

and below

tests/Examples/common/halevi_shoup_matvec.mlir
```diff
@@ -0,0 +1,8 @@
+module {
```
asraa: nit: rename this to matvec, since this isn't low level enough that it's an H-V matvec

tests/Transforms/mlir_to_openfhe_ckks/BUILD
```diff
@@ -7,8 +7,11 @@ glob_lit_tests(
     data = ["@heir//tests:test_utilities"],
     driver = "@heir//tests:run_lit.sh",
     # TODO(#1642): cross-level operation does not work well with bootstrapping
+    # FIXME: todo to re-enable tests
     exclude = [
         "bootstrap_waterline.mlir",
+        "matmul.mlir",
+        "naive_matmul.mlir",
```
asraa, line 14: actually do you want to delete this entirely? this does scalar encoding

lib/Dialect/Secret/Conversions/Patterns.h
```diff
@@ -0,0 +1,59 @@
...
+#include "mlir/include/mlir/IR/BuiltinTypeInterfaces.h"  // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"     // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+// Lower a client encryption function's secret.conceal op to lwe.rlwe_encode +
+// lwe.rlwe_encrypt. Modifies the containing function to add new secret key
```
asraa, line 16: how do we know this isn't a plaintext concealing? if this is a conceal op to reconcile secret/non-secret types in a loop, for e.g. the loop arg would be initialized as a plaintext

lib/Dialect/Secret/Conversions/Patterns.h
```diff
@@ -0,0 +1,59 @@
...
+// material args.
```
asraa, line 17: nit: add some method to specify this as an RLWE only lowering? we'll want to have lwe specific ops for CGGI

lib/Dialect/Secret/Conversions/Patterns.h
asraa: resolution: we should distinguish between the client-side conceal ops (fresh encryption) vs a server side conceal op (message plaintext encoding) either with an attr or distinct operation.

the case where there's a server side conceal op to resolve types in iter args is definitely a 0 noise encryption.

we can also ad a note + a follow-up issue

tests/Dialect/Secret/Conversions/secret_to_mod_arith/client_helpers.mlir
```diff
@@ -0,0 +1,29 @@
...
+    // CHECK: arith.extsi
+    // CHECK: mod_arith.encapsulate
+    %1 = secret.conceal %0 : tensor<1024xi16> -> !secret.secret<tensor<1024xi16>>
+    return %1 : !secret.secret<tensor<1024xi16>>
+  }
+  // CHECK: @dot_product__decrypt__result0
+  func.func @dot_product__decrypt__result0(%arg0: !secret.secret<tensor<1024xi16>>) -> i16 attributes {client.dec_func = {func_name = "dot_product", index = 0 : i64}} {
+    %c0 = arith.constant 0 : index
+    // CHECK: mod_arith.extract
+    // CHECK: arith.trunci
```
asraa, line 25: nit: make some check assertion on the flag (modulus size)

lib/Pipelines/ArithmeticPipelineRegistration.cpp
```diff
@@ -103,40 +126,45 @@ void mlirToSecretArithmeticPipelineBuilder(
...
-  // Note: LinalgToTensorExt requires that linalg.matmuls are the only operation
-  // within a secret.generic. This is to ensure that any tensor type conversions
-  // (padding a rectangular matrix to a square diagonalized matrix) can be
-  // performed without any type mismatches.
-  std::vector<std::string> opsToDistribute = {"linalg.matmul"};
-  auto distributeOpts = secret::SecretDistributeGenericOptions{
-      .opsToDistribute = llvm::to_vector(opsToDistribute)};
-  pm.addPass(createSecretDistributeGeneric(distributeOpts));
-  pm.addPass(createCanonicalizerPass());
-  auto linalgToTensorExtOptions = linalg::LinalgToTensorExtOptions{};
```
asraa, line 124: we will nuke this in a follow-up

lib/Transforms/ValidateNoise/ValidateNoise.cpp
```diff
@@ -92,13 +97,28 @@ struct ValidateNoise : impl::ValidateNoiseBase<ValidateNoise> {
     return success();
   }
 
+  bool shouldSkipValidation(Value value) {
+    // Client helpers have an implicit association between the return value of
+    // the entry function and the argument of the decryption helper. This isn't
+    // represented in the noise analysis directly (though perhaps it could be
+    // via DataFlowAnalysis::addDependency), so the noise analysis treats these
+    // values as fresh encryptions, but at a low level, which will exceed the
+    // noise bound if the main function does enough mod reduce ops.
```
asraa: nit: clarify that the noise is assigned fresh encryption while the noisevalue attr is copied over and there's a mismatch

asraa (commented):
wonderful!

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
j2kun: This now has an explicit kernel in ConvertToCiphertextSemantics, and the noise analysis runs afterward. This replaces the prior use of `bgv.extract`, etc., which modeled "slot" extraction (one-hot pt-ct mul + rotate)

lib/Dialect/ModuleAttributes.cpp
j2kun: `scheme.bgv` is a UnitAttr that signals to any pass that the target scheme is BGV (avoiding the need for pass options threaded through). It is set early in the pipeline. The latter is the attribute that stores the actual scheme parameters when they are known. It is set late in the pipeline during the parameter selection pass.

j2kun: Adding comments to the header clarify

tests/Transforms/mlir_to_openfhe_ckks/BUILD
j2kun: It's because the test case doesn't have packing. I moved the reinstatement to https://github.com/google/heir/pull/1876

tests/Examples/common/halevi_shoup_matvec.mlir
j2kun: Done

lib/Dialect/ModuleAttributes.cpp
j2kun: Done

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
j2kun: Added a todo to #1875

lib/Transforms/ValidateNoise/ValidateNoise.cpp
j2kun: Done

tests/Dialect/Secret/Conversions/secret_to_mod_arith/client_helpers.mlir
j2kun: Done

lib/Dialect/Secret/Conversions/Patterns.h
j2kun: Done

lib/Dialect/ModuleAttributes.cpp
asraa: Ahhhh thank you!!

PR #1634, opened 2025-03-31
Configure bazel to use a hermetic MacOS toolchain

j2kun:
This PR uses the https://github.com/bazelbuild/apple_support/ project to configure a proper CC toolchain for mac platforms. This PR follows the [analogous configuration of the Tensorflow project](https://github.com/tensorflow/tensorflow/blob/fde75f0fd00689b0731864a96efd81955897e9ba/.bazelrc#L196-L198) but using the simpler bazelmod configuration.

The _downside_ of this PR is that is requires mac users to have a full XCode installation (I tested it with XCode 16.2). The upside is that we should not expect to see any more mac versioning errors or linking errors when incompatible pieces of different toolchains are incorrectly selected by bazel via the default unix toolchain.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -22,9 +22,18 @@ build:linux --linkopt="-fuse-ld=lld"
 
 # macos specific options
 build:macos --apple_platform_type=macos
-build:macos --macos_minimum_os=10.13
-build:macos --macos_sdk_version=10.13
+build:macos --macos_minimum_os=10.15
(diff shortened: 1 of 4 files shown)


asraa (commented):
if i understand right, this doesn't add any system deps to mac os users, but will increase their build times since it'll use a bazel installed xcode? i greatly like less system deps for users, hermetic builds are good

asraa (approved):

Issue #1521, opened 2025-03-04
Fully support MacOS Github CI


j2kun:
A few problems remaining with the current workflow runner:

- The jit errors we see internally for mlir-runner are also showing up on macos
- OpenFHE compilation in the frontend JIT is failing, likely because the include paths are not set up properly.
- polynomial approximation unit tests are failing (!) with differences of actual values.

The build succeeds, but those tests fail, so I'm setting the workflow to run daily with the tests restricted to running `//tests/Dialect/...:all`

hieunch:
I was able to pass the tests with `mlir-runner` on a macOS machine by adding `alwayslink = True` to both the `_mlir_c_runner_utils` and `_mlir_runner_utils` library targets in the `BUILD.bazel` file within `mlir`.

```
cc_library(
    name = "_mlir_runner_utils",
    srcs = ["lib/ExecutionEngine/RunnerUtils.cpp"],
    hdrs = ["include/mlir/ExecutionEngine/RunnerUtils.h"],
    includes = ["include"],
    deps = [
        ":mlir_c_runner_utils",
        ":mlir_float16_utils",
    ],
    alwayslink = True,
)
```

j2kun:
Thank you!! I should be able to upstream this next week.

PR #1635, opened 2025-03-31
Make poly approx unit tests compatible with macos

j2kun:
Fixes the poly approximation failures from https://github.com/google/heir/issues/1521

The main issue seems to be that some primitives are implemented differently in macos, leading to slight differences (on the order of 1e-15) in some of the resulting values which are currently being tested for equality. This simply relaxes some of the equality checks.


ZenithalHourlyRate (approved):
LGTM!

PR #1638, opened 2025-04-01
Openfhe: support more param options

ZenithalHourlyRate:
For fine control of the openfhe backend, more options in `openfhe-configure-crypto-context` might be wanted. This is used when the information from LWE IR does not indicate all behavior (like mulDepth and number of moduli, moduli sizes, etc).

Changes

* Model almost all FHE (not threshold HE) related options in Openfhe
* Refactor openfhe-configure-crypto-context for readability
  + It used to mix finding config code and code-gen code, now it is separate
  + function args now become global object for easy reference

### DIFF
--- a/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
@@ -35,235 +35,306 @@ namespace mlir {
 namespace heir {
 namespace openfhe {
 
+struct Config {
+  // computed from IR
+  bool hasRelinOp;
+  bool hasBootstrapOp;
+  SmallVector<int64_t> rotIndices;
+  // inherited from IR mgmt.openfhe_params
+  int64_t evalAddCount;
+  int64_t keySwitchCount;
+  // inherited from IR LWE type
+  int64_t plaintextModulus;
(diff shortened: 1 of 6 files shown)


lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -38,232 +38,301 @@ namespace openfhe {
...
+      }
+      return WalkResult::advance();
+    });
+    return result;
   }
-  if (!rotIndices.empty()) {
-    builder.create<openfhe::GenRotKeyOp>(cryptoContext, privateKey, rotIndices);
+
+  // global pass object to store the configuration, for other functions
+  // to access easily. Good design?
```
j2kun, line 84: I think it is fine since this is a member variable of the pass instance, not a global variable or a static variable of the class. So each instance of this pass will get a different config object (instantiated with default-constructed values for all fields). You could make this slightly clearer by defining the struct outside this class, and then just having a private member variable declared under the class constructor.

ZenithalHourlyRate: Moved that outside the class and have a `Config config` in private member field

j2kun (approved):

PR #1639, opened 2025-04-02
Fix go_library_bzl

ZenithalHourlyRate:
This line is causing build failure on main.

I could not find this macro anywhere. The only place I found it is under https://github.com/google/saxml/blob/c13a410b64f30c4b4e201aebaeceb1605104c7e4/saxml/BUILD#L27 but under its own `//tools` there is no other macros, so it should come from other packages, but I did not find that package (not rules_go nor bazel_tools, bazel_tools should be a candidate as it has `tools/build_defs/` but it does not contain the `go` path)

### DIFF
--- a/tests/Examples/lattigo/BUILD
+++ b/tests/Examples/lattigo/BUILD
@@ -7,7 +7,6 @@ bzl_library(
     srcs = ["test.bzl"],
     visibility = ["//visibility:private"],
     deps = [
-        "//tools/build_defs/go:go_library_bzl",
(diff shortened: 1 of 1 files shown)


j2kun:
Sorry, this is a problem with my internal patching tool. I will add some extra guards against it

asraa (approved):
Thanks, I tested this internally and it doesn't seem to affect the builds either

j2kun (approved):

PR #1648, opened 2025-04-02
fix iter args bug, allow multi-test files

j2kun:
Fixes bug with frontend loop generation when there is only one iter arg.

Uses https://pybind11.readthedocs.io/en/stable/advanced/classes.html#module-local-class-bindings to allow multiple compiled functions in the same program.

Better long-term effort would be to cache the pybind11 module in the OpenFHE backend, but as I understand this should still have the local instantiations be interoperable.

### DIFF
--- a/frontend/heir/mlir_emitter.py
+++ b/frontend/heir/mlir_emitter.py
@@ -533,7 +533,7 @@ class TextualMlirEmitter:
         raise NotImplementedError("Nested loops are not supported")
 
     body_str = self.emit_block(loop_block, blocks_to_print)
-    if len(loop.inits) > 1:
+    if loop.inits:
       # Yield the iter args
       yield_vars = ", ".join([self.get_name(init) for init in loop.inits])
       ret_types = ", ".join(
--- a/frontend/loop_test.py
+++ b/frontend/loop_test.py
@@ -4,7 +4,7 @@ from heir.mlir import I64, Secret
 from absl.testing import absltest  # fmt: skip
 
 
-class EndToEndTest(absltest.TestCase):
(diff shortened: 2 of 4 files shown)


AlexanderViand-Intel:
uncommenting the second example in `example.py`, I get
```
  File "/home/sdp/code/heir/frontend/example.py", line 33, in <module>
    foo.setup()  # runs keygen/etc
    ^^^^^^^^^^^
  File "/home/sdp/code/heir/frontend/heir/backends/openfhe/backend.py", line 61, in setup
    self.compilation_result.setup_funcs["configure_crypto_context"](
RuntimeError: external/openfhe/src/pke/include/cryptocontext.h:l.371:ValidateKey(): Key is nullptr
```

However, the first and third example work fine together. I think this is because the first and second generate the *exact same * cryptocontext (there's maybe some uniquing going on within OpenFHE? or maybe pybind relies on `==`  which is defined on cryptocontexts?) because they ask for the same parameters, whereas the third one asks for different parameters (keyswitchcount./evaladdcount differ).


Note: here's why I think this is the case (temp debug output):

If `foo` is the second @compile:
```
Using crypto context gen func <built-in method foo__generate_crypto_context of PyCapsule object at 0x7e30cd7146c0> for foo, which produces <_heir_func.CryptoContext object at 0x7e30db3994b0>
```

but if `loop_test` is the second @compile:
```
Using crypto context gen func <built-in method loop_test__generate_crypto_context of PyCapsule object at 0x791257f04a80> for loop_testm which produces <_heir_loop_test.CryptoContext object at 0x791267191a70>
```

PR #1654, opened 2025-04-04
pre-commit: make search-and-replace effective for .mlir files

ZenithalHourlyRate:
Addresses https://github.com/google/heir/pull/1652#issuecomment-2777728931


asraa (approved):
Thanks for this fix

PR #1661, opened 2025-04-04
Use filegroup for referencing common mlir files in tests/Examples

j2kun:
Fixes https://github.com/google/heir/issues/1659


ZenithalHourlyRate (approved):

j2kun:
Turns out we don't even need the filegroup or data attribute, and can do this just using `exports_files` on the source directory. Internal google linters made me change it, so the PR will land with that change not reflected here.

PR #1668, opened 2025-04-04
Relax tensor_ext attributes to support layouts of scalars

j2kun:

This allows one to write 

- repeat a scalar in a ciphertext: `#alignment = #tensor_ext.alignment<in = [], out = [1], insertedDims = [0]>`
- put the scalar in the first position and pad with zeros: `#alignment = #tensor_ext.alignment<in = [], out = [1], insertedDims = [0], padding = [1023], paddingValue = 0 : i16>`

Followup PRs will update the layout passes to support assigning and materializing scalar layouts

### DIFF
--- a/tests/Dialect/TensorExt/IR/alignment_verifier.mlir
+++ b/tests/Dialect/TensorExt/IR/alignment_verifier.mlir
@@ -1,12 +1,12 @@
 // RUN: heir-opt --verify-diagnostics --split-input-file %s
 
-// expected-error@below {{in and out may not be empty arrays}}
-#align = #tensor_ext.alignment<in = [], out = [32]>
-func.func private @test_fn(tensor<16xi32> {foo.bar = #align})
+// in == [] implies the input is a scalar
+#align = #tensor_ext.alignment<in = [], out = [32], insertedDims = [0]>
+func.func private @test_fn(i32 {foo.bar = #align})
 
 // -----
 
-// expected-error@below {{in and out may not be empty arrays}}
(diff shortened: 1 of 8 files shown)


tests/Dialect/TensorExt/IR/alignment_verifier.mlir
```diff
@@ -1,12 +1,12 @@
 // RUN: heir-opt --verify-diagnostics --split-input-file %s
 
-// expected-error@below {{in and out may not be empty arrays}}
-#align = #tensor_ext.alignment<in = [], out = [32]>
-func.func private @test_fn(tensor<16xi32> {foo.bar = #align})
+// in == [] implies the input is a scalar
```
code-perspective, line 3: Quick QQ: How do you differentiate between empty array vs a scalar?

code-perspective (commented):
The change lgtm C++-wise except for one Q below.

tests/Dialect/TensorExt/IR/alignment_verifier.mlir
j2kun: I think a tensor with rank zero also represents a scalar in MLIR (e.g., `tensor<f32>`), cf. https://mlir.llvm.org/docs/Dialects/Builtin/#example-6, but I am not sure if they are treated as equivalent types by the MLIR type system. Either way, I don't think it's common to encounter a rank-0 tensor so I plan to use this to always mean a scalar.

code-perspective: Thanks

PR #1669, opened 2025-04-04
fix a bug in the emitter, emit cleartext binops

j2kun:

The OpenFHE emitter had a bug in ExtractSliceOp (line 899) using an offset starting from the iterator end as the end of the slice, rather than the start.

I also needed to add some cleartext binary ops which correspond to the lowering of `affine.apply` during a loop.


lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
```diff
@@ -848,7 +914,13 @@ LogicalResult OpenFhePkeEmitter::printOperation(tensor::EmptyOp op) {
 
 LogicalResult OpenFhePkeEmitter::printOperation(tensor::ExtractOp op) {
   // const auto& v1 = in[0, 1];
-  emitAutoAssignPrefix(op.getResult());
+  if (isa<lwe::NewLWECiphertextType, lwe::LWECiphertextType>(
```
asraa, line 917: how did you get the old lwe ciphertext types in the IR? was it meant to be a plaintext type?

asraa (approved):
approving, but qq - maybe there's a test that can emit cleartext tensor extracts in loops for that question?

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
j2kun: I added this mainly for completeness, though I admit it's probably not needed here.

PR #1671, opened 2025-04-05
propagate scale backward to plaintext mgmt.init ops

j2kun:


ZenithalHourlyRate (approved):
Indeed the code path of `mul` did not handle the ct-pt case, thanks for the fix!

PR #1627, opened 2025-03-30
Lattigo: Add Negate op and e2e CMUX example

ZenithalHourlyRate:
Fixes #1623

Depends on #1620.

This PR construct a Lattigo Negate Op via ring operation in Lattigo (they seem not to offer a high level one).

To better demonstrate its use a CMUX e2e example from #1620 is also added.

PR #1674, opened 2025-04-05
LattigoEmitter: add tensor.insert, improve tensor.extract, improve constant emitter

j2kun:
Extracted from https://github.com/google/heir/pull/1633/commits

Support more things in lattigo emitter needed for codegen of new layout engine IR.

### DIFF
--- a/lib/Target/Lattigo/LattigoEmitter.cpp
+++ b/lib/Target/Lattigo/LattigoEmitter.cpp
@@ -58,7 +58,7 @@ LogicalResult LattigoEmitter::translate(Operation &op) {
           // Arith ops
           .Case<arith::ConstantOp>([&](auto op) { return printOperation(op); })
           // Tensor ops
-          .Case<tensor::ExtractOp, tensor::FromElementsOp>(
+          .Case<tensor::ExtractOp, tensor::InsertOp, tensor::FromElementsOp>(
               [&](auto op) { return printOperation(op); })
           // Lattigo ops
           .Case<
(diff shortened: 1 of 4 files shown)


j2kun:
There is still one small bug here, the decryption looks correct in the debug but the extraction of the cleartext at the end is giving 0 for everything.

ZenithalHourlyRate:
> There is still one small bug here, the decryption looks correct in the debug but the extraction of the cleartext at the end is giving 0 for everything.

`int64[]` is needed when you pass the slice to `dec.Decrypt`, so more precise constant type could have some issue.

You could base the work on #1627, see https://github.com/google/heir/pull/1627#discussion_r2026485892

j2kun:
Rebased on https://github.com/google/heir/pull/1627 and https://github.com/google/heir/pull/1671 and integrated my (much more minimal, now) changes

j2kun:
Rebased, ready for final review. Note more emitters coming for tensor.extract_slice and tensor.splat.

lib/Target/Lattigo/LattigoEmitter.cpp
```diff
@@ -318,9 +316,23 @@ LogicalResult LattigoEmitter::printOperation(arith::ConstantOp op) {
 }
 
 LogicalResult LattigoEmitter::printOperation(tensor::ExtractOp op) {
-  // only support 1-dim tensor for now
   os << getName(op.getResult()) << " := " << getName(op.getTensor()) << "[";
-  os << getName(op.getIndices()[0]) << "]\n";
+  os << flattenIndexExpression(
+      op.getTensor().getType(), op.getIndices(),
+      [&](Value value) { return variableNames->getNameForValue(value); });
```
ZenithalHourlyRate, line 322: Use the `getName` wrapper for `variableNames->getNameForValue(value);`. Same below.

ZenithalHourlyRate (approved):

PR #1675, opened 2025-04-06
docs: openfhe emitter include path change

ZenithalHourlyRate:
OpenfhePkeEmitter now defaults to `install-relative`, and may cause trouble for new users running through the tutorial.


asraa (approved):
Thanks!

PR #1678, opened 2025-04-07
frontend: use bazel-built OpenFHE if no OpenFHE installed

AlexanderViand-Intel:
It's quite convenient to run the frontend directly from python (e.g.,`python frontend/example.py`) without using bazel. However, the OpenFHE backend would only work if either (a) one manually sets the same environment variables that bazel uses or (b) had OpenFHE installed to its default location. 

This PR extends the OpenFHE backend config so that, if no environment variables are found, and no default OpenFHE installation is found, it'll try to find the bazel-built OpenFHE.

### DIFF
--- a/frontend/testing.bzl
+++ b/frontend/testing.bzl
@@ -38,7 +38,7 @@ def frontend_test(name, srcs, deps = [], data = [], tags = []):
             "OPENFHE_INCLUDE_TYPE": "source-relative",
             "OPENFHE_LINK_LIBS": ":".join(libs),
             "OPENFHE_INCLUDE_DIR": ":".join(include_dirs),
-            "HEIR_REPO_ROOT_MARKER": "bazel-bin",
+            "HEIR_REPO_ROOT_MARKER": ".",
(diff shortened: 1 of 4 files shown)


frontend/testing.bzl
```diff
@@ -38,7 +38,7 @@ def frontend_test(name, srcs, deps = [], data = [], tags = []):
             "OPENFHE_INCLUDE_TYPE": "source-relative",
             "OPENFHE_LINK_LIBS": ":".join(libs),
             "OPENFHE_INCLUDE_DIR": ":".join(include_dirs),
-            "HEIR_REPO_ROOT_MARKER": "bazel-bin",
+            "HEIR_REPO_ROOT_MARKER": ".",
```
j2kun, line 41: Probably just don't set this at all?

j2kun (approved):

frontend/testing.bzl
AlexanderViand-Intel: I was mostly cargo-cult-coding here. This was introduced in the merge/patch to one of my frontend PRs, so not sure why it was needed in the first place?

PR #1683, opened 2025-04-08
tests: use export files for common src

ZenithalHourlyRate:
Followup of #1661 and #1627

* move `cmux.mlir` to `common` folder
* change visibility to `//test` folders only
* cleanup `bfv_debug.go` symlink using `export_files`
* use common `bgv_debug.go` and `ckks_debug.go`

I also intend to do so for openfhe debug handlers, but they would become `.cpp.inc` file instead of a standalone `cc_library`, unlike `go_library` here.


asraa (approved):
thanks

PR #1690, opened 2025-04-09
Add layout-propagation support for affine.for and tensor.extract, support scalar layouts

j2kun:

Rebased over https://github.com/google/heir/pull/1688 and https://github.com/google/heir/pull/1665

Changes:

- Support scalar layouts
- Add ciphertext-size as a pass option
- Improve default tensor layouts to add alignment (this part is a bit iffy)
- support affine.for and tensor.extract
- move add/mul to a default that passes layout through ops unchanged
- refactor assignDefaultLayoutForOpOperand to reuse the logic to insert assign_layout ops
- Modify alignment in linalg.reduce propagation
- Don't apply a layout for non-secret func args
- Don't require a layout for index types (also a bit iffy)
- Use setAttributeAssociatedWith in a few places for consistency


asraa (approved):
3rd commit lgtm!

j2kun:
Rebased over main

PR #1699, opened 2025-04-10
run the macos-ci action also if a PR is labeled to do so

j2kun:


.github/workflows/build_and_test_macos.yml
```diff
@@ -4,11 +4,14 @@ on:
   schedule:
     # daily at 3 AM
     - cron: "0 3 * * *"
+  pull_request_target:
```
asraa, line 7: why not use pull_request here?

pull_request_target is generally avoided https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

j2kun: Because I am an ignorant fool :) I will fix it.

j2kun:
@asraa fixed!

PR #1704, opened 2025-04-11
Emit affine loops in lattigo

j2kun:
This PR adds affine loop support for lattigo emitter.

### DIFF
--- a/lib/Target/Lattigo/LattigoEmitter.cpp
+++ b/lib/Target/Lattigo/LattigoEmitter.cpp
@@ -13,11 +13,12 @@
 #include "lib/Dialect/TensorExt/IR/TensorExtDialect.h"
 #include "lib/Target/Lattigo/LattigoTemplates.h"
 #include "lib/Utils/TargetUtils.h"
-#include "llvm/include/llvm/ADT/TypeSwitch.h"            // from @llvm-project
-#include "llvm/include/llvm/Support/CommandLine.h"       // from @llvm-project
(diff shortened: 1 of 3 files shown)


lib/Target/Lattigo/LattigoEmitter.cpp
```diff
@@ -228,6 +232,66 @@ LogicalResult LattigoEmitter::printOperation(func::CallOp op) {
...
+            variableNames->getNameForValue(operand)) {
+      // This occurs in cases where the loop is inserting into a tensor and
+      // passing it along as an inter arg.
+      continue;
+    }
+
+    os << getName(iterArg) << " := " << getName(operand) << "\n";
+  }
+
+  os << llvm::formatv("for {0} := {1}; {0} < {2}; {0} += 1 {{\n",
```
ZenithalHourlyRate, line 257: `+= 1` seems hard encoding but it is fine for now.

ZenithalHourlyRate (approved):

lib/Target/Lattigo/LattigoEmitter.cpp
j2kun: This is a good point, let me fix it. Also I think I could ensure all the iteration indices are `int64` because elsewhere `index` type is cast to `int64`.

j2kun: Fixed and force pushed

PR #1705, opened 2025-04-11
RNS: remove rns of poly

ZenithalHourlyRate:
After we repatriated `polynomial` dialect to support AnyType coefficient, to support `rns` we no longer need to have rns-of-poly `!rns<!poly, !poly>` as it is inconvenient to operate. Instead we should use `!poly<!rns<!mod_arith>>`.

This PR removes the stall rns of poly impl, and resolves the cyclic dependency among all these dialects.

Side Note: This is paving the way for a lowering of rns polynomial. Namely how `poly.add !poly<!rns<!mod_arith>>` lowers. My discussion with the intern is to make `mod_arith.add` supports `mod_arith.add !rns<!mod_arith>` as it is quite natural. `!rns` here could be viewed as `tensor`; another alternative is `rns.add !rns<!mod_arith>>` but this seems redefining a whole set of ops (does rns itself has the concept of addition?).

We will have a version that takes the first approach (first PR being just adding type support, without actual lowering) and we ask for feedback on the attempt.


j2kun:
Treating rns as if it were tensor seems fine to me.

j2kun (approved):

PR #1709, opened 2025-04-11
clear tensor_ext attributes when dropping secrets

j2kun:
This is for the plaintext backend, once the layout system is merged


asraa (approved):
do you not need to remove the tensor_ext.layout attrs from operations too? or is there a particular reason func can't have those attrs?

j2kun:
> do you not need to remove the tensor_ext.layout attrs from operations too? or is there a particular reason func can't have those attrs?

It's removed by convert-to-ciphertext-semantics, but the original type attribute is left over because it's needed for adding the client interface.

PR #1711, opened 2025-04-11
Openfhe and Lattigo emitter improvements

j2kun:
Added to lattigo:

- arith: addi, cmpi, divsi, extf, extsi, extui, index_cast, muli, remsi, select, subi
- tensor: concat, empty, insert_slice

Added to openfhe:

- arith: divsi, muli, subi
- tensor: extract_slice

Also fixes a bug with nested loop generation in extract_slice where the destination index was being initialized inside the loop instead of outside.

### DIFF
--- a/lib/Target/Lattigo/LattigoEmitter.cpp
+++ b/lib/Target/Lattigo/LattigoEmitter.cpp
@@ -68,9 +68,14 @@ LogicalResult LattigoEmitter::translate(Operation &op) {
           .Case<affine::AffineForOp, affine::AffineYieldOp>(
               [&](auto op) { return printOperation(op); })
           // Arith ops
-          .Case<arith::ConstantOp>([&](auto op) { return printOperation(op); })
+          .Case<arith::ConstantOp, arith::ExtSIOp, arith::ExtUIOp,
+                arith::IndexCastOp, arith::ExtFOp, arith::RemSIOp,
+                arith::AddIOp, arith::SubIOp, arith::MulIOp, arith::DivSIOp,
+                arith::CmpIOp, arith::SelectOp>(
+              [&](auto op) { return printOperation(op); })
           // Tensor ops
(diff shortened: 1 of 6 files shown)


lib/Target/Lattigo/LattigoEmitter.cpp
```diff
@@ -393,6 +398,235 @@ LogicalResult LattigoEmitter::printOperation(arith::ConstantOp op) {
...
+    os.unindent();
+    os << "} else {\n";
+    os.indent();
+    os << resultName << " = " << resultTy << "(0)\n";
+    os.unindent();
+    os << "}\n";
+  }
+
+  return success();
+  return success();
```
asraa, line 519: nit: remove

lib/Target/Lattigo/LattigoEmitter.cpp
```diff
@@ -393,6 +398,235 @@ LogicalResult LattigoEmitter::printOperation(arith::ConstantOp op) {
...
+    // }
+    auto res = convertType(op.getResult().getType());
+    if (failed(res)) {
+      return failure();
+    }
+    std::string resultName = getName(op.getResult());
+    std::string resultTy = res.value();
+
+    os << "var " << resultName << " " << resultTy << "\n";
+    os << "if " << inputVarName << " {\n";
```
asraa, line 507: I remember reviewing @ZenithalHourlyRate 's recent PR on the lattigo emitter and there's a similar boolean upcast to an integer in the encode operation.

I assume eventually these will converge once those encoding/repition go to the layout, but maybe it's worth tagging it with a TODO.
Mostly this is just repeated code and it reminded me of that other impl of boolToInt64 if/else statement

asraa (approved):

lib/Target/Lattigo/LattigoEmitter.cpp
j2kun: Now is a good a time as ever to generalize.

PR #1716, opened 2025-04-12
Only rotate-and-reduce post matvec when original matrix is non-square

j2kun:
I was using the packed matrix type, which may not be square even if the original matrix type is square.

[edit]: I had to pull in a few extra minor fixes from https://github.com/google/heir/pull/1633 related to the context-aware type conversion, since this new test was hitting those bugs as well


asraa (approved):

j2kun:
Please wait to merge until I add a test. This is not blocking the paper submission.

j2kun:
Ready for review

asraa (approved):
thanks!

Issue #1712, opened 2025-04-12
Yosys optimizer pass error


ludns:
Runnig the Yosys optimizer pass (`--yosys-optimizer=mode=Boolean`) on this IR fails
```
module {
  func.func @matmul(%arg0: !secret.secret<memref<10x10xi16>>, %arg1: !secret.secret<memref<10x10xi16>>, %arg2: !secret.secret<memref<10x10xi16>>) attributes {llvm.linkage = #llvm.linkage<external>} {
    secret.generic ins(%arg0, %arg1, %arg2 : !secret.secret<memref<10x10xi16>>, !secret.secret<memref<10x10xi16>>, !secret.secret<memref<10x10xi16>>) {
    ^body(%input0: memref<10x10xi16>, %input1: memref<10x10xi16>, %input2: memref<10x10xi16>):
      affine.for %arg3 = 0 to 10 {
        affine.for %arg4 = 0 to 10 {
          affine.for %arg5 = 0 to 10 {
            %0 = affine.load %input0[%arg3, %arg5] : memref<10x10xi16>
            %1 = arith.extsi %0 : i16 to i32
            %2 = affine.load %input1[%arg5, %arg4] : memref<10x10xi16>
            %3 = arith.extsi %2 : i16 to i32
            %4 = arith.muli %1, %3 : i32
            %5 = affine.load %input2[%arg3, %arg4] : memref<10x10xi16>
            %6 = arith.trunci %4 : i32 to i16
            %7 = arith.addi %5, %6 : i16
            affine.store %7, %input2[%arg3, %arg4] : memref<10x10xi16>
          }
        }
      }
      secret.yield
    }
    return
  }
}
```

Error:
```
<stdin>:5:7: error: 'affine.for' op unable to find printer for op
      affine.for %arg3 = 0 to 10 {
      ^
<stdin>:5:7: note: see current operation: 
"affine.for"() <{lowerBoundMap = affine_map<() -> (0)>, operandSegmentSizes = array<i32: 0, 0, 0>, step = 1 : index, upperBoundMap = affine_map<() -> (10)>}> ({
^bb0(%arg3: index):
  "affine.for"() <{lowerBoundMap = affine_map<() -> (0)>, operandSegmentSizes = array<i32: 0, 0, 0>, step = 1 : index, upperBoundMap = affine_map<() -> (10)>}> ({
  ^bb0(%arg4: index):
[...]

PR #1718, opened 2025-04-14
fix: fix path for yosys share files for bazel run

asraa:
Came up in the discussion of https://github.com/google/heir/issues/1712#issuecomment-2798449383

The yosys share files directory doesn't use the workspace prefix _main after the bazel upgrade - it was only intended to work in the root of the repository. I think I couldn't use the rootpath on a folder of files, so that's why i can't do something like "find the location in bazel-bin".

It would be so so much easier to hold the techmap in memory but I don't think I can..

### DIFF
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -6,10 +6,6 @@ package(
     default_visibility = ["//visibility:public"],
 )
 
-# changed from "heir" (or nothing) in bazel 8
-# https://bazel.build/remote/output-directories#layout-diagram
-WORKSPACE_PATH = "_main/"
-
 # Custom `mlir-opt` replacement that links our dialect and passes
 cc_binary(
     name = "heir-opt",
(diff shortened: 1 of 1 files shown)


j2kun:
> It would be so so much easier to hold the techmap in memory but I don't think I can..

we had discussed once upon a time having HEIR build the yosys IR in memory directly, and use the yosys API for running the yosys passes... that would probably eliminate this error entirely, but would be a lot of work.

j2kun:
> I think I couldn't use the rootpath on a folder of files 

maybe we could try packaging the files in a filegroup to achieve this?

j2kun (approved):

asraa:
> we had discussed once upon a time having HEIR build the yosys IR in memory directly, and use the yosys API for running the yosys passes..

Building the IR won't help this problem since it's the techmap that needs to be loaded in, and I went and investigated what it would look like to use the library to add the techmap and it's an absolute nightmare to separate out the pass call from the library calls :/

asraa:
> maybe we could try packaging the files in a filegroup to achieve this?

There isn't a way to get the base directory of a filegroup though. I could maybe try to put it all in one file though too, I just realized.

PR #1730, opened 2025-04-16
Move polynomial eval lowering into its own, isolated pass

j2kun:
After thinking a bit more about the prototype in https://github.com/google/heir/pull/1568, I realized that we don't actually need a polynomial to evaluate on ciphertext ops (yet) because we do polynomial approximation before lowering to scheme ops. So we only need to support float types and tensors of float types.

That may change if we need a lowering for `polynomial.eval` in polynomial-to-llvm, but we don't have anything that currently generates `polynomial.eval` in that context, as we lower it on data-semantic tensors long before we want to lower polynomial to llvm.

So this merely extracts the `polynomial.eval` lowering into its own pass and slightly cleans up the lowering for arith/tensor compatibility.


lib/Transforms/LowerPolynomialEval/LowerPolynomialEval.td
```diff
@@ -0,0 +1,45 @@
...
+  // TODO(#1565): Add support for Chebyshev-basis methods
+  //  - `"clenshaw"`: Clenshaw's method (Chebyshev basis)
+  //  - `"ps-cheb"`: Paterson-Stockmeyer (Chebyshev basis)
+  //  - `"bsgs"`: Baby Step Giant Step (Chebyshev basis)
+  }];
+  let dependentDialects = [
+    "::mlir::heir::polynomial::PolynomialDialect",
+  ];
+  let options = [
+    Option<"method", "method", "std::string",
```
asraa, line 34: you can also make this an enum too so that the helper string prints out the available options (with clEnumVal), or else maybe add this to the help string to show up on the CLI

lib/Transforms/LowerPolynomialEval/LowerPolynomialEval.td
```diff
@@ -0,0 +1,45 @@
...
+           /*default=*/"", "The method used to lower polynomial.eval.">,
+
+    // We allow the user to override the default dialect detection
+    // for cases like BGV/CKKS/BFV where the type used to trigger
+    // the dialect detection has a shared dialect (lwe).
+    Option<"dialect", "dialect", "std::string",
+           /*default=*/"", "The dialect to target; automatically detected if unset.">,
```
asraa, line 41: i don't _think_ this is actually used? it seems like it's force true/false if a method is set, but otherwise we auto detect

lib/Transforms/LowerPolynomialEval/Patterns.cpp
```diff
@@ -0,0 +1,201 @@
...
+LogicalResult LowerViaHorner::matchAndRewrite(EvalOp op,
+                                              PatternRewriter& rewriter) const {
+  Type evaluatedType = op.getValue().getType();
+  ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+  b.setInsertionPoint(op);
+
+  LLVM_DEBUG(llvm::dbgs() << "evaluatedType: " << evaluatedType << "\n");
+
+  auto attr =
+      dyn_cast<polynomial::TypedFloatPolynomialAttr>(op.getPolynomialAttr());
```
asraa, line 37: remind me, do we always expect poly.eval from the poly approx patterns to generated a typed float polynomial attr? (the verifier also allows int, but also "FloatPolynomialAttr") - could you add a quick comment at the return failure() below?

or maybe the intention was that another pattern will later match on other poly types?

lib/Transforms/LowerPolynomialEval/Patterns.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  // Start with the coefficient of the highest degree term
+  Value result =
+      b.create<arith::ConstantOp>(evaluatedType, attributeMap[maxDegree]);
+
+  // Apply Horner's method, accounting for possible missing terms
+  auto x = op.getOperand();
+  for (int64_t i = maxDegree - 1; i >= 0; i--) {
+    // Multiply by x
+    result = b.create<arith::MulFOp>(result, x);
```
asraa, line 60: quick comment: support we have a polynomial to evaluate that only has even integer powers. this code will generate the entire monomial basis and we'll be left with a bunch of odd x power terms, even though we could just fold those and simplify them to muls by x^2 powers

obviously this pass shouldn't handle that, but i'm curious if CSE / canonicalization will do that automagically?

lib/Transforms/LowerPolynomialEval/Patterns.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  auto terms = attr.getValue().getPolynomial().getTerms();
+
+  if (terms.empty()) {
+    // An empty polynomial evaluates to zero.
+    Value zeroConst = b.create<arith::ConstantOp>(evaluatedType,
+                                                  b.getZeroAttr(evaluatedType));
+    rewriter.replaceOp(op, zeroConst);
+    return success();
+  }
```
asraa, line 92: this is covered by the verifier, right, so shouldn't ever hit? (why isn't an empty poly eval not a folder?)

asraa (approved):
thanks!

lib/Transforms/LowerPolynomialEval/Patterns.cpp
j2kun: I think we can generalize later to include those options, but right now the polynomial approximation pass always produces a typed float polynomial attr

lib/Transforms/LowerPolynomialEval/Patterns.cpp
j2kun: > if CSE / canonicalization will do that automagically

I don't think so. And I feel it would be better to more generally identify that repeated muls can be replaced with iterated squaring. Or else that this pass should just emit iterated squaring directly.

lib/Transforms/LowerPolynomialEval/Patterns.cpp
j2kun: Correct, removed!

lib/Transforms/LowerPolynomialEval/LowerPolynomialEval.td
j2kun: Left over from the more general port. Removed.

lib/Transforms/LowerPolynomialEval/LowerPolynomialEval.td
j2kun: Done!

asraa (approved):
:tada:

PR #1737, opened 2025-04-17
add pybind11 libs for macos frontend

asraa:


j2kun (approved):

asraa:
OK the bazel test command - the linker can't find -lpython3.11, so i must be missing the right path to search - gonna see if i can fix that

PR #1745, opened 2025-04-21
[WIP]: add cibuildwheel to enable PyPI distribution of HEIR frontend

j2kun:


j2kun:
Update: fixed the test failure by doing a better pip-installed detection

frontend/heir/backends/util/clang.py
```diff
@@ -26,10 +26,16 @@ def __init__(self, path_to_clang: Optional[str] = None):
     self.compiler_binary_path = path_to_clang or self._find_clang()
 
   def _find_clang(self):
```
AlexanderViand, line 28: nit: rename this to `_find_cpp_compiler` or similar, since it'll happy return ANY compiler the user might have set up as their `CXX`

EDIT: and similarily, the error message should probably just talk about "a C++ compiler" instead of mentioning clang++?

pyproject.toml
```diff
@@ -1,3 +1,45 @@
...
+
+[project]
+name = "heir_py"
+version = "0.0.1"
+authors = [
+    { name = "Jeremy Kun", email = "jkun@google.com" },
+    { name = "Asra Ali", email = "asraa@google.com" },
+    { name = "Alexander Viand", email = "alexander.viand@intel.com" },
+]
+description = "Python frontend for the HEIR compiler"
```
AlexanderViand, line 13: Is it really correct to describe this package as "just" a frontend for the HEIR compiler when it actually ships with the compiler + backend(s)?

AlexanderViand (commented):
I wasn't able to do an editable install (`pip install -e  .`) - see https://gist.github.com/AlexanderViand/7e09b4f833e28a1e3e7a5a8425a55c3e  but I was able to `pip install .` though even after that, running `python frontend/example.py` from repo root would use the local stuff and say `is_pip_installed()` is `False`....

Some updates to frontend/README.MD for how a development setup is supposed to work would be very welcome! 

I'm also wondering how the pyproject.toml dependencies and requirements.txt/requirements-dev.txt interact?

j2kun:
> I wasn't able to do an editable install (`pip install -e  .`) - see https://gist.github.com/AlexanderViand/7e09b4f833e28a1e3e7a5a8425a55c3e  but I was able to `pip install .` though even after that, running `python frontend/example.py` from repo root would use the local stuff and say `is_pip_installed()` is `False`....

Wrt is_pip_installed, -e is a special case because it just points to the local files, and in that case you want it to use the same config as if it weren't pip installed. I am out for the day but I will test -e again before submitting and write some readme instructions.

> I'm also wondering how the pyproject.toml dependencies and requirements.txt/requirements-dev.txt interact? 

AFAIK they do not interact at all. I don't know if there's a way to make one a source of truth for the other.

pyproject.toml
j2kun: Updated

frontend/heir/backends/util/clang.py
j2kun: Done, along with more renaming of the module.

j2kun:
After 6baf813e you should be able to run `pip install -e .` from the repository root (not `frontend/`) and then it will work as expected. Note this will _not_ run the `bazel build` and expect the build to have already been run (or else you will see an error message like "couldn't find heir-opt, did you run bazel build?"). So `pip install -e .` will be fast.

By contrast, `pip install .` will run `bazel build` (using clang locally, and gcc in CI, cf. 3afacde173fa483a626ea18c536bb9b0490d4f0f) which may incur a rebuild since it has particular CLI flags chosen for the build. So this may be a lot slower.

AlexanderViand (approved):
Amazing, thanks - both worked flawlessly for me! 

one final nit: I think `frontend/requirements.txt` is no longer used/needed and can be removed

j2kun:
I agree with removing the requirements.txt, though doing so required a bit more reorganization of the bazel deps. Now there is one `heir_pip_deps` for the frontend, and one `heir_pip_dev_deps` for everything else, and notably the jaxite deps are in the dev deps.

j2kun:
The bazel rule name change broke copybara, so I will make that change internally after submitting this larger PR

Issue #1446, opened 2025-02-21
Migrate remaining dependencies to Bazelmod


j2kun:
The main dependencies that are still handled by the WORKSPACE:

- LLVM (Cf. https://github.com/llvm/llvm-project/pull/88927)
- LLVM dependencies, `zlib` and `zstd`
- or-tools and its deps (Cf. https://github.com/bazelbuild/bazel-central-registry/pull/3832)
- rules_hdl (Cf. https://github.com/hdl/bazel_rules_hdl/issues/336)
- OpenFHE and its dependencies (rapidjson, cereal)
- PocketFFT

The options are to wait for the dependencies to migrate to bazelmod, or else to add our own bazelmod extension (possibly with a `git_override`).

j2kun:
All that's left now is rules_hdl (Cf. Getting ready for bzlmod hdl/bazel_rules_hdl#336)

PR #1751, opened 2025-04-22
migrate or-tools to bazelmod

j2kun:
Part of https://github.com/google/heir/issues/1446

This also resolves a bunch of deprecation warnings for `cc_proto_library` which were fixed upstream.


asraa (approved):
:tada:

PR #1757, opened 2025-04-23
frontend: fix type inference for extended types

asraa:
Fixes type inference when extension is needed.

The emit instruction code assumed that the lhs_ssa value was always the long integer type that was being bumped up. Fixed to handle when the rhs_ssa value is the longer one.

### DIFF
--- a/frontend/heir/mlir_emitter.py
+++ b/frontend/heir/mlir_emitter.py
@@ -392,12 +392,12 @@ class TextualMlirEmitter:
         )
         return ""
       case ir.Expr(op="binop"):
-        emitted_expr, ext = self.emit_binop(assign.value)
+        emitted_expr, ext, ty = self.emit_binop(assign.value)
         name = self.get_or_create_name(assign.target)
         return (
             f"{ext}"
             f"{name} = {emitted_expr} :"
-            f" {mlirType(self.typemap.get(assign.value.lhs.name))} {mlirLoc(assign.loc)}"
+            f" {mlirType(ty)} {mlirLoc(assign.loc)}"
         )
       case ir.Expr(op="call"):
         func = assign.value.func
@@ -444,7 +444,7 @@ class TextualMlirEmitter:
 
     # Types agree: do nothing
     if lhs_type == rhs_type:
-      return self.get_name(lhs), self.get_name(rhs), ""
+      return self.get_name(lhs), self.get_name(rhs), "", lhs_type
 
     # types aren't integer types
     if not isinstance(lhs_type, types.Integer) or not isinstance(
@@ -460,7 +460,7 @@ class TextualMlirEmitter:
       # TODO (#1162): Support bitwidth extension for non-scalar types (e.g., tensors)
 
     if lhs_type.bitwidth == rhs_type.bitwidth:
-      return self.get_name(lhs), self.get_name(rhs), ""
+      return self.get_name(lhs), self.get_name(rhs), "", lhs_type
 
     # time to emit some extensions!
     short, long = lhs, rhs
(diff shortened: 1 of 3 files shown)


frontend/heir/mlir_emitter.py
```diff
@@ -476,40 +476,43 @@ def emit_ext_if_needed(self, lhs, rhs):
     )
 
     if lhs_type.bitwidth > rhs_type.bitwidth:
-      return self.get_name(lhs), tmp, ext
-    return tmp, self.get_name(rhs), ext
+      return self.get_name(lhs), tmp, ext, self.typemap.get(str(long))
+    return tmp, self.get_name(rhs), ext, self.typemap.get(str(long))
```
AlexanderViand, line 480: Why do you hardcode the returned type here? Shouldn't this just be `lhs_type` in the true case, and `rhs_type` in the false case?

frontend/heir/mlir_emitter.py
```diff
@@ -476,40 +476,43 @@ def emit_ext_if_needed(self, lhs, rhs):
...
-        return f"arith.cmp{suffix} ne, {lhs_ssa}, {rhs_ssa}", ext
+        return f"arith.cmp{suffix} ne, {lhs_ssa}, {rhs_ssa}", ext, ty
       case operator.add:
-        return f"arith.add{suffix} {lhs_ssa}, {rhs_ssa}", ext
+        return f"arith.add{suffix} {lhs_ssa}, {rhs_ssa}", ext, ty
       case operator.mul:
-        return f"arith.mul{suffix} {lhs_ssa}, {rhs_ssa}", ext
+        import ipdb
+
+        ipdb.set_trace()
```
AlexanderViand, line 502: leftover debug code?

frontend/type_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+from absl.testing import absltest  # fmt: skip
+
+
+class TypeTest(absltest.TestCase):
+
+  def test_type(self):
+
+    @compile()
+    def type_two(x: Secret[I64], y: Secret[I64]):
```
AlexanderViand, line 12: I must be missing something, because it seems like this test doesn't actually involve any type extension?

AlexanderViand (changes requested):
:see-no-evil: seems pretty obvious in hindsight but totally missed this at the time

frontend/heir/mlir_emitter.py
asraa: Ah, I thought it was easier to put the long type there since we want to emit the long type for the binary op.

A few lines earlier, lhs and rhs are appropriately places in short/long so what you believe should be happening is indeed what it's doing.

frontend/type_test.py
asraa: Constants need to be type extended - the constant 2 is initially an i32, and is the LHS SSA value for the `2 * x` - this was the regression test. 
(In a testcase we need a non block arg, e.g. a constant, to be mismatched with the input types)

frontend/heir/mlir_emitter.py
asraa: addressed! in hindsight it is nicer to just use lhs_type/rhs_type

frontend/type_test.py
asraa: added a comment here explaining

AlexanderViand: Oh, I'm blind, completely missed the constant `2` in there.

AlexanderViand (approved):
Thanks!

j2kun (approved):

PR #1760, opened 2025-04-23
properly convert constant argument attributes in ArithToModArith

j2kun:
The type conversions were preserving the original types of the underlying attributes, when they should have been converted to the new type of the specified modulus. This is only visible in filecheck when printing the generic form of the IR, but the test in this PR demonstrates the problem.

### DIFF
--- a/lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
+++ b/lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
@@ -102,17 +102,30 @@ struct ConvertConstant : public OpConversionPattern<mlir::arith::ConstantOp> {
       return failure();
     }
 
-    auto denseIntAttr = dyn_cast<DenseIntElementsAttr>(op.getValue());
-    if (denseIntAttr) {
-      auto result = b.create<mod_arith::ConstantOp>(
-          typeConverter->convertType(op.getType()), denseIntAttr);
+    Type newResultType = typeConverter->convertType(op.getResult().getType());
+    mod_arith::ModArithType newResultEltTy =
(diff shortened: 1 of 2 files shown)


lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
```diff
@@ -102,17 +102,30 @@ struct ConvertConstant : public OpConversionPattern<mlir::arith::ConstantOp> {
...
+    unsigned newWidth = storageType.getWidth();
+
+    if (auto denseIntAttr = dyn_cast<DenseIntElementsAttr>(op.getValue())) {
+      auto shapedType = cast<ShapedType>(newResultType);
+      SmallVector<APInt, 4> newValues;
+      for (const APInt &val : denseIntAttr.getValues<APInt>()) {
+        newValues.push_back(val.sextOrTrunc(newWidth));
+      }
+      TypedAttr newAttr = DenseIntElementsAttr::get(
+          RankedTensorType::get(shapedType.getShape(), storageType), newValues);
```
AlexanderViand, line 119: Sometimes, working with MLIR is just *beautiful* xD 

Surprisingly, it seems like this is the only place where we need to do this kind of `DenseIntElementsAttr`-to-`DenseIntElementsAttr` conversion, so it's probably not worth making this a helper...

AlexanderViand (approved):
Wow, that's a subtle bug/issue!

lib/Dialect/Arith/Conversions/ArithToModArith/ArithToModArith.cpp
j2kun: This issue will disappear if / when I upstream some helpers related to this that will allow us to construct dense int elements attributes using mod arith types directly.

PR #1761, opened 2025-04-23
dense attrs in mod_arith.constant folder

j2kun:
This shallowly supports dense attributes in the fold method of mod_arith::ConstantOp::fold, since without that the MLIR infra actually complains that fold() fails, but just on the constant op since it is has the ConstantLike trait.

Filed a followup to properly support dense attributes throughout mod_arith canonicalization and folding.


lib/Dialect/ModArith/IR/ModArithOps.cpp
```diff
@@ -214,17 +214,33 @@ void ConstantOp::print(OpAsmPrinter &p) {
...
 OpFoldResult ConstantOp::fold(FoldAdaptor adaptor) {
-  // Check if the value is an IntegerAttr
-  auto intAttr = dyn_cast_if_present<IntegerAttr>(adaptor.getValue());
-  if (!intAttr) return {};
-
-  auto modType = dyn_cast_if_present<ModArithType>(getType());
+  Type type = getElementTypeOrSelf(getType());
+  auto modType = dyn_cast<ModArithType>(type);
+  auto storageType = modType.getModulus().getType();
   if (!modType) return {};
```
AlexanderViand, line 220:
```suggestion
  if (!modType) return {};
  auto storageType = modType.getModulus().getType();
```
Not sure `modType` can actually ever be null (would make the op invalid w.r.t. to verifier?) but probably safer this way around.

lib/Dialect/ModArith/IR/ModArithOps.cpp
```diff
@@ -214,17 +214,33 @@ void ConstantOp::print(OpAsmPrinter &p) {
...
+           "non-shaped type with shaped attribute");
+    ShapedType shapedType = cast<ShapedType>(getType());
+    SmallVector<APInt, 4> values;
+    for (const auto value : denseElementsAttr.getValues<APInt>()) {
+      values.push_back(value.zextOrTrunc(modBitWidth).urem(modulus));
+    }
+    // Have to use an integer type here because DenseIntElementsAttr
+    // requires integer types.
+    return DenseIntElementsAttr::get(
+        RankedTensorType::get(shapedType.getShape(), storageType), values);
```
AlexanderViand, line 238: Once more, MLIR makes stuff like this truly *beautiful* ;)

AlexanderViand (approved):
LGTM (mod one suggested change) :+1:

lib/Dialect/ModArith/IR/ModArithOps.cpp
j2kun: Done

Issue #235, opened 2023-11-07
tfhe_rs: improve generality of e2e test runner


j2kun:
- Support generated functions with multiple return values
- Support a variable number of input arguments (not just 2 ints)
- Custom function name besides `fn_under_test`?

j2kun:
Variadic args seems hard in rust because rust doesn't have a splat operator to read in the variadic CLI arguments and convert them to a static function signature. I thought maybe macro_rules would provide a workaround, but didn't have the time to figure it out.

j2kun:
A colleague pointed me to this which might help support the variadic arguments issue: https://jsdw.me/posts/rust-fn-traits/

j2kun:
Now that I've got the openfhe e2e tests working, it seems reasonable to migrate to proper rust rules and a rust test harness instead of lit testing.

PR #1743, opened 2025-04-18
make tfhe-rs a bazel dependency

asraa:
* Adds a tfhe_rs_lib bazel macro that runs heir-opt / heir-translate / rust_library
* Updates all CPU tests to an end to end bazel macro

FPGA tests:
Left the Cargo.toml with the local tfhe-rs folder, and the RUN commands as documentation.

Fixes #235

PR #1765, opened 2025-04-25
demo: add AES demo for tfhe_rust CPU

asraa:
Run
```
bazel run -c opt tests/Examples/tfhe_rust:aes
```

The sbox takes quite a bit of time, so I'll update that, but at least it's working?

Depends on #1743

### DIFF
--- a/tests/Examples/common/aes/inv_mix_columns.mlir
+++ b/tests/Examples/common/aes/inv_mix_columns.mlir
@@ -1,0 +1,77 @@
+// Inverse mix columns step, computing a linear combination of the columns.
+
+#map = affine_map<(d0) -> (d0)>
+
+#mapA = affine_map<(d0, d1) -> (d0, d1)> // output of 2
+#map1A = affine_map<(d0, d1) -> (d1)> // output dim 1
+#map2A = affine_map<(d0, d1) -> (d0)> // output dim 1
+module {
+  func.func private @mul_gf256_4(%x: i8 {secret.secret}, %y: i8) -> (i8) {
+      %z = arith.constant 0 : i8
+      %c0 = arith.constant 0 : i8
+      %c1 = arith.constant 1 : i8
+      %2:2 = affine.for %4 = 0 to 4 iter_args(%5 = %x, %6 = %z) -> (i8, i8) {
+        // z = if y & (1 << i): z^x else z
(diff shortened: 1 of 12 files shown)


ghost:
Looks good to me

tests/Examples/tfhe_rust_hl/fpga/BUILD
j2kun, line 1: Do we have an alternative to a build file for these fpga tests? CC @WoutLegiest

j2kun (approved):
A Herculean effort, well done.

Fixes #235?

j2kun:
Oh, you could also remove the rust-specific step from the ci workflow now

tests/Examples/tfhe_rust_hl/fpga/BUILD
asraa: This came up on https://github.com/google/heir/pull/1743

The codegen'd output is the same as the CPU tests, so the only utility in the FPGA directories as I see it is a folder that has the main that we can scp to the server that has device access. I think we could remove the BUILD file

asraa:
@j2kun removing the CI step is handled in https://github.com/google/heir/pull/1743/files

This PR depends on it! 

That PR is awaiting one more internal update :/ :/

tests/Examples/common/aes/inv_mix_columns.mlir
```diff
@@ -0,0 +1,79 @@
+// Inverse mix columns step, computing a linear combination of the columns.
+
+// heir-opt -- tests/Examples/common/aes/inv_mix_columns.mlir --full-loop-unroll --inline --symbol-dce --tosa-to-boolean-tfhe
```
code-perspective: Probably the same thing but how about using: --mlir-to-cggi and  --scheme-to-tfhe-rs for consistency

I suspect this is a stray comment that needs removing?

code-perspective (approved):
Looks great!

tests/Examples/common/aes/inv_mix_columns.mlir
asraa: I actually did need the inlining and symbol dce. I can update the tosa-to-boolean-tfhe part though. We don't really have a story yet for multiple functions in the MLIR, but it's more readbale that way

PR #1775, opened 2025-04-30
Converting to Ctxt/Ctxt op of the TFHE compare ops

WoutLegiest:
Not shure what the MODULE.bazel.lock file is doing here. 
Could be add it to gitignore?


j2kun (approved):
The lock file can stay for now.  We should probably have something update it automatically...

asraa (approved):
I think you may need to rebase this so that the internal change can be submitted?

Issue #1603, opened 2025-03-22
LWE: add support for `lwe.r{add,sub,mul}_plain`


ZenithalHourlyRate:
We have `bgv/ckks.add_plain/sub_plain/mul_plain`, during `<scheme>-to-lwe` lowering, these ops does not have correspondence in LWE dialect, so those ops should be created.

johnbntn:
Is anyone currently working on this? I'd like to

ZenithalHourlyRate:
AFAIK no. Please go ahead.

Issue #1199, opened 2024-12-16
lwe-to-polynomial: migrate encrypt/decrypt lowering to support RNS


ZenithalHourlyRate:
PR #1190 migrated most of the types to NewLWEType, however, for lwe-to-polynomial, there are some non-trivial changes that needs to be done.

This also relates to #882 

`NewLWEPlaintext` now has a `PlaintextSpace.Ring` (for example `Z_65537 / (X^N + 1)`), which is different from `NewCiphertext`'s `CiphertextSpace.Ring` (typically RNS)
  + lowering should convert plaintext ring to rns ring
  + lowering should properly handle modulus switching from `Z_65537` to `Z_qi` (ciphertext rns modulus), which needs `mod_arith` dialect support, and it is non-trivial.
  + lowering should handle scaling factor for CKKS scheme.

j2kun:
IIUC @asraa is working on this, and note it's a blocker for https://github.com/google/heir/pull/2013/files

asraa:
The remaining part of this (that https://github.com/google/heir/pull/2036 pushes) is that RNS types are not supported in polynomial-to-mod-arith now. Since poly coeff types can be RNS, poly operations are still operating on the RNS polynomial types. A further pass, e.g. polynomial-to-mod-arith should decompose the RNS polynomials into the individual limbs.

j2kun:
> A further pass, e.g. polynomial-to-mod-arith should decompose the RNS polynomials into the individual limbs.

It seems reasonable to define mod_arith ops to work on rns types as if they were vector types... Then maybe we could get away with inserting an "rns-unroll" type pass before mod-arith-to-arith. Thoughts?

AlexanderViand:
> > A further pass, e.g. polynomial-to-mod-arith should decompose the RNS polynomials into the individual limbs.
> 
> It seems reasonable to define mod_arith ops to work on rns types as if they were vector types... Then maybe we could get away with inserting an "rns-unroll" type pass before mod-arith-to-arith. Thoughts?

Yep, something like LimbwiseMappable would be great :)

asraa:
Oooh I like this! 

I'm just curious - does hardware natively support a vector of different modular operations? Or do they break them down into individual single mod Z_p operations?

AlexanderViand:
> I'm just curious - does hardware natively support a vector of different modular operations? Or do they break them down into individual single mod Z_p operations?

[...]

PR #1776, opened 2025-04-30
Add rlwe plain ops

johnbntn:
Added lwe ciphertext-plaintext ops and their lowerings for bgv/ckks  (#1603)  

Also I can't fully test the `lwe-to-polynomial` lowering until #1199 but I did my best to add the functionality and tests.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -343,12 +343,11 @@
     "https://bcr.bazel.build/modules/rules_java/7.2.0/MODULE.bazel": "06c0334c9be61e6cef2c8c84a7800cef502063269a5af25ceb100b192453d4ab",
     "https://bcr.bazel.build/modules/rules_java/7.3.2/MODULE.bazel": "50dece891cfdf1741ea230d001aa9c14398062f2b7c066470accace78e412bc2",
     "https://bcr.bazel.build/modules/rules_java/7.6.1/MODULE.bazel": "2f14b7e8a1aa2f67ae92bc69d1ec0fa8d9f827c4e17ff5e5f02e91caa3b2d0fe",
-    "https://bcr.bazel.build/modules/rules_java/8.11.0/MODULE.bazel": "c3d280bc5ff1038dcb3bacb95d3f6b83da8dd27bba57820ec89ea4085da767ad",
-    "https://bcr.bazel.build/modules/rules_java/8.11.0/source.json": "302b52a39259a85aa06ca3addb9787864ca3e03b432a5f964ea68244397e7544",
     "https://bcr.bazel.build/modules/rules_java/8.3.2/MODULE.bazel": "7336d5511ad5af0b8615fdc7477535a2e4e723a357b6713af439fe8cf0195017",
     "https://bcr.bazel.build/modules/rules_java/8.5.1/MODULE.bazel": "d8a9e38cc5228881f7055a6079f6f7821a073df3744d441978e7a43e20226939",
     "https://bcr.bazel.build/modules/rules_java/8.6.1/MODULE.bazel": "f4808e2ab5b0197f094cabce9f4b006a27766beb6a9975931da07099560ca9c2",
     "https://bcr.bazel.build/modules/rules_java/8.6.3/MODULE.bazel": "e90505b7a931d194245ffcfb6ff4ca8ef9d46b4e830d12e64817752e0198e2ed",
+    "https://bcr.bazel.build/modules/rules_java/8.6.3/source.json": "8330cc5d277085bbcf93e9f1c85c24d06975585606a1215df4faf886a8d3cc9e",
     "https://bcr.bazel.build/modules/rules_jvm_external/4.4.2/MODULE.bazel": "a56b85e418c83eb1839819f0b515c431010160383306d13ec21959ac412d2fe7",
     "https://bcr.bazel.build/modules/rules_jvm_external/5.1/MODULE.bazel": "33f6f999e03183f7d088c9be518a63467dfd0be94a11d0055fe2d210f89aa909",
     "https://bcr.bazel.build/modules/rules_jvm_external/5.2/MODULE.bazel": "d9351ba35217ad0de03816ef3ed63f89d411349353077348a45348b096615036",
@@ -446,6 +445,24 @@
   },
   "selectedYankedVersions": {},
   "moduleExtensions": {
+    "@@apple_rules_lint+//lint:extensions.bzl%linter": {
+      "general": {
+        "bzlTransitiveDigest": "g7izj5kLCmsajh8IospHh4ZQ35dyM0FIrA8D4HapAsM=",
+        "usagesDigest": "CZ3zTNUXS/3G0ZyuPiHDOU3Fqzwk8N59OOLNI26OIuo=",
+        "recordedFileInputs": {},
+        "recordedDirentsInputs": {},
+        "envVariables": {},
+        "generatedRepoSpecs": {
+          "apple_linters": {
+            "repoRuleId": "@@apple_rules_lint+//lint/private:register_linters.bzl%register_linters",
+            "attributes": {
+              "linters": {}
+            }
+          }
+        },
+        "recordedRepoMappingEntries": []
+      }
+    },
     "@@apple_support+//crosstool:setup.bzl%apple_cc_configure_extension": {
       "general": {
         "bzlTransitiveDigest": "gv4nokEMGNye4Jvoh7Tw0Lzs63zfklj+n4t0UegI7Ms=",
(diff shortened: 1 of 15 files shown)


lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
```diff
@@ -400,16 +400,13 @@ struct LWEToOpenfhe : public impl::LWEToOpenfheBase<LWEToOpenfhe> {
...
 
         // AddPlain
-        ConvertCiphertextPlaintextOp<bgv::AddPlainOp, openfhe::AddPlainOp>,
-        ConvertCiphertextPlaintextOp<ckks::AddPlainOp, openfhe::AddPlainOp>,
+        ConvertCiphertextPlaintextOp<lwe::RAddOp_Plain, openfhe::AddPlainOp>,
 
         // SubPlain
-        ConvertCiphertextPlaintextOp<bgv::SubPlainOp, openfhe::SubPlainOp>,
-        ConvertCiphertextPlaintextOp<ckks::SubPlainOp, openfhe::SubPlainOp>,
+        ConvertCiphertextPlaintextOp<lwe::RSubOp_Plain, openfhe::SubPlainOp>,
```
j2kun, line 406: You have defined sub_plain in LWE to allow the plaintext in either operand, but IIRC OpenFHE requires it is only in one operand. Could you double-check this, and then maybe this needs a custom rule or canonicalization pattern to ensure the plaintext operand is in the right place?

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -358,6 +358,21 @@ struct ConvertRAdd : public OpConversionPattern<RAddOp> {
...
+  ConvertRAddPlain(mlir::MLIRContext *context)
+      : OpConversionPattern<RAddOp_Plain>(context) {}
+
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      RAddOp_Plain op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOpWithNewOp<::mlir::heir::polynomial::AddOp>(
+        op, adaptor.getOperands()[0], adaptor.getOperands()[1]);
```
j2kun, line 371: not sure why the other patterns use `getOperands()[0]` in this file, but the `OpAdaptor` should have named methods corresponding to the tablegen names (`getLhs()`, `getRhs()`) that you can use here.

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -467,6 +497,52 @@ struct ConvertRMul : public OpConversionPattern<RMulOp> {
...
+  // verify num elements, verify order of ciphertext-plaintext
+  LogicalResult matchAndRewrite(
+      RMulOp_Plain op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    auto x = adaptor.getLhs();
+    auto xT = cast<RankedTensorType>(x.getType());
+    auto y = adaptor.getRhs();
+    auto yT = cast<RankedTensorType>(y.getType());
+
+    if (xT.getNumElements() != 2 || yT.getNumElements() != 1) {
```
j2kun, line 515: If the plaintext op can be in either operand, you may want to add a step before this that checks the types of `op.getLhs()` and `op.getRhs()` and then swaps `x` and `y` appropriately.

Or we could:

- make the tablegen force the LHS to be ciphertext and RHS to be plaintext
- add a general canonicalization rule moves the plaintext to the RHS. 

The lowerings to LWE may already ensure this, but new lowerings to LWE may not be aware of this requirement.

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -467,6 +497,52 @@ struct ConvertRMul : public OpConversionPattern<RMulOp> {
...
+    auto x1 =
+        b.create<tensor::ExtractOp>(xT.getElementType(), x, ValueRange{i1});
+
+    auto y0 =
+        b.create<tensor::ExtractOp>(yT.getElementType(), y, ValueRange{i0});
+
+    auto z0 = b.create<::mlir::heir::polynomial::MulOp>(x0, y0);
+    auto z1 = b.create<::mlir::heir::polynomial::MulOp>(x1, y0);
+
+    auto z = b.create<tensor::FromElementsOp>(ArrayRef<Value>({z0, z1}));
```
j2kun, line 539: `polynomial::MulOp` is supposed to also allow tensor-typed inputs and apply mul elementwise. If you weren't aware of that, you could try simplifying this pattern by doing

```
auto repeated = b.create<tensor::FromElementsOp>({y, y});
auto prod  = b.create<polynomial::MulOp>(x, repeated);
```

j2kun (commented):
Generally looks good! I have some suggested simplifications and then I would leave time for @ZenithalHourlyRate or @AlexanderViand to comment before final approval.

AlexanderViand (changes requested):
As Jeremy already pointed out, some backends restrict which side of the op the plaintext can appear on. This is most annoying for `sub_plain`. I think @ZenithalHourlyRate knows more on what, if any, restrictions there really are, though.

Even if it turns out not to be an issue, I think all the tests, including the various `<scheme>_to_<backend>.mlir` deserve to have a pt-lhs and a pt-rhs version of each op, just to make sure this is explicitly tested for.

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
j2kun: in an OpConversionPattern `OpAdaptor` is defined as `SourceOp::Adaptor` where `SourceOp` is the template argument (`RAddOp_Plain` here). Cf. https://github.com/llvm/llvm-project/blob/5dc3cd0ee40c00d9fb542488fa5a54ff70273112/mlir/include/mlir/Transforms/DialectConversion.h#L586

MODULE.bazel.lock
```diff
@@ -1610,7 +1610,7 @@
     },
     "@@rules_python+//python/extensions:pip.bzl%pip": {
       "general": {
-        "bzlTransitiveDigest": "I2tEsg/D0/3B4CIPYr4NzeUfrj2ODZtxU4zM8Z7TzJU=",
+        "bzlTransitiveDigest": "vKDNEc7MJLQpnsEBRqgLsrNo+zskN5XVJzAg1j5LtPs=",
```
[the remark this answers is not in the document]
j2kun: I think if you rebase this will go away. It is fine to commit changes to the lockfile though. We will eventually set up something that does this periodically so it doesn't show up in random PRs.

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
AlexanderViand: Oh, I see - I guess I was thinking of the (now removed) OneToNConversionPattern case, where it's really just a GenericAdaptor... that *does* raise the question (unrelated to this PR) of why all the patterns here use the same style 🙈

lib/Dialect/LWE/IR/LWEOps.td
```diff
@@ -159,13 +165,21 @@ def LWE_RAddOp : LWE_BinOp<"radd", [SameOperandsAndResultRings, SameOperandsAndR
   }];
 }
 
+def LWE_RAddOp_Plain : LWE_CiphertextPlaintextOp<"radd_plain", [Commutative, SameOperandsAndResultPlaintextTypes, InferTypeOpAdaptor]> {
```
ZenithalHourlyRate, line 168:
```suggestion
def LWE_RAddPlainOp : LWE_CiphertextPlaintextOp<"radd_plain", [Commutative, SameOperandsAndResultPlaintextTypes, InferTypeOpAdaptor]> {
```

In HEIR we prefer a name like `LWE_RAddPlainOp` so the cpp op name will become `lwe::RAddPlainOp` instead of `lwe::RAddOp_Plain`. I do not think we have op name that has underscore in it. A mnemonic with underscore is accepted.

ZenithalHourlyRate: Also, please add canonicalizer on ct-pt ordering to LWE_RAddPlainOp and LWE_RMulPlainOp, it is just copy-pasting what 948dd4f66 has done in BGVOps.td and BGVOps.cpp.

lib/Dialect/LWE/Conversions/LWEToOpenfhe/LWEToOpenfhe.cpp
ZenithalHourlyRate: BGV_SubPlainOp also allows pt in either operand, openfhe has two overloads for `cc->EvalSub` that accepts both of them. It is Lattigo that only accepts `evaluator.Sub(ct, pt)`, but that reordering is handled in `LWEToLattigo` by `Add(Negate(ct), pt)`, where Negate is only present in `lattigo` dialect, so it should not be a concern for `lwe` dialect.

lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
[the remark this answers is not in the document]
ZenithalHourlyRate: ConvertRlweCommutativePlainOp and ConvertRlweSubPlainOp have already taken care of it so it should not be a concern for at `lwe` dialect level.

ZenithalHourlyRate (changes requested):
Generally LGTM! Just some op naming convention suggestion.

About the ct-pt ordering problem, openfhe/lattigo lowerings for lwe already handle them so it should not be an issue for lwe dialect itself. I think the polynomial lowering should be aware of it instead of enforcing lwe dialect to have some constraint.

ZenithalHourlyRate:
> For some reason when I was adding bgv/ckks-to-openfhe tests I continued to fail them even though my test log showed the expected output. When I added -dump-input=help as a FileCheck flag they passed. Has anyone else ever seen this behavior?

No, it should mean the CHECK directive you wrote does not match the input.

Also, according to https://llvm.org/docs/CommandGuide/FileCheck.html#cmdoption-FileCheck-dump-input, `-dump-input=help` means `Explain input dump and quit`, so it should not fail. And bazel test asserts PASSED or not by checking the return value of executable, so you will always get a PASSED with such flag.

johnbntn:
I think I added all of the requested changes:
- Removed underscore in op names
- Added canonicalizers for add/mul plain
- Simplified the poly lowering
- Fixed the openfhe lowering tests by pattern matching just the plaintext instead of the plaintext and return types
- Added some more tests for orderings and canonicalizers

Please let me know if I missed anything or if you have any more requested changes

ZenithalHourlyRate (approved):
LGTM!

There are still some comments not addressed

https://github.com/google/heir/pull/1776#discussion_r2069310687

on usage of adaptor

https://github.com/google/heir/pull/1776#discussion_r2069593135

on `rlwe_ops.mlir` containing IR syntax test for pt-ct. I know there are some tests on them on the lowering part (e.g. bgv-to-openfhe), but the IR syntax test itself should also reflect that.

ZenithalHourlyRate:
Also remember to squash the commits so that the PR merge tool can pick it up.

j2kun:
> I believe I finished making all of the requested changes but please let me know if I missed anything.

Due to our strange project setup in re Google, you also need to squash your commits before we can merge it.

Issue #1789, opened 2025-05-05
[frontend] support type hints in python code


asraa:
I had some code for the AES transciphering demo - I know it's on my table to write a redux of the full thing, but I wanted to post this issue so i can work on it.

There's an efficient sbox lookup implementation by bayer-poralta that uses bitwise operators. Here's an implementation; https://www.bearssl.org/gitweb/?p=BearSSL;a=blob;f=src/symcipher/aes_ct.c;h=66776d9e206c92cbbf3fa799c651a3d3652bd75a;hb=HEAD#l29

I could almost copy and paste the body (the bit assignments were more complicated), but the annoying part was that all the arithmetic ended up in i64 in the frontend. Could be pretty easily solvable if I could give a type hint to the x0, ... , x7's as i1s. Something like

```
def sub_bytes(x: Secret[I8]):
    x0: I1 = (x >> 7) & 1
    x1: I1 = (x >> 6) & 1
    x2: I1 = (x >> 5) & 1
    x3: I1 = (x >> 4) & 1
...
```

AlexanderViand:
Mh, that's odd - independently of the `I1` thing,  I thought I had overridden the constant behavior in numba to where it'd set integer constants to the smallest size in [`i8`,`i16`,`i32`,`i64`] that they'd fit into, so I'd have expected to see the constant be an `i8`, but it's an ` %0 = arith.constant 7 : i32` ❓

AlexanderViand:
Found the bug, I'm an absolute idiot, PR incoming...

Concerning the actual type-hint/issue: what do you want the `I1` type hint to do here? It seems like you want `I8` (which my fix would give you) but if you actually want this to be cast down to Bool, I don't think a type hint would be sufficient for that, we'd need something like `x0 = cast( (x >> 7) & 1, I1)`

AlexanderViand:
From the PR:

@asraa:
> I'm extracting the 8th bit of an i8 here, and want to truncate that to an i1 MLIR type so that the following bitwise arithmetic would happen under i1. Instead the python frontend that thinks x is i8 will keep all arithmetic in i8 (and then sometimes to i64 which is probably another bug)

Oh, I see - I'm not sure type hints are the right way to go about this (a custom numba intrinsic `cast(<value>,<type>)` might be easier), but I'll see what I can do!

asraa:
[...]

PR #1794, opened 2025-05-06
frontend: fix incorrect type inference for constants

AlexanderViand:
🤦‍♂️ 

Let me know if this already "fixes" #1789 for you, or if you actually need a way to override the automated type inference with a type hint.


asraa (approved):
wonderful! haha, honestly it took me a second to realize typ wasn't defined earlier and why this fixed anything.

also - does not totally fix everything, but does improve the situation for constants! the issue from the PR is mostly when we have some python code like

```
x0 = (x >> 7) & 1
```

I'm extracting the 8th bit of an i8 here, and want to truncate that to an i1 MLIR type so that the following bitwise arithmetic would happen under i1. Instead the python frontend that thinks x is i8 will keep all arithmetic in i8 (and then sometimes to i64 which is probably another bug)

Issue #1782, opened 2025-05-02
Investigate "Accurate BGV Parameters Selection: Accounting for Secret and Public Key Dependencies in Average-Case Analysis"


AlexanderViand:
**Accurate BGV Parameters Selection: Accounting for Secret and Public Key Dependencies in Average-Case Analysis**
_Beatrice Biasioli, Chiara Marcolla, Nadir Murru, Matilda Urani_
> The Brakerski-Gentry-Vaikuntanathan (BGV) scheme is one of the most significant fully homomorphic encryption (FHE) schemes. It belongs to a class of FHE schemes whose security is based on the presumed intractability of the Learning with Errors (LWE) problem and its ring variant (RLWE). Such schemes deal with a quantity, called noise, which increases each time a homomorphic operation is performed. Specifically, in order for the scheme to work properly, it is essential that the noise remains below a certain threshold throughout the process. For BGV, this threshold strictly depends on the ciphertext modulus, which is one of the initial parameters whose selection heavily affects both the efficiency and security of the scheme. In this paper, we provide a new method to estimate noise growth, closely aligning with experimental results and forming the basis for parameter selection that ensures correctness and improves efficiency.

https://arxiv.org/abs/2504.18597

Upon skimming it, I wonder to what the claim of being implementation-independent really means here?
"In this paper, we propose the first average-case noise analysis for BGV that does not provide underestimates, taking in account the dependencies introduced by the common secret and public key. [...] A related average-case analysis for the BGV scheme is presented in [Optimisations and Tradeoffs for HElib], where the authors develop a noise estimation method tailored to the specific implemen-tation of BGV in HElib [..]. In contrast, our work proposes a general analysis that does not depend on the specific library and instead focuses on capturing the structural dependencies among the errors."

ZenithalHourlyRate:
> A related average-case analysis for the BGV scheme is presented in [Optimisations and Tradeoffs for HElib], where the authors develop a noise estimation method tailored to the specific implemen-tation of BGV in HElib [..].

For circuit Enc -> Mult -> Ms -> (Mult -> Ms -> ...)

In this paper the author provides a bound for the second and later mult and modulus switching, as the behavior of the noise is good after the first modulus switching. This is a common pattern for BGV in HElib, so they claim it is implementation-specific.

In the view of 2504.18597, the first Ms would reduce the contribution of `e` term (common error in publickey) negligible so the former method works, however for general circuit this should not work, so it is not implementation-specific, but the limited scope of former analysis itself.

PR #1807, opened 2025-05-12
NoiseAnalysis: track log-scale noise for deep circuit

ZenithalHourlyRate:
For deeper circuit (e.g. L=10 or L=20), the noise bound could be 500 or 1000 in log scale (so the variance could be 1000 and 2000 in log scale), which a `double` could not hold.

This PR rewrites the `NoiseState` so that is has `logValue` as its storage but its `operator+` and `operator*` still pretends it is the normal addition and multiplication, so that programmer could manipulate it easier.

Also it also leaves space for #1782, which also need `NoiseWithDegree` so the degree field is merged into `NoiseState`.

I've compared noise analysis traces of previous version and this version and they matches. For variance based method, I set alpha from 0.001 to 2^-32 to make failing probability smaller, though this parameter alone is not enough for making noise analysis correct.


lib/Analysis/NoiseAnalysis/Noise.h
```diff
@@ -40,13 +41,25 @@ class NoiseState {
   static NoiseState of(double value) {
```
j2kun, line 41: Is there a reason to keep `NoiseState::of` anymore? If not, it may be worth marking it as deprecated so the compiler can point out where to migrate any remaining uses. If so, we should figure out a way to remove the need for it, as having a class with multiple semantics for its underlying storage is risky.

j2kun (approved):
LGTM!

lib/Analysis/NoiseAnalysis/Noise.h
ZenithalHourlyRate: Removed the original `NoiseState::of` with normal scale semantic and renamed `NoiseState::ofLog2` with log scale semantic to `::of`. Now the user should always expect to manipulate it in normal scale while its storage is in log scale. The only time the user becomes aware it is log-scale is when they call `getValue`, which is now carefully handled with comments saying it is in log-scale.

j2kun (approved):

PR #1816, opened 2025-05-15
NoiseAnalysis: Refactor noise to string

ZenithalHourlyRate:
The old API `toLogXXXString` is quite redundant given there are now so many noise models in HEIR, this PR refactors them out.

### DIFF
--- a/lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
+++ b/lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
@@ -10,6 +10,7 @@
 #include "lib/Dialect/Mgmt/IR/MgmtOps.h"
 #include "lib/Dialect/Secret/IR/SecretOps.h"
 #include "lib/Dialect/TensorExt/IR/TensorExtOps.h"
+#include "lib/Utils/Utils.h"
 #include "llvm/include/llvm/ADT/TypeSwitch.h"             // from @llvm-project
 #include "llvm/include/llvm/Support/Debug.h"              // from @llvm-project
 #include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"     // from @llvm-project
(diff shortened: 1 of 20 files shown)


lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
```diff
@@ -55,10 +55,15 @@ LogicalResult NoiseAnalysis<NoiseModel>::visitOperation(
...
-               << " to " << value << "\n");
+    [[maybe_unused]] auto doubleToString = [](double d) {
+      std::stringstream stream;
+      stream << std::fixed << std::setprecision(2) << d;
+      return stream.str();
+    };
+    LLVM_DEBUG(llvm::dbgs() << "Propagating "
+                            << doubleToString(noiseModel.toLogBound(
+                                   getLocalParam(value), noise))
+                            << " to " << value << "\n");
```
j2kun, line 66: Rather than define a `maybe_unused` function, put the function in a utils class somewhere (or in Noise.h?) to avoid redefining it where you use it. Or just use the block form of `LLVM_DEBUG` to skip the function definition altogether.

```suggestion
    LLVM_DEBUG({
      std::stringstream stream;
      stream << std::fixed << std::setprecision(2)
             << noiseModel.toLogBound(
                  getLocalParam(value), noise);
      llvm::dbgs() << "Propagating "
                            << stream.str()
                            << " to " << value << "\n";
    });
```

j2kun (changes requested):

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
ZenithalHourlyRate: Turned out `RLWEParams.cpp` also needs this function and I put it in `lib/Utils/Utils.h`

j2kun (approved):

PR #1817, opened 2025-05-15
BGV: Symbolic Noise Analysis for Dependency Tracking

ZenithalHourlyRate:
### Status

This PR is intended for the symbolic part `Symbolic.h/Symbolic.cpp`, as this part is big enough.

Also it should be rebased after #1816 is in.

The BGV adoption of it is not comprehensive as it only supports addition/multiplication but it is enough to demonstrate along with (#1782, which would be another PR). Its complete version supporting modulus switching/key switching needs another PR regarding `SelectVariableNames`. Its extension to BFV is already in a branch (also need `SelectVariableNames`). CKKS noise analysis could also benefit from it but lets wait until all the TODOs in #1685.

### Backgrounds

Recent attacks like [GNSJ24](https://www.usenix.org/conference/usenixsecurity24/presentation/guo-qian), [CCP+24](eprint.iacr.org/2024/127), [CSBB24](eprint.iacr.org/2024/116) demonstrated that the ability for noise analysis to handle dependent ciphertext is crucial for security.

These attacks exploited the fact that dependency among ciphertext will lead to faster noise growth, and former average-case analysis is not so able to handle them (strong independency assumption).

This approach could handle it but there are some issues. One obvious issue is that is could not scale-up as the number of symbolic terms grows exponentially with the number of multiplication (though there are some optimizations).

HEIR formerly has a pass called `openfhe-count-add-and-key-switch` in #1254 that could in spirit handle those attacks but it is standalone and not connected to the NoiseAnalysis inside HEIR.

#### Multiplication of Independent ciphertext

```mlir
func.func @mul(%arg0 : i16 {secret.secret}, %arg1 : i16 {secret.secret}) -> i16 {
  %1 = arith.muli %arg0, %arg1 : i16
  return %1 : i16
}
```

MP24 model

```
Propagating 28.05 to <block argument> of type 'i16' at index: 0
Propagating 28.05 to <block argument> of type 'i16' at index: 1
Propagating 58.94 to %1 = arith.muli %input0, %input1
Propagating 58.94 to %2 = mgmt.relinearize %1
Propagating 24.08 to %3 = mgmt.modreduce %2
```

Symbol model: able to handle common dependency like secret key and error in public key (namely #1782)

```
Propagating 28.05 ... block arg
Propagating 28.05 ... block arg
Propagating 59.23 ... muli
```

#### Multiplication of same ciphertext

```mlir
func.func @mul(%arg0 : i16 {secret.secret}) -> i16 {
[...]

PR #1820, opened 2025-05-16
SelectVariableNames: Make a Wrapper for DataflowAnalysis

ZenithalHourlyRate:
NoiseAnalysis (and also other analysis) may want to get unique name of a `Value`, e.g. for symbolic way of analysis (see #1817)

I do not know if this way of providing name for analysis is favorable (so requesting opinions from @j2kun before more technical details). I used to have another `NameAnalysis` so that NoiseAnalysis could access `NameLattice`, turned out that is a duplicate of `SelectVariableNames` so I modified this analysis to have a dataflow framework wrapper.

Notice that other ways of solving it include

* Pass `SelectVariableName` struct to every analysis, which is inconvenient if we get more analysis. Passing it to the DataFlowSolver might be more convenient.
* Pass it in `LocalParam`, which is essentially constructed from the IR, which requires annotating the IR with HEIR generated name (quite strange)

<summary>solved now</summary>
<details>
It was me that used reference for destroyed struct. 
</details>

Other problem involves loop-carried variable which has the same name but does not have the same analysis state, which we (and also MLIR) are quite unable to handle. One way is to force the invariant for these variables via bootstrapping, but then we need to think about what "reset the Symbolic State" means for bootstrapping inside a loop.

### DIFF
--- a/lib/Analysis/SelectVariableNames/SelectVariableNames.h
+++ b/lib/Analysis/SelectVariableNames/SelectVariableNames.h
@@ -4,14 +4,53 @@
 #include <cassert>
 #include <string>
 
-#include "llvm/include/llvm/ADT/DenseMap.h"          // from @llvm-project
-#include "mlir/include/mlir/IR/BuiltinAttributes.h"  // from @llvm-project
-#include "mlir/include/mlir/IR/Operation.h"          // from @llvm-project
-#include "mlir/include/mlir/IR/Value.h"              // from @llvm-project
-#include "mlir/include/mlir/Support/LLVM.h"          // from @llvm-project
+#include "llvm/include/llvm/ADT/DenseMap.h"  // from @llvm-project
+#include "mlir/include/mlir/Analysis/DataFlow/SparseAnalysis.h"  // from @llvm-project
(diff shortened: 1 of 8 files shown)


lib/Analysis/SelectVariableNames/SelectVariableNames.h
```diff
@@ -4,15 +4,54 @@
...
+  }
+
+  // For Lattice in DataFlowFramework
+  static VariableName join(const VariableName &lhs, const VariableName &rhs) {
+    if (!lhs.isInitialized() || lhs.getName().empty()) return rhs;
+    if (!rhs.isInitialized() || rhs.getName().empty()) return lhs;
+
+    if (lhs.getName() == rhs.getName()) return lhs;
+
+    // If the names are different, we can just return the first one
```
j2kun, line 37: I'm guessing this line never actually executes, if the transfer function below is a no-op... Maybe it would be worth adding an assertion here? Otherwise, I would question the use of this analysis that trivially ignores half the information.

j2kun (changes requested):
Personally I think that because the variable name selection is purely syntactic (evidenced by a no-op transfer function), it doesn't need a dataflow analysis and should just be passed as an extra argument to the analysis constructor.

PR #1783, opened 2025-05-02
frontend: add cggi scheme support up to cggi dialect

asraa:
no backend is supported right now, so the cggi_test will fail at:

In the future, we should decide how to pass options to the compiler - I would actually just like to wait until the backend is selected and decide based on that. 
  - e.g. for tfhe-rs, use lookup tables
  - for binFHE, use boolean gates

```
asraa@asraali:~/git/heir/frontend$ python -m cggi_test
Running tests under Python 3.12.9: /usr/local/google/home/asraa/git/heir/frontend/venv/bin/python
[ RUN      ] EndToEndTest.test_simple_arithmetic
HEIR Debug: Writing raw input MLIR to /tmp/tmpbwe19k0o/foo.raw.mlir
HEIR Debug: Writing input MLIR w/ shape inference to /tmp/tmpbwe19k0o/foo.in.mlir
HEIR Debug: Writing secretness-annotated MLIR to /tmp/tmpbwe19k0o/foo.annotated.mlir
HEIR Debug: Writing secretness-annotated graph to /tmp/tmpbwe19k0o/foo.annotated.dot
HEIR Debug: Running heir-opt --canonicalize --mlir-to-cggi --mlir-print-debuginfo --view-op-graph
HEIR Debug: Writing output MLIR to /tmp/tmpbwe19k0o/foo.out.mlir
HEIR Debug: Writing output graph to /tmp/tmpbwe19k0o/foo.out.dot
HEIR Debug: Leaving workspace_dir /tmp/tmpbwe19k0o for manual inspection.

[  FAILED  ] EndToEndTest.test_simple_arithmetic
======================================================================
ERROR: test_simple_arithmetic (__main__.EndToEndTest.test_simple_arithmetic)
EndToEndTest.test_simple_arithmetic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/asraa/git/heir/frontend/cggi_test.py", line 10, in test_simple_arithmetic
    @compile(
     ^^^^^^^^
  File "/usr/local/google/home/asraa/git/heir/frontend/heir/pipeline.py", line 268, in decorator
    return run_pipeline(
           ^^^^^^^^^^^^^
  File "/usr/local/google/home/asraa/git/heir/frontend/heir/pipeline.py", line 194, in run_pipeline
    raise NotImplementedError(f"Backend compilation is unsupported for CGGI scheme, check CGGI output at {mlirpath}")
[...]

PR #1825, opened 2025-05-16
frontend: support "casting" between MLIR types

AlexanderViand:
Tries to address the use case presented in #1789 though it does so not via type hints but via adding support for casts:

```llvm
def sub_bytes(x: Secret[I8]):
    x0 = I1((x >> 7) & 1)
    x1 = I1((x >> 6) & 1)
    x2 = I1((x >> 5) & 1)
    x3 = I1((x >> 4) & 1)
    ...
```

Still WIP:
* There are some open questions around what the expected behavior should be for arithmetic FHE (see discussion in the issue)
* If we want to only support it for CGGI for now, this might depend on #1783 (not sure if that's 100% necessary, though)
* There should probably be some tests once we determined what exactly it should be doing/not doing in the C++ side of things.
* For practical reasons, this also depends on #1824

### DIFF
--- a/frontend/cggi_test.py
+++ b/frontend/cggi_test.py
@@ -1,14 +1,16 @@
 from heir import compile
 from heir.mlir import I8, Secret
+from heir.backends.cleartext import CleartextBackend
 
 
 from absl.testing import absltest  # fmt: skip
 class EndToEndTest(absltest.TestCase):
 
-  def test_simple_arithmetic(self):
+  def test_simple_cggi_arithmetic(self):
 
     @compile(
         scheme="cggi",
+        backend=CleartextBackend(),
         debug="True",
     )
     def foo(a: Secret[I8], b: Secret[I8]):
--- a/frontend/heir/mlir_emitter.py
+++ b/frontend/heir/mlir_emitter.py
@@ -12,7 +12,8 @@ from numba.core import bytecode
 from numba.core import controlflow
 from numba.core.types import Type as NumbaType
 
-from heir.interfaces import InternalCompilerError
+from heir.mlir.types import MLIRType, MLIR_TYPES, I1, I8, I16, I32, I64, F32, F64
+from heir.interfaces import CompilerError, DebugMessage, InternalCompilerError
 
 
 def mlirType(numba_type: NumbaType) -> str:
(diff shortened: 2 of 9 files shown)


AlexanderViand:
@j2kun / @asraa: I see the cggi-frontend tests fail when doing `bazel test //frontend:cggi_test` but they work fine when running directly (`python frontend/cggi_test.py`). I'd guess this isn't related to this PR but to running CGGI from python under bazel more generally, since it happens even with the standard CGGI test (which wasn't enabled in bazel before)...

The error message I get when running with bazel is
```
HEIR Error at /home/sdp/.cache/bazel/_bazel_sdp/e1a6590193554bff6e6c8ff665423111/sandbox/processwrapper-sandbox/10673/execroot/_main/bazel-out/k8-dbg/bin/frontend/cggi_test.runfiles/_main/frontend/cggi_test.py (11): HEIR compilation for function foo failed:
running `heir-opt --canonicalize --mlir-to-cggi --mlir-print-debuginfo --view-op-graph` produced these errors:
ERROR: Can't open input file `/home/sdp/.cache/bazel/_bazel_sdp/e1a6590193554bff6e6c8ff665423111/sandbox/processwrapper-sandbox/10673/execroot/_main/bazel-out/k8-dbg/bin/frontend/cggi_test.runfiles/_main/techmap.v' for reading: No such file or directory
```

j2kun:
> @j2kun / @asraa: I see the cggi-frontend tests fail when doing `bazel test //frontend:cggi_test` but they work fine when running directly (`python frontend/cggi_test.py`). I'd guess this isn't related to this PR but to running CGGI from python under bazel more generally, since it happens even with the standard CGGI test (which wasn't enabled in bazel before)...
> 
> The error message I get when running with bazel is
> 
> ```
> HEIR Error at /home/sdp/.cache/bazel/_bazel_sdp/e1a6590193554bff6e6c8ff665423111/sandbox/processwrapper-sandbox/10673/execroot/_main/bazel-out/k8-dbg/bin/frontend/cggi_test.runfiles/_main/frontend/cggi_test.py (11): HEIR compilation for function foo failed:
> running `heir-opt --canonicalize --mlir-to-cggi --mlir-print-debuginfo --view-op-graph` produced these errors:
> ERROR: Can't open input file `/home/sdp/.cache/bazel/_bazel_sdp/e1a6590193554bff6e6c8ff665423111/sandbox/processwrapper-sandbox/10673/execroot/_main/bazel-out/k8-dbg/bin/frontend/cggi_test.runfiles/_main/techmap.v' for reading: No such file or directory
> ```

The techmap error looks like it's an issue with your build file. You'd need something like `data = ["@heir//lib/Transforms/YosysOptimizer/yosys:techmap.v"]`, but even then it should be looking for this file in `HEIR_YOSYS_SCRIPTS_DIR` which may need to be configured properly (it's in `heir_cli` and `tools/BUILD`...).

Ignoring that, it is true that we can't enable CGGI tests right now because we don't have a frontend-compatible backend for it. HEIR doesn't have a CGGI backend via OpenFHE's binfhe yet, and the frontend doesn't support rust for `tfhe-rs` yet (nor Python for `jaxite`). @raghav198 has a draft from his paper in https://github.com/raghav198/heir/tree/cggi-to-openfhe-draft, but I doubt we need full CGGI support in the frontend for this PR. Can it wait?

AlexanderViand:
> Ignoring that, it is true that we can't enable CGGI tests right now because we don't have a frontend-compatible backend for it.

Yep, but this is using the CleartextBackend (so just testing down to `cggi` dialect) and works fine when being called outside the bazel environment, so these tests should work even w/o CGGI backend support.

I'll look into the BUILD file thing you mentioned!
EDIT: Looks like it was indeed just missing the `HEIR_YOSYS_SCRIPTS_DIR` (and `HEIR_ABC_BINARY`) paths in `testing.bzl`, thanks for the hint!

frontend/cggi_test.py
```diff
@@ -1,14 +1,16 @@
 from heir import compile
-from heir.mlir import I8, Secret
+from heir.mlir import I1, I8, Secret
```
j2kun:
```suggestion
from heir.mlir import I8, Secret
```

frontend/heir/mlir_emitter.py
```diff
@@ -36,6 +37,72 @@ def mlirType(numba_type: NumbaType) -> str:
...
+def isFloatLike(typ: NumbaType | MLIRType) -> bool:
+  if isinstance(typ, type) and issubclass(type, MLIRType):
+    return typ in {F32, F64}
+  if isinstance(typ, NumbaType):
+    return isinstance(typ, types.Float)
+  raise InternalCompilerError(f"Encountered unexpected type {type}")
+
+
+# Needed because, e.g. Boolean doesn't have a bitwidth
+def getBitwidth(typ: NumbaType | MLIRType) -> int:
```
j2kun, line 57: nit: the following may be more readable, and note the fix `type(type)` to `type(typ)`

```python
def getBitwidth(typ: NumbaType | MLIRType) -> int:
    match typ:
        case types.Integer(bitwidth=bw):
            return bw
        case types.Boolean():
            return 1
        case I1 | I8 | I16 | I32 | I64:
            # e.g., <class '...I32'>.__name__ -> "I32"
            return int(typ.__name__[1:])
        case _:
            raise InternalCompilerError(f"Unexpected type {typ} ({type(typ)})")
```

Same for `is*Like` (also note line 53 uses `type` instead of `typ`)

frontend/heir/mlir_emitter.py
```diff
@@ -446,7 +540,10 @@ def emit_assign(self, assign):
           self.forward_name_to_id(assign.target, name.strip("%"))
         return const_str
       case ir.Global():
-        self.globals_map[assign.target.name] = assign.value.name
+        self.globals_map[assign.target.name] = (
+            assign.value.name,
+            assign.value.value,
```
j2kun, line 549: Please add a comment at the definition

```
self.globals_map = {}
```

To explain what is in this map. Now that it's more than string->string it's not clear what `assign.value.value` actually refers to. What is the type of `global_` when the name is "bool"?

AlexanderViand:
Tests pass w/ bazel now, so should be good to review :)

j2kun (approved):
LGTM, a few nits

frontend/heir/mlir_emitter.py
AlexanderViand: Unfortunately, this doesn't seem to work, since Python thinks the `case I1 | ... ` is just an always-matching variable capture to a variable named `I1` 😕

PR #1827, opened 2025-05-18
Simplify some lit tests

j2kun:
This PR was extracted from https://github.com/google/heir/pull/1633.

As I replaced the old add-client-interface with the new pass, some tests (such as noise analysis passes) were unable to handle the existing test inputs due to new client helper functions being present at a different level of the hierarchy. This does not affect the end to end tests, because these tests were picking subsets of the full pipeline, while starting from the entry point of the compiler (secret-annotated arith) instead of just testing the individual pass. So this PR either:

- Replaces the test command to just exercise the individual pass,
- Replaces the test command to run the full pipeline
- Replaces the test input with the result of running the pipeline until just before the pass being tested, and then runs just the one pass under test


ZenithalHourlyRate (approved):
I like the idea of testing individual pass. Though I am bit concerned how much of these _intermediate_ IR would change when the packing pass is in. Maybe we could maintain a original copy of it in comments so later migration could be smoother, or just let them find it in git history.

asraa (approved):

j2kun:
> I like the idea of testing individual pass. Though I am bit concerned how much of these _intermediate_ IR would change when the packing pass is in. Maybe we could maintain a original copy of it in comments so later migration could be smoother, or just let them find it in git history.

FWIW, the tests in this PR were unchanged by the packing pass, except that some still used `secret<scalar>` instead of tensor (though it was not material to the test itself)

PR #1830, opened 2025-05-18
annotate enc/dec funcs for cross-pass matching

j2kun:
This allows one to use the annotation on the encryption func to identify the original function the cilent helper is for, as well as the argument/result index.

I needed this in https://github.com/google/heir/pull/1633/files because I need a way to copy mgmt attributes from the main function to the client helpers before lowering secret to scheme.

### DIFF
--- a/lib/Dialect/ModuleAttributes.h
+++ b/lib/Dialect/ModuleAttributes.h
@@ -48,6 +48,25 @@ void moduleClearBackend(Operation *moduleOp);
 void moduleSetOpenfhe(Operation *moduleOp);
 void moduleSetLattigo(Operation *moduleOp);
 
+// Func attributes for client helpers
+//
+// This corresponds to a named attribute client.enc_func whose
+// value is a dictionary {func_name = "foo", index = 2 : i64}
+//
+// This means that the function with this attribute is an encryption
+// helper for the function "foo" and the argument at index 2.
+
+constexpr const static ::llvm::StringLiteral kClientEncFuncAttrName =
+    "client.enc_func";
+constexpr const static ::llvm::StringLiteral kClientDecFuncAttrName =
+    "client.dec_func";
+
+// The name of the function this client helper is made for.
+constexpr const static ::llvm::StringLiteral kClientHelperFuncName =
+    "func_name";
+// The argument or operand index the client helper function is for.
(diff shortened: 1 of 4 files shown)


lib/Dialect/ModuleAttributes.h
```diff
@@ -48,6 +48,19 @@ void moduleClearBackend(Operation *moduleOp);
...
+    "client_enc_func";
+constexpr const static ::llvm::StringLiteral kClientDecFuncAttrName =
+    "client_dec_func";
+
+// The name of the function this client helper is made for.
+constexpr const static ::llvm::StringLiteral kClientHelperFuncName =
+    "func_name";
+// The argument or operand index the client helper function is for.
+constexpr const static ::llvm::StringLiteral kClientHelperIndex = "index";
+
```
ZenithalHourlyRate, line 63: Bit surprised that funcOp itself allows attr name with no `.` in it. Still, maybe we can get separate it into namespaces like `client.enc_func` so other passes can tell whether this attr is their concern?

As for `func_name` and `index`, better describe that they are inside the dictionAttr like an example here

```mlir
client.enc_func = {func_name = "name", index = 0 : i64}
```

ZenithalHourlyRate (approved):

lib/Dialect/ModuleAttributes.h
j2kun: Done

lib/Transforms/AddClientInterface/AddClientInterface.cpp
```diff
@@ -45,46 +49,58 @@ using tensor_ext::UnpackOp;
...
+                                     BlockArgument funcArgument,
                                      ImplicitLocOpBuilder &builder,
                                      int64_t ciphertextSize) {
-  SmallVector<Type> funcArgTypes(argTypes.begin(), argTypes.end());
+  std::string encFuncName("");
+  llvm::raw_string_ostream encNameOs(encFuncName);
+  encNameOs << op.getSymName() << "__encrypt__arg"
+            << funcArgument.getArgNumber();
+  auto originalTypeAttr = op.getArgAttrOfType<OriginalTypeAttr>(
+      funcArgument.getArgNumber(), kOriginalTypeAttrName);
```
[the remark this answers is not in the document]
j2kun: Done

PR #1837, opened 2025-05-20
Copy mgmt attr to client helpers

j2kun:
Copies the mgmt attr from the main compiled function to the client helpers. This is required to lower them from secret to scheme in https://github.com/google/heir/pull/1633

The main idea is to copy the mgmt attrs to the function arg attrs and result attrs, and then rely on the attribute propagation utilities to populate the client helper function bodies with mgmt attrs. This last step is necessary for two reasons:

- secret-to-scheme requires mgmt attrs for type conversion, and at least, the last op in the client encryption step (secret.conceal) produces a secret type that needs to be lowered
- the various analysis passes work on the entire IR and will fail if mgmt attrs are not found. It's simpler to ensure the mgmt attrs are consistent across the entire IR than to have the analyses have special cases to know about client helpers.


asraa (approved):
thanks!

PR #1843, opened 2025-05-21
implement RegionBranchOpInterface for secret.generic

j2kun:
This PR implements RegionBranchOpInterface for `secret.generic`. Though `generic` does not have any "real" branching (single-directional control flow), this interface is the easiest way to enhance the dataflow analysis passes, so that the lattice operands of a `secret.yield` op are propagated to the results of the enclosing `secret.generic` op.

Note, the part of the interface that has `secret.yield` say which operands are forwarded is automatically generated by the `ReturnLike` trait on `secret.yield`.

### DIFF
--- a/lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
+++ b/lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
@@ -28,7 +28,15 @@ namespace heir {
 // explicit specialization of NoiseAnalysis for NoiseByBoundCoeffModel
 template <typename NoiseModel>
 void NoiseAnalysis<NoiseModel>::setToEntryState(LatticeType *lattice) {
-  // At an entry point, we have no information about the noise.
+  if (isa<secret::SecretType>(lattice->getAnchor().getType())) {
+    Value value = lattice->getAnchor();
+    auto localParam = LocalParamType(&schemeParam, getLevelFromMgmtAttr(value),
+                                     getDimensionFromMgmtAttr(value));
+    NoiseState encrypted = noiseModel.evalEncrypt(localParam);
+    this->propagateIfChanged(lattice, lattice->join(encrypted));
+    return;
+  }
+
   this->propagateIfChanged(lattice, lattice->join(NoiseState::uninitialized()));
 }
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -31,7 +31,10 @@ bazel_dep(name = "rules_python", version = "1.2.0")
 bazel_dep(name = "googletest", version = "1.16.0")
 bazel_dep(name = "google_benchmark", version = "1.9.1")
 bazel_dep(name = "abseil-cpp", version = "20250127.0", repo_name = "com_google_absl")
-bazel_dep(name = "eigen", version = "4.0.0-20241125.bcr.1")
+
+# eigen 4.0.0-20241125.bcr.1 has issues with gitlab changing the checksum
+# Cf. https://github.com/google/heir/issues/1840
+bazel_dep(name = "eigen", version = "3.4.0.bcr.3")
 bazel_dep(name = "or-tools", version = "9.12", repo_name = "com_google_ortools")
 
 # needed as transitive deps of yosys
--- a/lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
+++ b/lib/Analysis/AddAndKeySwitchCountAnalysis/AddAndKeySwitchCountAnalysis.cpp
@@ -35,14 +35,6 @@ LogicalResult CountAnalysis::visitOperation(
   };
 
   llvm::TypeSwitch<Operation &>(*op)
-      .Case<secret::GenericOp>([&](auto genericOp) {
-        Block *body = genericOp.getBody();
-        for (auto i = 0; i != body->getNumArguments(); ++i) {
-          auto blockArg = body->getArgument(i);
-          // one Vfresh
-          propagate(blockArg, CountState(1, 0));
-        }
-      })
       .Case<arith::AddIOp, arith::SubIOp, arith::AddFOp, arith::SubFOp>(
           [&](auto &op) {
             // condition on result secretness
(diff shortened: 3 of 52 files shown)


j2kun:
Ok, I wasn't able to separate the implementation of the interface from much of the rest of the analysis pass cleanup I wanted to do in other PRs (so there's a lot of duplication across my open PRs right now).

The main problem is that when you implement this interface for `GenericOp`, the transfer function we defined is no longer called, which that generic block arguments are not instantiated in the way the analysis passes had previously relied on. Instead, it uses the `setToEntryState` along with the transfer defined by the new `RegionBranchOpInterface` implementation.

So as a result, the changes in this PR include:

- The implementation of RegionBranchOpInterface in SecretOps
- `setToEntryState` updates in all analysis passes; they now set non-secret values to have uninitialized lattices (as before), but secret values get set to an appropriate default value. Most of these are obvious, but notably the noise analysis defaults to the noise of a fresh encryption.
- A new `walkValues` helper that replaces the ad-hoc IR traversal required to persist analysis results as attributes. All analysis passes now use this during their `annotateFoo` helpers.
- Convert analysis passes to properly analyze/annotate function arguments, so that DistributeGeneric no longer copies mgmt attributes to the func. This is now guaranteed by the analysis passes themselves. Cf. AnnotateMgmt.cpp which ensures function args/results are annotated properly.
- More consistent handling of when mgmt propagation should happen (cf. `shouldHaveMgmtAttribute`)
- Tests are only updated insofar as they require extra mgmt attributes to be persisted on func arguments, etc.

j2kun:
The secretness performance regression has been fixed by replacing the secretness analysis with a persisted attribute that is read when the loop is processed.

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
```diff
@@ -28,7 +28,15 @@ namespace heir {
...
-  // At an entry point, we have no information about the noise.
+  if (isa<secret::SecretType>(lattice->getAnchor().getType())) {
+    Value value = lattice->getAnchor();
+    auto localParam = LocalParamType(&schemeParam, getLevelFromMgmtAttr(value),
+                                     getDimensionFromMgmtAttr(value));
+    NoiseState encrypted = noiseModel.evalEncrypt(localParam);
+    this->propagateIfChanged(lattice, lattice->join(encrypted));
+    return;
+  }
+
```
ZenithalHourlyRate, line 39: I am a bit concerned about setting all secret type to noise of fresh computation.

For example,

```
%b = secret.generic(%a : secret<i32>) {
  %c = modreduce %a
  yield %c
} -> secret<i32>
```

`%c` will have smaller noise than `%a` in BGV, and BranchOpInterface will propagate `%c` to `%b` but that propagation will fail (not reset the noise) because NoiseState is monotonic. The same thing will happen for bootstrap, though we may actually never care about the noise outside of an secret.generic, or if in the future we are to analyse a function with multiple secret.generic body (like loop or function call), we then need to care.

ZenithalHourlyRate (commented):
The monotonic requirement of the Lattices may make the propagation incorrect, and affect the actual propagation for the mgmt attrs.

We may try the following the example and see if `annotate-mgmt` correctly propagate things to the end

```
%b = secret.generic(%a {level=1}) {
  %c = modreduce %a {level = 0}
  yield %c {level = 0}
}
return %b
// %b here should have level _annotation_ of 0 instead of 1.
```

Note that for LevelState, it means LevelState(1) and LevelState(0) because it is reversed. If we propagate LevelState(1) to join a LevelState(0), the result is LevelState(0), which is not expected. In the past this resetting behavior is resolved by fine control of order of insertion of mod reduce and level analysis in `secret-insert-mgmt-bgv`, and I do have concern about this implicit behavior.

lib/Analysis/NoiseAnalysis/BFV/NoiseAnalysis.cpp
j2kun: So when I run that example through validate-noise (with a small change to annotate stuff outside of secret.generic), I get the following

```mlir
module attributes {bgv.schemeParam = #bgv.scheme_param<logN = 12, Q = [4294991873, 1073153], P = [4295049217], plaintextModulus = 65537>, scheme.bgv} {
  func.func @main(%arg0: !secret.secret<tensor<8xi8>> {mgmt.mgmt = #mgmt.mgmt<level = 1>, noise.bound = "32.47"}) -> (!secret.secret<tensor<8xi8>> {mgmt.mgmt = #mgmt.mgmt<level = 0>}) {
    %0 = secret.generic(%arg0: !secret.secret<tensor<8xi8>> {mgmt.mgmt = #mgmt.mgmt<level = 1>, noise.bound = "32.47"}) {
    ^body(%input0: tensor<8xi8>):
      %1 = mgmt.modreduce %input0 {mgmt.mgmt = #mgmt.mgmt<level = 0>, noise.bound = "28.50"} : tensor<8xi8>
      secret.yield %1 : tensor<8xi8>
    } -> (!secret.secret<tensor<8xi8>> {mgmt.mgmt = #mgmt.mgmt<level = 0>, noise.bound = "28.50"})
    return %0 : !secret.secret<tensor<8xi8>>
  }
}
```

This also looks right to me. I think what is maybe confusing here is when `setToEntryState` is called. It is called for all block arguments of every region at the start of the computation, as well as whenever control-flow operands can't be inferred (which is never for `secret.generic`). So the only things that will be set to "fresh noise" is a secret-typed block argument. Note the block arguments _inside_ a `secret.generic` are never secret, so these will be set to uninitialized. And because the join for noise takes a max, the case of a loop iter arg would be handled correctly: 

- The first loop iteration would default to fresh encryption
- When joined with the actual init value, it would be the max of a fresh encryption noise and the init's noise.
- Each loop iteration will join the above to the yielded iter arg's noise.
 
As I understand, fresh encryption is always the smallest possible noise, so something like a ciphertext bootstrapped at the end of every loop iteration would produce larger noise in the yielded value, and the max would correctly override the initial entry-state lattice value.

ZenithalHourlyRate: > what is maybe confusing here is when setToEntryState is called. It is called for all block arguments of every region at the start of the computation

That make sense!

> As I understand, fresh encryption is always the smallest possible noise

Unfortunately no, for BGV, modulus switching will produce even smaller noise, but that is a bit off-topic.

ZenithalHourlyRate (approved):

j2kun:
Ah, I had to copy 2a5711864534d3f79d3aa29012eb745d0057a72c from https://github.com/google/heir/pull/1837

PR #1856, opened 2025-05-28
fix secret.generic syntax in comments/demos

j2kun:
```
sed -i 's/secret.generic ins/secret.generic/g'
```


asraa (approved):
Ah thank you! I remember I only checked mlir and td files :P

Issue #1797, opened 2025-05-07
MacOs Build Issues


kragall:
So I've got a new Mac and while setting up heir there I ran into build problems. After spending some time there it turns out that even with my old Mac (where the build environment was working previously) I'm not able to build heir anymore. I'm not sure, when it broke exactly.

The three errors I get all seem to occur in macOS sdk files:
```
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/stdio.h:61:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/_stdio.h:318:7: error: expected identifier or '('
  318 | FILE    *fdopen(int, const char *) __DARWIN_ALIAS_STARTING(__MAC_10_6, __IPHONE_2_0, __DARWIN_ALIAS(fdopen));
      |          ^

usr/include/stdio.h:61:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/_stdio.h:318:7: error: expected ')'

In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/stdio.h:61:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/_stdio.h:318:7: error: expected ')'
  318 | FILE    *fdopen(int, const char *) __DARWIN_ALIAS_STARTING(__MAC_10_6, __IPHONE_2_0, __DARWIN_ALIAS(fdopen));
      |          ^
external/net_zlib/zutil.h:140:33: note: expanded from macro 'fdopen'
  140 | #        define fdopen(fd,mode) NULL /* No fdopen() */
      |

In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/stdio.h:61:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/_stdio.h:318:7: error: expected ')'
[...]

PR #1878, opened 2025-06-04
bump zlib version and migrate it to bazelmod

j2kun:
I believe this fixes https://github.com/google/heir/issues/1797, but perhaps @kragall could verify since I didn't experience this error on my Mac machine.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -32,6 +32,19 @@ bazel_dep(name = "googletest", version = "1.16.0")
 bazel_dep(name = "google_benchmark", version = "1.9.1")
 bazel_dep(name = "abseil-cpp", version = "20250127.0", repo_name = "com_google_absl")
 
+# Yosys dependencies that are already on the bazel central registry
+# abc can't be used in place because it changes the names of the targets from
+# abc-lib -> abc and abc -> abc_bin
+#
+# bazel_dep(name = "abc", version = "0.0.0-20250408-yosyshq.bcr.1", repo_name = "edu_berkeley_abc")
+
+# The versions of yosys deps zlib and libffi that come with rules_hdl don't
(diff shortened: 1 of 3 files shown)


MODULE.bazel
```diff
@@ -32,6 +32,11 @@ bazel_dep(name = "googletest", version = "1.16.0")
 bazel_dep(name = "google_benchmark", version = "1.9.1")
 bazel_dep(name = "abseil-cpp", version = "20250127.0", repo_name = "com_google_absl")
 
+# Yosys dependencies that are already on the bazel central registry
+# abc can't be used in place because it changes the names of the targets from abc-lib -> abc and abc -> abc_bin
+# bazel_dep(name = "abc", version = "0.0.0-20250408-yosyshq.bcr.1", repo_name = "edu_berkeley_abc")
```
asraa, line 39: Oh! I might be able to handle this, in doing the fully-homomorphic-encryption stuff I am again familiarizing myself with the dep issues.

kragall:
After adding this line in `MODULE.bazel`
`bazel_dep(name = "libffi", version = "3.4.7.bcr.3", repo_name ="org_sourceware_libffi")`

and removing these lines in `WORKSPACE.bzlmod`
`load("@rules_hdl//dependency_support/org_sourceware_libffi:org_sourceware_libffi.bzl", "org_sourceware_libffi")`
`org_sourceware_libffi()`

the new error in #1797 disappears and the build works again (at least on my Mac).

MODULE.bazel
j2kun: the difficulty here is that we need to override the BUILD file in the BCR, and there's no good way to do that locally. You have to maintain a fork of the project with your own BUILD file and point the bazel_dep to it with a git_override.

j2kun:
Hooray! Let me add that ffi dep as well

Issue #402, opened 2024-01-30
Bug: secret-forget-secrets does not work after yosys optimizer


asraa:
@j2kun - I thought it would be nice to check the Yosys optimizer pass for correctness by chaining yosys-optimizer with secret-forget-secrets and then lowering to LLVM to execute (this would have caught the bug)

The issue here seems to be that the `secret.cast` that transforms the multi-bit secrets to the tensor of single-bits has no conversion pattern. We already have code that extracts bits from plaintext multi-bits so we can use that here for an input conversion.

j2kun:
I lost the train of thought on this bug. Is there a simple reproducer for the error?

asraa:
Sure! Maybe the add_one example is the easiest. It hits this bug on anything with a multi-bit integer value.

```
bazel run -c dbg @heir//tools:heir-opt --  --yosys-optimizer --canonicalize --cse --secret-forget-secrets $(pwd)/tests/yosys_optimizer/arith_ops.mlir
```

j2kun:
Reproducer for `secret-forget-secrets`

```mlir
module {
  func.func private @internal_generic_17345143092562439778(%arg0: i3, %arg1: i3, %arg2: i3, %arg3: i3) -> i3 {
    %0 = arith.subi %arg0, %arg1 : i3
    %1 = arith.muli %0, %arg2 : i3
    %2 = arith.andi %1, %arg3 : i3
    return %2 : i3
  }
  func.func @ops(%arg0: !secret.secret<i3>, %arg1: !secret.secret<i3>, %arg2: !secret.secret<i3>, %arg3: !secret.secret<i3>) -> !secret.secret<i3> {
    %c2 = arith.constant 2 : index
    %c1 = arith.constant 1 : index
    %c0 = arith.constant 0 : index
    %false = arith.constant false
    %0 = secret.cast %arg0 : !secret.secret<i3> to !secret.secret<memref<3xi1>>
    %1 = secret.cast %arg1 : !secret.secret<i3> to !secret.secret<memref<3xi1>>
    %2 = secret.cast %arg2 : !secret.secret<i3> to !secret.secret<memref<3xi1>>
    %3 = secret.cast %arg3 : !secret.secret<i3> to !secret.secret<memref<3xi1>>
    %4 = secret.generic(%0: !secret.secret<memref<3xi1>>, %1: !secret.secret<memref<3xi1>>, %2: !secret.secret<memref<3xi1>>, %3: !secret.secret<memref<3xi1>>) {
[...]

PR #1883, opened 2025-06-05
Lower secret.cast in secret-forget-secrets

j2kun:
Fixes https://github.com/google/heir/issues/402


asraa (approved):
great! happy those got moved to helpers for reuse

PR #1889, opened 2025-06-05
Support layout-propagation for MNIST model

j2kun:
WIP progress toward https://github.com/google/heir/issues/1888

Issue #1888, opened 2025-06-05
Get a simple MNIST model e2e through a SIMD FHE pipeline


j2kun:
Tracking work on getting the following input MLIR through the entire pipeline:

```mlir
module {
  func.func @main(
      %layer1_weights: tensor<512x784xf32> {debug.name = "layer1_weights"},
      %layer1_bias: tensor<512xf32> {debug.name = "layer1_bias"},
      %layer2_weights: tensor<10x512xf32> {debug.name = "layer2_weights"},
      %layer2_bias: tensor<10xf32> {debug.name="layer2_bias"},
      %input: tensor<1x784xf32> {secret.secret, debug.name = "input"}
  ) -> (tensor<1x10xf32> {debug.name = "result"}) {
    %c0_10 = arith.constant dense<0.000000e+00> : tensor<10xf32>
    %c0_512 = arith.constant dense<0.000000e+00> : tensor<512xf32>
    %0 = tensor.empty() : tensor<784x512xf32>
    %transposed = linalg.transpose ins(%layer1_weights : tensor<512x784xf32>) outs(%0 : tensor<784x512xf32>) permutation = [1, 0]
    %collapsed = tensor.collapse_shape %input [[0, 1]] : tensor<1x784xf32> into tensor<784xf32>
    %1 = linalg.vecmat ins(%collapsed, %transposed : tensor<784xf32>, tensor<784x512xf32>) outs(%c0_512 : tensor<512xf32>) -> tensor<512xf32>
    %2 = arith.addf %layer1_bias, %1 : tensor<512xf32>
    %3 = arith.maximumf %2, %c0_512 : tensor<512xf32>
    %4 = tensor.empty() : tensor<512x10xf32>
    %transposed_1 = linalg.transpose ins(%layer2_weights : tensor<10x512xf32>) outs(%4 : tensor<512x10xf32>) permutation = [1, 0]
    %5 = linalg.vecmat ins(%3, %transposed_1 : tensor<512xf32>, tensor<512x10xf32>) outs(%c0_10 : tensor<10xf32>) -> tensor<10xf32>
    %6 = arith.addf %layer2_bias, %5 : tensor<10xf32>
    %expanded = tensor.expand_shape %6 [[0, 1]] output_shape [1, 10] : tensor<10xf32> into tensor<1x10xf32>
    return %expanded : tensor<1x10xf32>
  }
}
```

j2kun:
[...]

PR #1892, opened 2025-06-06
Add layout-optimization infrastructure to handle kernel changes during hoisting

j2kun:
Rebased on https://github.com/google/heir/pull/1889

Partial progress toward https://github.com/google/heir/issues/1888

This PR adds:

- An attribute to specify the chosen kernel for layout-optimizer
- A new op interface that allows layout-optimizer to handle all convert_layout op hoisting generically
- converting existing hoisters for arith ops to use the interface
- expansion of the layout-optimizer to incorporate the cost of the kernel

For next PR:

- an implementation of the interface for matvec when alignment is unchanged
- baseline selection of kernel costs

### DIFF
--- a/lib/Dialect/TensorExt/IR/TensorExtDialect.cpp
+++ b/lib/Dialect/TensorExt/IR/TensorExtDialect.cpp
@@ -5,6 +5,7 @@
 #include "lib/Dialect/TensorExt/IR/TensorExtAttributes.h"
 #include "llvm/include/llvm/ADT/ArrayRef.h"              // from @llvm-project
 #include "llvm/include/llvm/ADT/STLFunctionalExtras.h"   // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
 #include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
 #include "mlir/include/mlir/IR/Diagnostics.h"            // from @llvm-project
 #include "mlir/include/mlir/IR/DialectImplementation.h"  // from @llvm-project
--- a/lib/Kernel/Kernel.cpp
+++ b/lib/Kernel/Kernel.cpp
@@ -1,0 +1,40 @@
+#include "Kernel.h"
+
+#include <set>
+#include <string>
+#include <unordered_map>
+
+#include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"         // from @llvm-project
+#include "mlir/include/mlir/IR/OperationSupport.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+namespace {
+static std::unordered_map<KernelName, std::string> correspondingOp = {
+    {KernelName::MatvecNaive, "linalg.matvec"},
+    {KernelName::MatvecDiagonal, "linalg.matvec"},
+};
+
+std::set<std::string> requiredNontrivial = {"linalg"};
+}  // namespace
+
+bool isSupportedKernel(Operation *op, KernelName name) {
+  std::string dialect = std::string(op->getDialect()->getNamespace());
+  if (name == KernelName::Trivial) {
+    return requiredNontrivial.count(dialect) == 0;
+  }
+
+  if (correspondingOp.find(name) == correspondingOp.end()) {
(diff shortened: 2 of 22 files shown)


lib/Dialect/TensorExt/IR/TensorExtDialect.cpp
```diff
@@ -37,6 +38,12 @@ void TensorExtDialect::initialize() {
 #define GET_OP_LIST
 #include "lib/Dialect/TensorExt/IR/TensorExtOps.cpp.inc"
       >();
+
+  // FIXME: determine a better place to put these!
+  // declarePromisedInterface<LayoutConversionHoistableOpInterface,
```
asraa: do you need these since the registration is done in HEIRInterfaces?

lib/Kernel/Kernel.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "mlir/include/mlir/IR/Operation.h"              // from @llvm-project
+#include "mlir/include/mlir/IR/OperationSupport.h"       // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+namespace {
+static std::unordered_map<KernelName, std::string> correspondingOp = {
+    {KernelName::MatvecNaive, "linalg.matvec"},
+    {KernelName::MatvecDiagonal, "linalg.matvec"},
```
asraa, line 17: How come the verifier for the the layout hoistable op interface isn't failing that the trivial kernels for addi operations aren't in this table?

lib/Transforms/LayoutOptimization/Hoisting.h
```diff
@@ -0,0 +1,33 @@
+#ifndef LIB_TRANSFORMS_LAYOUTOPTIMIZATION_HOISTRESULT_H_
```
asraa: nit: update HOISTRESULT -> HOISTING

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
```diff
@@ -396,26 +396,65 @@ LogicalResult LayoutPropagation::visitOperation(CollapseShapeOp op) {
 
   auto tensor = op.getSrc();
   LayoutAttr inputLayout = assignedLayouts.at(tensor);
+  // TODO(#1593): Handle inserted dims when propagating through collapse_shape
```
asraa: hahahahaha i thought about this problem a lot when i was implementing the broadcast -> collapse shape folder, so maybe eventually i can take this

tests/Transforms/layout_propagation/mnist.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: heir-opt --layout-propagation %s | FileCheck %s
```
asraa: could you attach the TODO #1888 here too so that we can add some meaningful checks in here when more support is in the pipeline?

asraa (approved):
Thank you for pushing more of this infra in!

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
j2kun: Maybe we won't have to if we change the system to use relations...

tests/Transforms/layout_propagation/mnist.mlir
j2kun: Let's do that in https://github.com/google/heir/pull/1889

j2kun: Added to https://github.com/google/heir/pull/1889

lib/Dialect/TensorExt/IR/TensorExtDialect.cpp
j2kun: Removed

lib/Kernel/Kernel.cpp
j2kun: Because nothing actually sets this yet. I will add a test to ensure it's triggered. This comment also makes me realize the map should be a multimap.

j2kun: I can't get the op interface verifier to actually run in ODS, so I added it to a precondition of `layout-optimization` and left a todo. Also slightly changed it so that a missing attribute corresponds to "trivial kernel"

asraa (approved):
Curious about the attribute verifier...

PR #1893, opened 2025-06-08
CKKS: Add EXTENDED encryption technique in SchemeParam

ZenithalHourlyRate:
This PR adds the EXTENDED encryption technique flag to the SchemeParamAttr and corresponding C++ struct. To do so the encryptionTechniqueExtended flag is moved from BGVSchemeParam to RLWESchemeParam so it could be shared among all RLWE schemes.

This is mainly meant for CKKS noise models (e.g. #1685). By default Lattigo uses the EXTENDED encryption since we always use HYBRID keyswitching for now, meanwhile OpenFHE does not support EXTENDED mode for CKKS, so there is a difference in noise modelling.

This should have no other effect on the backend, so tests are not added.


lib/Dialect/CKKS/IR/CKKSEnums.td
```diff
@@ -10,4 +10,11 @@ def CKKS_EncryptionTypeEnum : I32EnumAttr<"CKKSEncryptionType", "An enum attribu
     let cppNamespace = "::mlir::heir::ckks";
 }
 
+def CKKS_EncryptionTechniqueEnum : I32EnumAttr<"CKKSEncryptionTechnique", "An enum attribute representing an encryption technique", [
+  I32EnumAttrCase<"standard", 0>,
+  I32EnumAttrCase<"extended", 1>
```
j2kun, line 18:
```suggestion
  // The extended encryption technique encrypts at Qp then mod reduces to Q.
  // This has the benefit of smaller encryption noise.
  I32EnumAttrCase<"extended", 1>
```

IMO the tablegen will be the first place someone looks for a docstring.

lib/Parameters/CKKS/Params.h
```diff
@@ -28,7 +28,8 @@ class SchemeParam : public RLWESchemeParam {
 
   static SchemeParam getConcreteSchemeParam(std::vector<double> logqi,
                                             int logDefaultScale, int slotNumber,
-                                            bool usePublicKey);
+                                            bool usePublicKey,
```
j2kun, line 31: Not needed for now, but the next time one of these options changes, it would make sense to just store the enum defined in CKKSEnums.td directly instead of bools.

lib/Dialect/CKKS/IR/CKKSEnums.td
j2kun: Also, is there a canonical citation we can refer to?

lib/Transforms/GenerateParam/GenerateParam.td
```diff
@@ -228,6 +228,8 @@ def GenerateParamCKKS : Pass<"generate-param-ckks"> {
            "coefficient modulus to use for the ciphertext space.">,
     Option<"usePublicKey", "use-public-key", "bool", /*default=*/"true",
            "If true, uses a public key for encryption.">,
+    Option<"encryptionTechniqueExtended", "encryption-technique-extended", "bool", /*default=*/"false",
+           "If true, uses EXTENDED encryption technique for encryption.">,
```
j2kun: May also want to add a canonical citation to the CLI option (e.g., if there's a short IACR link)

lib/Dialect/CKKS/IR/CKKSEnums.td
ZenithalHourlyRate: Since we are using OpenFHE terminology, I just refer to https://ia.cr/2022/915 the OpenFHE paper.

Issue #1844, opened 2025-05-22
Update OpenFHE backend to v1.3


ZenithalHourlyRate:
OpenFHE recently made their 1.3 release, with the following changes relatively notable for HEIR

* composite scaling modes to CKKS
* complex-number arithmetic in CKKS

https://github.com/openfheorg/openfhe-development/releases/tag/v1.3.0

j2kun:
Looks like the internal needs some small tweaks for v1.3. Mainly, in CKKS we still codegen SetPlaintextModulus for CKKS and that causes a crash that somehow is not reflected in the OSS repo (maybe because `-fno-exceptions`?)

PR #1905, opened 2025-06-11
update openfhe to v1.3

j2kun:
Fixes https://github.com/google/heir/issues/1844

Don't submit yet, it depends on an internal CL to update the openFHE version in Google's monorepo

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -2433,7 +2433,7 @@
         "bzlTransitiveDigest": "pBYZfL5VYF4BT5jARY3p7l3d7Uknixz0stBBKWj2iaw=",
         "usagesDigest": "fJqnUFr/GgZxd/VmNa3e4q3P8xv3wi51f6kXAa9+tsg=",
         "recordedFileInputs": {
-          "@@//requirements-dev.txt": "26da3dcbf88e8d5600d4bbedf32ac5d38e6992ba510cb49d693891310b5d473c",
+          "@@//requirements-dev.txt": "54fd907ca5b52c0522f9c479b4fe17cf8af6843d1d5b58f62a14705a0552c692",
           "@@//requirements.txt": "7efa66b1fcb36e4aac632075969146e3f087ec5049cd926b84ffdabaa58b8c49",
           "@@grpc+//requirements.bazel.txt": "95a27c3f9a46b8114d464c70ba93cda18cfe8c02004db81028f9306b2691701e",
           "@@or-tools+//bazel/notebook_requirements.txt": "ca78fad693f1b35eed8bb7c54e5ddf7ad255c4b9d94ce18efa55859759a8fb70",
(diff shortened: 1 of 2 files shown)


j2kun:
Looks like the bug is that SetMultiplicativeDepth(0) (which for box_blur_64x64 is correct) produces bad params. Setting the mul depth to 1 works. The other error is fixed in the internal CL.

ZenithalHourlyRate:
> Looks like the bug is that SetMultiplicativeDepth(0) (which for box_blur_64x64 is correct) produces bad params. Setting the mul depth to 1 works. The other error is fixed in the internal CL.

So OpenFHE v1.3 generates a more aggresive parameter, and for box_blur, it generates n=4096, whereas previously it generates n=8192. Then 4096 slots could not be packed into it (note that we only half pack to ensure 1-dim behavior).

It turns out that we have no other way to specify it. See https://github.com/ZenithalHourlyRate/heir/commit/8a05338d65cdf35a5eb67325f960104ca38770d7.
This commit fixes the CKKS failure, but for BGV, it does not work. Specifically, this time box_blur is correct, but roberts_cross is incorrect.

Notably we have the following two exceptions:

```
C++ exception with description "external/openfhe/src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp:l.73:computeRingDimension(): The specified ring dimension (8192) does not comply with HE standards recommendation (16384)." thrown in the test body.
```

```
C++ exception with description "external/openfhe/src/pke/lib/scheme/bfvrns/bfvrns-parametergeneration.cpp:l.367:ParamsGenBFVRNSInternal(): The batch size cannot be larger than the ring dimension." thrown in the test body.
```

If we specify nothing, we does not get the guaranteed slot number.

If we specify slot number, OpenFHE may generate a ringDim that is smaller than slot number (I think this is a bug)

If we specify slot number and ringDim, OpenFHE may find that ringDim is too small for its computed parameter.

There is no "minimum ring dim" option for OpenFHE.

See also https://github.com/google/heir/pull/1439#issue-2866208950 where ptm plays some role...

Regarding https://github.com/google/heir/pull/1916#issuecomment-2980967117, I did not observe "more" failures caused by 1916.

ZenithalHourlyRate:
If we just want to make CI passing, we can specify ringDim=8192 for box_blur relying on the assumption that _OpenFHE generates secure parameter for that ringDim_. However, this is not future-proof and once OpenFHE changes we need to change again.

Or we specify mulDepth=1 for box_blur.

See https://github.com/google/heir/commit/f86acd73b56a958635d8ebcd8fe9c2d3dd26d2d6 on how to specifying these parameters.

j2kun:
Internal Google openfhe has been updated, so as long as this passes CI we can merge it

j2kun:
I went with Hongren's override option to get it merged.

ZenithalHourlyRate (approved):

PR #1906, opened 2025-06-12
Mgmt: Add MemoryEffect for InitOp to prevent CSE

ZenithalHourlyRate:
Inspired by #1904. The problem is that when mgmt.init has multiple uses, the back-prop should not propagate to the same mgmt.init; so from the begining we need to ensure mgmt.init only has one use.

Currently, this invariant is implicitly maintained by

* `UseInitOpForPlaintextOperand` in `insert-mgmt` that create mgmt.init for each _Use_
* `ConvertAdjustScaleToMulPlain` in `populate-scale` that create ad-hoc mgmt.init for each mulconst.

However, CSE does not respect this and merge them together since the op was marked as `Pure`. We could not judge the equivalence among init ops unless we have finished all the chain `--insert-mgmt --optimize-relinearation --populate-scale`, where all these MgmtAttr attached to the mgmt.init op could change.

So instead we could manually manage init op by registering bogus momery effect, and manually clean them up in `secret-to-<scheme>`.

Another idea would be attach `id` attr to each mgmt init op, but this counter should be maintained across passes so it is not so convenient.

Also, for secret.generic canonicalization, mgmt.init op attaches a ConditionallySpeculatable trait to tell the canonicalizer that it could be moved out of the secret generic body.

### DIFF
--- a/lib/Dialect/Mgmt/IR/MgmtOps.cpp
+++ b/lib/Dialect/Mgmt/IR/MgmtOps.cpp
@@ -7,6 +7,10 @@ namespace mlir {
 namespace heir {
 namespace mgmt {
 
+//===----------------------------------------------------------------------===//
+// Canonicalization Patterns
+//===----------------------------------------------------------------------===//
+
 // Kept inside a namespace because it generates a function called
 // populateWithGenerated, which can conflict with other generated patterns.
 #include "lib/Dialect/Mgmt/IR/MgmtCanonicalization.cpp.inc"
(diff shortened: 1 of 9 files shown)


j2kun:
I noticed this as well when working on #1904, but I didn't realize CSE would combine them. We could avoid running CSE in between when `mgmt.init` is created and when the attributes are back-propagated to them, since once they have the attributes attached they would be safe to merge.

However, I agree that:

1. Merging them manually would be a very simple pattern to implement.
2. Even if we don't merge them manually, after we lower to scheme, cse would clean them up at that step.

j2kun (approved):

ZenithalHourlyRate:
> We could avoid running CSE in between when mgmt.init is created and when the attributes are back-propagated to them, since once they have the attributes attached they would be safe to merge.

Just an additional note, the attributes are not finalized after `insert-mgmt`. We only get the whole information after `populate-scale`. If we merge them before `populate-scale` using CSE (highly likely), it will cause trouble. Note that it is possible for two pt to have same level/dimension but different scale.

PR #1911, opened 2025-06-13
fix cggi test

j2kun:
Looks like there was a race condition for Alex's casting changes and the support for floordiv that both modified this test.

### DIFF
--- a/frontend/cggi_test.py
+++ b/frontend/cggi_test.py
@@ -19,7 +19,7 @@ class EndToEndTest(absltest.TestCase):
       return y
 
     # Test cleartext functionality
-    self.assertEqual(-15, foo.original(7, 8))
+    self.assertEqual(-8, foo.original(7, 8))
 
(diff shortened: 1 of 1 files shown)


asraa (approved):
how does this test execute at all if there's no cggi backend?

AlexanderViand:
> how does this test execute at all if there's no cggi backend?

In my PR, I enabled the tests by using `CleartextBackend`. That way, we're at least somewhat exercising the frontend + `--mlir-to-cggi` and making sure nothing errors out.

asraa:
But CGGI doesn't have a cleartext backend?

AlexanderViand:
> But CGGI doesn't have a cleartext backend?

It's a **Clear**textBackend defined in Python. With it specified, the frontend runs `heir-opt` as normal but then just uses a wrapper around the original python function to supply ClientInterface/etc instead of actually dispatching to generated code.

PR #1913, opened 2025-06-13
Add SimFHE Emitter

AlexanderViand:
@ai-mannamalai I played around with [SimFHE ](https://github.com/bu-icsg/SimFHE) a bit and as I suspected, it was indeed pretty straight-forward to put together a basic emitter. 

**Using this:**
* Either use [my SimFHE fork](https://github.com/alexanderviand/SimFHE) or add `"generated"` to `DECORATION_LIST` in SimFHE's `profiler.py`
* Take your favorite HEIR/MLIR program and run it through `heir-opt --mlir-to-ckks` or equivalent.
* Run `heir-translate -emit-simfhe > genrated.py`  and place the resulting file in the SimFHE directory
* Running `python generated.py` in the SimFHE directory should (assuming you've installed the requirements) output the same kind of table you'd get from running SimFHE's `experiment.py` 

**Caveats:**
I have _no idea_ how SimFHE is supposed to be used, and the available code for it is somewhat messy and nearly entirely non-documented. As a result, I have no idea if the code this emitter generates uses SimFHE correctly. I also don't really understand how to interpret the output SimFHE presents, so I can't even say if it produces the expected output for the test cases I tried. 
Finally, I've used this as a playground to experiment with OpenAI's [Codex](https://openai.com/index/introducing-codex/) feature so a lot of this is AI generated, with all the implications that has...

If someone is interested in using this for anything real, I'd highly recommend reaching out to the SimFHE folks first!

### DIFF
--- a/lib/Target/SimFHE/SimFHEEmitter.cpp
+++ b/lib/Target/SimFHE/SimFHEEmitter.cpp
@@ -1,0 +1,179 @@
+#include "lib/Target/SimFHE/SimFHEEmitter.h"
+
+#include "lib/Analysis/SelectVariableNames/SelectVariableNames.h"
+#include "lib/Dialect/CKKS/IR/CKKSDialect.h"
+#include "lib/Dialect/LWE/IR/LWEDialect.h"
+#include "lib/Dialect/LWE/IR/LWETypes.h"
+#include "lib/Dialect/ModArith/IR/ModArithDialect.h"
+#include "lib/Dialect/Polynomial/IR/PolynomialDialect.h"
+#include "lib/Dialect/RNS/IR/RNSDialect.h"
+#include "lib/Dialect/RNS/IR/RNSTypeInterfaces.h"
+#include "llvm/include/llvm/ADT/StringRef.h"           // from @llvm-project
+#include "llvm/include/llvm/ADT/TypeSwitch.h"          // from @llvm-project
+#include "llvm/include/llvm/Support/FormatVariadic.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/DialectRegistry.h"      // from @llvm-project
+#include "mlir/include/mlir/Tools/mlir-translate/Translation.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace simfhe {
+
+void registerToSimFHETranslation() {
+  TranslateFromMLIRRegistration reg(
+      "emit-simfhe", "translate ckks dialect to SimFHE python code",
+      [](Operation *op, llvm::raw_ostream &output) {
+        return translateToSimFHE(op, output);
+      },
+      [](DialectRegistry &registry) {
+        registry.insert<func::FuncDialect, ckks::CKKSDialect, lwe::LWEDialect,
+                        polynomial::PolynomialDialect,
+                        mod_arith::ModArithDialect, rns::RNSDialect>();
+        rns::registerExternalRNSTypeInterfaces(registry);
+      });
+}
+
+LogicalResult translateToSimFHE(Operation *op, llvm::raw_ostream &os) {
+  SelectVariableNames variableNames(op);
+  SimFHEEmitter emitter(os, &variableNames);
+  return emitter.translate(*op);
+}
+
+SimFHEEmitter::SimFHEEmitter(llvm::raw_ostream &os,
+                             SelectVariableNames *variableNames)
+    : os(os), variableNames(variableNames) {}
+
+LogicalResult SimFHEEmitter::translate(Operation &op) {
+  LogicalResult status =
+      llvm::TypeSwitch<Operation &, LogicalResult>(op)
+          .Case<ModuleOp, func::FuncOp, func::ReturnOp, ckks::AddOp,
+                ckks::AddPlainOp, ckks::SubOp, ckks::SubPlainOp, ckks::MulOp,
+                ckks::MulPlainOp, ckks::NegateOp, ckks::RotateOp,
+                ckks::RelinearizeOp, ckks::RescaleOp, ckks::LevelReduceOp,
+                ckks::BootstrapOp>(
+              [&](auto innerOp) { return printOperation(innerOp); })
+          .Default([&](Operation &) {
+            return op.emitOpError("unable to find printer for op");
+          });
+  if (failed(status)) {
+    return op.emitOpError(
+        llvm::formatv("Failed to translate op {0}", op.getName()));
+  }
+  return success();
+}
+
+LogicalResult SimFHEEmitter::printOperation(ModuleOp moduleOp) {
+  os << kModulePrelude << "\n";
+  SmallVector<func::FuncOp> funcs;
+  for (Operation &op : moduleOp) {
+    if (auto func = dyn_cast<func::FuncOp>(op)) {
+      funcs.push_back(func);
+    }
+  }
+  for (Operation &op : moduleOp) {
+    if (failed(translate(op))) {
+      return failure();
+    }
+  }
+  os << kMainPrelude << "\n";
+  os.indent();  // for if __name__ == "__main__":
+  os.indent();  // for internal for loop
+
+  for (auto func : funcs) {
+    os << "targets.append(Target(\"generated." << func.getName() << "\",1, [";
+    for (auto arg : func.getArguments()) {
+      if (isa<lwe::NewLWECiphertextType>(arg.getType()))
+        os << "scheme_params.fresh_ctxt,";
+      else if (isa<lwe::NewLWEPlaintextType>(arg.getType()))
+        os << "params.PolyContext(scheme_params.fresh_ctxt.logq, "
+              "scheme_params.fresh_ctxt.logN, "
+              "scheme_params.fresh_ctxt.dnum,1), ";
+      else
+        return func.emitOpError("Unsupported argument type for SimFHEEmitter: ")
+               << arg.getType();
+    }
+    os << "scheme_params]))\n";
+  }
+  os.unindent();  // exit internal for loop
+  os << kMainPostlude << "\n";
+  return success();
+}
+
+LogicalResult SimFHEEmitter::printOperation(func::FuncOp funcOp) {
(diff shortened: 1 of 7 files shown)


lib/Target/SimFHE/SimFHEEmitter.h
j2kun, line 1: Please add some breadcrumbs, either here or in the BUILD file, for what SimFHE is and the caveats expressed in the PR notes about it being experimental.

tests/Emitter/SimFHE/emit_simfhe.mlir
```diff
@@ -0,0 +1,72 @@
...
+// CHECK: stats += evaluator.multiply([[ARG3:[A-Za-z0-9_]+]],
+// scheme_params.arch_param) CHECK-NEXT: [[RES3:[A-Za-z0-9_]+]] = [[ARG3]]
+// CHECK: stats += evaluator.rotate([[ARG4:[A-Za-z0-9_]+]],
+// scheme_params.arch_param) CHECK-NEXT: [[RES4:[A-Za-z0-9_]+]] = [[ARG4]]
+// CHECK: stats += evaluator.add_plain([[ARG5:[A-Za-z0-9_]+]],
+// scheme_params.arch_param) CHECK-NEXT: [[RES5:[A-Za-z0-9_]+]] = [[ARG5]]
+// CHECK: stats += evaluator.subtract_plain([[ARG6:[A-Za-z0-9_]+]],
+// scheme_params.arch_param) CHECK-NEXT: [[RES6:[A-Za-z0-9_]+]] = [[ARG6]]
+// CHECK: stats += evaluator.multiply_plain([[ARG7:[A-Za-z0-9_]+]],
+// scheme_params.arch_param) CHECK-NEXT: [[RES7:[A-Za-z0-9_]+]] = [[ARG7]]
```
j2kun: This looks malformed CHECK-NEXT are mid line

lib/Target/SimFHE/SimFHEEmitter.cpp
```diff
@@ -0,0 +1,179 @@
...
+    if (failed(translate(op))) {
+      return failure();
+    }
+  }
+  os << kMainPrelude << "\n";
+  os.indent();  // for if __name__ == "__main__":
+  os.indent();  // for internal for loop
+
+  for (auto func : funcs) {
+    os << "targets.append(Target(\"generated." << func.getName() << "\",1, [";
```
j2kun, line 82: Could you assert this part in the output? It seems relatively detailed and important.

lib/Target/SimFHE/SimFHEEmitter.cpp
```diff
@@ -0,0 +1,179 @@
...
+BINARY_EMIT(ckks::AddPlainOp, getLhs, "add_plain");
+BINARY_EMIT(ckks::SubOp, getLhs, "add");
+BINARY_EMIT(ckks::SubPlainOp, getLhs, "add_plain");
+BINARY_EMIT(ckks::MulOp, getLhs, "multiply");
+BINARY_EMIT(ckks::MulPlainOp, getLhs, "multiply_plain");
+UNARY_EMIT(ckks::NegateOp, getInput, "multiply_plain");
+UNARY_EMIT(ckks::RotateOp, getInput, "rotate");
+LogicalResult SimFHEEmitter::printOperation(ckks::RelinearizeOp op) {
+  std::string name = getName(op.getInput());
+  os << "stats += evaluator.key_switch(" << name << ", " << name
```
j2kun: Why does key switch duplicate the argument, and why can't this be an instance of the binary op case?

lib/Target/SimFHE/SimFHEEmitter.cpp
```diff
@@ -0,0 +1,179 @@
...
+  os << "\n";
+  return success();
+}
+
+LogicalResult SimFHEEmitter::printOperation(func::ReturnOp op) {
+  // return is handled via accumulating stats; nothing to do
+  return success();
+}
+
+#define UNARY_EMIT(OPTYPE, GETTER, FUNCNAME)                 \
```
j2kun: Not a huge fan of these macros. Better would be to just have a helper method that takes the SSA values for the operands as inputs, and the func name as a string.

For example of why I don't like it: BINARY_EMIT somehow doesn't use both operands (is this a bug?), but the MACRO kind of makes that hard to see.

lib/Target/SimFHE/SimFHEEmitter.cpp
```diff
@@ -0,0 +1,179 @@
...
+    os << getName(op.getResult()) << " = " << input << "\n"; \
+    return success();                                        \
+  }
+
+#define UNARY_EMIT_DROP(OPTYPE, GETTER, FUNCNAME)                   \
+  LogicalResult SimFHEEmitter::printOperation(OPTYPE op) {          \
+    std::string input = getName(op.GETTER());                       \
+    os << "stats += evaluator." FUNCNAME "(" << input               \
+       << ", scheme_params.arch_param)\n";                          \
+    os << getName(op.getResult()) << " = " << input << ".drop()\n"; \
```
j2kun, line 141: what is `drop`?

lib/Target/SimFHE/SimFHEEmitter.cpp
```diff
@@ -0,0 +1,179 @@
...
+#define BINARY_EMIT(OPTYPE, GETTER, FUNCNAME)                \
+  LogicalResult SimFHEEmitter::printOperation(OPTYPE op) {   \
+    std::string input = getName(op.GETTER());                \
+    os << "stats += evaluator." FUNCNAME "(" << input        \
+       << ", scheme_params.arch_param)\n";                   \
+    os << getName(op.getResult()) << " = " << input << "\n"; \
+    return success();                                        \
+  }
+
+// Since we only care about the cost, add/subtract are treated the same
```
j2kun: Is this implying that there is no `sub` in the SimFHE API? Better would be to make that explicit.

```suggestion
// SimFHE doesn't have `sub`. Since we only care about the cost it doesn't matter.
```

j2kun (changes requested):

tests/Emitter/SimFHE/emit_simfhe.mlir
AlexanderViand: Yep, sorry, forgot to add the real test 🙈

lib/Target/SimFHE/SimFHEEmitter.cpp
AlexanderViand: The second argument is the key argument, using the input again was a quick hack to get valid code. I've replaced it with a `fresh_ctxt` and a warning that we don't really know the exact key details from the IR.

lib/Target/SimFHE/SimFHEEmitter.cpp
AlexanderViand: drop() is how you drop a level/limb from a SimFHE PolyContext object (which they use to represent input sizes/etc)

lib/Target/SimFHE/SimFHEEmitter.cpp
AlexanderViand: assert how? If you mean check for it in the test, the updated test has that

lib/Target/SimFHE/SimFHEEmitter.cpp
AlexanderViand: The macros are mostly a result of this being heavily AI-generated - I'll remove the redundant binary one, as in SimFHE, you only supply the size of the operands with one input, and it seems to assume the operands will both have exactly the same size/etc

lib/Target/SimFHE/SimFHEEmitter.cpp
AlexanderViand: This can't use the normal macro, since it's one of the rare case where there's actually two PolyContext things being fed into the evaluator API

ai-mannamalai:
Thanks, @AlexanderViand - I have time to review only by Monday. This is super fast :)

lib/Target/SimFHE/SimFHEEmitter.cpp
ai-mannamalai:
```suggestion
    os << "assert generated."<<func.getName()<<", 'cannot be empty'\n";
    os << "targets.append(Target(\"generated." << func.getName() << "\",1, [";
```
maybe like this ?

lib/Target/SimFHE/SimFHEEmitter.cpp
```diff
@@ -0,0 +1,173 @@
...
+              "scheme_params.fresh_ctxt.logN, "
+              "scheme_params.fresh_ctxt.dnum,1), ";
+      else
+        return func.emitOpError("Unsupported argument type for SimFHEEmitter: ")
+               << arg.getType();
+    }
+    os << "scheme_params]))\n";
+  }
+  os.unindent();  // exit internal for loop
+  os << kMainPostlude << "\n";
```
ai-mannamalai:
```suggestion
  os << kMainEpilogue << "\n";
```

lib/Target/SimFHE/SimFHETemplates.h
```diff
@@ -0,0 +1,43 @@
...
+  for scheme_params in [
+      params.Alg_benchmark_baseline,
+      params.Alg_benchmark_mod_down_merge,
+      params.Alg_benchmark_mod_down_hoist,
+      params.BEST_PARAMS,
+  ]:
+    print(scheme_params)
+)""";
+
+constexpr std::string_view kMainPostlude = R"""(
```
ai-mannamalai:
```suggestion
constexpr std::string_view kMainEpilogue = R"""(
```

lib/Target/SimFHE/SimFHEEmitter.cpp
j2kun: Yes I meant just assert something about its correctness in a test.,

j2kun:
LGTM. Rebase and squash and maybe @code-perspective will be able to approve the internal change?

code-perspective:
I see internal CI test failures that need fixing. On it!

PR #1915, opened 2025-06-14
types.py: use match/case

j2kun:
The main issues not allowing this to be done in the review for #1825:

- Bare words are treated as a variable capture unless scoped (with a module prefix, e.g., `mt.I1`) or instantiated (`I1()`).
- The "MLIRType" values passed to these functions are not actually instances of MLIRType (the type hint is still incorrect in this PR), but rather instances of `class`. We cannot rectify this by instantiating the MLIRType to simplify the match syntax (e.g., use `I1()` instead of `mt.I1`) because #1825 added a member variable on MLIRType so the type class could be used as a cast.

Possibly an improvement would be to add a no-value constructor to MLIRType so that `I1()` can be instantiated as a standalone type, and then call these helpers with actual instances of MLIRType rather than class constructor objects. @AlexanderViand let me know if you think this should be done in this PR or after, but at least not doing it before sending it for review makes the problem clearer.

Side note: This PR also fixes a bug in isFloatLike, where a condition was `issubclass(type, MLIRType)` when it should have been `issubclass(typ, MLIRType)`

### DIFF
--- a/frontend/heir/mlir/types.py
+++ b/frontend/heir/mlir/types.py
@@ -1,10 +1,10 @@
 """Defines Python type annotations for MLIR types."""
 
 from abc import ABC, abstractmethod
-from typing import Generic, Self, TypeVar, TypeVarTuple, get_args, get_origin
+from typing import Generic, TypeVar, TypeVarTuple, get_args, get_origin
 from numba.core.types import Type as NumbaType
 from numba.core.types import boolean, int8, int16, int32, int64, float32, float64
-from numba.extending import typeof_impl, type_callable
+from numba.extending import type_callable
 
 T = TypeVar("T")
 Ts = TypeVarTuple("Ts")
--- a/frontend/heir/mlir_emitter.py
+++ b/frontend/heir/mlir_emitter.py
@@ -4,70 +4,87 @@ from collections import deque
 from dataclasses import dataclass
 import operator
 import textwrap
-from typing import Any, NewType
+from typing import Any
 
 from numba.core import ir
-from numba.core import types
 from numba.core import bytecode
 from numba.core import controlflow
-from numba.core.types import Type as NumbaType
+import numba.core.types as nt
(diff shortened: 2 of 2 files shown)


AlexanderViand (approved):
👍 Thanks, this looks really nice! 

I agree that the "numba types are instances, mlir types are classes" is annoying... it's a result of where they come from  - the numba type inference system uses instances of numba types, while the casting/etc system is grabbing the global symbol, which is just, e.g., `I1`. But if you've got an idea for how to get around that and make this all look nicer, I'm all for it :)

j2kun:
@AlexanderViand PTAL at 4c7c83b0490a9fe15a608889e94216735a7c112e to see how I tried aligning it.

frontend/heir/mlir/types.py
```diff
@@ -22,8 +22,19 @@ def check_for_value(a: "MLIRType"):
 
 class MLIRType(ABC):
 
-  def __init__(self, value: int):
-    self.value = value
+  def __init__(self, value: Optional[int] = None):
+    # MLIRType subclasses are sed in two ways:
```
AlexanderViand:
```suggestion
    # MLIRType subclasses are used in two ways:
```

frontend/heir/mlir_emitter.py
```diff
@@ -514,9 +523,12 @@ def emit_assign(self, assign):
             )
           target_ssa = self.get_or_create_name(assign.target)
           ssa_id = self.get_or_create_name(assign.value.args[0])
+
+          # Construct an instance of the MLIR type in question
+          mlir_type_instance = global_()
```
AlexanderViand, line 528: This isn't super pretty, but seems like a great tradeoff for making the matching code so much nicer!

AlexanderViand (approved):
Nice! I agree this makes a lot more sense :)

PR #1916, opened 2025-06-16
OpenFHE: BFV: Count Add and KeySwitch Op

ZenithalHourlyRate:
We used to correctly set EvalAddCount and KeySwitchCount for BGV in OpenFHE *for security reasons*. See #1254 for background.

For BFV, the case when mulDepth = 0 also needs setting EvalAddCount and KeySwitchCount. However, the requirement is that

* Only one of them could be non-zero

```
// throw an exception if at least 2 variables are not zero
```

* Only one of them takes the responsibility to ensure correctness

```
mulDepth != 0
// Only EvalMult operations are used in the correctness constraint
// the correctness constraint from Section 3.1 of https://eprint.iacr.org/2021/204.pdf
// is used

keyswitchCount != 0
// this case supports automorphism w/o any other operations
// base for relinearization

addCount != 0
// only public key encryption and EvalAdd (optional when evalAddCount = 0)
// operations are supported the correctness constraint from section 3.5 of
// https://eprint.iacr.org/2014/062.pdf is used
```

The case when there is no mul but there are mixed adds/keyswitchs remains unclear.

See https://github.com/openfheorg/openfhe-development/blob/v1.3.0/src/pke/lib/scheme/bfvrns/bfvrns-parametergeneration.cpp for detail.

### DIFF
--- a/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
@@ -341,6 +341,23 @@ struct ConfigureCryptoContext
     config.scalingTechniqueFixedManual = scalingTechniqueFixedManual;
     config.levelBudgetDecode = levelBudgetDecode;
     config.levelBudgetEncode = levelBudgetEncode;
+
+    // for BFV, keep only one of MulDepth/EvalAddCount/KeySwitchCount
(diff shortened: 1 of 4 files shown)


j2kun (approved):
What a strange constraint for OpenFHE to impose. Thanks for looking into this!

j2kun:
I did try patching https://github.com/google/heir/pull/1905 with this PR and it creates more failures. I wonder if there's an issue with how this interacts with the update to v1.3?

Issue #1919, opened 2025-06-17
Add an option to bypass HEIR layout assignment and optimizations


edwjchen:
To integrate automatic packing systems as a frontend to HEIR, we should explore and document the pipeline that disables HEIR's layout assignment. @AlexanderViand pointed to disabling the mlir-to-secret-arithmetic pass.

AlexanderViand:
> [@AlexanderViand](https://github.com/AlexanderViand) pointed to disabling the mlir-to-secret-arithmetic pass.

Unfortunately , simply skipping the arithmetization stage causes some issues, see #1922 😅

PR #1922, opened 2025-06-17
allow skipping arithmetization (and, therefore, packing) pipeline

AlexanderViand:
Work towards
* #1919 

As a a first step, I added a `skip-arithmetization` flag to `--mlir-to-<bgv/ckks>` to skip the entire `--mlir-to-secret-arithmetic` pipeline. However, this runs into two issues

1. The compiler just crashes with 
```heir-opt: lib/Analysis/NoiseAnalysis/Noise.cpp:30: NoiseState mlir::heir::NoiseState::operator+(const NoiseState &) const: Assertion `isKnown() && rhs.isKnown()' failed``` so it seems like there's some noise-analysis initialization happening in the arithmetization pipeline that we're missing? @ZenithalHourlyRate ?

2. Currently, the generation of (the MLIR version of) the client-side functions (encrypt/decrypt/etc) happens at the end of the arithmetization phase, but in most cases we'll still need those. That should be simple to fix with another flag, though.

### DIFF
--- a/lib/Pipelines/ArithmeticPipelineRegistration.h
+++ b/lib/Pipelines/ArithmeticPipelineRegistration.h
@@ -28,6 +28,12 @@ void heirSIMDVectorizerPipelineBuilder(OpPassManager &manager,
                                        bool disableLoopUnroll);
 
 struct MlirToRLWEPipelineOptions : public SimdVectorizerOptions {
+  PassOptions::Option<bool> enableArithmetization{
+      *this, "enable-arithmetization",
+      llvm::cl::desc(
+          "If false, skip the arithmetization pipeline and try to directly "
+          "lower to RLWE scheme (default to true)"),
+      llvm::cl::init(true)};
   PassOptions::Option<int> ciphertextDegree{
       *this, "ciphertext-degree",
       llvm::cl::desc("The degree of the polynomials to use for ciphertexts; "
--- a/lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
+++ b/lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
@@ -233,9 +233,11 @@ FailureOr<std::string> getWeightType(Type type) {
 
 LogicalResult translateToOpenFhePke(Operation *op, llvm::raw_ostream &os,
                                     const OpenfheImportType &importType,
-                                    const std::string &weightsFile) {
+                                    const std::string &weightsFile,
+                                    bool skipVectorResizing) {
   SelectVariableNames variableNames(op);
-  OpenFhePkeEmitter emitter(os, &variableNames, importType, weightsFile);
+  OpenFhePkeEmitter emitter(os, &variableNames, importType, weightsFile,
+                            skipVectorResizing);
   LogicalResult result = emitter.translate(*op);
   return result;
 }
@@ -1191,16 +1193,21 @@ LogicalResult OpenFhePkeEmitter::printOperation(
 LogicalResult OpenFhePkeEmitter::printOperation(
     openfhe::MakePackedPlaintextOp op) {
   std::string inputVarName = variableNames->getNameForValue(op.getValue());
-  std::string filledPrefix =
-      variableNames->getNameForValue(op.getResult()) + "_filled";
-  std::string inputVarFilledName = filledPrefix;
-  std::string inputVarFilledLengthName = filledPrefix + "_n";
-
   FailureOr<Value> resultCC = getContextualCryptoContext(op.getOperation());
   if (failed(resultCC)) return resultCC;
   std::string cc = variableNames->getNameForValue(resultCC.value());
 
+  if (skipVectorResizing_) {
+    emitAutoAssignPrefix(op.getResult());
+    os << cc << "->MakePackedPlaintext(" << inputVarName << ");\n";
+    return success();
+  }
   // cyclic repetition to mitigate openfhe zero-padding (#645)
+  std::string filledPrefix =
+      variableNames->getNameForValue(op.getResult()) + "_filled";
+  std::string &inputVarFilledName = filledPrefix;
+  std::string inputVarFilledLengthName = filledPrefix + "_n";
+
   os << "auto " << inputVarFilledLengthName << " = " << cc
      << "->GetCryptoParameters()->GetElementParams()->GetRingDimension() / "
         "2;\n";
@@ -1212,25 +1219,30 @@ LogicalResult OpenFhePkeEmitter::printOperation(
   os << "  " << inputVarFilledName << ".push_back(" << inputVarName << "[i % "
      << inputVarName << ".size()]);\n";
   os << "}\n";
-
   emitAutoAssignPrefix(op.getResult());
   os << cc << "->MakePackedPlaintext(" << inputVarFilledName << ");\n";
+
   return success();
 }
 
 LogicalResult OpenFhePkeEmitter::printOperation(
     openfhe::MakeCKKSPackedPlaintextOp op) {
   std::string inputVarName = variableNames->getNameForValue(op.getValue());
-  std::string filledPrefix =
-      variableNames->getNameForValue(op.getResult()) + "_filled";
-  std::string inputVarFilledName = filledPrefix;
-  std::string inputVarFilledLengthName = filledPrefix + "_n";
-
   FailureOr<Value> resultCC = getContextualCryptoContext(op.getOperation());
   if (failed(resultCC)) return resultCC;
   std::string cc = variableNames->getNameForValue(resultCC.value());
 
+  if (skipVectorResizing_) {
+    emitAutoAssignPrefix(op.getResult());
+    os << variableNames->getNameForValue(resultCC.value())
+       << "->MakeCKKSPackedPlaintext(" << inputVarName << ");\n";
+    return success();
+  }
   // cyclic repetition to mitigate openfhe zero-padding (#645)
+  std::string filledPrefix =
+      variableNames->getNameForValue(op.getResult()) + "_filled";
+  std::string inputVarFilledName = filledPrefix;
+  std::string inputVarFilledLengthName = filledPrefix + "_n";
   os << "auto " << inputVarFilledLengthName << " = " << cc
      << "->GetCryptoParameters()->GetElementParams()->GetRingDimension() / "
         "2;\n";
@@ -1245,6 +1257,7 @@ LogicalResult OpenFhePkeEmitter::printOperation(
   emitAutoAssignPrefix(op.getResult());
   os << variableNames->getNameForValue(resultCC.value())
      << "->MakeCKKSPackedPlaintext(" << inputVarFilledName << ");\n";
+
   return success();
 }
 
@@ -1481,11 +1494,13 @@ LogicalResult OpenFhePkeEmitter::emitType(Type type, Location loc,
 OpenFhePkeEmitter::OpenFhePkeEmitter(raw_ostream &os,
                                      SelectVariableNames *variableNames,
                                      const OpenfheImportType &importType,
-                                     const std::string &weightsFile)
+                                     const std::string &weightsFile,
+                                     bool skipVectorResizing)
     : importType_(importType),
       os(os),
       variableNames(variableNames),
-      weightsFile_(weightsFile) {}
+      weightsFile_(weightsFile),
+      skipVectorResizing_(skipVectorResizing) {}
 }  // namespace openfhe
 }  // namespace heir
 }  // namespace mlir
--- a/tests/Regression/issue_1919.mlir
+++ b/tests/Regression/issue_1919.mlir
@@ -1,0 +1,23 @@
+// RUN: heir-opt %s "--mlir-to-bgv=enable-arithmetization=false" | FileCheck %s
+
+// A program that does not require any kind of arithmetization (i.e., already consists purely of operations that we can lower directly to FHE)
+// should be able to be lowered to FHE successfully without running the arithmetization stage.
+
+// CHECK: @foo
+func.func @foo(%x: tensor<1024xi16> {secret.secret}, %y: tensor<1024xi16> {secret.secret}) -> tensor<1024xi16> {
+    // CHECK: bgv.add
+    %add = arith.addi %x, %y : tensor<1024xi16>
+    // CHECK: bgv.sub
+    %sub = arith.subi %x, %y : tensor<1024xi16>
+    // CHECK: bgv.mul
+    %mul = arith.muli %add, %sub : tensor<1024xi16>
+    return %mul : tensor<1024xi16>
+}
+
+// CHECK: @foo__encrypt__
+// CHECK-NOT: affine.for
+// CHECK-NOT: tensor.extract
+
+// CHECK: @foo__decrypt__
+// CHECK-NOT: affine.for
+// CHECK-NOT: tensor.insert
--- a/lib/Pipelines/ArithmeticPipelineRegistration.cpp
+++ b/lib/Pipelines/ArithmeticPipelineRegistration.cpp
@@ -194,7 +194,16 @@ void mlirToPlaintextPipelineBuilder(OpPassManager &pm,
 void mlirToRLWEPipeline(OpPassManager &pm,
                         const MlirToRLWEPipelineOptions &options,
                         const RLWEScheme scheme) {
-  mlirToSecretArithmeticPipelineBuilder(pm, options);
+  if (options.enableArithmetization) {
+    mlirToSecretArithmeticPipelineBuilder(pm, options);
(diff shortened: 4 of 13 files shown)

### CODE REVIEW

**Comment by ZenithalHourlyRate** [comment]
> 1. `` heir-opt: lib/Analysis/NoiseAnalysis/Noise.cpp:30: NoiseState mlir::heir::NoiseState::operator+(const NoiseState &) const: Assertion `isKnown() && rhs.isKnown()' failed `` so it seems like there's some noise-analysis initialization happening in the arithmetization pipeline that we're missing? @ZenithalHourlyRate ?

It does not call wrap-generic

**Comment by AlexanderViand** [comment]
🤦 Great catch! That should hopefully be easy to fix xD

**Comment by AlexanderViand** [comment]
I've now changed it so that if the `skip-arithmetization-stage` flag is set, we still `--wrap-generic` which works great 🎉 

I've tried to have it also do `--add-client-interface` but that failed with 
``` error: function argument at index 0 missing original type attribute func.func @foo(%x: i16 {secret.secret}, %y: i16 {secret.secret}) -> i16 {``` so I added an option to that pass to just assume the current type is the intended one.

However, I also needed to supply a layout, and while trying to create a dummy 1-1 layout, I noticed that the compiler generally assumes a LayoutAttr has all its components (e.g. `alignment`), despite having a constructor that lets the alignment be nullptr, so that send me down a bit of a rabbit hole xD 

The current state works, but I need to clean up my additions to AddClientInterface before this is ready for review 😅

**Comment by j2kun** [comment]
We could make add-client-interface have an option to not insert any assign_layout ops, rather than try to assign it a layout explicitly.

**Comment by AlexanderViand** [comment]
> We could make add-client-interface have an option to not insert any assign_layout ops, rather than try to assign it a layout explicitly.

Oh, that's much cleaner, I switched to that approach! 

EDIT: updated to remove unrelated file change that had somehow gotten into this commit

**Comment by AlexanderViand** [comment]
@edwjchen  Let me know if this helps your use case in Rotom! 

For reference, here's the *.cpp output produced from the test case added in this PR:
Note: it'll still generate the "repeat to fill up all slots" code, but if the tensors have the correct size already, it'll just be a 1:1 copy.
```cpp
#include "openfhe/pke/openfhe.h"  // from @openfhe

using namespace lbcrypto;
using CiphertextT = ConstCiphertext<DCRTPoly>;
using MutableCiphertextT = Ciphertext<DCRTPoly>;
using CCParamsT = CCParams<CryptoContextBGVRNS>;
using CryptoContextT = CryptoContext<DCRTPoly>;
using EvalKeyT = EvalKey<DCRTPoly>;
using PlaintextT = Plaintext;
using PrivateKeyT = PrivateKey<DCRTPoly>;
using PublicKeyT = PublicKey<DCRTPoly>;

CiphertextT foo(CryptoContextT cc, CiphertextT ct, CiphertextT ct1) {
  const auto& ct2 = cc->EvalAdd(ct, ct1);
  const auto& ct3 = cc->EvalSub(ct, ct1);
  const auto& ct4 = cc->EvalMultNoRelin(ct2, ct3);
  const auto& ct5 = cc->Relinearize(ct4);
  const auto& ct6 = cc->ModReduce(ct5);
  return ct6;
}
CiphertextT foo__encrypt__arg0(CryptoContextT cc, std::vector<int16_t> v0, PublicKeyT pk) {
  std::vector<int64_t> v1(std::begin(v0), std::end(v0));
  auto pt_filled_n = cc->GetCryptoParameters()->GetElementParams()->GetRingDimension() / 2;
  auto pt_filled = v1;
  pt_filled.clear();
  pt_filled.reserve(pt_filled_n);
  for (unsigned i = 0; i < pt_filled_n; ++i) {
    pt_filled.push_back(v1[i % v1.size()]);
  }
  const auto& pt = cc->MakePackedPlaintext(pt_filled);
  const auto& ct = cc->Encrypt(pk, pt);
  return ct;
}
CiphertextT foo__encrypt__arg1(CryptoContextT cc, std::vector<int16_t> v0, PublicKeyT pk) {
  std::vector<int64_t> v1(std::begin(v0), std::end(v0));
  auto pt_filled_n = cc->GetCryptoParameters()->GetElementParams()->GetRingDimension() / 2;
  auto pt_filled = v1;
  pt_filled.clear();
  pt_filled.reserve(pt_filled_n);
  for (unsigned i = 0; i < pt_filled_n; ++i) {
    pt_filled.push_back(v1[i % v1.size()]);
  }
  const auto& pt = cc->MakePackedPlaintext(pt_filled);
  const auto& ct = cc->Encrypt(pk, pt);
  return ct;
}
std::vector<int16_t> foo__decrypt__result0(CryptoContextT cc, CiphertextT ct, PrivateKeyT sk) {
  PlaintextT pt;
  cc->Decrypt(sk, ct, &pt);
  pt->SetLength(1024);
  const auto& v0_cast = pt->GetPackedValue();
  std::vector<int16_t> v0(std::begin(v0_cast), std::end(v0_cast));
  return v0;
}
CryptoContextT foo__generate_crypto_context() {
  CCParamsT params;
  params.SetMultiplicativeDepth(1);
  params.SetPlaintextModulus(65537);
  params.SetEvalAddCount(2);
  params.SetKeySwitchCount(1);
  params.SetKeySwitchTechnique(HYBRID);
  CryptoContextT cc = GenCryptoContext(params);
  cc->Enable(PKE);
  cc->Enable(KEYSWITCH);
  cc->Enable(LEVELEDSHE);
  return cc;
}
CryptoContextT foo__configure_crypto_context(CryptoContextT cc, PrivateKeyT sk) {
  cc->EvalMultKeyGen(sk);
  return cc;
}
```

EDIT: now with 100% more C++

AlexanderViand:
I'm also thinking about adding a flag to the OpenFHE emitter to suppress the emission of the "fill-up-all-slots" logic... That way, you'd get a hard error from OpenFHE if your tensor size was wrong, rather than potentially confusing unexpected behavior.

AlexanderViand:
Here's the *.cpp output with `--skip-vector-resizing`:

```cpp
#include "openfhe/pke/openfhe.h"  // from @openfhe

using namespace lbcrypto;
using CiphertextT = ConstCiphertext<DCRTPoly>;
using MutableCiphertextT = Ciphertext<DCRTPoly>;
using CCParamsT = CCParams<CryptoContextBGVRNS>;
using CryptoContextT = CryptoContext<DCRTPoly>;
using EvalKeyT = EvalKey<DCRTPoly>;
using PlaintextT = Plaintext;
using PrivateKeyT = PrivateKey<DCRTPoly>;
using PublicKeyT = PublicKey<DCRTPoly>;

CiphertextT foo(CryptoContextT cc, CiphertextT ct, CiphertextT ct1) {
  const auto& ct2 = cc->EvalAdd(ct, ct1);
  const auto& ct3 = cc->EvalSub(ct, ct1);
  const auto& ct4 = cc->EvalMultNoRelin(ct2, ct3);
  const auto& ct5 = cc->Relinearize(ct4);
  const auto& ct6 = cc->ModReduce(ct5);
  return ct6;
}
CiphertextT foo__encrypt__arg0(CryptoContextT cc, std::vector<int16_t> v0, PublicKeyT pk) {
  std::vector<int64_t> v1(std::begin(v0), std::end(v0));
  const auto& pt = cc->MakePackedPlaintext(v1);
  const auto& ct = cc->Encrypt(pk, pt);
  return ct;
}
CiphertextT foo__encrypt__arg1(CryptoContextT cc, std::vector<int16_t> v0, PublicKeyT pk) {
  std::vector<int64_t> v1(std::begin(v0), std::end(v0));
  const auto& pt = cc->MakePackedPlaintext(v1);
  const auto& ct = cc->Encrypt(pk, pt);
  return ct;
}
std::vector<int16_t> foo__decrypt__result0(CryptoContextT cc, CiphertextT ct, PrivateKeyT sk) {
  PlaintextT pt;
  cc->Decrypt(sk, ct, &pt);
  pt->SetLength(1024);
  const auto& v0_cast = pt->GetPackedValue();
  std::vector<int16_t> v0(std::begin(v0_cast), std::end(v0_cast));
  return v0;
}
CryptoContextT foo__generate_crypto_context() {
  CCParamsT params;
  params.SetMultiplicativeDepth(1);
  params.SetPlaintextModulus(65537);
  params.SetEvalAddCount(2);
  params.SetKeySwitchCount(1);
  params.SetKeySwitchTechnique(HYBRID);
  CryptoContextT cc = GenCryptoContext(params);
  cc->Enable(PKE);
  cc->Enable(KEYSWITCH);
  cc->Enable(LEVELEDSHE);
  return cc;
}
CryptoContextT foo__configure_crypto_context(CryptoContextT cc, PrivateKeyT sk) {
  cc->EvalMultKeyGen(sk);
  return cc;
}
```

lib/Pipelines/ArithmeticPipelineRegistration.h
```diff
@@ -28,6 +28,12 @@ void heirSIMDVectorizerPipelineBuilder(OpPassManager &manager,
                                        bool disableLoopUnroll);
 
 struct MlirToRLWEPipelineOptions : public SimdVectorizerOptions {
+  PassOptions::Option<bool> skipArithmetization{
+      *this, "skip-arithmetization",
```
j2kun: Generally speaking it's better to have all flags be phrased in terms of enabling a feature, rather than have some flags enable and some flags disable. Could we change this to `enable-automatic-arithmetization` with a default of true?

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
```diff
@@ -1191,60 +1193,73 @@ LogicalResult OpenFhePkeEmitter::printOperation(
...
-  os << inputVarFilledName << ".clear();\n";
-  os << inputVarFilledName << ".reserve(" << inputVarFilledLengthName << ");\n";
-  // inputVarFilledLengthName is unsigned
-  os << "for (unsigned i = 0; i < " << inputVarFilledLengthName << "; ++i) {\n";
-  os << "  " << inputVarFilledName << ".push_back(" << inputVarName << "[i % "
-     << inputVarName << ".size()]);\n";
-  os << "}\n";
+  if (skipVectorResizing_) {
+    emitAutoAssignPrefix(op.getResult());
+    os << cc << "->MakePackedPlaintext(" << inputVarName << ");\n";
```
j2kun, line 1202: Readability nit: prefer to `return success()` in the "early exit" branch and avoid an indented block in the else case. Less indentation -> easier to read.

Same below on line 1240

tests/Regression/issue_1919.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: heir-opt %s "--mlir-to-bgv=skip-arithmetization=true" | FileCheck %s
+
+// A program that does not require any kind of arithmetization (i.e., already consists purely of operations that we can lower directly to FHE)
+// should be able to be lowered to FHE successfully without running the arithmetization stage.
+
+// CHECK: @foo
```
j2kun, line 6: Can we add an assertion somewhere about the absence of an op that would suggest an automated layout was used? Otherwise it's not clear that this test case would fail if the flag were flipped.

Since this PR also touches the openfhe emitter, I would expect either a unit test of the openfhe emitter or (better) an e2e test of the entire pipeline using this new flag. At least, to ensure OpenFHE doesn't err when the correct slot size is used.

In particular, I was under the impression that Edward's example would cause OpenFHE to err, and that some extra setting on the crypto context was necessary to make it ignore the implied insecurity when using a small slot size; perhaps that should be a different flag. Could we port Edward's working example to an e2e test? cc @edwjchen

j2kun (changes requested):

tests/Regression/issue_1919.mlir
edwjchen: @j2kun should I add the e2e rotom matmul example to this pr or create a new one?

Without skipping the arithmetization pass, my test will fail since the input layout aren't aligned properly with my circuit. However, I've fixed the example to use the correct slot size from OpenFHE (16k or more); I just had to repeat the  values in the packing.

Lastly, although my example only has a single multiplication depth, the example only works when setting `params.SetMultiplicativeDepth` to be 2 or greater. Can this parameter be set with existing flags?

j2kun: If it's simple enough, can we add it to this PR, @AlexanderViand ? 

> I just had to repeat the values in the packing.

This is exactly what the OpenFHE changes in this PR _revert_. Should we instead keep this repetition behavior in the OpenFHE emitter, if we're skipping it only to have you do it manually?

j2kun: > the example only works when setting params.SetMultiplicativeDepth to be 2 or greater. Can this parameter be set with existing flags?

We have passes that should correctly identify the multiplicative depth, but sometimes it seems OpenFHE will not generate correct parameters when the multiplicative depth provided is too small. If we put your example in this PR I think we should be able to get to the bottom of it; last I checked there were a few commits that went in recently that improve the situation. Also note we're in the process of upgrading to OpenFHE v1.3 which should also have some differences in parameter settings.

edwjchen: @AlexanderViand the test is very simple! it's similar to adding the dot_product example in tests/Examples/openfhe/ckks/dot_product_8f

edwjchen: > This is exactly what the OpenFHE changes in this PR revert. Should we instead keep this repetition behavior in the OpenFHE emitter, if we're skipping it only to have you do it manually?

@j2kun the other change that I would need to avoid (apart for repeating values) is the layout packing currently determined by HEIR. Specifically, it's the code used to set the layout before filling in `pt_filled`.

for example, rather than using `v4135` to create `pt_filled`, I would directly pass in the repeated packed vector from the function arguments. 
```c++
for (auto v4135 = 0; v4135 < 16384; ++v4135) {
    float v4137 = v7[v4135];
    size_t v4138 = v4135 % v5;
    bool v4139 = v4138 < v4;
    size_t v4140 = v4138 + v5;
    size_t v4141 = v4139 ? v4140 : v4138;
    v4134[v4141] = v4137;
  }
  ```

j2kun: Right, I think we're in agreement, but just for context: there are two different unrelated changes being controlled by this flag: the packing pipeline and a fallback mechanism in the openfhe emitter. The packing pipeline may produce a packing using, say, 4096 slots, while openfhe requires 8192 to meet its security requirement. In this case, for both encrypted inputs and server side plaintexts, there is a default behavior of cyclic repetition to ensure we use as many slots as OpenFHE requires. I'm writing this from my phone so I don't recall all the details, but my main point here is that we could only skip the HEIR packing step and leave in the "fallback" step, and then you wouldn't have to manually cyclically replicate your inputs. Alex has a point I think that making the user replicate manually as an acknowledgement that this is compatible with the user's manual layout, but also I haven't seen an example where cyclic repetition isn't the right thing to do.

edwjchen: I see! that makes sense to me too

AlexanderViand: I've addressed most of the comments, except for adding a test for the (separate) OpenFHE emitter flag and, more importantly, actually looking at the rotom e2e test and setting that up properly.

EDIT: found the OpenFHE emitter tests, turns out they're in tests/Dialect/CKKS/Emitter instead of tests/Emitter 🤷‍♂️

j2kun: > turns out they're in tests/Dialect/CKKS/Emitter instead of tests/Emitter

Yeah that annoyed me as well. We should reorganize them to match the `lib/` organization

asraa: https://github.com/google/heir/pull/1932 

although i would be down to put emitters in lib/Dialect/Openfhe/Emitters as well maybe

AlexanderViand:
Ok, this should now work for @edwjchen's use case:

 `heir-opt --mlir-to-ckks="enable-arithmetization=false ciphertext-degree=<tensor-size> --scheme-to-openfhe`.

In addition to the new arithmetization flag, you also need to set the (poorly named) ciphertext-degree to the tensor size you're using in the program. Otherwise, you'll get a somewhat obscure error message about not being able to lower `secret.conceal` which is caused by the tensor size not matching the number of slots HEIR expected (1024 by default, iirc?).

You should then be able to generate OpenFHE code as usual (`heir-translate --emit-openfhe-pke`), which will generate encryption helpers that don't try to do any kind of fancy packing, but will still do cyclical repetition. Should the tensor size you're using happen to already match the size OpenFHE expects, you can do `heir-translate --emit-openfhe-pke --skip-vector-resizing` to avoid the redundant copying.

PS: I moved the rotom test to `tests/Examples/openfhe/ckks/custom_arithmetization` and hooked it into our usual e2e testing setup

j2kun (approved):

PR #1927, opened 2025-06-24
Refactor Log2Arithmetic into standalone file

ZenithalHourlyRate:
This PR refactors the arithmetic in log scale code in `Noise.h` into a standalone file so that others can benefit from it, e.g. CKKS Range Analysis which also needs to record the range in log scale.

### DIFF
--- a/lib/Utils/LogArithmetic.cpp
+++ b/lib/Utils/LogArithmetic.cpp
@@ -1,0 +1,44 @@
+#include "lib/Utils/LogArithmetic.h"
+
+namespace mlir {
+namespace heir {
+
+Log2Arithmetic Log2Arithmetic::operator+(const Log2Arithmetic &rhs) const {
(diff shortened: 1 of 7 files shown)


lib/Utils/LogArithmetic.cpp
j2kun, line 1: It may not be necessary, but you could potentially add unit tests of the functionality here, rather than state the expected behavior (e.g., for negative infinity) in comments.

j2kun (approved):

lib/Utils/LogArithmetic.cpp
j2kun: Let me know if you plan to make any changes based on this, otherwise we can merge it.

ZenithalHourlyRate: Unit tests added.

PR #1928, opened 2025-06-24
ModArith: Add CRT support

znssong:
This PR introduces support for Chinese Remainder Theorem (CRT) decomposition and interpolation. 

Key changes:
1. Extended the `mod_arith.mod_switch` operation to support conversion between `ModArith` type and `RNS` type, i.e. CRT decomposition and interpolation.
2. Enhanced the `mod_switch.encapsulate` and `mod_switch.extract` ops to handle conversion between (tensor of) RNS types and tensor of integers.


lib/Dialect/ModArith/IR/ModArithOps.td
```diff
@@ -20,68 +20,76 @@ class ModArith_Op<string mnemonic, list<Trait> traits = [Pure]> :
...
-def ModArith_ModSwitchOp : ModArith_Op<"mod_switch", [Pure, SameOperandsAndResultShape]> {
+def ModArith_ModSwitchOp : ModArith_Op<"mod_switch", [Pure]> {
   let summary = "change the modulus of a mod_arith";
 
   let description = [{
-    "mod_switch" operation to change the modulus of a mod_arith type to a
-    bigger space.
+    The "mod_switch" operation performs either modulus switching (changing the
+    modulus of a mod_arith type to a new value by reducing modulo the new
+    modulus) or CRT decomposition/interpolation.
```
ZenithalHourlyRate, line 79: We may add some comments here about the behavior of changing modulus of a mod_arith type to smaller space; possibly citing https://github.com/openfheorg/openfhe-development/blob/02a8e9c76c3e2eff53392530199c63e4da53eb65/src/core/lib/math/hal/bigintdyn/mubintvecdyn.cpp#L139-L165

The reason is that in FHE people often use the range [-q/2, q/2] but in actual implementation [0, q] is more convenient for storage. An example is that $-1 \equiv q - 1$ in the storage. When we switch from large $q$ to smaller $p$, if we do not carefully handle it, the new value would be $q-1 \pmod{p}$ and is not -1 anymore.

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -250,6 +269,113 @@ struct ConvertMac : public OpConversionPattern<MacOp> {
...
+      bigModulus = bigModulus.zext(width);
+      auto integerType = IntegerType::get(op.getContext(), width);
+      auto intConst = getInteger(&b, integerType);
+      auto bigModulusConst = intConst(bigModulus);
+      Value sum = intConst(APInt::getZero(width));
+      for (int64_t i = 0; i < inputRNS.getBasisTypes().size(); i++) {
+        auto modArithType = cast<ModArithType>(inputRNS.getBasisTypes()[i]);
+        auto modulus = modArithType.getModulus().getValue();
+        modulus = modulus.zext(width);
+        auto tmp = bigModulus.udiv(modulus);
```
ZenithalHourlyRate, line 372: We may need a verifier that ensures bigModulus = \prod smaller RNS Modulus

tests/Dialect/ModArith/Conversions/mod_arith_to_arith/runner/lower_mod_switch_decompose.mlir
ZenithalHourlyRate, line 1: Besides the runner tests, corresponding `mod-arith-to-arith.mlir` tests are also needed.

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -250,6 +269,113 @@ struct ConvertMac : public OpConversionPattern<MacOp> {
...
+        rewriter.replaceOp(op, tensor);
+      } else {
+        llvm_unreachable("Verifier should make sure this doesn't happen.");
+      }
+    } else if (auto inputRNS = dyn_cast<rns::RNSType>(inputType)) {
+      auto outputModArith = cast<ModArithType>(outputType);
+      auto bigModulus = outputModArith.getModulus().getValue();
+      auto outputWidth = bigModulus.getBitWidth();
+      auto width =
+          2 * outputWidth + llvm::Log2_64_Ceil(inputRNS.getBasisTypes().size());
```
ZenithalHourlyRate: Bit confused how such bitwidth is calculated.

The following discussion use symbols from [here](https://github.com/openfheorg/openfhe-development/blob/02a8e9c76c3e2eff53392530199c63e4da53eb65/src/core/include/lattice/hal/default/dcrtpoly-impl.h#L779-L787). OpenFHE uses arbitrary precision integer, yet in LLVM we only have APInt that is not adaptive on integer width.

The width of `qt/qi` is bounded by `outputWidth`; The width of `[(qt/qi)^(-1) mod qi]` is bouded by `qi`, and the width of `M[j]` (`extract = b.create<tensor::ExtractOp>(adaptor.getInput(), ValueRange{index})` in our case) is bounded by `qi`. So the width for multiplication result should be `outputWidth + 2 * width(qi)`, often `2 * outputWidth` is a good bound but it is not guaranteed that `2 * width(qi) < outputWidth`.

`llvm::Log2_64_Ceil(inputRNS.getBasisTypes().size());` comes from addition of all RNS limbs and is reasonable.

ZenithalHourlyRate (approved):
LGTM! Though CI needs fixing.

j2kun:
Looks like some small linter errors. pre-commit should fix these automatically

j2kun:
The segfault in roberts cross can be ignored. Cf. https://github.com/google/heir/issues/1988

znssong:
> The segfault in roberts cross can be ignored. Cf. #1988

OK, thanks.

lib/Dialect/ModArith/IR/ModArithOps.td
```diff
@@ -20,68 +20,76 @@ class ModArith_Op<string mnemonic, list<Trait> traits = [Pure]> :
...
   let arguments = (ins
     SignlessIntegerLike:$input
   );
-  let results = (outs ModArithLike:$output);
+  let results = (outs ModArithOrRNSLike:$output);
+  let hasVerifier = 1;
   let assemblyFormat = "operands attr-dict `:` type($input) `->` type($output)";
 }
 
-def ModArith_ExtractOp : ModArith_Op<"extract", [Pure, ElementwiseMappable]> {
```
j2kun, line 43: Nit: is there a reason to remove ElementwiseMappable here, since the type constraints still allow tensor types? Also for EncapsulateOp above.

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -250,6 +269,127 @@ struct ConvertMac : public OpConversionPattern<MacOp> {
...
+  ConvertModSwitch(mlir::MLIRContext *context)
+      : OpConversionPattern<ModSwitchOp>(context) {}
+
+  using OpConversionPattern::OpConversionPattern;
+
+  auto getInteger(ImplicitLocOpBuilder *b, IntegerType integerType) const {
+    return [=](const APInt &value) mutable {
+      return b->create<arith::ConstantOp>(IntegerAttr::get(integerType, value));
+    };
+  }
```
j2kun, line 282: This can be simplified (perhaps enough to inline) via `b.create<arith::ConstantIntOp>(integerType, value)`

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -250,6 +269,127 @@ struct ConvertMac : public OpConversionPattern<MacOp> {
...
+        auto diffConst = intConst(newModulus - oldModulus.urem(newModulus));
+        auto remuAdd = b.create<arith::AddIOp>(remu, diffConst);
+        auto remuAddMod =
+            b.create<arith::RemUIOp>(remuAdd, intConst(newModulus));
+        Value result = b.create<arith::SelectOp>(cmp, remu, remuAddMod);
+        if (newWidth < oldWidth) {
+          auto outputType = IntegerType::get(op.getContext(), newWidth);
+          result = b.create<arith::TruncIOp>(outputType, result);
+        }
+        rewriter.replaceOp(op, result);
```
j2kun, line 318: I think this would be easier to read if either the different cases were separated into different functions, or else the nesting was reduced by returning early at the end of each block (with slightly more verbose conditions). E.g.,

```
if (auto inputModArith = dyn_cast<ModArithType>(inputType) && auto outputModArith = dyn_cast<ModArithType>(outputType)) {
   ...
   return success();
}

if (auto inputModArith = dyn_cast<ModArithType>(inputType) && auto outputRns = dyn_cast<rns::RNSType>(outputType)) {
   ...
   return success();
}

...
```

lib/Dialect/ModArith/Conversions/ModArithToArith/ModArithToArith.cpp
```diff
@@ -250,6 +269,127 @@ struct ConvertMac : public OpConversionPattern<MacOp> {
...
+      bigModulus = bigModulus.zext(width);
+      auto integerType = IntegerType::get(op.getContext(), width);
+      auto intConst = getInteger(&b, integerType);
+      auto bigModulusConst = intConst(bigModulus);
+      Value sum = intConst(APInt::getZero(width));
+      for (int64_t i = 0; i < inputRNS.getBasisTypes().size(); i++) {
+        auto modArithType = cast<ModArithType>(inputRNS.getBasisTypes()[i]);
+        auto modulus = modArithType.getModulus().getValue();
+        modulus = modulus.zext(width);
+        auto tmp = bigModulus.udiv(modulus);
```
j2kun, line 372: It would help slightly to have the variable names match the formulas above, or to have a comment saying which is which. `bigModulus` = `qt` and `modulus` = `qi`

j2kun (approved):
Some minor feedback if you have time before we merge it later this week.

PR #1931, opened 2025-06-25
Mention missing pre-commit dependencies in development.md

AlexanderViand:
At some point the pre-commit hooks were expanded and they now require ruby and go, as well as being hardcoded to python 3.11 but neither was mentioned in the documentation.

Noticed this because I forgot to renew my dev machine reservation *again* and had to run through set-up again🙈

Keeping this as draft in case I notice more stuff 😉


docs/content/en/docs/development.md
```diff
@@ -34,7 +34,7 @@ For the best experience, we recommend following these steps:
   Just as with bazel, you will want to move this somewhere on your PATH, e.g.:
 
   ```bash
-  mkdir ~/bin
+  mkdir -p ~/bin
```
j2kun, line 37: Does this change imply there are systems with no home directory, but for which `~` has a meaningful resolution in the shell?

AlexanderViand: Probably not xD -  I'm just using it because it won't throw a "cannot create directory <...>: File exists" (we tell people to mkdir ~/bin both for bazel in getting_started.md and for buildifier in development.md)

PR #1921, opened 2025-06-17
BFV: Add Canonical Embedding based Noise Model

ZenithalHourlyRate:
This PR adds a noise model for BFV using canonical embedding model. It is similar to the BGV one.

PR #1948, opened 2025-07-02
Cleanup include guard

ZenithalHourlyRate:
After the change of directory structure, the include guard of some header file is not changed accordingly. For these Noise related headers, they were made pre-date that change.

See also https://github.com/google/heir/pull/1921#discussion_r2178380640


j2kun (approved):

asraa:
There might be some rebase conflicts with this once #1927 gets in so I'll ping if there are

Issue #1490, opened 2025-02-27
Make the python frontend pip-installable


j2kun:
This requires us to set up a hatch build, and ship the built openfhe dependencies, and then publish it on PyPI and test that it works on a bunch of platforms.

j2kun:
This is mostly done (except for a CI workflow to actually publish the package), but then I learned that rules_python has packaging rules.

https://rules-python.readthedocs.io/en/latest/api/rules_python/python/packaging.html

It says that the `py_wheel` rule only supports pure-python wheels, but then I found  https://github.com/EnzymeAD/Enzyme-JAX/tree/main which somehow packages a python wheel for an MLIR project with python bindings of the underlying MLIR dialects... Something to look into.

PR #1952, opened 2025-07-03
More improvements in preparation for publishing to PyPI

j2kun:
- Enable the PyPI upload step of the workflow (using trusted publishers on PyPI). Workflow will be triggered on a release event from `main`.
- Rewrite the README to point people to the python package.
- Copy binaries to the `bin/` folder of the wheel so users can access the underlying `heir-opt` more easily.
- Moves cibuildwheel config to pyproject.toml.
- Adds windows/mac support.

Fixes https://github.com/google/heir/issues/1490


j2kun:
I already manually pushed 0.0.1 to https://pypi.org/project/heir-py/0.0.1/, so you can test it out if you like.

j2kun:
This is ready for review. I think we should have a separate conversation about the release process we'd like to institute before we start pushing releases.

asraa (approved):

PR #1961, opened 2025-07-07
remove cmake build

j2kun:
According to Muthu, Cornami (the main people interested in having the CMake build) have migrated to using bazel to keep up to date with HEIR at head easier. So there's no reason to keep the CMake build files around.


asraa (approved):
I'm in favor of this - any reason to keep this @AlexanderViand ? (CI was disabled too anyway right?)

AlexanderViand (approved):
Yep, cornami was the only cmake user afaik

Was cmake ever mentioned in the documentation?

j2kun:
I grepped the codebase and only found mention of it in the context of openfhe and LLVM

PR #1966, opened 2025-07-08
frontend: fix support for reassigning values in binary ops

asraa:
This fixes an issue with parsing the numba IR to MLIR. Vars in numba IR can be reassigned and this would cause MLIR to complain that SSA values were re-defined.

I'm not sure why this only happens with some reassignments, but I added a testcase where this does happen.

### DIFF
--- a/frontend/heir/mlir_emitter.py
+++ b/frontend/heir/mlir_emitter.py
@@ -475,6 +475,19 @@ class TextualMlirEmitter:
   def forward_name_to_id(self, from_var, to_str):
     self.numba_names_to_ssa_var_names[from_var.name] = to_str
 
+  def reassign_and_forward_name(self, var, expr):
+    # returns a string that sets var equal to the expr. this helper
+    # handles assigning var to a new SSA value if it was already defined
+    is_assigned = self.has_name(var)
+    if is_assigned:
+      name = self.get_next_name()
+    else:
+      name = self.get_or_create_name(var)
+    assign_str = f"{name} = {expr}"
+    if is_assigned:
+      self.forward_name_to_id(var, name.strip("%"))
+    return assign_str
+
   def emit_assign(self, assign):
     match assign.value:
       case ir.Arg():
@@ -489,12 +502,10 @@ class TextualMlirEmitter:
         return ""
       case ir.Expr(op="binop"):
         emitted_expr, ext, ty = self.emit_binop(assign.value)
-        name = self.get_or_create_name(assign.target)
-        return (
-            f"{ext}"
-            f"{name} = {emitted_expr} :"
-            f" {mlirType(ty)} {mlirLoc(assign.loc)}"
-        )
+        expr = f"{emitted_expr} : {mlirType(ty)} {mlirLoc(assign.loc)}"
+        # if the var is being reassigned, then create a new SSA var
+        assign_str = self.reassign_and_forward_name(assign.target, expr)
+        return f"{ext}{assign_str}"
       case ir.Expr(op="call"):
         func = assign.value.func
         # if assert fails, variable was undefined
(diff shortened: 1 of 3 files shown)


AlexanderViand:
Doesn't numba literally call their IR class/object `SSA IR`? I guess it's not as SSA as that'd imply?

j2kun:
Happy to merge it in, but I would love to see the offending numba IR if you have it handy.

j2kun (approved):

asraa:
It's not quite SSA IR. The testcase that I committed is one example where it fails, and that is
```
label 0:
    x = arg(0, name=x)                       ['x']
    a = x                                    ['a', 'x']
    $8load_global.1 = global(range: <class 'range'>) ['$8load_global.1']
    $const18.3.1 = const(int, 2)             ['$const18.3.1']
    $20call.4 = call $8load_global.1($const18.3.1, func=$8load_global.1, args=[Var($const18.3.1, loop_test.py:14)], kws=(), vararg=None, varkwarg=None, target=None) ['$20call.4', '$8load_global.1', '$const18.3.1']
    $28get_iter.5 = getiter(value=$20call.4) ['$20call.4', '$28get_iter.5']
    $phi30.0 = $28get_iter.5                 ['$28get_iter.5', '$phi30.0']
    jump 30                                  []
label 30:
    $30for_iter.1 = iternext(value=$phi30.0) ['$30for_iter.1', '$phi30.0']
    $30for_iter.2 = pair_first(value=$30for_iter.1) ['$30for_iter.1', '$30for_iter.2']
    $30for_iter.3 = pair_second(value=$30for_iter.1) ['$30for_iter.1', '$30for_iter.3']
    $phi34.1 = $30for_iter.2                 ['$30for_iter.2', '$phi34.1']
    branch $30for_iter.3, 34, 62             ['$30for_iter.3']
label 34:
    _ = $phi34.1                             ['$phi34.1', '_']
    a0 = a                                   ['a', 'a0']
    $const42.4.2 = const(int, 1)             ['$const42.4.2']
    a1 = a + $const42.4.2                    ['$const42.4.2', 'a', 'a1']
    a = a0 + a1                              ['a', 'a0', 'a1']
    jump 30                                  []
label 62:
    $66return_value.3 = cast(value=a)        ['$66return_value.3', 'a']
    return $66return_value.3                 ['$66return_value.3']
```

You can see the re-assigning of the var a in label 34

We seem to call it SSA IR but in numba it's a "core.ir.FunctionIR" object.

(Also, I had to previously fixed the same issue when re-assigning constants, hence why i got to factor out that logic, so I can also post an offending testcase from that).

asraa:
Here's a simpler example of reassigning to a constant
```
    @compile()
    def reassign(x: Secret[I64]):
      a = 5
      if x == 0:
        a = 2
      return a
```

gives the numba IR
```
label 0:
    x = arg(0, name=x)                       ['x']
    a = const(int, 5)                        ['a']
    $const10.2.2 = const(int, 0)             ['$const10.2.2']
    $12compare_op.3 = x == $const10.2.2      ['$12compare_op.3', '$const10.2.2', 'x']
    bool16 = global(bool: <class 'bool'>)    ['bool16']
    $16pred = call bool16($12compare_op.3, func=bool16, args=(Var($12compare_op.3, loop_test.py:14),), kws=(), vararg=None, varkwarg=None, target=None) ['$12compare_op.3', '$16pred', 'bool16']
    branch $16pred, 18, 22                   ['$16pred']
label 18:
    a = const(int, 2)                        ['a']
    jump 22                                  []
label 22:
    $24return_value.1 = cast(value=a)        ['$24return_value.1', 'a']
    return $24return_value.1                 ['$24return_value.1']
```

I played around a tiny bit with this. My hypothesis is that within a label vars are unique, but label 18 can reassign `a`. (so if I just did `a = 5` and then `a = 2` in the same block the numba vars would be unique, e.g. `a` and `a.1`)

AlexanderViand:
> It's not quite SSA IR.

> We seem to call it SSA IR but in numba it's a "core.ir.FunctionIR" object.

Oh, interesting! Maybe it *used* to be truly SSA at some point? I think I've seen references to "SSA IR" in numba code (especially around type inference) and [old documentation](https://numba.pydata.org/numba-doc/0.8/ir.html) seems to call it SSA, but the [current documentation](https://numba.readthedocs.io/en/stable/proposals/type-inference.html?highlight=mostly+static-single+assignment+(SSA)
) does indeed call Numba IR a "mostly" SSA IR 🤷

Issue #1595, opened 2025-03-20
Implement a real layout conversion cost model


asraa:
In PR #1588 we simply use a binary cost indicating whether there is a layout conversion necessary or not. We should implement a real cost model that can estimate the cost of converting layouts taking into account the number of rotations / adds / muls.

j2kun:
An initial approach can be to refactor the code in implement-shift-networks so that the shift network construction is simulated and the cost derived from the number of low-level rotate+mul ops produced.

johnbntn:
Is anyone currently working on this? I'd like to

j2kun:
Nobody yet as far as I know. I may be making some changes to the layout system more generally in the near future, but this part shouldn't be affected too much. Also open to any other ideas you have for a cost model.

johnbntn:
Sounds good! I think I'll start by looking at the shift network code because I also want to understand the repo a bit more but I'll definitely try to think of other ways to implement it as well

j2kun:
For what it's worth, the shift network pass hasn't been integrated into the larger pipeline yet. So that may be a low hanging fruit to start if you want to study that part of the codebase. It also probably needs some more testing and bug fixing as a consequence of not being integrated.

PR #1967, opened 2025-07-09
Implement a layout conversion cost analysis

johnbntn:
Implemented a basic layout conversion cost analysis for the `layout-optimization` pass. It reuses the logic for creating a permutation/shift network from the `convert-to-ciphertext-semantics` and `implement-shift-network` passes, respectively.

To estimate the cost, I compare the number of rounds required in every rotation group and return the largest.

I moved two functions `getMinArgUnusedTarget()` and `getMinUnusedInput()` to `TransformUtils` and a lot of the code from `implement-shift-network` to the header file so I'm not duplicating too much.

This is from #1595

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
+++ b/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
@@ -1,7 +1,11 @@
 #ifndef LIB_DIALECT_TENSOREXT_TRANSFORMS_IMPLEMENTSHIFTNETWORK_H_
 #define LIB_DIALECT_TENSOREXT_TRANSFORMS_IMPLEMENTSHIFTNETWORK_H_
 
-#include "mlir/include/mlir/Pass/Pass.h"  // from @llvm-project
+#include "lib/Utils/ADT/FrozenVector.h"
+#include "lib/Utils/AffineMapUtils.h"
+#include "lib/Utils/Graph/Graph.h"
+#include "llvm/include/llvm/Support/Debug.h"  // from @llvm-project
+#include "mlir/include/mlir/Pass/Pass.h"      // from @llvm-project
 
 namespace mlir {
 namespace heir {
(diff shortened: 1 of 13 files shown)


lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -1110,6 +1075,8 @@ struct ConvertToCiphertextSemantics
 
     if (failed(applyContextAwarePartialConversion(module, target,
                                                   std::move(patterns)))) {
+      LLVM_DEBUG(llvm::dbgs() << "ts failed bruh\n");
+
```
ZenithalHourlyRate: clean up random debug message.

MODULE.bazel.lock
ZenithalHourlyRate, line 1: the lock file change should not be checked in (I know it is annoying).

ZenithalHourlyRate (commented):
Just some nitpicking comments for making this PR look better.

Please squash the commits into one commit. We require single commit upon approval.

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
johnbntn: Ah my bad, thought I got all of them

lib/Transforms/LayoutOptimization/LayoutOptimization.cpp
```diff
@@ -54,10 +55,8 @@ auto &kLayoutAttrName = tensor_ext::TensorExtDialect::kLayoutAttrName;
 
 using Cost = int64_t;
 
-// TODO(#1595): Implement a more accurate cost model.
-Cost computeCostOfLayoutConversion(LayoutAttr fromLayout, LayoutAttr toLayout) {
-  return (fromLayout == toLayout) ? 0 : 1;
-}
+// keep track of tensor with largest # of slots for layout analysis
+int64_t maxSlots = 0;
```
j2kun: We have the total number of slots available in the pipeline at the time this pass is constructed (cf. https://github.com/google/heir/blob/main/lib/Pipelines/ArithmeticPipelineRegistration.cpp#L143, the preceding `options.ciphertextDegree` is given to the pass just before layout-optimization), so you could get this by adding the pass option to `LayoutOptimization.td` in the same way it's passed to `LayoutPropagation.td`.

lib/Utils/TransformUtils.h
```diff
@@ -26,6 +26,13 @@ Value convertIntegerValueToMemrefOfBits(Value integer, OpBuilder &b,
 Value convertMemrefOfBitsToInteger(Value memref, Type resultType, OpBuilder &b,
                                    Location loc);
 
+// Return the first unused input index not mapped from by the partial
+// permutation.
```
j2kun: Since this is now in a more public place, maybe add a note how the array ref is interpreted as a "partial permutation" (it's a permutation on `perm.size()`, with `perm[i] = j` mapping `i` to `j`, and the permutation is unset if and only if `j == -1`)

lib/Utils/TransformUtils.cpp
```diff
@@ -100,5 +106,41 @@ Value convertMemrefOfBitsToInteger(Value memref, Type resultType, OpBuilder &b,
   return result;
 }
 
+// Return the first output index not mapped to by the partial permutation.
```
j2kun: The comments are not needed here since they're already in the header.

```suggestion
```

lib/Utils/TransformUtils.cpp
```diff
@@ -100,5 +106,41 @@ Value convertMemrefOfBitsToInteger(Value memref, Type resultType, OpBuilder &b,
...
+      llvm::dbgs() << i << " ";
+    }
+    llvm::dbgs() << "\n";
+  });
+
+  return *unmappedOutputs.begin();
+}
+
+// Return the first unused input index not mapped from by the partial
+// permutation.
```
j2kun:
```suggestion
```

lib/Transforms/LayoutOptimization/LayoutConversionCost.cpp
```diff
@@ -0,0 +1,88 @@
...
+        llvm::dbgs() << val << ' ';
+      };
+    } llvm::dbgs() << "\n";);
+
+    iterateIndices(dataSemanticShape, evaluateNextIndex);
+    minUnusedTarget = getMinUnusedTarget(permutation);
+    minUnusedInput = getMinUnusedInput(permutation);
+  }
+
+  for (int64_t rotation : uniqueRotations) {
```
j2kun: So the above creates the permutation, but once you have the permutation it is possible to represent it much more efficiently than just counting the total number of rotations used in the above (naive) decomposition.

One of the more recent methods, which we implemented in `VosVosErkinShiftNetworks.cpp` (and I blogged about [here](https://www.jeremykun.com/2024/09/02/shift-networks/) though I think there was a bug in my python code I never fixed), is used in the [implement-shift-networks pass](https://github.com/google/heir/blob/e23475ece7467500d226913963b4a19e5abec411/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp#L351-L373).

I think it should be straightforward to extract the part that converts a permutation to a set of rotation groups into a utility function, so that both that pass and this cost function can use it. Then the cost could be (tentatively, it would depend on the backend in reality) the size of the largest rotation group, which would be something like the depth of the shift network, which would correlate to the latency if you were able to run all shift network rotation groups in parallel.

lib/Transforms/LayoutOptimization/LayoutOptimization.cpp
johnbntn: Totally missed that, thanks! Also modified one of the `layout-optimization` tests to use  the option.

johnbntn:
I think I have everything implemented, lmk if you have any questions

j2kun:
Thanks! I'll take a final look at it tomorrow or Friday.

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
```diff
@@ -10,6 +16,216 @@ namespace tensor_ext {
...
+};
+
+class ShiftStrategy {
+ public:
+  ShiftStrategy() = default;
+
+  SmallVector<ShiftRound> getRounds() const { return rounds; }
+
+  // Run the shifting strategy and populate the `rounds` member variable.
+  void evaluate(const Permutation &permutation, const RotationGroup &group) {
```
j2kun: Would it be possible to just put the declarations in the header file and the implementations still in the `cpp` file? It's not a blocker but I feel it's a better organizational setup, and avoids issues like importing this header file overwriting the `DEBUG_TYPE`.

johnbntn:
Any updates on this?

asraa:
Ahh! Oops - reviewing this and I'll merge it internally :)

(Update: just waiting on someone to approve my internal change to fix some includes)

PR #1994, opened 2025-07-12
Reorganize/extend developer docs

j2kun:
- Separate IDE page + neovim config
- Separate bazel tips page
- Separate page about boilerplate tools
- Add independent sections about tools like ibazel, buildifier, heir-lsp
- merge the contributing.md page into the Developer/_index.md


docs/content/en/docs/Development/_index.md
```diff
@@ -0,0 +1,101 @@
...
+
+We use [pre-commit](https://pre-commit.com/) to manage a series of git
+pre-commit hooks for the project; for example, each time you commit code, the
+hooks will make sure that your C++ is formatted properly. If your code isn't,
+the hook will format it, so when you try to commit the second time you'll get
+past the hook. Note that spelling mistakes identified by the codespell hook will
+not be auto-corrected and require manual resolution, rather than simply
+re-running pre-commit.
+
+All hooks are defined in `.pre-commit-config.yaml`. To install these hooks, run
```
AlexanderViand: to an inattentive reader, this might sound like `pip install -r ...` is all they need to do, so maybe add a `first`?
```suggestion
All hooks are defined in `.pre-commit-config.yaml`. To install these hooks, first run
```

docs/content/en/docs/Development/_index.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Development
+weight: 30
+---
+
```
AlexanderViand, line 5: Maybe add a short intro pargrapg with a reference back to getting_started here, saying that this assumes you've completed the steps from https://heir.dev/docs/getting_started/#building-from-source?

AlexanderViand: In the intro paragraph, it might also be a good idea to explain what this page will cover and to link out to /bazel, /boilerplate and /ide. I know they'll be in the navigation, too, but people don't always explore those...

Oh, thinking about it, why don't we move the general parts of "Contributing to HEIR"  (everything above https://heir.dev/docs/contributing/#contributing-code-to-heir) to the https://heir.dev "landing page" and move the info about "Contributing code to HEIR"  here (Well, below ## Pre-Commit)? That way, heir.dev/development becomes a more-or-less "one stop shop" for all things development?

On that note: I think `research_with_heir` could also move into a development sub-page (and get a mention/link in the intro paragraph).

docs/content/en/docs/Development/_index.md
```diff
@@ -0,0 +1,101 @@
...
+```
+   -DLLVM_TARGETS_TO_BUILD="Native" \
+```
+
+### OpenFHE
+
+A simple way to build OpenFHE is to follow the instructions in the
+[openfhe-configurator](https://github.com/openfheorg/openfhe-configurator)
+repository. This allows to build the library with or without support for the
+Intel [HEXL library](https://github.com/intel/hexl). First, clone the repository
```
AlexanderViand:
```suggestion
Intel [HEXL library](https://github.com/intel/hexl) which adds AVX512 support.
First, clone the repository
```

docs/content/en/docs/Development/ide.md
```diff
@@ -0,0 +1,316 @@
...
+dialects.
+
+Build the LSP binary, then move it to a location on your path.
+
+```bash
+bazel build //tools:heir-lsp
+cp bazel-bin/tools/heir-lsp /usr/local/bin
+```
+
+Then point your IDE to the LSP in the normal manner.
```
AlexanderViand: I don't recommend this approach, as it means you'll need to remember to re-copy the lsp whenever a new op is added, either upstream or in HEIR. What I end up doing is just pointing the IDE to the bazel-bin versions of heir-lsp.

Also, the docs should include the .vscode/settings.json snippet for this (also for buildifier location, using the ~/bin location we recommend?)

docs/content/en/docs/Development/ide.md
```diff
@@ -0,0 +1,316 @@
+---
+title: IDE configuration
+weight: 50
+---
+
```
AlexanderViand, line 5: Maybe just start with #VSCode? I get that `heir-lsp` and `clangd` are technically more generic (And maybe add a statement to that extent to the "we only support vscode disclaimer"?) but I think it'd be easier for most people if the page worked well top-down as a way to set-up vscode with all the necessary extensions, depenencies and settings.json snippets, with non-vscode users simply reading the same instructions and skipping stuff not relevant for their setup.

AlexanderViand (changes requested):
This is great! I have a feeling most of my comments aren't even about the specific changes made, but just come from re-reading the docs!

I'm also wondering if we should add macOS instructions, too, wherever we have ubuntu/debian-specific instructions (i.e., add `brew install <...>` where we have `apt-get install ...`, especially if the package names differ - otherwise, we can just add a general note that things should work with brew, too.

docs/content/en/docs/Development/_index.md
j2kun: Made a few changes to this effect, though I left the research page alone for now (added a link from the homepage)

docs/content/en/docs/Development/ide.md
j2kun: Updated to give both options.

docs/content/en/docs/Development/ide.md
j2kun: Forcing everyone to read through a bunch of VSCode config to find the stuff relevant for their IDE seems... a bit egocentric :) I can't make sense of the VSCode config, and I don't expect you to make sense of the neovim config.

Also, I don't think the HEIR docs is the right place to teach people about their IDE (e.g., how to run a debugger, or how to use the bazel extension), and it would be better to just link to the dedicated documentation for those tools, or write a blog post that makes it clear when it was published so that people don't expect it to stay up to date as various external tools change beyond our control.

Especially since most people I've seen are more than comfortable running `bazel test` from the VSCode terminal....

AlexanderViand:
Oh, and because I literally just ran into it: can you add some documentation on how to add an exception to codespell in the precommit/codespell part? I thought I remembered how to do that but can't find the example I had anymore

docs/content/en/docs/Development/ide.md
AlexanderViand: Fair enough! I actually don't even use the bazel integration in vscode (it's waaay too slow to update the target list) and run bazel from the terminal, too, but it's still nice to have the formatting/lsp/etc :) 

Maybe we should (in the next iteration, let's get this merged first) just collect all the settings.json stuff into a single "dump this into your settings.json" thing. In addition to the clangd stuff, there's overriding the python formatter with pyink and (not yet documented but highly recommended)  excluding /bazel and /external from python linting/analysis and the files watcher, as well as setting the lsp path.

AlexanderViand (approved):
Thanks!!

j2kun:
Done and squashed

Issue #1992, opened 2025-07-11
TBD


AlexanderViand:

AlexanderViand:
Note: with the addition of IndexType support in #1999, the example here actually passes.
However, this issue is about the unexpected segfault that occurred here when the typeconverter does NOT have a conversion for IndexType. 

I'm not sure how to create a new reproducer, probably need to introduce some unrelated type?

ai-mannamalai:
One solution: https://github.com/google/heir/pull/2073

Issue #1929, opened 2025-06-25
Adding support for for-loops with secret length


MarkusBever:
In general, it is not possible to compile for-loops, where the length of the loop is secret. To avoid this issue one can annotate an upper and lower bound for for-loops. Here is an example to compute the power function by applying multiplications in a for-loop.

```
module {
  func.func @power(%arg0: i32, %arg1: i32) -> i32 {
    %c0 = arith.constant 0 : index
    %c1 = arith.constant 1 : index
    %c1_i32 = arith.constant 1 : i32
    %0 = arith.index_cast %arg1 : i32 to index
    %1 = scf.for %arg2 = %c0 to %0 step %c1 iter_args(%arg3 = %c1_i32) -> (i32) {
      %2 = arith.muli %arg3, %arg0 : i32
      scf.yield %2 : i32
    } {lower = 0, upper = 6}
    return %1 : i32
  }
}
```

Currently, when the program is compiled with:
`bazel run @heir//tools:heir-opt -- --secretize --mlir-to-bgv --mlir-print-ir-after-all --mlir-print-ir-tree-dir=/tmp/mlir $PWD/test_ks_test/power.mlir`

Unrolling the loop into multiple if statements does work fine, but there are two other issues:
- arith.index_cast is not supported during lowering
- arith.cmpi/arith.select have no proper implementation

AlexanderViand:
Hey! As a side note, this example should work as-is for CGGI but I see you're using BGV here (and it'd be the same for BFV or CKKS).  Regarding the two issues: I think the first one (`artih.index_cast`) should be easy to support, as it's essentially a no-op in FHE, while the second one might be a bit harder, though `arith.select` does have a lowering ([SelectRewrite](https://github.com/google/heir/blob/main/lib/Transforms/SelectRewrite/SelectRewrite.td)) already.

Really, it's `arith.cmpi` that's the actually complicated one to implement in the arithmetic schemes, as it requires either polynomial interpolation (not practical for the size of plaintext spaces that OpenFHE lets you use) or polynomial approximation (which has its own issues). Jeremy did some initial work on automatic polynomial approximation for univariate functions, and we can rewrite comparison that way (using sign(a-b)), but I don't think the polynomial approximation flow is fully up and running yet. More importantly, in order to actually do a good job with the approximation, we'd likely need some kind of input range analysis, which is a bigger TODO (but also necessary for automating various other things, so already on the TODO list).

[...]

PR #1999, opened 2025-07-12
WIP: Lowering for comparisons in arithmetic pipeline

AlexanderViand:
This is work towards fixing #1929 (for loop with secret bound) which primarily hinges on the ability to actually lower a comparison to a polynomial approximation.

So far, this (draft) PR has been work to get to the point where we can rewrite the code from #1929 all the way down to just having `math_ext.sign` that needs to be arithmetized/lowered to a polynomial.

On the way there, I ended up touching a bunch of places, though:
* Various fixes to `--convert-secret-to-static-for` to make it more robust/handle more edge cases
* Made `--add-client-interface` capable of handling multiple `func.func` in the same module 
  (purely to avoid needing to split into multiple teot files, not directly related to the comparison issue)
* added `IndexType` to the type conversions in `--convert-to-ciphertext-semantics`  
  (this will appear to "fix" #1992 but really that issue isn't about the inability to lower `arith.index_cast` but about the compiler segfaulting after encountering an op it cannot lower)
*  Adds a bit of  an **UGLY HACK** to `--layout-propagation`:
  The pass previously assumed that index type'd operands don't need to have a layout assigned (i.e., index valued scalars don't need to be converted to ciphertext-sized tensors) which is true for _most_ index type values (e.g., the offset in a rotation, the indices in a tensor.extract/insert, etc). However, in the case of index values appearing in an `arith.cmpi`, we _do_ need to assign them a layout (and they _do_ need to be turned into ciphertext-sized vectors) because we will be doing math to them as part of the lowering. There was already a comment to suggest defining an op interface that can be used to decide whether a given op operand needs to have a layout assigned, and I think this would indeed be the correct solution. My current ugly hack simply hard-codes an exception for `arith.cmpi` into the index type skipping code.
* Adds a `math_ext` dialect with a single op `math_ext.sign`, with plans to move this upstream to `math`
* Adds a new pass `--comparison-to-sign-rewrite` that rewrites `arith.cmpi`/`arith.cmpf`:
  Currently, the pass only supports a < b or b > a, no equality. It also assumes that you can multiply with 0.5, so really only makes sense in a CKKS pipeline, though I've currently just added it to `--mlir-to-secret-arithmetic`.  For BGV/BFV, comparison usually involves approximating `mod t` rather than `sign`....
  
 
 The next step would be to try and actually enable polynomial approximation for `math_ext.sign` and see if the resulting programs to anything useful :)


AlexanderViand:
Ok, so after re-enabling OpenMP support for OpenFHE and allowing the python frontend to actually use OpenFHE in multi-threaded mode, it was at least somewhat more feasible to try and run the generated program. Unfortunately, it'll eventually fail with an OpenFHE Error (see below). Nevertheless, I think this is a good point to start getting some reviews and try and to get this PR merged, since it's already doing _a lot_ of things, and I assume the fix for whatever noise/scale/etc management thing is going wrong here will be its own issue and PR.

> RuntimeError: external/openfhe/src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp:l.489:EvalBootstrap(): Degree [10] must be less than or equal to the correction factor [7].

EDIT: I think I'll actually pull the OpenMP thing out of this PR, as it seems like the least related and easiest to extract + I want to make sure it works on a variety of machines before we merge it and accidentally break the OpenFHE backend

lib/Dialect/MathExt/IR/MathExtOps.td
```diff
@@ -0,0 +1,23 @@
...
+
+class MathExt_Op<string mnemonic, list<Trait> traits = []> :
+        Op<MathExt_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::math_ext";
+}
+
+def MathExt_SignOp : MathExt_Op<"sign", [Pure, ElementwiseMappable,SameOperandsAndResultType]> {
+  let summary = "Returns the sign of the input value";
+  let description = [{
+    Returns -1 if the input is negative, 0 if it is zero, and 1 if it is positive.
```
j2kun, line 17: State NaN inputs are undefined behavior or something.

lib/Transforms/CompareToSignRewrite/CompareToSignRewrite.cpp
```diff
@@ -0,0 +1,129 @@
...
+                                  PatternRewriter &rewriter) const override {
+      Value lhs = op.getLhs();
+      Value rhs = op.getRhs();
+      Value sub;  // will be set to a - b for a < b or b - a for b > a
+
+      if (op.getPredicate() == arith::CmpFPredicate::OLT ||
+          op.getPredicate() == arith::CmpFPredicate::ULT)
+        sub = rewriter.create<arith::SubFOp>(op.getLoc(), lhs, rhs);
+      else if (op.getPredicate() == arith::CmpFPredicate::OGT ||
+               op.getPredicate() == arith::CmpFPredicate::UGE)
```
j2kun:
```suggestion
               op.getPredicate() == arith::CmpFPredicate::UGT)
```

lib/Dialect/MathExt/IR/MathExtDialect.td
```diff
@@ -0,0 +1,17 @@
+#ifndef LIB_DIALECT_MATHEXT_IR_MATHEXTDIALECT_TD_
+#define LIB_DIALECT_MATHEXT_IR_MATHEXTDIALECT_TD_
+
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
+
+def MathExt_Dialect : Dialect {
+  let name = "math_ext";
+  let description = [{
+    // FIXME: add documentation
```
j2kun: Add docs

tests/Regression/issue_1992.mlir
```diff
@@ -0,0 +1,16 @@
+// RUN: heir-opt %s --convert-to-ciphertext-semantics | FileCheck %s
+// FIXME: with the addition of IndexType support, this test actually passes.
```
j2kun: Resolve FIXME

lib/Transforms/CompareToSignRewrite/CompareToSignRewrite.cpp
```diff
@@ -0,0 +1,129 @@
...
+    LogicalResult matchAndRewrite(arith::CmpIOp op,
+                                  PatternRewriter &rewriter) const override {
+      Location loc = op.getLoc();
+      Value lhs = op.getLhs();
+      Value rhs = op.getRhs();
+
+      if (op.getPredicate() != arith::CmpIPredicate::slt &&
+          op.getPredicate() != arith::CmpIPredicate::ult &&
+          op.getPredicate() != arith::CmpIPredicate::sgt &&
+          op.getPredicate() != arith::CmpIPredicate::uge)
```
j2kun:
```suggestion
          op.getPredicate() != arith::CmpIPredicate::ugt)
```

lib/Transforms/CompareToSignRewrite/CompareToSignRewrite.td
```diff
@@ -0,0 +1,19 @@
+#ifndef LIB_TRANSFORMS_COMPARETOSIGNREWRITE_COMPARETOSIGNREWRITE_TD_
+#define LIB_TRANSFORMS_COMPARETOSIGNREWRITE_COMPARETOSIGNREWRITE_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def CompareToSignRewrite : Pass<"compare-to-sign-rewrite"> {
+  let summary = "Rewrites arith.cmpi/arith.cmpf to a math_ext.sign based expression";
+  let description = [{
+    "This pass rewrites arith.cmpi/cmpf <pred> %a, %b to some combination of add/mul and sign operations.
+     TODO(#1929): provide detailed description of the expression for each predicate."
```
j2kun:
```suggestion
    This pass rewrites arith.cmpi/cmpf <pred> %a, %b to some combination of add/mul and sign operations.
     TODO(#1929): provide detailed description of the expression for each predicate.
```

lib/Transforms/PolynomialApproximation/PolynomialApproximation.cpp
```diff
@@ -211,6 +215,10 @@ struct ConvertUnaryOp : public OpRewritePattern<OpTy> {
         rewriter.replaceOpWithNewOp<EvalOp>(op, polyAttr, op.getOperand());
     evalOp->setAttr("domain_lower", domainLowerAttr);
     evalOp->setAttr("domain_upper", domainUpperAttr);
+
+    emitWarning(op.getLoc(),
```
j2kun: This warning is not actionable. Please remove it here and below.

If you're worried, maybe omit it from the pipeline until we have end-to-end testing.

lib/Transforms/PolynomialApproximation/PolynomialApproximation.td
```diff
@@ -47,6 +47,7 @@ def PolynomialApproximation : Pass<"polynomial-approximation"> {
     - `tan`
     - `tanh`
     - `trunc`
+    - `sign`
```
j2kun, line 52: This is not in the math dialect.

tests/Regression/issue_1929.mlir
```diff
@@ -0,0 +1,62 @@
+// RUN: heir-opt %s --mlir-to-ckks | FileCheck %s
+// TODO (#1929): Improve test with better CHECKs
+// TODO (#1641): enable i32-based tests once CKKS pipeline allows integer types
```
j2kun:
```suggestion
```

tests/Regression/issue_1929.mlir
```diff
@@ -0,0 +1,62 @@
...
+    // CHECK-NOT: math_ext.sign
+    // CHECK-NOT: polynomial.eval
+    %1 = scf.for %arg2 = %c0_i32 to %0 step %c1_i32 iter_args(%arg3 = %c1_f32) -> (f32) : i32 {
+        %2 = arith.mulf %arg3, %arg0 : f32
+        scf.yield %2 : f32
+    } {upper = 6}
+    return %1 : f32
+}
+
+// func.func @secret_loop_index_step(%arg0: i32 {secret.secret}, %arg1: i32 {secret.secret}) -> i32 {
```
j2kun, line 44:
```suggestion
// TODO (#1641): enable i32-based tests once CKKS pipeline allows integer types
// func.func @secret_loop_index_step(%arg0: i32 {secret.secret}, %arg1: i32 {secret.secret}) -> i32 {
```

tests/Regression/issue_1992.mlir
AlexanderViand: I think I'll just remove the test case here, since it's not really related to this PR

j2kun:
Also note I'm working on upstreaming the sign ops in https://github.com/j2kun/llvm-project/pull/8

lib/Dialect/MathExt/IR/MathExtDialect.td
```diff
@@ -0,0 +1,17 @@
...
+#define LIB_DIALECT_MATHEXT_IR_MATHEXTDIALECT_TD_
+
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
+
+def MathExt_Dialect : Dialect {
+  let name = "math_ext";
+  let description = [{
+    // Math-related operations we require for HEIR
+    // which do not (yet) exists in upstream `math`.
```
j2kun:
```suggestion
    Math-related operations we require for HEIR
    which do not (yet) exist in upstream `math`.
```

j2kun (approved):

kragall:
I've just noticed that there's another open issue regarding lowering of arith.cmpi #827. Would there be a benefit to having this as an alternative approach to the `--comparison-to-sign-rewrite` for BGV/BFV?

AlexanderViand:
> I've just noticed that there's another open issue regarding lowering of arith.cmpi #827. Would there be a benefit to having this as an alternative approach to the `--comparison-to-sign-rewrite` for BGV/BFV?

Right, now that we have a LattiGo backend, the interpolation based approaches might indeed be viable! I'd say it's a lower priority right now but I'd be happy to offer my assistance if someone wants to try :)

kragall:
The comparisons have been bugging me for a while now. I'll have a look, but I can't commit much time for them right now. Why does the LattiGo backend make that approach more viable? 

Meanwhile, I managed to crash the `--compare-to-sign-rewrite` pass with this example:

```
  func.func @scr01(%arg0: f32 {secret.secret}) -> f32 {
    %cst = arith.constant 0.000000e+00 : f32
    %cst_0 = arith.constant 1.000000e+00 : f32
    %cst_1 = arith.constant 3.000000e+00 : f32
    %cst_2 = arith.constant 1.000000e+01 : f32
    %0 = arith.cmpf ugt, %arg0, %cst_1 : f32
    %1 = arith.cmpf ult, %arg0, %cst_2 : f32
    %2 = arith.select %1, %cst_0, %cst : f32
    %3 = arith.select %0, %2, %cst : f32
    return %3 : f32
  }
 ```

**Comment by AlexanderViand** [comment]
> The comparisons have been bugging me for a while now. I'll have a look, but I can't commit much time for them right now. Why does the LattiGo backend make that approach more viable?

For this, you need the plaintext modulus $p$ to be extremely small since you need to evaluate a depth-$p$ polynomial to interpolate, and OpenFHE doesn't let you set the plaintext sufficiently small, but iirc LattiGo does.

> Meanwhile, I managed to crash the --compare-to-sign-rewrite pass with this example:

Thanks, good catch! I wasn't taking into account different float types when creating the `1.0` and `0.5`, should be fixed now!

File: lib/Dialect/MathExt/IR/MathExtOps.td [hunk 12]
```diff
@@ -0,0 +1,24 @@
...
+    The behavior is undefined for NaN inputs.
+  }];
+  let arguments = (ins SignlessIntegerOrFloatLike:$value);
+  let results = (outs SignlessIntegerOrFloatLike:$result);
```
asraa, line 21: should the result just be SignlessIntegerLike? I would expect that the result shouldn't be a float

edit: after finishing my review i realize it's floatlike too because the result types probably need to be float for CKKS

PR #2006, opened 2025-07-14
ci: try to fix the pr label

asraa:
avoid the if statement so we can get better debugging for the workflow

### DIFF
--- a/.github/workflows/pull_ready.yml
+++ b/.github/workflows/pull_ready.yml
@@ -12,7 +12,7 @@ permissions:
 jobs:
   add-label-on-maintainer-approval:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_ready.yml
```diff
@@ -37,5 +37,10 @@ jobs:
         GH_REPO: ${{ github.repository }}
         NUMBER: ${{ github.event.pull_request.number }}
         LABELS: pull_ready
+        COMMITS: ${{ github.event.pull_request.commits }}
       run: |
+          if (( COMMITS == 1 )); then
```
j2kun:
```suggestion
          if (( COMMITS != 1 )); then
```

j2kun (approved):

PR #2009, opened 2025-07-15
fix openfhe include_dirs in python fronend

AlexanderViand:
fixes an issue with the frontend's OpenFHE bazel-based config include dirs, noticed during the HEIR Hackathon today


asraa (approved):
nice catch. i realized i didn't even have a good understanding of all the output dirs (https://bazel.build/remote/output-directories)

Issue #1057, opened 2024-10-15
Evaluation key management for lwe dialect


ZenithalHourlyRate:
`bgv.relinearize` / `bgv.rotate` currently can not be lowered to lwe and further polynomial dialect. The problem here is that the lwe dialect lacks the information of the evaluation key for multiplication/rotation.

For lowering to openfhe, it is solved by a `--openfhe-configure-crypto-context` pass (#696) that creates a `!openfhe.crypto_context` and it was further passed to other openfhe arithmetic ops.

If we follow the same methodology, it can be breaks down into the following steps.

* Add `LWEContext` and `LWEEvaluationKey`
* Configure `LWEContext` via a pass like the openfhe one
  * Add ops like `lwe.gen_mulkey` and `lwe.gen_rotkey`
* Add `LWEContext` as argument to the entry function and `bgv-to-lwe` will extract `LweEvaluationKey` from it

I wonder if this is a desired approach to the problem and want to know maintainer's thoughts about this (I searched about how to materialize relinearize/keyswitching in the issues but found little discussion on it).

asraa:
Hey! Sorry for the delay last week at the conferences. I'm going to place this on our next meeting notes agenda as well.

> LWEEvaluationKey

I agree that we should add key-switching keys to our LWE dialect's types for both relinearization and key-switching after rotation. I think they key should have an identifier on it that describes the key-switching mechanism (it can generally be a key-basis transform like (1, s, s^2) -> (1, s) or we can make it either a relinearization key for multiplications or an evaluation key after rotations that describes (1, s^i) -> (1, s). The latter is probably more simple to read and for the later conversion passes to handle).

> Add LWEContext

I'm not totally convinved  we need a global context object - but let's figure out what would go in it. I do think we wit ill need to "configure" some global cryptosystem information... but it's not clear to me that it needs a separate object vs can be held in the types for the ciphertext and keys. For e.g, maybe the attributes on the `lwe.relinearization_key` can hold all of the information we need.

Materializing a key-switching key will unfortunately require making a decision on the algorithm for key-switching used (whether this is an RNS variant of a scheme, the base used to decompose the digits of the ciphertext whether 2 or other low norm base, and the technique). This paper https://eprint.iacr.org/2018/117.pdf (section 2.4) has some descriptions. 

[...]

Issue #252, opened 2023-11-14
bgv: incorporate key switching optimizations in Mono-Güneysu 2023


j2kun:
https://eprint.iacr.org/2023/1642

A New Perspective on Key Switching for BGV-like Schemes

> Furthermore, we revisit an idea by Gentry, Halevi, and Smart [19] switching primes in and out of the ciphertext and find novel opportunities for constant folding

PR #2013, opened 2025-07-15
First step of lowering ckks.relinearize

j2kun:
This PR does the first step in lowering ckks.relinearize (implementing the crypto in the compiler). To do this, it requires:

A new `polynomial.key_switch_inner` op, since key switch operates on the actual polynomial components of a ciphertext it cannot easily be represented at a higher level of abstraction. If we _had_ to, we could add higher-level `lwe` ops that extract different components of a ciphertext, but we felt this was easier until we had a strong reason to do that.

An optional key-switching-key operand to `ckks.relinearize`, which is represented as a tensor of ciphertexts. I think we may ultimately need to add a higher-level relin key or key switching key type, since we may need to identify which function argument is the key switching key for the purpose of setting up the client interfaces, and right now if it's just a tensor of ciphertext it may be hard to identify it (you'd have to find a relin op and trace it back up to the function argument, but the relin op itself may have been lowered by then?). Punting on this for now though because it's not clear we definitely need it.

This PR depends on #2036 which adds RNS support to `lwe-to-polynomial`.

### DIFF
--- a/lib/Dialect/CKKS/IR/CKKSOps.td
+++ b/lib/Dialect/CKKS/IR/CKKSOps.td
@@ -168,15 +168,33 @@ def CKKS_RelinearizeOp : CKKS_Op<"relinearize", [SameOperandsAndResultRings,
   let arguments = (ins
     LWECiphertext:$input,
     DenseI32ArrayAttr:$from_basis,
-    DenseI32ArrayAttr:$to_basis
+    DenseI32ArrayAttr:$to_basis,
+    Optional<RankedTensorOf<[LWECiphertext]>>:$keySwitchingKey
   );
 
   let results = (outs
     LWECiphertext:$output
   );
 
+  let builders = [
+    OpBuilder<(ins
+      "Value":$input,
+      CArg<"::mlir::DenseI32ArrayAttr">:$fromBasis,
+      CArg<"::mlir::DenseI32ArrayAttr">:$toBasis), [{
+      return build($_builder, $_state, input,
+                   fromBasis, toBasis, {});
+    }]>
+  ];
+
   let hasVerifier = 1;
-  let assemblyFormat = "operands attr-dict `:` qualified(type($input)) `->` qualified(type($output))" ;
+  // This format is a bit weird because even if the key is omitted, you still
+  // have to declare the type with parentheses around the operands (e.g.,
+  // `(!ct1) -> !ct`). I can't get the declarative format to work with two
+  // different type signatures because... well mlir-tblgen just doesn't do
+  // optional stuff well.
+  let assemblyFormat = [{
+    $input (`,` $keySwitchingKey^)? attr-dict `:` functional-type(operands, results)
+  }];
 }
 
 def CKKS_RescaleOp : CKKS_Op<"rescale"> {
--- a/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
@@ -2,6 +2,7 @@
 
 #include <utility>
 
+#include "lib/Dialect/CKKS/IR/CKKSOps.h"
 #include "lib/Dialect/LWE/IR/LWEAttributes.h"
 #include "lib/Dialect/LWE/IR/LWEOps.h"
 #include "lib/Dialect/LWE/IR/LWETypes.h"
@@ -35,6 +36,12 @@ namespace mlir::heir::lwe {
 #define GEN_PASS_DEF_LWETOPOLYNOMIAL
 #include "lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.h.inc"
 
+RankedTensorType convertCiphertextType(lwe::LWECiphertextType type) {
+  auto ring = type.getCiphertextSpace().getRing();
+  auto polyTy = polynomial::PolynomialType::get(type.getContext(), ring);
+  return RankedTensorType::get({type.getCiphertextSpace().getSize()}, polyTy);
+}
+
 class CiphertextTypeConverter : public TypeConverter {
  public:
   // Convert ciphertext to tensor<#dim x !poly.poly<#rings[#level]>>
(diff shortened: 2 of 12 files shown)


AlexanderViand:
@j2kun Any chance we can set up a call/meeting with Eric, you and me to continue what we started at the hackathon? (maybe after we do some of the remaining boring infrastructure work?)

j2kun:
Yes, definitely. I have vacation mid next week so let's aim for Monday or Tuesday? I will try to clean up what we started and get it to a readable point tomorrow

j2kun:
Ignoring the format CI failure (https://github.com/google/heir/pull/2030) and some random cleanup (https://github.com/google/heir/pull/2029), this is at a point where the basic lowering for relin -> key_switch_inner is working.

AlexanderViand:
Just linking relevant issue(s) that might or might not be "fixed" by this PR:

* #252
* #499
* #1057

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -373,5 +373,20 @@ def Polynomial_EvalOp : Polynomial_Op<"eval", [AllTypesMatch<["value", "output"]
   let hasVerifier = 1;
 }
 
+def Polynomial_KeySwitchInnerOp : Polynomial_Op<"key_switch_inner",
+    [AllTypesMatch<["value", "constantOutput", "linearOutput"]>]> {
+  let summary = "Key switch on an RNS polynomial component";
+  let description = [{
+  }];
```
AlexanderViand: Still needs a description (though I'm honestly not sure exactly how to describe it, beyond listing what it does :shrug: )

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -373,5 +373,20 @@ def Polynomial_EvalOp : Polynomial_Op<"eval", [AllTypesMatch<["value", "output"]
...
+  let arguments = (ins
+      PolynomialLike:$value,
```
AlexanderViand: Didn't we say yesterday that we'd drop this to just `Polynomial` for now? Maybe add a TODO for re-enabling `PolynmoialLike` and making it elementwisemappable-ish?

AlexanderViand: Same for the results, of course

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -373,5 +373,20 @@ def Polynomial_EvalOp : Polynomial_Op<"eval", [AllTypesMatch<["value", "output"]
   let hasVerifier = 1;
 }
 
+def Polynomial_KeySwitchInnerOp : Polynomial_Op<"key_switch_inner",
+    [AllTypesMatch<["value", "constantOutput", "linearOutput"]>]> {
```
AlexanderViand, line 400: I feel like there should be some kind of same ring-level constraint between the keys and the input, too?

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -528,12 +544,48 @@ struct ConvertRMulPlain : public OpConversionPattern<RMulPlainOp> {
...
+    Value two = rewriter.create<arith::ConstantIndexOp>(op.getLoc(), 2);
+    Value input0 = rewriter.create<tensor::ExtractOp>(op.getLoc(),
+                                                      adaptor.getInput(), zero);
+    Value input1 = rewriter.create<tensor::ExtractOp>(op.getLoc(),
+                                                      adaptor.getInput(), one);
+    Value input2 = rewriter.create<tensor::ExtractOp>(op.getLoc(),
+                                                      adaptor.getInput(), two);
+    polynomial::KeySwitchInnerOp ksInnerOp =
+        rewriter.create<polynomial::KeySwitchInnerOp>(
+            op.getLoc(), input2, adaptor.getKeySwitchingKey());
```
AlexanderViand, line 522: Won't this crash if the ckks.relinearize op doesn't have the optional key? The pattern should probably fail to match on such ops?

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -168,15 +168,33 @@ def CKKS_RelinearizeOp : CKKS_Op<"relinearize", [SameOperandsAndResultRings,
...
   let hasVerifier = 1;
-  let assemblyFormat = "operands attr-dict `:` qualified(type($input)) `->` qualified(type($output))" ;
+  // This format is a bit weird because even if the key is omitted, you still
+  // have to declare the type with parentheses around the operands (e.g.,
+  // `(!ct1) -> !ct`). I can't get the declarative format to work with two
+  // different type signatures because... well mlir-tblgen just doesn't do
+  // optional stuff well.
+  let assemblyFormat = [{
+    $input (`,` $keySwitchingKey^)? attr-dict `:` functional-type(operands, results)
+  }];
```
AlexanderViand, line 197: Nice workaround!

AlexanderViand (approved):
A few minor things, but I'd also be fine with merging as is, as we'll probably end up tweaking this again later anyway

lib/Dialect/Polynomial/IR/PolynomialOps.td
crockeea: "Generates a linear ciphertext that, when added to a ciphertext encrypted under the input key, switches it to a ciphertext encrypted under the output key."

tests/Dialect/CKKS/Conversions/ckks_to_openfhe/ckks_to_openfhe.mlir
```diff
@@ -62,7 +62,7 @@ module {
     // CHECK: %[[v6:.*]] = openfhe.relin [[C]], %[[x6:.*]]: ([[S]], [[T]]) -> [[T2:.*]]
     %relin = ckks.relinearize %x  {
       from_basis = array<i32: 0, 1, 2, 3>, to_basis = array<i32: 0, 1>
-    }: !ct_D4 -> !ct
+    }: (!ct_D4) -> !ct
```
crockeea, line 65: Is this a linter change, or is there some reason we need parens?

crockeea (approved):

tests/Dialect/CKKS/Conversions/ckks_to_openfhe/ckks_to_openfhe.mlir
AlexanderViand: I think Jeremy mentioned somewhere that this was the best he could do for making relin_key optional w/o writing a full custom parser/printer for the op. This is using one of the "automagic" printer/parsers, which properly supports optional arguments but requires the parentheses even if there's just a single argument type.

j2kun: Yes this is very annoying, because the declarative custom format syntax supports optional stuff but is quite buggy in how it handles optional and disjunctions involving optionals.

j2kun:
Comments addressed in 602044784667f5e7a78c5cb039beb5c9d2fdc607

PR #2017, opened 2025-07-16
ValidateNoise: make error message less verbose

ZenithalHourlyRate:
`getOperation()->emitOpError()` would dump the whole IR, making it quite verbose. Using `emitError(loc)` would make it less verbose.


AlexanderViand:
I recently noticed the `--mlir-print-stacktrace-on-diagnostic` and `--mlir-print-op-on-diagnostic` flags which might allow us to have the best of both worlds (concise error messages by default but lots of context for debugging purposes)? I don't know how they interact with the two different emitError/Warning approaches, though.

Issue #1700, opened 2025-04-10
Explore tools for annotating intermediate layers of ML models with range bounds


j2kun:
From the meeting notes on parameter selection https://docs.google.com/document/d/1ASmm8UiQisMyk1EQYVMSw--I095nsTswZEPJ6BlcrIQ/edit?usp=sharing

There seem to be two approaches:

## Runtime analysis using a validation set

For these, most techniques are ad-hoc. One uses something like pytorch's [register_module_forward_hook](https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_hook.html) to dump intermediate values, and run it on a validation set to get the values dumped to some format.

We can do this and parse the results back to annotations on the pytorch model, and try to get those preserved when dumping to MLIR. A colleague of mine at Google is going to tinker with a nice way to do this, and next week I can try to convert his work to produce MLIR annotations.

## Formal methods

There are tools like https://github.com/Verified-Intelligence/alpha-beta-CROWN (this one seems the best and most actively developed) and https://github.com/vas-group-imperial/VeriNet that are purely based on static analysis and provide interval-based propagation through a model for worst-case bounds.

I haven't seen these tools before today, but alpha-beta-CROWN seems to support pytorch and ONNX and there is a usage doc here: https://github.com/Verified-Intelligence/alpha-beta-CROWN/blob/main/complete_verifier/docs/abcrown_usage.md

j2kun:
CC @ZenithalHourlyRate maybe you would want to try using alpha-beta-CROWN on some test model in the mean time to see what it outputs?

j2kun:
While working on a potential torch integration, filed https://github.com/pytorch/xla/issues/8993 to ask about attribute propagation.

j2kun:
Here is my prototype for saving intermediate layer values https://github.com/j2kun/torch-intermediate

I have not tried this in conjunction with torch-mlir, just the stableHLO export. https://github.com/pytorch/xla/issues/8993#issuecomment-2813595343 has an input/output example

CC @AlexanderViand

AlexanderViand:
WIP here: https://github.com/llvm/torch-mlir/compare/main...belfortlabs:torch-mlir:preserve-metadata

PR #2018, opened 2025-07-16
CKKS: Add RangeAnalysis

ZenithalHourlyRate:
In CKKS, knowing the range of the plaintext inside each ciphertext is crucial for (worst-case) noise analysis (e.g. #1685), and further the selection of parameters.

Currently we have the following ways to know the range

1. Plaintext backend: it can tell the range of all intermediate ciphertexts of _one specific input_
2. Tools like #1700: it can tell the range of some ciphertext in ML model
3. Assuming all ciphertexts are in range [-1, 1]: commonly used in paper, not practical.
4. User annotates the range of the input and compiler does analysis on intermediate ciphertexts. This PR enables the analysis part; the annotation part needs further discussion (e.g. integration with python frontend and packing pipeline).

As an example, the current parameter generation process is modified to use the analysed range to select first-mod-bits when the user specifies the first-mod-bits to be 0; other usage will be explored in further noise model PRs.

### Example

The dot_product_8f example is an inner product of `[-1, 1] : <8xf32>` with addition of `0.1 : f32`, so maximal possible value is 8.1, and the analyse result is

```
Propagate CKKSRangeState to %22 = mgmt.modreduce %21 {mgmt.mgmt = #mgmt.mgmt<level = 0>} : tensor<8xf32>: CKKSRangeState(normal: 8.10, log2: 3.02)
```

Note that if we use plaintext backend, it will infer that the maximal value is 2.5, and select a smaller parameter; when the user changes the input, the selected parameter may not be valid.

Then the parameter selection process chose 51 as its first mod bits
```
First modulus bits not specified, using 51 bits.
Scheme Param:
logDefaultScale: 45
ringDim: 16384
level: 2
logqi: 51.00 45.00 45.00 
```

### Limitation

For now only the dot_product example (and the cross_level example, which requires 12 extra bits) can work through the analysis; other example like hv_matmul does not have the range result due to complex IR structure.

### DIFF
--- a/lib/Analysis/DimensionAnalysis/DimensionAnalysis.h
+++ b/lib/Analysis/DimensionAnalysis/DimensionAnalysis.h
@@ -126,8 +126,6 @@ class DimensionAnalysisBackward
   void visitCallOperand(OpOperand &operand) override {}
 };
 
-// this function will assert false when Lattice does not exist or not
-// initialized
 std::optional<DimensionState::DimensionType> getDimension(
     Value value, DataFlowSolver *solver);
 
--- a/lib/Analysis/RangeAnalysis/BUILD
+++ b/lib/Analysis/RangeAnalysis/BUILD
@@ -1,0 +1,23 @@
+package(
+    default_applicable_licenses = ["@heir//:license"],
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "RangeAnalysis",
+    srcs = ["RangeAnalysis.cpp"],
+    hdrs = ["RangeAnalysis.h"],
+    deps = [
+        "@heir//lib/Analysis:Utils",
+        "@heir//lib/Analysis/SecretnessAnalysis",
+        "@heir//lib/Dialect/Mgmt/IR:Dialect",
+        "@heir//lib/Dialect/Secret/IR:Dialect",
+        "@heir//lib/Utils",
+        "@heir//lib/Utils:LogArithmetic",
+        "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:Analysis",
+        "@llvm-project//mlir:CallOpInterfaces",
+        "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Support",
+    ],
+)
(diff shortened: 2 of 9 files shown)


lib/Analysis/CKKSRangeAnalysis/CKKSRangeAnalysis.cpp
j2kun, line 1: This doesn't appear to be specific to CKKS, and so could we rename the file/structs involved to just be, perhaps, "RangeAnalysis"?

lib/Analysis/CKKSRangeAnalysis/CKKSRangeAnalysis.h
```diff
@@ -0,0 +1,112 @@
...
+      public SecretnessAnalysisDependent<CKKSRangeAnalysis> {
+ public:
+  using SparseForwardDataFlowAnalysis::SparseForwardDataFlowAnalysis;
+  friend class SecretnessAnalysisDependent<CKKSRangeAnalysis>;
+
+  void setToEntryState(CKKSRangeLattice *lattice) override {
+    // For double input, default range is [-1, 1]
+    // This handles both secret input and plaintext func arg
+    propagateIfChanged(lattice,
+                       lattice->join(CKKSRangeState({Log2Arithmetic::of(1)})));
```
j2kun: Would it be possible to make this configurable via a pass flag? I feel the -1, 1 default is a surprise and it should be at least made more visible by being configurable at the pass level.

j2kun (changes requested):

lib/Analysis/CKKSRangeAnalysis/CKKSRangeAnalysis.cpp
ZenithalHourlyRate: Name changed to RangeAnalysis.

lib/Analysis/CKKSRangeAnalysis/CKKSRangeAnalysis.h
ZenithalHourlyRate: Added a `input-range` parameter to `generate-param-ckks`

j2kun (approved):

PR #2020, opened 2025-07-17
Migrate op builders to Op::create() for better IDE awareness

j2kun:
Cf. https://discourse.llvm.org/t/psa-opty-create-now-with-100-more-tab-complete/87339/1 and https://github.com/llvm/llvm-project/pull/147168

This is going to be broken since the LLVM integration is a week or so behind. Leaving this PR in draft form until then.

This PR was implemented by running (GNU) sed

```bash
sed -E -i 's/(\w+)\.create<(.*?)>\(/\2::create(\1,/g' **/*.cpp
```

then applying the formatter.


AlexanderViand:
We're now up-to-date enough with LLM for this and I'm looking forward to no longer reading *.inc files to figure this stuff out xD

The automated version of the migration (#2058) somehow only touched a single file, but I guess this PR will fix it more generally 😍

j2kun:
I'm out til tomorrow so feel free to patch/rebase

PR #2021, opened 2025-07-17
Updates to Gcp scripts and examples

code-perspective:
Use jax version 0.6.0 since it works with the jaxite version
Read client key and server key from a saved file if it exists


j2kun (approved):
Squash and let's merge it!

PR #2032, opened 2025-07-19
Add deepwiki badge

AlexanderViand:
Deepwiki analyzses public GH repos and creates summaries of them which are human readable but primairly designed to be used with RAG or agentic AI coding assistants

See https://deepwiki.com/google/heir for the information it has compiled on HEIR, which (at first glance) seems pretty nice.
In fact, it seems to have some info we should have in our human-written docs, too, such as info on the Context Aware dialect conversion (unless we do have that somewhere and I just couldn't find it?)

By default, it does not seem to refresh the information very often (e.g., it was currently from April), and while one can manually get it to refresh (I just requested a refresh), if you add the deepwiki badge to a repo's README, it will auto-refresh on a weekly basis.

Whether or not WE end up using deepwiki, it's very likely it'll end up "consumed" by various AI things people are using, so it's probably a good idea to have it be more up to date.

PS: let me try to request an AI review on this AI-focused PR, for maximum AI-ness xD
EDIT: formatting fails because of unrelated failure on main
EDIT2: Ironically, after re-generation, the info is arguably _worse_ even if it is more up-to-date (for example, it decided aginst generating a page on context aware dialect conversion this time?!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 ![GitHub Discussions](https://img.shields.io/github/discussions/google/heir)
 ![GitHub License](https://img.shields.io/github/license/google/heir)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/google/heir/badge)](https://securityscorecards.dev/viewer/?uri=github.com/google/heir)
+[![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/google/heir)
 
 An MLIR-based toolchain for
 [homomorphic encryption](https://en.wikipedia.org/wiki/Homomorphic_encryption)


j2kun:
I hate to be a naysayer, but I don't like this PR for a few reasons.

Adding a badge is an endorsement of the content of the generated docs, and I'd prefer not to endorse something we can't vet or edit. (see the end for a sample of things I don't like about the Wiki's content)

Having the badge is an advertisement for the project, and I don't think HEIR's Github readme is the right place to advertise random AI projects.

I don't buy the reasoning that we should spend any effort to maintain the quality of third-party projects just because further third-party projects are blindly consuming it. It implies an unbounded maintenance burden, and we've got too much else to do. The fact that it's AI in a world of ever more AI-generated slop makes it worse: the authors of this tool don't even know we exist nor care about our project, and they are just one of likely dozens. Would we have to advertise and integrate with them all?

If you care about this DeepWiki being up to date, and you can manually request a refresh, then surely you can automate the manual requesting of the refresh without putting a badge here. Write a cron job that posts to the same API as the browser.

The documentation generated is, in my opinion, quite bad, and it doesn't seem like freshness is the main problem.

  - It gets basic concepts wrong in subtle ways, for example that "The secret.yield operation is used ... to specify the results that should be encrypted." and "secret.conceal: Converts a plaintext value to a secret value". From the tfhe-rust backend page: "High-Level Emitter (TfheRustHLEmitter) Generates more readable and maintainable Rust code using tfhe-rs high-level APIs" and "The TFHE-Rust code generation integrates with HEIR's Bazel build system through specialized test macros and CI workflows" are just misleading. From the contributing guidelines it says to "download bazel 8.1.0 via install_cibuildwheel_deps.sh")
  - It omits most of the interesting parts of the codebase. For example:
    - Only three pages of transformations are there, including the nitpicky secret.generic canonicalization patterns, yosys optimizer, and a page on dialect conversion.
    - The dialect conversion page combines different aspects of dialect conversion in a very haphazard manner (mixing information about dialect conversion in general with information about the dialect conversion passes we provide, which are all lumped together even though they are quite distinct). And what it does include about DialectConversion ignores plenty of dialect conversion passes anyway (no lwe-to-polynomial or convert-to-ciphertext-semantics for example).
    - There is little on the site about analysis passes, nothing about parameter selection, noise models, or anything that is really specific to FHE.
    - The pipeline page has only stuff about CGGI and tosa
  - It exposes way too much internal implementation details (for example the "python frontend" starts with all of this stuff about numba and very low level details about instruction generation).
  - It reads like a badly written paper trying to hype up the capabilities of the system, rather than describe what is there, how it works, and why it was done the way it was.

AlexanderViand:
I think being a naysayer is the right call here, and I now actually agree that we shouldn't be sending people there! While I liked some of the stuff I saw, after re-generation, the quality of the summaries has gone way down and not having control over that is very annoying.

PR #2048, opened 2025-07-22
Start implementing new layout notation

j2kun:


lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -396,12 +408,74 @@ Value implementUnpackOpForScalar(
...
+           << "Expected layout to be an IntegerRelation-style layout";
+  }
+
+  RankedTensorType dataSemanticType =
+      cast<RankedTensorType>(op.getValue().getType());
+  llvm::SmallVector<int64_t> dataSemanticShape;
+
+  IntegerRelation rel = layout.getIntegerRelation();
+  for (unsigned varPos = rel.getVarKindOffset(VarKind::Range);
+       varPos < rel.getNumRangeVars(); ++varPos) {
```
asraa: I just noticed this bug: you are starting at the range offset and then the upper bound of this for loop is the numrangevars, so this loop will have size 0.

asraa:
```suggestion
       varPos < rel.getVarKindEnd(VarKind::Range); ++varPos) {
```

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -396,12 +408,74 @@ Value implementUnpackOpForScalar(
...
+             << "Required all range variables to have an inferable upper "
+                "bound, "
+                "but found no upper bound for variable at position "
+             << varPos << " in relation " << printRelation(rel);
+    }
+    dataSemanticShape.push_back(dimBound.value());
+  }
+
+  // Last axis is always slot count in one ciphertext. This will need to be
+  // changed if/when we support hypercube slot structures.
```
asraa: since the range (at least in the example) seems to be (ct, slot), then isn't the slot count already pushed into the shape above?

tests/Transforms/add_client_interface/add_client_interface_new_layout.mlir
```diff
@@ -0,0 +1,23 @@
+// TODO(#2047): re-enable
+// CHANGE BACK: heir-opt --add-client-interface="ciphertext-size=1024" %s | FileCheck %s
+// RUN: heir-opt %s
+
+// Data is 32x64, being packed into ciphertexts of size 1024 via Halevi-Shoup
+// diagonal layout.
+!ct_ty = !secret.secret<tensor<32x1024xi16>>
+#layout = #tensor_ext.new_layout<
+  domainSize=2,
+  relation="(row, col, ct, slot) : ((slot mod 32) - row == 0, (ct + slot) mod 64 - col == 0, row >= 0, col >= 0, ct >= 0, slot >= 0)"
```
asraa: when the relation is written this way (range -> domain kinda style) there's something missing bounding ct.
even if I add in constraints according to the data semantic type (row < 32, col < 64) and the ciphertext slot size (slot < 1024), then the two inequalities don't give me info on the size of ct.

in `(ct + slot) mod 64 - col == 0`, the `mod 64` operation means that ct can be as large as possible.

Adding or writing the constraint like `col - row = ct` with `col < 64` and `row < 32` implies `ct < 64`.

So I feel like something is missing in the relation written as is (or maybe the mod operation is misplaced?)

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
j2kun: True. Removed.

tests/Transforms/add_client_interface/add_client_interface_new_layout.mlir
j2kun: I think we can add upper bounds to all of these in the relation.

asraa: Well, the UB on the number of ciphertexts is computed based on the relation and kernel. I don't think that's the problem, I think that we're missing a constraint actually in the way H-V diagonalization was written

j2kun: > Well, the UB on the number of ciphertexts is computed based on the relation and kernel

Yes, but in this test at least the kernel is already chosen, so it makes sense that the info is materialized in the layout.

j2kun:
Addressed comments in 2955b48add5cf8cf14a99cfdd573c35ea3b68e76

tests/Transforms/add_client_interface/add_client_interface_new_layout.mlir
asraa: Right so `(ct + slot) mod 64 - col == 0` can actually just be written as `ct + slot - col == 0` without the mod. 

Combined with `0 <= slot < 1024` and `0 <= col < 64` will imply `0 <= ct < 64`. The mod is what allows ct to be unbounded.

asraa: Either way! I guess I just needed to convince myself that there was missing info as is.

j2kun: The mod provides repetition of the layout in the slots I believe. I think this is required for the rotations to make sense if there is not enough data to fill up the whole ciphertext.

asraa: Ah you're right! OK noted to add the bounds earlier in the pipeline. I'll see if I can attach it to an existing issue.

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -398,12 +410,71 @@ Value implementUnpackOpForScalar(
...
+    std::optional<int64_t> dimBound =
+        rel.getConstantBound64(BoundType::UB, varPos);
+    if (!dimBound) {
+      return op.emitError()
+             << "Required all range variables to have an inferable upper "
+                "bound, "
+                "but found no upper bound for variable at position "
+             << varPos << " in relation " << printRelation(rel);
+    }
+    dataSemanticShape.push_back(dimBound.value());
```
asraa: bounds are inclusive, so you'll want a +1 here

PR #2056, opened 2025-07-29
Update CGGI Pipeline for e2e loop tests

WoutLegiest:
- Patch + maintenance on the previous arith to tfhe-rs-hl pipeline
- Extended the emitter to support Affine ops
- Cleanup of CGGI related test
- Adding an additional test to test the HL Loop functions


lib/Pipelines/BooleanPipelineRegistration.cpp
```diff
@@ -209,6 +209,58 @@ CGGIBackendPipelineBuilder toTfheRsPipelineBuilder() {
...
+        createExpandCopyPass(ExpandCopyPassOptions{.disableAffineLoop = true}));
+    pm.addPass(memref::createFoldMemRefAliasOpsPass());
+    pm.addPass(createForwardStoreToLoad());
+    pm.addPass(createCanonicalizerPass());
+    pm.addPass(createCSEPass());
+    pm.addPass(createSCCPPass());
+  };
+}
+
+CGGIBackendPipelineBuilder toTfheRsHLPipelineBuilder() {
```
asraa: could this be combined with https://github.com/google/heir/blob/97f20718031a235f0125c6d2aac4a8a9454599a1/lib/Pipelines/BooleanPipelineRegistration.cpp#L192?

lib/Pipelines/BooleanPipelineRegistration.cpp
```diff
@@ -209,6 +209,58 @@ CGGIBackendPipelineBuilder toTfheRsPipelineBuilder() {
   };
 }
 
+CGGIBackendPipelineBuilder arithToTfheRsHLPipelineBuilder() {
```
asraa: is there any reason why mlir-to-cggi=data-type=Integer won't work? 

https://github.com/google/heir/blob/97f20718031a235f0125c6d2aac4a8a9454599a1/lib/Pipelines/BooleanPipelineRegistration.cpp#L119-L121

Right now we are trying to split the pipelines into the input to scheme and then scheme to library (so mlir-to-cggi will follow --scheme-to-tfhe-rs instead of one mlir-to-tfhe-rs)

WoutLegiest: I checked, and the mlir-to-cggi pipeline does loop unrolling. 
Thinking about parallelisation in the future (and larger network) that becomes a bottleneck. iirc, you already cannot do the mnist model with loopunrolling

lib/Pipelines/BooleanPipelineRegistration.cpp
WoutLegiest: Indeed, can be removed!

lib/Pipelines/BooleanPipelineRegistration.cpp
asraa: ahhh! i see. maybe can you only enable loop unrolling when it's a yosys (boolean path) in mlir-to-cggi?

WoutLegiest: Updated the mlir-to-cggi pipeline and removed the one I made

tests/Examples/tfhe_rust_hl/cpu/src/main_hello_world.rs
```diff
@@ -0,0 +1,36 @@
...
+        let input: u8 = 31;
+
+        let a: tfhe::FheUint<tfhe::FheUint8Id> = FheUint8::encrypt(input, &client_key);
+        let input_vec = core::array::from_fn(|_1| core::array::from_fn(|_1| a.clone()));
+
+        set_server_key(server_key);
+
+        // let t = Instant::now();
+        let result = hello_world_clean_xsmall_test_rs_lib::fn_under_test(&input_vec);
+        // let elapsed = t.elapsed();
```
code-perspective, line 24: Uncomment or remove the commented lines

j2kun (approved):

PR #2060, opened 2025-07-30
migrate to new builder style

AlexanderViand:
Using Jeremy's regex from #2020 (+ two manual fixes for cases with nested < > inside the call). New PR because I don't think I have write access to his fork.


j2kun (changes requested):
Can we not include the module.bazel changes in this PR? I think something is screwy that's causing back and forth changes when people build on different platforms. We should turn off the eager updates.

AlexanderViand-Intel:
> Can we not include the module.bazel changes in this PR? I think something is screwy that's causing back and forth changes when people build on different platforms. We should turn off the eager updates.

Oops, I staged all changes and forgot to unstage the module.bazel afterwards. Fixed now, but not having to constantly fight git over this file would be great!

PR #2062, opened 2025-07-31
update lockfile

j2kun:
I don't see a good way to make lockfile updates less frequent without turning off the lockfile entirely. @AlexanderViand can you try patching this CL and see if `bazel mod deps --lockfile_mode=update` causes any diff on your system?


j2kun:
@asraa thoughts? Maybe we could add the lockfile to `.gitignore` and exclusively update it via automated GH actions (once we figure out how to get that action to work?)

asraa:
> @asraa thoughts? Maybe we could add the lockfile to .gitignore and exclusively update it via automated GH actions (once we figure out how to get that action to work?)

I was just thinking we should remove the lockfile from version control entirely, but it's not a bad idea to populate it from GH actions. (I do need your online help to debug the actions though so I can have someone approve / run >.>)

AlexanderViand:
> I don't see a good way to make lockfile updates less frequent without turning off the lockfile entirely. @AlexanderViand can you try patching this CL and see if `bazel mod deps --lockfile_mode=update` causes any diff on your system?

No diffs with this one 👍

AlexanderViand (approved):

PR #2063, opened 2025-08-01
Add a basic layout codegen (without IR materialization)

j2kun:
This PR creates a helper library for doing code generation for a layout expressed as an integer relation. This does not include the materialization of the generated AST as MLIR, but Asra has an in-progress PR that does this step and just needs to be updated to use the ISL API.

### DIFF
--- a/lib/Utils/Layout/Codegen.cpp
+++ b/lib/Utils/Layout/Codegen.cpp
@@ -1,0 +1,161 @@
+#include "lib/Utils/Layout/Codegen.h"
+
+#include "isl/ast_build.h"                                          // from @isl
+#include "isl/ast_type.h"                                           // from @isl
+#include "isl/constraint.h"                                         // from @isl
+#include "isl/ctx.h"                                                // from @isl
+#include "isl/map.h"                                                // from @isl
+#include "isl/set.h"                                                // from @isl
+#include "isl/space.h"                                              // from @isl
+#include "isl/union_map.h"                                          // from @isl
+#include "mlir/include/mlir/Analysis/Presburger/IntegerRelation.h"  // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+using presburger::IntegerRelation;
+using presburger::VarKind;
+
+static __isl_give isl_union_map* convertRelationToUnionMap(
+    const IntegerRelation& rel, isl_ctx* ctx) {
+  // ISL has these variables types, which map to MLIR FPL variable types:
+  // isl_dim_param -> VarKind::Symbol
+  // isl_dim_in -> VarKind::Domain
+  // isl_dim_out -> VarKind::Range
+  // isl_dim_div -> VarKind::Local
+  //
+  // ISL puts parameters first, but since we have zero symbols, we can ignore
+  // the implied change to the offsets.
+  unsigned numDomain = rel.getNumVarKind(VarKind::Domain);
+  unsigned numRange = rel.getNumVarKind(VarKind::Range);
+  unsigned numLocal = rel.getNumVarKind(VarKind::Local);
+
+  // This represents the mapping from all four variables to just iterate over
+  // the range variables. This is specific to FHE because we know the packing
+  // must be a partial function from range to domain.
+  //
+  // E.g., this is specifying the `S[row,col,ct,slot] -> [ct,slot]` part of a
+  // union map like
+  //
+  //   {
+  //      S[row,col,ct,slot] -> [ct,slot] :
+  //      0 <= row,ct < 4
+  //      and 0 <= col < 8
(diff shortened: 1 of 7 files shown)


j2kun:
@asraa removed the attempt at eliminating the domain variables, so it's ready for review now

j2kun:
Asra is out Friday, so maybe @AlexanderViand will have time to review it tomorrow? :)

lib/Utils/Layout/CodegenTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+  MLIRContext context;
+  IntegerRelation relation = relationFromString(
+      "(d0, d1) : ((d0 - d1) mod 2 == 0, d0 >= 0, d1 >= 0, 10 >= d0, 30 >= d1)",
+      1, &context);
+  // std::string dataId = "data";
+  // std::string slotId = "slot";
+  // relation.resetIds();
+  // relation.setId(VarKind::Domain, 0, Identifier(&dataId));
+  // relation.setId(VarKind::Range, 0, Identifier(&slotId));
+
```
AlexanderViand: ?

lib/Utils/Layout/Codegen.cpp
```diff
@@ -0,0 +1,178 @@
...
+  //   } else if (rel.getVarKindAt(i) == VarKind::Range) {
+  //     dimKind = isl_dim_out;
+  //     offset = rel.getVarKindOffset(VarKind::Range);
+  //   } else if (rel.getVarKindAt(i) == VarKind::Local) {
+  //     dimKind = isl_dim_div;
+  //     offset = rel.getVarKindOffset(VarKind::Local);
+  //   }
+  //   char *name = "???";
+  //   space = isl_space_set_dim_name(space, dimKind, offset, name);
+  // }
```
AlexanderViand: ?

lib/Utils/Layout/Codegen.cpp
AlexanderViand, line 1: Though this is called "codegen", it seems like the vast majority of the code here is actually about presburger -> ISL conversion (+a thin wrapper around ISL's AST codegen)?

lib/Utils/Layout/CodegenTest.cpp
j2kun: There's a means in FPL to associate a variable in an IntegerRelation with some data object, in our case it would be an SSA value, so that later you can reconnect things back to the source IR. I was testing that and left this comment in.

lib/Utils/Layout/Codegen.cpp
j2kun: Same as above, I want to figure out how to keep track of where the variables came from (it seems we can set a string in ISL)

lib/Utils/Layout/Codegen.cpp
j2kun: Correct, though it didn't start as a thin wrapper, and there will be a followup by Asra on converting ISL's AST back to MLIR.

lib/Utils/Layout/Parser.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "llvm/include/llvm/ADT/StringRef.h"  // from @llvm-project
+#include "mlir/include/mlir/Analysis/Presburger/IntegerRelation.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/MLIRContext.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+
+presburger::IntegerRelation relationFromString(llvm::StringRef integerSetStr,
+                                               int numDomainVars,
```
asraa, line 18: Why is there not need set (some of) the rest of the vars as range vars explicitly?
(I was reading about SetDim being the kind of dimensions that come from integer polyhedron, and those default to range vars. So is it that the rest of the vars after the domain vars are implicitly range vars? A comment when I forget will be useful)

I realize this is mostly for unit tests? But the functionality for creating an ISL tree relies on vars in the relation being set as domain/range/local. So just curious

lib/Utils/Layout/Codegen.h
```diff
@@ -0,0 +1,21 @@
...
+namespace heir {
+
+// Generate a LoopNest from an IntegerRelation.
+//
+// This is intended to support the case where the IntegerRelation defines a
+// single polyhedron representing a ciphertext layout, and the code generated
+// for the packing can be expressed as a single perfect loop nest with a single
+// assignment operator in the innermost loop body.
+//
+// `tree` is an outparam to hold the generated ISL AST tree.
```
asraa: did you remove the `tree` out param in favor of a return value? (+1 to that)

j2kun: Done!

lib/Utils/Layout/Parser.h
j2kun: Yes, for unit tests only. Added a comment.

Issue #2047, opened 2025-07-22
Generalize affine-map layout notation to use IntegerRelations


j2kun:
This is the parent issue for the task of replacing our existing tensor_ext.layout attribute with a new one based on integer relations.

j2kun:
I'm on vacation for a week so this is a brain dump of what I was working on and where to go next.

#2048 does the basic job of adding a `tensor_ext.new_layout` attribute that is based on `IntegerSet` (and `IntegerRelation`) rather than `AffineMap`. It also adds a shell of the step where we need to take a layout attribute and generate a loop nest that actually populates the plaintext slots with data.

I have not yet been able to find upstream code that implements the code generation. It seems like there are some places upstream where something is done, but it either doesn't rely on IntegerRelation simplification, or it doesn't generate code (just analysis, like 'is the relation empty').

To follow up on codegen, I asked questions to Tobias and Kunwar, who are experts in this, but haven't heard back yet. I posted on [MLIR discourse](https://discourse.llvm.org/t/simplest-possible-loop-generation-from-integerset-integerrelation/87486) which got a response with an outline of what the algorithm would look like if we implemented it ourselves (especially [this paper](https://icps.u-strasbg.fr/~bastoul/research/papers/Bas04-PACT.pdf)). [At some point in the past](https://discourse.llvm.org/t/about-mlir-polyhedral-optimization/1268/4?u=j2kun) someone mentioned you may need an ILP solver, which we have but that implies we'd have to do it in HEIR rather than upstream. I was also pointed to [this paper on AST polyhedral codegen](https://dl.acm.org/doi/10.1145/2254064.2254123) which appears independent of ISL.

I tried looking for out of tree projects that do AST codegen, using Presburger as a subroutine in doing so, but all the ones I could find (e.g., polygeist) depend on ISL instead. So that is one option, depending directly on ISL and using its generic ast code gen features (as [demonstrated here](http://playground.pollylabs.org/)), with the downside being that we have to pull in ISL as a dependency, and it's ~10k lines of inscrutable C code.

However, we can ignore codegen for the moment and still make progress for the next week. If we get #2048 merged, then we can start implementing kernels and other steps in the pipeline (e.g., layout hoisting) that use the new layout attribute, and even migrate tests. The only thing we won't be able to do is migrate end-to-end tests until we have codegen, or delete the old codepaths until we get the full pipeline using the new attributes.

j2kun:
[...]

PR #2072, opened 2025-08-06
Add ISL as a dependency to HEIR

j2kun:


j2kun:
Part of https://github.com/google/heir/issues/2047

asraa (approved):
Approved, +1 to alex's comment

j2kun:
Done

j2kun:
This one is going to need some internal work to get ISL into third_party as well, so the PR won't land for a bit due to that.

j2kun:
Actually, we can merge this but we can't use ISL in HEIR code until the internal work is done. I may wait on that until I get the open source version of the PR I want to write working (hopefully today/tomorrow) and then do the internal work in conjunction with that.

j2kun:
One of the complexities here I haven't yet figured out is the multiprecision math backend required by ISL. We have two options that: `gmp` and `imath`.

`gmp` (GNU multiprecision arithmetic) has issues in that the bazel configuration is notoriously complex due to having a lot of dynamically-generated assembly code in the build process. On linux, however, you can just install `libgmp-dev` and `#include <gmp.h>` will use the system installation. MacOS does not appear to be so simple, and despite being able to `brew install gmp` I can't seem to get bazel to discover the headers and link it properly. Depending on the system dep also incurs an extra setup step in the initial HEIR dev setup, which is already cumbersome. I don't know how to get a pure bazel, cross-platform dependency on `gmp`.

The other option is [`imath`](https://github.com/creachadair/imath), which has a much simpler build and I can build ISL with it on both MacOS and Linux. However, it's not as fast as `gmp` and it is yet another dependency that is neither on the bazel central registry, nor in Google's monorepo. It's also a much smaller project with less widespread use compared to `gmp`. One option is to just vendor `imath` inside my `isl` fork, as it's simple enough that I can just copy the source files to a subdirectory. (This may be beneficial because there is a second, unrelated project called `imath` that is already in Google's monorepo)

@AlexanderViand @asraa thoughts?

AlexanderViand:
I'm in favor of whatever is fastest in human time - oh, and if the performance difference between the two math backends ever becomes a significant issue for HEIR, we should really reconsider how much we're asking ISL to do during a compilation 😉

j2kun:
I'll go with imath for now.

j2kun:
According to Tobias about gmp/imath performance:

> Fig. 17 in https://dl.acm.org/doi/pdf/10.1145/3428263 shows the difference. imath is slower than gmp. However, we implemented small-integer-optimization, which only works with imath and which outperforms gmp by a factor of  about very roughly 2x.

j2kun:
My ISL fork is now updated to vendor imath as a dependency.

PR #2109, opened 2025-08-18
Hoist integer-relation-based layout through matvec

j2kun:
Rebased over https://github.com/google/heir/pull/2117

### DIFF
--- a/tests/Transforms/layout_optimization/verifier.mlir
+++ b/tests/Transforms/layout_optimization/verifier.mlir
@@ -28,13 +28,12 @@ func.func @main(%arg0: tensor<512x512xf32>, %arg1: tensor<512xf32>) -> tensor<51
 }
 // -----
 
-// TODO(#1888): re-enable when matvec is hoistable
 // Missing required kernel
-// func.func @main(%arg0: tensor<512x512xf32>, %arg1: tensor<512xf32>) -> tensor<512xf32> {
-//   %cst = tensor.empty() : tensor<512xf32>
-//   // expected-error@REENABLEME {{has unsupported kernel}}
-//   %0 = linalg.matvec
-//     ins(%arg0, %arg1 : tensor<512x512xf32>, tensor<512xf32>)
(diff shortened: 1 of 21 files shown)


tests/Transforms/layout_optimization/verifier.mlir
```diff
@@ -28,13 +28,12 @@ func.func @main(%arg0: tensor<512x512xf32>, %arg1: tensor<512xf32>) -> tensor<51
...
-//   %cst = tensor.empty() : tensor<512xf32>
-//   // expected-error@REENABLEME {{has unsupported kernel}}
-//   %0 = linalg.matvec
-//     ins(%arg0, %arg1 : tensor<512x512xf32>, tensor<512xf32>)
-//     outs(%cst : tensor<512xf32>) -> tensor<512xf32>
-//   func.return %0 : tensor<512xf32>
-// }
+func.func @main(%arg0: tensor<512x512xf32>, %arg1: tensor<512xf32>) -> tensor<512xf32> {
+  %cst = tensor.empty() : tensor<512xf32>
+  // expected-error@below {{has unsupported kernel}}
```
asraa, line 34: The error message was a little confusing - this is because secret.kernel attribute is missing and required for linalg operations? If I were a user it would indicate to me that matvec is unsupported.

But I think secret.kernel attrs are more meant for users to provide a kernel, and otherwise our internal pipeline will handle it? I don't think layout-propagation adds these attrs right now (is that a follow-up issue I can track / add myself to)

j2kun: If a kernel attribute is not found, it uses "Trivial" (which is different from MatvecNaive), and the matvec op specifically forbids the Trivial kernel.

I mostly put this in because I wasn't sure what to do here. If layout-propagation sets a kernel for all matvecs (which we plan to do), then this seems fine, it would mainly indicate someone is running layout-optimization without having first run layout-propagation, which, ok, that's your choice.

Issue #2081, opened 2025-08-11
Plaintext backend tests segfault on MacOS


asraa:
There seems to be a segfault when running the generated binary (roberts_cross, dot_product_8f_debug, etc)

Time0o:
I believe I have figured this out, I could reproduce and fix on my machine. I'll describe the issue for `dot_product_8f_debug` but I haven't tested this for all cases because this would require modifications in a bunch of files (I can make that PR if you'd like).

This is related to this: https://github.com/google/heir/pull/2111 which suggest using `llvm.emit_c_interface`. These are emitted now but (for this test) just not used. I think maybe @j2kun may already be aware of this but this confirms it for macOS (function tries to return in `x0`, `x1`, ... while caller expects result on stack).

Currently `tests/Examples/plaintext/dot_product_f_debug/dot_product_8f_debug_test.c` makes the following declaration:

```c
memref_1d_f32 dot_product__encrypt__arg0(
    /* arg*/
    float* allocated, float* aligned, int64_t offset, int64_t size,
    int64_t stride);
```

And then calls this function which is generated from MLIR. This is however not correct, for my system there is an ABI mismatch in how the caller expects the struct type to be returned and how this function actually does it. I believe MLIR intends you to instead use `_mlir_ciface_dot_product__encrypt__arg0` which circumvents calling conventions issues. So e.g. this function would have to be called like:

```
typedef struct {
  void* arg0;
  void* arg1; 
  int64_t arg2;
  int64_t arg3;
  int64_t arg4;
+} encrypt_input;

// ...

void _mlir_ciface_dot_product__encrypt__arg0(
    memref_1d_f32 *, encrypt_input const *);

// ...

memref_1d_f32 encArg0;

encrypt_input inputs0 = {arg0, arg0, 0, 8, 1};

_mlir_ciface_dot_product__encrypt__arg0(&encArg0, &inputs0);
```

This returns 0, 8, 1 in `encArg0` for me while the current implementation returns garbage values.

If I am right this means that technically all calls to C functions should be changed to `_mlir_ciface_...` although I assume this is only a problem when passing or returning structure types by value.

j2kun:
We had done this for a number of tests already (look for test directories under tests/ called runner/), and so I think the migration would be straightforward by applying the same pattern. I was going to do this Friday and got sidetracked 😅

Time0o:
@j2kun I halfway did this now for the tests in `plaintext` because those were failing my build, I will open a PR once I have the rest.

[...]

Issue #1988, opened 2025-07-11
Roberts cross plaintext tests segfault on MacOS (and maybe linux??)


j2kun:
```bash
$ bazel test tests/Examples/plaintext/roberts_cross:all
INFO: Analyzed 15 targets (204 packages loaded, 25994 targets configured).
FAIL: //tests/Examples/plaintext/roberts_cross:roberts_cross_cc_test (Segmentation fault) (see /private/var/tmp/_bazel_jkun/2e10cb98584cfc46a7fac3bd366c3dee/execroot/_main/bazel-out/darwin_arm64-dbg/testlogs/tests/Examples/plaintext/roberts_cross/roberts_cross_cc_test/test.log)
INFO: From Testing //tests/Examples/plaintext/roberts_cross:roberts_cross_cc_test:
==================== Test output for //tests/Examples/plaintext/roberts_cross:roberts_cross_cc_test:
================================================================================
FAIL: //tests/Examples/plaintext/roberts_cross:roberts_cross_mod_786433_cc_test (Segmentation fault) (see /private/var/tmp/_bazel_jkun/2e10cb98584cfc46a7fac3bd366c3dee/execroot/_main/bazel-out/darwin_arm64-dbg/testlogs/tests/Examples/plaintext/roberts_cross/roberts_cross_mod_786433_cc_test/test.log)
INFO: From Testing //tests/Examples/plaintext/roberts_cross:roberts_cross_mod_786433_cc_test:
==================== Test output for //tests/Examples/plaintext/roberts_cross:roberts_cross_mod_786433_cc_test:
================================================================================
FAIL: //tests/Examples/plaintext/roberts_cross:roberts_cross_mod_536903681_cc_test (Segmentation fault) (see /private/var/tmp/_bazel_jkun/2e10cb98584cfc46a7fac3bd366c3dee/execroot/_main/bazel-out/darwin_arm64-dbg/testlogs/tests/Examples/plaintext/roberts_cross/roberts_cross_mod_536903681_cc_test/test.log)
INFO: From Testing //tests/Examples/plaintext/roberts_cross:roberts_cross_mod_536903681_cc_test:
==================== Test output for //tests/Examples/plaintext/roberts_cross:roberts_cross_mod_536903681_cc_test:
================================================================================
INFO: Found 12 targets and 3 test targets...
INFO: Elapsed time: 58.931s, Critical Path: 56.69s
INFO: 12 processes: 8645 action cache hit, 14 darwin-sandbox.
[...]

PR #2111, opened 2025-08-19
PlaintextBackend: fix ptm=0 cases and encoding

ZenithalHourlyRate:
As detailed in https://github.com/google/heir/issues/1988#issuecomment-3072485274, after the packing system is in (notably the encoding/decoding process happens before lwe dialect), the plaintext backend should be changed accordingly.

* When plaintext modulus is 0, namely totally in original i16/i32 arithmetic, `-to-mod-arith` part should be skipped
* Plaintext backend tests should call encoder/decoder

As for #2081, I wonder how would segment fault happen after this PR.

```
// We can use both signatures
//   result_memref = func(arg_memref)
//   func(&result_memref, arg_memref)
```
Also it becomes quite astonishing for me that both signatures work.


j2kun:
> So I believe a more proper fix is to tag llvm.emit_c_interface somewhere in the pipeline,

If you feel strongly here, please file an issue and we can investigate (we're having a techdebt week at Google soon)

j2kun (approved):

PR #2112, opened 2025-08-19
Isolate and improve ISL conversion step

j2kun:
The previous way we were converting IntegerRelation to isl basic/union map was buggy because ISL was overthinking and creating duplicate local variables that were not being properly constrained (in `isl_basic_map_add_constraint`). A more direct way to do this (h/t Enzyme) is to use `isl_basic_map_from_constraint_matrices` and directly construct the constraint matrices from FPL.

I also took the opportunity to separate the FPL -> ISL conversion step from the schedule generation needed for codegen, as the schedule involves adding new variables to the domain (essentially duplicating range variables). That part is unrelated to the actual conversion, and was confounding me during bug hunting.

Next, it would be nice to have a backwards conversion from ISL -> FPL, so we could use ISL for simplification while FPL's simplification routines are incomplete.

### DIFF
--- a/lib/Utils/Layout/IslConversionTest.cpp
+++ b/lib/Utils/Layout/IslConversionTest.cpp
@@ -1,0 +1,79 @@
+#include <cstdlib>
+#include <string>
+
+#include "gmock/gmock.h"  // from @googletest
+#include "gtest/gtest.h"  // from @googletest
+#include "lib/Utils/Layout/IslConversion.h"
+#include "lib/Utils/Layout/Parser.h"
+#include "llvm/include/llvm/ADT/StringRef.h"  // from @llvm-project
+#include "mlir/include/mlir/Analysis/Presburger/IntegerRelation.h"  // from @llvm-project
+#include "mlir/include/mlir/AsmParser/AsmParser.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Affine/Analysis/AffineAnalysis.h"  // from @llvm-project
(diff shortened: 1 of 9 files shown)


lib/Utils/Layout/IslConversionTest.cpp
```diff
@@ -0,0 +1,83 @@
...
+      "(d0, d1, d2, d3, d4, d5, d6, d7) : (-d1 + d3 == 0, d0 - d4 * 1024 - d5 "
+      "== 0, d2 - d6 * 32 - d7 == 0, d5 - d7 == 0, -d0 + 31 >= 0, d0 >= 0, d1 "
+      "== 0, d2 >= 0, -d2 + 1023 >= 0, -d0 + d3 * 1024 + 1023 >= 0, d0 - d3 * "
+      "1024 >= 0, -d0 + d4 * 1024 + 1023 >= 0, d0 - d4 * 1024 >= 0, -d2 + d6 * "
+      "32 + 31 >= 0, d2 - d6 * 32 >= 0)";
+  IntegerRelation relation =
+      affine::FlatAffineValueConstraints(parseIntegerSet(relStr, &context));
+  relation.convertVarKind(VarKind::SetDim, 3, 8, VarKind::Local);
+  relation.convertVarKind(VarKind::SetDim, 0, 1, VarKind::Domain);
+  relation.dump();
```
asraa: nit: remove these dumps

j2kun: Done

PR #2117, opened 2025-08-20
[Utils/Layout]: Hoist vec layout conversion through matvec

j2kun:


lib/Utils/Layout/Hoisting.cpp
```diff
@@ -0,0 +1,119 @@
...
+  // relation (ct, slots) -> (ct, slots))
+  //
+  // 2. Compose (1) with the (ct, slot) dims of the matrix packing.
+
+  IntegerRelation fromClone(fromVecLayout);
+  IntegerRelation toClone(toVecLayout);
+
+  // Project out the ciphertext dimension, though this will need to change when
+  // we get a larger vector than can fit in one ciphertext. This is where the
+  // assumption about the vector packing fitting into one ciphertext comes from.
```
asraa, line 54: Can you assert this assumption in debug mode (kinda for documentation) by doing something like fromClone/toClone's range 0 upper bound is 0?

lib/Utils/Layout/Hoisting.cpp
```diff
@@ -0,0 +1,119 @@
...
+  fromClone.projectOut(1);
+  toClone.projectOut(1);
+
+  fromClone.inverse();
+  fromClone.compose(toClone);
+  fromClone.removeRedundantConstraints();
+  fromClone.simplify();
+
+  // Put the ct dim back in with bounds from the matrix layout, and equality
+  // between the domain ct var and the range ct var.
```
asraa:
```suggestion
  // At this stage, fromClone models the re-packing relation on the vector (slots) -> (slots). 
  // Put the ct dim back in with bounds from the matrix layout, and equality
  // between the domain ct var and the range ct var.
```

lib/Utils/Layout/Hoisting.cpp
```diff
@@ -0,0 +1,119 @@
...
+  fromClone.insertVar(VarKind::Domain, 0, 1);
+  fromClone.insertVar(VarKind::Range, 0, 1);
+  std::optional<int64_t> ctUb =
+      matrixLayout.getConstantBound64(BoundType::UB, 2);
```
asraa:
```suggestion
      matrixLayout.getConstantBound64(BoundType::UB, matrixLayout.getVarKindOffset(VarKind:Range);
```

should we just make helpers as part of the tensor_ext dialect like `getCtOffset(relation)` and `getSlotOffset(relation)`?

asraa (approved):
YAY

lib/Utils/Layout/Hoisting.cpp
j2kun: Fixed

lib/Utils/Layout/Hoisting.cpp
j2kun: Done

lib/Utils/Layout/Hoisting.cpp
j2kun: Done. I don't think the helpers are needed right now, but maybe in the future it would make sense to do this at a higher level (e.g., a wrapper type for a CiphertextPacking around IntegerRelation that could expose a method like getNumCiphertexts)

PR #2127, opened 2025-08-21
Support round-tripping from IntegerRelation to ISL and back

j2kun:
A simple matter of copying over the matrix coefficients. Love the ISL API that let's you pick the order of the variables when converting to the underlying matrix.


lib/Utils/Layout/IslConversion.cpp
```diff
@@ -7,8 +7,10 @@
 #include "include/isl/mat.h"                  // from @isl
 #include "include/isl/space.h"                // from @isl
 #include "include/isl/space_type.h"           // from @isl
+#include "include/isl/val.h"                  // from @isl
```
asraa, line 10: curious, do you not need to separate the ISL headers anymore?

asraa (approved):
So simple!

lib/Utils/Layout/IslConversion.cpp
j2kun: we'll see when we copybara

Issue #2125, opened 2025-08-21
Print/parse new layout attribute using ISL basic map string representation


asraa:
Right now we use ISL basic map to represent an integer relation and then compute the AST for codegen using that. The ISL basic map representation looks something like this
```
{ [i0] -> [o0, o1] : o0 = 0 and o1 = 2i0 and i0 >= 0 and i0 <= 15 }
```

Where the `i` dimensions are inputs (and we can set the identifiers ourselves) and the `o` are outputs (which we can set to `ct, slot` for e.g.).

Our NewLayoutAttr has a very verbose format, with a flat set of dimensions and locals:
```
#new_layout = #tensor_ext.new_layout<domainSize=1, relation="(d0, d1, d2) : (d2 - d0 * 2 == 0, d0 >= 0, d2 >= 0, -d0 + 15 >= 0, -d2 + 31 >= 0, d1 == 0)">
```
and because of that we also have to add the # of domain vars and local vars to parse it properly.

Using the ISL dump would be a lot nicer! (And it should be possible to upstream a prettier format like that too eventually).

For this we would have to rewrite the attribute parsers and printers to go through ISL, which can also help simplify them before printing

PR #2130, opened 2025-08-22
ISL-string based layout attribute

j2kun:
The layout attribute is too messy when printing it as a flat IntegerSet, so instead we convert to and from ISL to use its string representation, which represents the formulas more readably.

Rebased on https://github.com/google/heir/pull/2127

Fixes https://github.com/google/heir/issues/2125


j2kun:
Almost done here, just need to update test syntax and assertions

j2kun:
> so cool! while i'm sure it works can you add a test and/or add to the td file what the attr looks like with locals? just as documentation

Done!

PR #2137, opened 2025-08-22
separate polynomial approximation into subpipeline and test atomically

j2kun:
This one looks like it was already pretty well tested. So I just isolated the passes into a subpipeline, added a distinguished place for pipeline tests (which I think we should migrate many of our other sub-pipeline tests to like mlir-to-cggi).

I did notice the small constants in the ReLU approximation in the test. I think as a P2 we should add an option to skip operations like mul's with such a small coefficient, since they will have problems with the scaling factor and are negligible anyway (this is what everyone does in the literature)


asraa (approved):
RE the small constants, I remember when reviewing the paterson stockmeyer that there is a threshold value that can be set for the coeffs. https://github.com/google/heir/blob/5d886ad274ecd9b3c7ebe835485fb13316368780/lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h#L20

but i don't know about the other methods but it would be nice to move that out to a parameter

PR #2141, opened 2025-08-22
[wip] Layout folding analysis

j2kun:


lib/Analysis/LayoutFoldingAnalysis/LayoutFoldingAnalysis.h
```diff
@@ -0,0 +1,109 @@
...
+class LayoutIsFreeLattice : public dataflow::Lattice<LayoutIsFree> {
+ public:
+  using Lattice::Lattice;
+};
+
+// An analysis that identifies and assigns the Secretness of an SSA value in a
+// program. This is used by other passes to selectively apply transformations on
+// operations that evaluate secret types. We use a forward dataflow analysis
+// because the Secretness state propagates forward from the input arguments of a
+// function down to values produced from operations that use these arguments.
```
asraa: nit: stale comment

j2kun:
> LGTM as is! I have a small concern about why the layout solver doesn't need to be run if hoisting changes the lattice, but I think it may be because any new convert layouts introduced by the solver would be immediately cleaned up if they can. 

Though I haven't investigated this in depth, my heuristic justification is that the hoisting pass walks the IR in reverse, so any changes it makes won't affect the IR upstream from the point of modification. This would not be true of other layout optimizers.

lib/Analysis/LayoutFoldingAnalysis/LayoutFoldingAnalysis.h
j2kun: Fixed

Issue #2136, opened 2025-08-22
Add infrastructure for testing for tensor_ext rewrites to validate kernel implementations


asraa:
https://github.com/google/heir/pull/2124#pullrequestreview-3142014819

Testing the kernel rewrites is a little tricky. This came up in a Baby Step Giant Step transform in #2124 that rewrites the reduction with adds/muls/rotates. Validating this is a little tricky, since we'll need to simulate the operations on vectors with data in the slots.

I think it probably does make sense to add support for rotations and vector types in the arithmetic DAG so we can start to simulate kernels that utilize the SIMD slots rotations (and aren't just elementwise).

We'll probably need to do a bunch of testing anyway on this since we need to know when we can/can't pad zeros, and things like that. Probably testing is the first step doing this analysis that can inform passes to optimize packing or padding options.

PR #2154, opened 2025-08-27
Test rotate_and_reduce kernel implementations

j2kun:
Migrates to using ArithmeticDag for rotate_and_reduce kernel implementation. Fixes https://github.com/google/heir/issues/2136

Also adds a kernel for "sum all slots" `rotate_and_reduce`, including partial reductions via a stride.

The main design change is that `KernelImplementation.h` switches from using `std::variant` to a less verbose option: an abstract base class with subclasses that handle SSA values vs literal values. This allows us to nearly seamlessly use the kernel-generating function for both literal values (for testing) and SSA values (for IR materialization). The operands just need to be cast (sometimes this can be done implicitly) like `SSAValue(::mlir::Value)` and `LiteralValue(std::vector<int>(...))`. However, I still have to use templates in the kernel-generating code implementation, because you can't store an abstract class as member data in the dag, you need the concrete class. So I added 

```cpp
template <typename T>
std::enable_if_t<std::is_base_of<AbstractValue, T>::value,
                 std::shared_ptr<ArithmeticDagNode<T>>>
implementRotateAndReduce(const T& vector, std::optional<T> plaintexts,
                         int64_t period, int64_t steps)
```

Which allows the implementation to use the concrete class, as well as restricts the allowable choices to be subclasses of `AbstractValue`.

This PR includes some significant reorganization:

- ArithmeticDag is now part of lib/Kernel
- IRMaterializingVisitor is moved to lib/Kernel and generalized to work for all ArithmeticDags whose leaves are SSAValue
- new Visitors are added for evaluating an ArithmeticDag of tensors and printing a dag as a string (in TestingUtils.cpp)
- LowerPolynomialEval and ChebyshevPatersonStockmeyer are migrated to use the new visitors
- `convertFloatToSemantics` is moved to MathUtils so it can be reused


lib/Dialect/TensorExt/Transforms/ImplementRotateAndReduce.cpp
```diff
@@ -4,53 +4,52 @@
...
 
+using ::mlir::heir::kernel::ArithmeticDagNode;
+using ::mlir::heir::kernel::implementRotateAndReduce;
+using ::mlir::heir::kernel::IRMaterializingVisitor;
+using ::mlir::heir::kernel::SSAValue;
+
 #define GEN_PASS_DEF_IMPLEMENTROTATEANDREDUCE
 #include "lib/Dialect/TensorExt/Transforms/Passes.h.inc"
 
 // TODO(#2136): Add a better way to test the correctness of this kernel.
```
asraa: remove this!

lib/Kernel/KernelImplementation.h
```diff
@@ -1,42 +1,182 @@
...
+    }
+    return vectorDag;
+  }
+
+  auto plaintextsDag = NodeTy::leaf(*plaintexts);
+
+  // Use a value of sqrt(n) as the baby step / giant step size.
+  int64_t numBabySteps = static_cast<int64_t>(std::floor(std::sqrt(steps)));
+  int64_t giantStepSize = numBabySteps;
+  int64_t numGiantSteps = steps / numBabySteps;
```
asraa, line 149: ok so one of your tests is with n = 8, but this still worked but i think thats because floor(sqrt(8)) = 2 and 4 * 2 = 8. fine

but what about 10? then floor(sqrt(10)) = 3 and 10 isn't divisible by 3

lib/Kernel/IRMaterializingVisitor.h
```diff
@@ -0,0 +1,56 @@
...
+        });
+  }
+
+  Value operator()(const ConstantNode& node) override;
+  Value operator()(const LeafNode<SSAValue>& node) override;
+  Value operator()(const AddNode<SSAValue>& node) override;
+  Value operator()(const SubtractNode<SSAValue>& node) override;
+  Value operator()(const MultiplyNode<SSAValue>& node) override;
+  Value operator()(const LeftRotateNode<SSAValue>& node) override;
+  Value operator()(const ExtractNode<SSAValue>& node) override;
```
asraa, line 45: propagating errors with these visitors isn't possible because of the Value return type - is it expected that issues like extracting on a scalar value or something will occur via an op verification issue. i guess the other way is to stop the ArithmeticDAG from being able to extract on certain ssa value types, but then that class will also need to propagate errors :/

j2kun: > extracting on a scalar value

This would be caught in the same way that we catch someone giving `ExtractOp::create` SSA values of the wrong type: at IR verification time.

lib/Kernel/KernelImplementation.h
j2kun: So I tried it with this:

```diff
diff --git a/lib/Kernel/RotateAndReduceImplTest.cpp b/lib/Kernel/RotateAndReduceImplTest.cpp
index 60a50ab6b..fe2285c8e 100644
--- a/lib/Kernel/RotateAndReduceImplTest.cpp
+++ b/lib/Kernel/RotateAndReduceImplTest.cpp
@@ -95,6 +95,25 @@ TEST(RotateAndReduceImplTest, TestUnitPeriodWithPlaintexts) {
   EXPECT_EQ(expected, actual);
 }

+TEST(RotateAndReduceImplTest, TestUnitPeriodWithIndivisibleN) {
+  int64_t n = 10;
+  int64_t period = 1;
+
+  std::vector<int> vector = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+  std::vector<std::vector<int>> plaintexts = {
+      {1, 2, 3, 4, 5, 6, 7, 8, 9, 10},      {2, 3, 4, 5, 6, 7, 8, 9, 10, 11},
+      {3, 4, 5, 6, 7, 8, 9, 10, 11, 12},     {4, 5, 6, 7, 8, 9, 10, 11, 12, 13},
+      {5, 6, 7, 8, 9, 10, 11, 12, 13, 14},   {6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
+      {7, 8, 9, 10, 11, 12, 13, 14, 15, 16}, {8, 9, 10, 11, 12, 13, 14, 15, 16, 17},
+      {9, 10, 11, 12, 13, 14, 15, 16, 17, 18}, {10, 11, 12, 13, 14, 15, 16, 17, 18, 19},
+  };
+
+  std::vector<int> expected = runNaive(vector, plaintexts, period, n);
+  std::vector<int> actual = runImpl(vector, plaintexts, period, n);
+  EXPECT_EQ(expected, actual);
+}
+
+
 TEST(RotateAndReduceImplTest, TestPeriod2WithPlaintexts) {
   int64_t n = 8;
   int64_t period = 2;

```

And I got

```
Expected equality of these values:
  expected
    Which is: { 330, 330, 340, 360, 390, 430, 480, 540, 610, 690 }
  actual
    Which is: { 240, 330, 328, 334, 336, 346, 364, 378, 400, 430 }
```

So it's wrong, but then again I was primarily porting the kernel that was there. How about we submit this as-is and update the kernel to support not-perfectly-divisible inputs tomorrow?

PR #2155, opened 2025-08-27
keyswitch inner lowering prep

AlexanderViand:
Nearly forgot about this 😅 
Prep for the next step of lowering relin/keyswitch with @j2kun  and @crockeea 

Cf. #2157

### DIFF
--- a/lib/Dialect/ModArith/IR/ModArithOps.td
+++ b/lib/Dialect/ModArith/IR/ModArithOps.td
@@ -230,4 +230,25 @@ def ModArith_SubIfGEOp : ModArith_Op<"subifge", [SameOperandsAndResultType]> {
   let assemblyFormat = "operands attr-dict `:` qualified(type($output))";
 }
 
+def ModArith_FastBasisConversionOp : ModArith_Op<"fast_basis_conv", [Pure]> {
+  let summary = "Fast RNS basis conversion using precomputed constants";
+  let description = [{
+    Converts values between RNS bases using the fast basis conversion algorithm.
+
+    Example:
+    ```mlir
+    %converted = mod_arith.fast_basis_conv %input : !mod_arith.mod_int<65537:i32>
+      -> !mod_arith.mod_int<32768:i32>
+    ```
+  }];
+
+  let arguments = (ins
+    ModArithLike:$input
+  );
+
+  let results = (outs
+    ModArithLike:$output
+  );
+}
+
 #endif  // LIB_DIALECT_MODARITH_IR_MODARITHOPS_TD_
--- a/lib/Dialect/ModArith/IR/ModArithOps.td
+++ b/lib/Dialect/ModArith/IR/ModArithOps.td
@@ -230,6 +230,7 @@ def ModArith_SubIfGEOp : ModArith_Op<"subifge", [SameOperandsAndResultType]> {
   let assemblyFormat = "operands attr-dict `:` qualified(type($output))";
 }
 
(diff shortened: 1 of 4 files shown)


AlexanderViand:
@crockeea

j2kun:
I'll start by extracting the partition op from this PR into its own.

lib/Dialect/ModArith/IR/ModArithOps.td
```diff
@@ -230,4 +230,57 @@ def ModArith_SubIfGEOp : ModArith_Op<"subifge", [SameOperandsAndResultType]> {
   let assemblyFormat = "operands attr-dict `:` qualified(type($output))";
 }
 
+// TODO (Cleanup): Move this to RNS dialect?
+def ModArith_FastBasisConversionOp : ModArith_Op<"fast_basis_conv", [Pure]> {
+  let summary = "Fast RNS basis conversion using precomputed constants";
```
crockeea, line 235: Some details on what this function accomplishes:
First, given x = (x0, x1, x2, ...) mod (q0, q1, q2, ...), we can use the Chinese Remainder Theorem to reconstruct x mod q0*q1*q2*... from (x0, x1, x2, ...). Let Q=q0*q1*q2*... and denote the CRT reconstruction by X=crt(x0,x1,x2...,q0,q1,q2).
The next piece we need is to "lift" x from mod Q to the integers. For this, we need a *representative*. It's convenient to use the 'standard representative' in [0, Q), but better for FHE noise control to use the 'canonical representative' in [-floor(Q/2), floor(Q/2)]. I recommend fixing the canonical representative for this operation. We can denote this by Y=lift(X).
Given an integer (not mod anything), we can then reduce it mod each modulus in the target basis, say (p0, p1, p2, ...) by computing (Y mod p0, Y mod p1, ...). Denote this by reduce(y, p0, p1, p2, ...)

So FastBasisConversion outputs reduce(lift(crt(x0,x1,x2,...,q0,q1,q2,...)), p0, p1, p2, ...). However, it does so without going through all the steps above. In particular, we want to avoid the CRT reconstruction because it requires large words to represent X. FastBasisConversion computes Y mod p_i directly, without a CRT reconstruction.

We are implementing Garner's algorithm; there are other FBC algorithms which achieve the same goal in different ways (e.g., using floating point arithmetic).

tests/Dialect/Polynomial/Conversions/polynomial_to_mod_arith/lower_keyswitch_inner.mlir
```diff
@@ -0,0 +1,55 @@
...
+  %c, %l = polynomial.key_switch_inner %p, %ksk : (!poly, tensor<10x2x!poly>) -> (!poly, !poly)
+  return %c, %l : !poly, !poly
+}
+
+// Expected Code
+func.func @test_keyswitch_inner_lowered(%x: !poly, %ksk: tensor<10x2x!poly>) -> (!poly, !poly) {
+  %xAsTensor = polynomial.to_tensor %xParts : !poly -> tensor<Nx!rns_L1>
+  // x has L RNS terms
+  // L is number of normal primes
+  %xParts = mod_arith.partition %xAsTensor {partition_size L/beta} : tensor<Nx!rns_L1> -> tensor<Nx??x!rns...>
```
crockeea, line 27: The [largest] partition size is the number of key-switch primes, which I call `k`. If \ell<=L is the number of primes on the input polynomial, then beta = #number of partitions = ceil(\ell/k). I don't think this yields a nice equation for k, though.

PR #2156, opened 2025-08-27
Lower the partition step of key_switch_inner

j2kun:
One of the ops for lowering `key_switch_inner` is to partition the coefficients of a polynomial. This op will eventually lower to a combination of

```
tensor.extract_slice (extract the divisible part)
tensor.reshape (reshape the divisible part)
tensor.extract_slice (extract the extra piece)
```

This actually kind of makes me think we don't need a dedicated op at all. We can just lower directly to the three ops this will lower to. Is there an advantage to having this intermediate op? I suppose that would depend on how other methods of implementing keyswitch might differ with this one.

Part of #2157

CC @crockea @AlexanderViand

### DIFF
--- a/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
+++ b/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
@@ -10,12 +10,14 @@
 #include <utility>
 #include <vector>
 
+#include "lib/Dialect/CKKS/IR/CKKSAttributes.h"
 #include "lib/Dialect/ModArith/IR/ModArithOps.h"
 #include "lib/Dialect/ModArith/IR/ModArithTypes.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialAttributes.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialDialect.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialOps.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialTypes.h"
+#include "lib/Parameters/CKKS/Params.h"
 #include "lib/Utils/APIntUtils.h"
 #include "lib/Utils/ConversionUtils.h"
 #include "lib/Utils/Polynomial/Polynomial.h"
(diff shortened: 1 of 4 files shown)


j2kun:
OK I added a second commit (so we can see them side by side) that shows what the lowering would look like if we skipped having a dedicated `partition` op and lowered to what `tensor_ext.partition` would have lowered to anyway. It seems tidy enough that I think we should just do that.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -1246,6 +1246,67 @@ struct ConvertINTT : public OpConversionPattern<INTTOp> {
...
+    RankedTensorType coefficientTensorType =
+        cast<RankedTensorType>(adaptor.getValue().getType());
+
+    if (coefficientTensorType.getRank() != 1) {
+      return rewriter.notifyMatchFailure(
+          op,
+          "Can only lower key_switch_inner when input is a single polynomial");
+    }
+
+    int64_t partSize = 7;  // FIXME: how to select this?
```
crockeea: Part of the CKKS parameters is the number of key-switch primes `k`. The number of keyswitch primes determines the size of each partition, i.e., `partSize=k`. 

Roughly: For a security level n, the sum of the bit-lengths of the `L` "normal" primes q_0,...,q_{L-1} and the `k` key-switch primes p_0,...,p_{k-1} must be smaller than some bound B(n). Though it varies from application to application, one common parameterization is that all p_i are 60 bits, q_0 is 60 bits, and all other q_i are 40 bits, thus 60(k+1) + 40*(L-1) <= B(n)

With this constraint, you get to pick k and L however you like, subject to 1<= k<= L. This results in a performance/utility tradeoff: when k is small (say 1), we can allocate more bits to the normal primes. This means we get more computation levels prior to bootstrapping. However, it has the effect of making key-switching very slow (because the decomposition step divides the input into groups of size 1, so there are many groups). When k is large, say k~L, key switching is much faster (because now there's only one group output from decomposition), but you used a huge amount of your modulus budget on key-switch primes, so you don't have nearly as many levels for computation.

TL;DR, `partSize=k` is not something you would pick statically, this is a scheme parameter.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -1246,6 +1246,67 @@ struct ConvertINTT : public OpConversionPattern<INTTOp> {
...
+    int64_t numPieces = coefficientTensorType.getShape()[0] / partSize;
```
crockeea: Recommend naming this "numFullPartitions" or similar.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
j2kun: For the purpose of this pass, then, we would expect to find this value on the ckks parameters, which would be attached to the module op as a static attribute (static here just means "known at compile time"). I will have to check if we populate the key switch prime count already or not in our parameter selection pass.

crockeea (approved):

j2kun:
Ok I'm going to address Eric's comments and otherwise remove the new op, just go straight to the lowered partition. I'll also update this PR to have  some tests.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
j2kun: So we have the number of keyswitch primes on the parameters already (as "pi.size()") but I can't say how well it is chosen. I think most of our CKKS parameter selection stuff is default-ish right now because we don't have a good checked-in CKKS noise model.

However, I did want to see to what extent this is specific to CKKS or whether it applies equally to BGV/BFV. We have the same parameter for those schemes and we could reuse it here, though in this PR so far I have just hard-coded CKKS.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
j2kun: Done

j2kun:
Updated and ready to merge.

For this IR:

```mlir
!Z1032955396097_i64 = !mod_arith.int<1032955396097 : i64>
!Z1095233372161_i64 = !mod_arith.int<1095233372161 : i64>
!rns_L1 = !rns.rns<!Z1095233372161_i64, !Z1032955396097_i64>
!rns_new = !rns.rns<!Z1095233372161_i64>
#ring_rns_L1_1_x1024 = #polynomial.ring<coefficientType = !rns_L1, polynomialModulus = <1 + x**1024>>
#ring_new = #polynomial.ring<coefficientType = !rns_new, polynomialModulus = <1 + x**1024>>
!poly = !polynomial.polynomial<ring = #ring_rns_L1_1_x1024>
!poly_new = !polynomial.polynomial<ring = #ring_new>

module attributes {
    ckks.schemeParam = #ckks.scheme_param<
      logN = 14,
      Q = [36028797019389953, 35184372121601, 35184372744193, 35184373006337, 35184373989377, 35184374874113],
      P = [36028797019488257, 36028797020209153],
      logDefaultScale = 45
    >
} {
  // CHECK: @test_keyswitch_inner_lowering
  func.func @test_keyswitch_inner_lowering(%p: !poly, %ksk: tensor<10x2x!poly>) -> (!poly, !poly) {
    // CHECK-NOT: polynomial.key_switch_inner
    %c, %l = polynomial.key_switch_inner %p, %ksk : (!poly, tensor<10x2x!poly>) -> (!poly, !poly)
    return %c, %l : !poly, !poly
  }
}
```

We get the following, where the `...` hides the original key_switch_inner op that cannot yet be replaced (hence the lowering currently fails because it's incomplete).

```mlir
"func.func"() <{function_type = (tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>, tensor<10x2x1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>) -> (tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>, tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>), sym_name = "test_keyswitch_inner_lowering"}> ({
^bb0(%arg0: tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>, %arg1: tensor<10x2x1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>):
  %0 = "tensor.extract_slice"(%arg0) <{operandSegmentSizes = array<i32: 1, 0, 0, 0>, static_offsets = array<i64: 0>, static_sizes = array<i64: 1024>, static_strides = array<i64: 1>}> : (tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>) -> tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>
  %1 = "arith.constant"() <{value = dense<[512, 2]> : tensor<2xindex>}> : () -> tensor<2xindex>
  %2 = "tensor.reshape"(%0, %1) : (tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>, tensor<2xindex>) -> tensor<512x2x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>
  %3 = "tensor.extract_slice"(%arg0) <{operandSegmentSizes = array<i32: 1, 0, 0, 0>, static_offsets = array<i64: 1024>, static_sizes = array<i64: 0>, static_strides = array<i64: 1>}> : (tensor<1024x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>) -> tensor<0x!rns.rns<!mod_arith.int<1095233372161 : i64>, !mod_arith.int<1032955396097 : i64>>>
  ...
}) : () -> ()
```

PR #2167, opened 2025-08-29
Refactor other lower-polynomial-eval methods to use ArithmeticDag

j2kun:


lib/Transforms/LowerPolynomialEval/LowerPolynomialEval.cpp
```diff
@@ -28,16 +28,20 @@ struct LowerPolynomialEval
     if (method == PolynomialApproximationMethod::Automatic) {
       patterns.add<LowerViaHorner, LowerViaPatersonStockmeyerChebyshev,
                    LowerViaPatersonStockmeyerMonomial>(context,
-                                                       /*force=*/false);
+                                                       /*force=*/false,
+                                                       minCoefficientThreshold);
```
j2kun: On second thought, should I even allow dropping coefficients on monomial-basis methods?

asraa: I don't think so - at high degrees anyway there's going to be a lot of error build up from rounding or dropping coefficients. I think monomial basis methods should probably be represented as close as possible (and probably won't even be used that much anyway)

asraa: I think the same holds true about horner's/P-S method too? all of those (except the chebyshev p-s) are still a monomial basis approximation

j2kun: Fixed

Issue #1504, opened 2025-03-03
Remove remaining uses of mlir-runner from the test suite


j2kun:
Cf. the comments on https://github.com/google/heir/pull/1466

mlir-runner adds a ton of build dependencies because the upstream bazel rules pull in every MLIR dialect and transform, which also pulls in a lot of extra LLVM code (e.g., for gpu targets and whatnot).

If we convert the remaining uses of mlir-runner to lower all the way to LLVM and run C-based tests, we will probably reduce the time to build tests by maybe 30% (10 minutes)?

PR #2175, opened 2025-09-02
Remove mlir-runner from the last two critical testsuites

j2kun:
mod_arith_to_arith and heir_polynomial_to_llvm are the last places in the codebase that we aren't planning to delete that use mlir-runner.

After this PR goes in, we are planning to delete `tests/Transforms/memref_to_arith/memref_global_replace/` and then all uses of `mlir-runner` can be deleted, which should greatly reduce build/test times.

Part of https://github.com/google/heir/issues/1504

#vc4a


tests/Dialect/Arith/Transforms/runner/lower_split.mlir
```diff
@@ -1,17 +0,0 @@
...
-// 1e82868a74 + 9f
-// CHECK: 82 86 8b 13
-func.func @test_lowera_quarter_mul(%arg0: i32) -> i32{
-  %c1 = arith.constant 522067228: i32 // Hex 1f1e1d1c
-  %c2 = arith.constant 159 : i8
-  %3 = arith.extui %c2 : i8 to i32
-  %4 = arith.muli %c1, %arg0 : i32
-  %5 = arith.addi %4, %3 : i32
-  return %5 : i32
-}
```
asraa, line 17: this is really weird but git isn't showing this as a deleted file and is just showing this as deleting the content?

tests/Dialect/ModArith/Conversions/mod_arith_to_arith/runner/BUILD
```diff
@@ -1,10 +1,198 @@
-load("//bazel:lit.bzl", "glob_lit_tests")
+load("@heir//bazel:lit.bzl", "glob_lit_tests")
+load("@heir//tests/llvm_runner:llvm_runner.bzl", "llvm_runner_test")
 
 package(default_applicable_licenses = ["@heir//:license"])
 
 glob_lit_tests(
```
asraa: you can just remove this glob_lit_test macro now

tests/llvm_runner/memref_types.h
```diff
@@ -0,0 +1,23 @@
+#ifndef TESTS_LLVM_RUNNER_MEMREF_TYPES_H_
+#define TESTS_LLVM_RUNNER_MEMREF_TYPES_H_
+
+#include <cstdint>
+
+// StridedMemRef descriptor type with static rank.
+template <typename T, int N>
```
asraa, line 9: It looks like these types are also defined in the cpp source file (memrefCopy.cpp) so its interesting to me that the compiler doesn't complain

tests/Dialect/ModArith/Conversions/mod_arith_to_arith/runner/BUILD
j2kun: Done

tests/llvm_runner/memref_types.h
j2kun: Cleaned it up a bit more.

tests/Dialect/Arith/Transforms/runner/lower_split.mlir
j2kun: Hmm it shows up as deleted for me. I pushed a few changes maybe try again

lib/Transforms/EmitCInterface/EmitCInterface.td
```diff
@@ -0,0 +1,10 @@
+#ifndef LIB_TRANSFORMS_EMITCINTERFACE_EMITCINTERFACE_TD_
+#define LIB_TRANSFORMS_EMITCINTERFACE_EMITCINTERFACE_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def EmitCInterface : Pass<"emit-c-interface"> {
+  let summary = "Adds llvm.emit_c_interface to each external function.";
```
ZenithalHourlyRate: Maybe non-external?

lib/Transforms/EmitCInterface/EmitCInterface.cpp
ZenithalHourlyRate, line 1: I would recommend merging `--emit-c-interface` into the `heir-polynomial-to-llvm` pipeline, as it only adds a _new_ function (actually a wrapper) without modifying the original function's signature when lowered to `llvm`, and old code can use the signature of the original.

j2kun: Done

lib/Transforms/EmitCInterface/EmitCInterface.td
j2kun: Done

PR #2179, opened 2025-09-03
fix clang matrix and ensure lld is installed

j2kun:
For clang-20:

> E: Unable to locate package clang-20

For clang-19:

> clang: error: invalid linker name in argument '-fuse-ld=lld'

For clang-17:

> external/openfhe/src/core/include/utils/parallel.h:40:14: fatal error: 'omp.h' file not found
>   40 |     #include <omp.h>
>      |              ^~~~~~~

Ubuntu's default set of packages seem to be a mess, so I'm going to try installing from llvm's apt list, which seems more consistent.


.github/ensure_clang_version.sh
asraa, line 1: delete this?

.github/workflows/docs.yml
```diff
@@ -16,10 +16,15 @@ jobs:
...
     steps:
+    - name: Install dependencies
+      run: |
+        ./.github/install_clang_version.sh 19
+
     - name: Checkout
       uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # pin@v3
       with:
         fetch-depth: 0         # Fetch all history for .GitInfo and .Lastmod
+
```
asraa, line 27: wanna remove fastbuild in this file?

j2kun: Done

j2kun:
OK @asraa this should be ready now

.github/workflows/build_and_test.yml
```diff
@@ -13,42 +13,39 @@ concurrency:
...
+          # the cache, so if the compiler changes but the cache does not, it
+          # will try to use the old compiler path.
+          which -a clang | xargs ls -al
+          ls -al /etc/alternatives/clang
+          bazel cquery --output=starlark --starlark:expr="str(providers(target))" @bazel_tools//tools/cpp:current_cc_toolchain | sed 's/,/\n/g' | grep -C 5 clang
 
       - name: "Run `bazel build`"
         run: |
-          bazel build -c opt //... --explain=bazel-explain.log --verbose_explanations
+          bazel build -c opt //...
```
asraa, line 47: why not cache these builds?

j2kun: They are cached. It's weird because the earlier cache action sets up a post-action cache save when there's a cache miss. The action below that was removed is a forced cache save that errs when the cache key is present. I was earlier experimenting with trying to override the github cache because I wasn't sure it was working properly and I was kind of fumbling.

Issue #2162, opened 2025-08-28
Determine the right tradeoff to use when baby-step-giant-step step sizes are not evenly divisible


j2kun:
Filing issue for pending TODO.

The problem is that when we are lowering rotate-and-reduce via baby-step-giant-step, if the `numBabySteps` does not divide the total number of `steps`, we have two options:

1. Add some extra ops at the end to account for the leftover
2. Find the closest number to sqrt(n) that does divide `steps` and use that.

Not sure which is best, and I decided to punt on doing the most optimal thing for now and implementing a naive version of (2). It seems like (2) is suboptimal in extreme cases like a prime number of steps, so the question is how to select between (1) and (2)

lawrencekhlim:
I may be able to provide some insight from my experience:

1. Let's say you want to do 11 periodic rotations (that all sum up) with BSGS. Ideally, you perform 3-4 baby steps and 3-4 giant steps. For instance, you can do 4 baby steps and 3 giant steps as follows:
Baby steps: [0, 1, 2, 3]
Giant Step 1: [0, 1, 2, 3]
Giant Step 2: [4, 5, 6, 7]
Giant Step 3: [8, 9, 10]

Note that in Giant Step 3, you do not need to add in the fourth baby step to achieve correctness. I guess a simple algorithm would be to round up from the square root of number of rotations (ceil(sqrt(N)), which would be the number of baby steps, and exclude leftovers from the final giant step.

In fact, in general, it's okay to skip summing some of the baby steps if they aren't needed.

2. If you (in the future) have hoisting or even [double hoisting](https://eprint.iacr.org/2020/1203) BSGS implemented, then the optimal ratio between N1 (number of baby steps) and N2 (number of giant steps) (for N1 * N2 = N) will differ. Single Hoisting will prefer N1 to be larger around 2 times N2, while double hoisting will prefer N1 to be larger around 8-16 times N2.

EDIT: Corrected some wrong information about double hoisting.

seydang:
The [OpenFHE default parameter selection](https://github.com/openfheorg/openfhe-development/blob/aa8a86e1143f1e47d4354bdd757080e903ba5875/src/pke/lib/scheme/ckksrns/ckksrns-utils.cpp#L700) for BSGS in bootstrapping favors the N2 (giant step) to be larger than N1 (baby step). I believe, the N1 and N2 in our discussion correspond to the b and g parameters in the permalink, respectively.
[...]

PR #2196, opened 2025-09-08
Add fuzz testing framework

j2kun:
I'd like to try adding a fuzz testing framework to help with the testing of our polynomial and kernel work. This is the boilerplate for getting the dep added. Before merging this I'd like to try porting the polynomial eval lowering methods to use this framework


asraa (commented):
ooh cool! i used to work on oss-fuzz and i'd be happy to integrate it so we have open source automated fuzzing

j2kun:
@asraa let's revive this one for #2162

fuzztest.bazelrc
```diff
@@ -0,0 +1,84 @@
+### DO NOT EDIT. Generated file.
+#
+# To regenerate, run the following from your project's workspace:
+#
+#  bazel run @com_google_fuzztest//bazel:setup_configs > fuzztest.bazelrc
```
asraa: I know this is generated, but isn't the dep `@fuzztest`?

lib/Kernel/RotateAndReduceFuzzTest.cpp
```diff
@@ -0,0 +1,175 @@
...
+  if (vector.empty() || steps <= 0 || period <= 0) return;
+
+  std::vector<int> expected = runNaiveNoPlaintexts(vector, period, steps);
+  std::vector<int> actual = runImpl(vector, {}, period, steps);
+
+  // Verify result matches the naive implementation
+  EXPECT_EQ(expected, actual);
+}
+
+// Property: Different reduce operations should produce different results
```
asraa: The test doesn't end up asserting this property, but I wonder is there even any value in this test? Maybe a test with a muli reduce op should just be handled comparing it to a naive implementation

fuzztest.bazelrc
j2kun: Agreed. fixed.

lib/Kernel/RotateAndReduceFuzzTest.cpp
j2kun: I agree, a fuzz test is not the right place for testing this anyway. Removed.

PR #2199, opened 2025-09-08
Remove old layout code

j2kun:
This PR removes the old layout code (based on plain affine maps and alignment attributes) and migrates all existing code paths to the new layout system based on integer relations.

I will attempt to summarize the many changes that went into this. The short version is: I had to clean up a lot of technical debt from our prior lack of support for multi-packed ciphertexts, and this impacted pretty much every part of the arithmetic pipelines.

## Pipeline changes

- Following from https://github.com/google/heir/pull/2277, now that most scheme-level ops are elementwise mappable, the central mlir-to-rlwe pipeline now has a `elementwise-to-affine` step that converts batch scheme ops to loops.
- `implement-shift-network` has been added to the `mlir-to-secret-arithmetic` subpipeline
- As the new layout system uses the new `implement-rotate-and-reduce` baby-step giant-step optimization, this can produce rotations of server-side constant plaintexts, and those are not lowered in secret-to-scheme because they have no secret operands or results. So the arithmetic pipelines now have a step of `tensor-ext-to-tensor` after all secret tensor-ext ops are lowered. 
- Similarly, a new canonicalization pattern, `FoldRotateOfConst`, was added to fold a rotated dense constant (e.g., a plaintext mask) to the corresponding pre-rotated dense constant.
- Minor: a new step for `fold-constant-tensors` was added to help simplify extract_slice of tensor.splat which occurred as an edge case of the `cmux` test; the then/else branches were dynamically-valued cleartext scalars but the test was a ciphertext, so the then/else values got layouts implemented as a splat, and then were immediately slice-extracted for plaintext-ciphertext ops. So this pattern folds that to another splat and removes the original splat.

### General

- Most conversion passes needed extra support in their type converters for tensors-of-secret things to accommodate the new elementwise versions of each op.
- Many places in the codebase assumed a type was a scalar ciphertext/plaintext type, and needed to have some version of `getElementTypeOrSelf` to replace it.
- Various validations checked the shape of a tensor was exactly `{numSlots}` and this was changed to check the trailing axis.
- Patterns for various tensor ops (insert/insert_slice, extract/extract_slice, empty, splat, from_elements) were added in many lowerings. This is needed mainly because some APIs weren't sensible to make elementwise (e.g., openfhe make_packed_plaintext can't be elementwise because the OpenFHE API doesn't support this)
- Some analysis passes required slight tweaks to support tensor operations that had no impact on noise.
- `addTensorConversionPatterns` was added to generically type-convert operands and results of "blind" tensor ops in lowerings like `lwe-to-lattigo`.
- Many `llvm_runner` tests were using `StridedMemRefType<int64_t>` for the C API, which defaults to a rank-1 tensor, and so for multi-packed ciphertexts these were updated to `StridedMemRefType<int64_t, 2>` or similar.
- A variety of tests were reorganized and renamed to better match the lib/ structure.

### `secret-to-<scheme>`

- Support for insert/extract slice in `secret-to-scheme` conversions are limited to operating on full ciphertexts (i.e., you can only extract a slice from a ciphertext-semantic tensor if you're extracting one or more full rows).
- "encode" ops are not elementwise, so the way to handle encoding a multi-packed plaintext is to extract-slice all the rows, encode each one, and from_elements them back together.

### `secret-to-mod-arith`

- A bug was fixed in that, due to a namespace conflict, `secret-to-mod-arith` was using _context-aware_ conversion patterns when it should not have been.

### `secret-to-ckks`

This pass had a number of hard-coded handling of "multi-packed" ciphertexts that were added purely to make the initial halevi-shoup matvec demo work from Summer of 2024. This code is all removed and the corresponding ops are supported in generality.

### `tensor-ext-to-tensor`

(submitted separately as https://github.com/google/heir/pull/2291)

### `layout-propagation`

- `OperandLayoutRequirementOpInterface` got a new `isSecret` option because `tensor.insert` has two regimes: if it's inserting a cleartext scalar into a plaintext or ciphertext, the scalar does not require a layout. If it's inserting a ciphertext scalar, then the scalar does require a layout.
- A hook for `rectifyIncompatibleOperandLayouts` was added for `tensor.insert`, which handles the case of inserting a secret scalar into a secret destination by layout-converting the scalar to the same layout as the dest tensor. This is bad but it's also bad to insert scalars into secret tensors so I don't expect this kernel to be used all that much. (It was in a test that did a manual matvec mul by a nested loop with inner body extract, mul, sum, insert)
- The default layout for the result of a `tensor.extract` op now does not require replicating the scalar to all slots. This allows us to avoid some layout conversions that are unnecessary when the tensor.extract is the last operation in the IR, and in other cases where the result is used in a way that doesn't strictly require full replication of the scalar in every slot.

### `secret-insert-mgmt`

All these passes now add `mgmt.init` ops for relevant tensor operations that have plaintext operands (e.g., insert_slice into a plaintext `tensor.empty`)

## Emitter Changes

- Lattigo and OpenFHE emitters are completely converted to "flattened" tensor types (e.g., all code-genned vector-like types are 1D and all access indices are code-genned in a flattened form). This brings both emitters in line, such that more codegen can be shared between the two. All restrictions on tensor shapes are removed.
- SCF codegen was added to the lattigo emitter
- A bug was fixed in the lattigo emitCopySlice where assignment done inside a loop was causing it to  shadow (via lexical scope) the actual destination of the copy.

## Issues

- The debug logging mechanism is now unable to unpack, and so logged results are _packed_ cleartexts Cf. #2280 for places where this occurs.
- To tests (matmul_loop.mlir and matmul_unrolled.mlir) were removed because, while they compiled successfully with the new code, the generated code was quite large and the tests became slow. Asra and I agreed we don't need these tests anymore.
-


lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
j2kun, line 1: Just sorting here

lib/Dialect/LWE/IR/LWEOps.h
```diff
@@ -32,15 +32,19 @@ namespace lwe {
 //===----------------------------------------------------------------------===//
 // Op verifiers
 //===----------------------------------------------------------------------===//
-
-inline LWECiphertextType getCtTy(Value v) {
-  return cast<LWECiphertextType>(getElementTypeOrSelf(v.getType()));
+//
```
asraa:
```suggestion
```

lib/Dialect/ModuleAttributes.h
```diff
@@ -70,10 +70,13 @@ constexpr const static ::llvm::StringLiteral kClientEncFuncAttrName =
     "client.enc_func";
 constexpr const static ::llvm::StringLiteral kClientDecFuncAttrName =
     "client.dec_func";
+constexpr const static ::llvm::StringLiteral kClientPackFuncAttrName =
+    "client_pack_func";
```
asraa:
```suggestion
    "client.pack_func";
```

for consistency with above?

tests/Dialect/Secret/Conversions/secret_to_bgv/ciphertext_plaintext.mlir
```diff
@@ -35,6 +35,38 @@ module attributes {bgv.schemeParam = #bgv.scheme_param<logN = 14, Q = [67239937,
...
   }
 
+  // CHECK: func @test_mul_plain_elementwise
+  // CHECK-SAME: %[[arg0:.*]]: tensor<2x!lwe.lwe_ciphertext
+  // CHECK-SAME: %[[arg1:.*]]: tensor<2x1024xi1>
+  func.func @test_mul_plain_elementwise(%arg0 : !secret.secret<tensor<2x1024xi1>> {mgmt.mgmt = #mgmt}, %arg1 : tensor<2x1024xi1>) -> (!secret.secret<tensor<2x1024xi1>> {mgmt.mgmt = #mgmt}) {
+    %arg1_attr = mgmt.init %arg1 {mgmt.mgmt = #mgmt} : tensor<2x1024xi1>
+    %0 = secret.generic(%arg0 : !secret.secret<tensor<2x1024xi1>>) {
+    // CHECK: %[[extracted:.*]] = tensor.extract_slice %[[arg1]]
+    // CHECK: %[[v0:.*]] = lwe.rlwe_encode %[[extracted]]
```
asraa, line 45: assert two of these for each of the slices?

lib/Dialect/Secret/Conversions/SecretToCKKS/SecretToCKKS.cpp
```diff
@@ -136,25 +133,35 @@ class SecretToCKKSTypeConverter
...
+    SmallVector<IntegerAttr, 6> modulusChain;
     for (auto modArithType :
          cast<rns::RNSType>(ring_.getCoefficientType()).getBasisTypes()) {
       auto modulus = cast<mod_arith::ModArithType>(modArithType).getModulus();
-      moduliChain.push_back(modulus);
+      modulusChain.push_back(modulus);
+    }
+
+    Type messageType = type.getValueType();
+    if (tensorValueType && tensorValueType.getRank() > 1) {
```
asraa, line 152: oh THANK YOU for simplifying our need for all that logic. It's so much nicer when we know our types are going to be (ct, slot) tensors

lib/Dialect/Secret/Conversions/Patterns.h
```diff
@@ -55,6 +57,43 @@ struct ConvertClientReveal
...
+  using SecretGenericOpConversion<tensor::InsertSliceOp,
+                                  tensor::InsertOp>::SecretGenericOpConversion;
+
+  FailureOr<Operation*> matchAndRewriteInner(
+      secret::GenericOp op, TypeRange outputTypes, ValueRange inputs,
+      ArrayRef<NamedAttribute> attributes,
+      ContextAwareConversionPatternRewriter& rewriter) const override;
+};
+
+// An empty ciphertext-semantic tensor cab ne used as the initializer of a
```
asraa:
```suggestion
// An empty ciphertext-semantic tensor can be used as the initializer of a
```

lib/Dialect/Secret/Conversions/SecretToModArith/SecretToModArith.cpp
```diff
@@ -348,6 +350,57 @@ class SecretGenericOpCipherPlainConversion
...
+    if (!innerOp) {
+      return failure();
+    }
+
+    ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+    Value source = innerOp.getSource();
+    Value dest = innerOp.getDest();
+
+    if (op.getOpOperandForBlockArgument(dest) != nullptr) {
+      // A secret dest is already handled
```
asraa: could you add a comment here clarifying? i'm a little uncertain where it's already handled (not by another pattern I think,  unless you mean it would be correctly covered by ConvertAnyNestedGeneric?)

lib/Dialect/Secret/Conversions/Patterns.cpp
```diff
@@ -243,5 +246,177 @@ LogicalResult ConvertClientReveal::matchAndRewrite(
   return success();
 }
 
+static std::optional<int64_t> getConstantInt(OpFoldResult ofr) {
```
asraa: use `std::optional<int64_t> getConstantIntValue(OpFoldResult ofr);` from StaticValueUtils.h?

lib/Dialect/Secret/Conversions/Patterns.cpp
```diff
@@ -243,5 +246,177 @@ LogicalResult ConvertClientReveal::matchAndRewrite(
...
+  if (llvm::any_of(strides, [&](OpFoldResult ofr) {
+        return getConstantInt(ofr) != 1;
+      })) {
+    return rewriter.notifyMatchFailure(op,
+                                       "expected slice to have unit stride");
+  }
+
+  // The slice's size must match the ciphertext size.
+  // E.g., sizes[-1] == 4096.
+  if (inputTy.getDimSize(lastDim) != getConstantInt(sizes[lastDim])) {
```
asraa: the pattern also says that it should match on extracting one ciphertext - should you also check that the first size is exactly 1?

lib/Dialect/Secret/Conversions/Patterns.cpp
```diff
@@ -243,5 +246,177 @@ LogicalResult ConvertClientReveal::matchAndRewrite(
...
+
+  auto validationRes =
+      validateSliceAlignment(inputs[0], offsets, sizes, strides, op, rewriter);
+  if (failed(validationRes)) {
+    return validationRes;
+  }
+
+  // Now convert the op to a tensor.extract op that extracts one ciphertext.
+  SmallVector<Value> indices;
+  for (size_t i = 0; i < offsets.size() - 1; ++i) {
```
asraa: why loop when the validation asserts that there's only 2 indices (corresponding to ct slot)?

lib/Dialect/Secret/Conversions/Patterns.cpp
```diff
@@ -243,5 +246,177 @@ LogicalResult ConvertClientReveal::matchAndRewrite(
...
+    if (auto idx = getConstantInt(offsets[i]); idx.has_value()) {
+      indices.push_back(
+          arith::ConstantIndexOp::create(rewriter, op.getLoc(), idx.value()));
+    } else {
+      return rewriter.notifyMatchFailure(
+          op, "expected extract_slice to have constant offsets");
+    }
+  }
+
+  auto resultCtTy = cast<lwe::LWECiphertextType>(outputTypes[0]);
```
asraa, line 335:
```suggestion
  // Exactly one ciphertext output is guaranteed since validateSliceAlignment ensures that the result size is exactly a ciphertext size
  auto resultCtTy = cast<lwe::LWECiphertextType>(outputTypes[0]);
```

lib/Dialect/Secret/Conversions/Patterns.cpp
j2kun: Good call!

lib/Dialect/Secret/Conversions/SecretToModArith/SecretToModArith.cpp
j2kun: Indeed, the core of the pattern here is that we need to insert the encode op for a cleartext destination, and the other cases are handled by ConvertAnyNestedGeneric because the generic type converter does the right thing.

Added comments clarifying.

tests/Dialect/Secret/Conversions/secret_to_bgv/ciphertext_plaintext.mlir
j2kun: Done

lib/Dialect/Secret/Conversions/Patterns.cpp
j2kun: Good point. I also slightly generalized it to allow offsets[0] to fold to a Value (because nothing here strictly requires that first offset to be statically known)

lib/Dialect/Secret/Conversions/Patterns.cpp
j2kun: Done

Issue #1662, opened 2025-04-04
ConvertToCiphertextSemantics: improve kernels for tensor.extract/insert


j2kun:
These ops are here for completeness mainly, but we have lots of tests that do loops with extract/insert. The new layout system isn't yet equipped to mix the HECO-style heuristic passes with the explicit layout management, so for now we run the HECO passes first and have default lowerings for insert/extract ops in the new layout pipeline.

As a stopgap, convert-to-ciphertext-semantics has kernels for tensor.insert/extract, but they are incomplete because they do not handle the full space of possible layout attributes, just the default case of full replication.

Issue #1542, opened 2025-03-10
Support packing data across multiple ciphertexts in a tensor


j2kun:
The layout system supports describing a layout that packs data among multiple ciphertexts (e.g., `(x, y, z) -> (x, F(x, y, z))` would use the first dimension as a ciphertext selector while the slot within a ciphertext is chosen according to `F`.

The initial `convert-to-ciphertext-semantics` pass supports only packing into a single ciphertext for now.

PR #2204, opened 2025-09-09
Add a tensor.insert kernel for new layouts

j2kun:
tensor.insert kernel is broken into four cases:

  - insertion indices are static/dynamic
  - scalar `v` to insert is secret/cleartext

In all cases, this PR has the kernel generate a mask for the ciphertext-semantic scalar and dest tensors. The mask for the scalar has 1's or `v`'s for the secret/cleartext cases, and the mask for the dest has 1's everywhere except 0's in the target of the insertion.

If the indices are static, then the insertions can be done directly: take the layout, add constraints for the static indices on the domain, then simplify and (using ISL) enumerate the points of the range. Each point becomes one insertion. A new utility `getRangePoints` was added to support this, which can be generalized as needed.

If the indices are dynamic, then we have to use the codegen, where the body of the loop does the relevant mask construction.

This improves over the previous insert kernel in that:

- Rotations are no longer required
- It supports dynamic insertion indices (previous one required static)
- It supports arbitrary layouts for the scalar (including multi-packed ciphertexts, fixing the tensor.insert parts of #1542 and #1662)

This one feels like another win for the new layout system, since it is so straightforward to express the assembly of a special plaintext mask.

### DIFF
--- a/lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
+++ b/lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
@@ -26,6 +26,7 @@
 #include "lib/Utils/ContextAwareConversionUtils.h"
 #include "lib/Utils/ContextAwareDialectConversion.h"
 #include "lib/Utils/ContextAwareTypeConversion.h"
+#include "lib/Utils/Layout/Codegen.h"
 #include "lib/Utils/Layout/Utils.h"
 #include "lib/Utils/MathUtils.h"
 #include "lib/Utils/TransformUtils.h"
@@ -69,10 +70,12 @@ namespace mlir {
 namespace heir {
 
 namespace {
-using ::mlir::heir::kernel::ArithmeticDagNode;
-using ::mlir::heir::kernel::implementHaleviShoup;
-using ::mlir::heir::kernel::IRMaterializingVisitor;
-using ::mlir::heir::kernel::SSAValue;
+using kernel::ArithmeticDagNode;
+using kernel::implementHaleviShoup;
+using kernel::IRMaterializingVisitor;
+using kernel::SSAValue;
+using presburger::IntegerRelation;
+using presburger::VarKind;
 using tensor_ext::LayoutAttr;
 using tensor_ext::NewLayoutAttr;
 
@@ -1140,13 +1143,18 @@ class ConvertTensorInsert
       setMaterializedAttr(op);
       return success();
     }
-    LayoutAttr tensorLayout = cast<LayoutAttr>(tensorLayoutResult.value());
     FailureOr<Attribute> resultLayoutResult =
         getTypeConverter()->getContextualAttr(op.getResult());
     if (failed(resultLayoutResult)) {
       return op.emitError() << "failed to fetch layout attribute for input";
     }
-    LayoutAttr resultLayout = cast<LayoutAttr>(resultLayoutResult.value());
+
+    LayoutAttr tensorLayout = dyn_cast<LayoutAttr>(tensorLayoutResult.value());
+    LayoutAttr resultLayout = dyn_cast<LayoutAttr>(resultLayoutResult.value());
+
+    if (!tensorLayout || !resultLayout) {
+      return failure();  // code will be deleted soon, no need to give message
+    }
 
     // The indices at which to insert must be materialized via the layout
     // mapping, which corresponds to inserting an affine.apply.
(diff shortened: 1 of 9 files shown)


j2kun:
ok I did realize I missed a big part of the "dynamic mask generation" step. I need to add an if statement to make sure the dynamic index values are actually hit during the iteration

lib/Transforms/ConvertToCiphertextSemantics/TypeConversion.cpp
```diff
@@ -72,7 +72,7 @@ Type materializeLayout(RankedTensorType type, LayoutAttr attr,
 
 Type materializeScalarNewLayout(Type type, NewLayoutAttr attr,
                                 int ciphertextSize) {
-  return RankedTensorType::get({ciphertextSize}, type);
+  return RankedTensorType::get({1, ciphertextSize}, type);
```
asraa: ugh shoot this is fixed in https://github.com/google/heir/pull/2200 but CI was annoying. i'll bump it

lib/Utils/Layout/Codegen.cpp
```diff
@@ -381,6 +381,9 @@ FailureOr<scf::ForOp> MLIRLoopNestGenerator::generateForLoop(
   isl_ast_expr_free(expr);
   isl_ast_node_free(node);
 
+  // Restore the insertion point to after the outermost loop.
```
asraa: would setting an insertion guard at the start of the func work?

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -1231,6 +1239,353 @@ class ConvertTensorInsert
...
+  SmallVector<int64_t> staticIndices;
+  for (Value v : indices) {
+    OpFoldResult staticIndex = getAsOpFoldResult(v);
+    if (auto attr = dyn_cast<Attribute>(staticIndex)) {
+      staticIndices.push_back(cast<IntegerAttr>(attr).getInt());
+    } else {
+      return failure();
+    }
+  }
+  return staticIndices;
```
asraa: you can also use `getConstantIntValues(getAsOpFoldResult(indices)`

tests/Transforms/convert_to_ciphertext_semantics/tensor_insert.mlir
```diff
@@ -1,49 +1,97 @@
...
+// CHECK:    ^body(%[[tensor_clear:.*]]: tensor<1x16xi16>, %[[scalar_clear:.*]]: tensor<1x16xi16>):
+
+// This test differs from other tests in that the scalar mask has 1's and is then multiplied with the scalar_clear
+// CHECK:      %[[scalar_mask1:.*]] = tensor.insert %[[c1]] into %[[dense0]][%[[c0_index]], %[[c3]]]
+// CHECK:      tensor.insert %[[c1]] into %[[scalar_mask1]][%[[c0_index]], %[[c11]]]
+// CHECK:      arith.muli
+// CHECK:      arith.muli
+// CHECK:      arith.addi
+// CHECK:    return %[[res]]
+func.func @insert_secret_into_secret(%arg0: !secret.secret<tensor<8xi16>> {tensor_ext.layout = #layout}, %arg1: !secret.secret<i16> {tensor_ext.layout = #scalar_layout}) -> (!secret.secret<tensor<8xi16>> {tensor_ext.layout = #layout}) {
```
asraa, line 90: what happens when it's secret into cleartext? in the matchandrewrite code we assume that the layout would be assigned to a cleartext, so i think it would follow the same lowering as secret into secret? and then masks would cleared up on the plaintext if they can.

asraa (approved):

lib/Transforms/ConvertToCiphertextSemantics/TypeConversion.cpp
j2kun: Fixed

lib/Utils/Layout/Codegen.cpp
j2kun: Yes! I forgot about that.

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
j2kun: Great call!

tests/Transforms/convert_to_ciphertext_semantics/tensor_insert.mlir
j2kun: I agree. The cleartext tensor would be given a layout and treated as secret. I'm willing to let that one slide and deal with it when it becomes a problem if it's not working properly with this.

PR #2205, opened 2025-09-09
tensor.extract kernel with new layout

j2kun:
Rebased over #2204

TODO:

- add tests
- think more about dynamic index case


j2kun:
@asraa this is ready for review now!

lib/Utils/Layout/Utils.h
```diff
@@ -61,19 +61,31 @@ presburger::IntegerRelation expandDimensions(
...
-// Returns a new relation produced by constraining the data index dimensions of
-// the given relation to the provided values.
+presburger::IntegerRelation fixVars(const presburger::IntegerRelation& relation,
+                                    ArrayRef<int64_t> fixedValues,
+                                    presburger::VarKind varKind);
+
+// Returns a new relation produced by constraining the domain variables of the
+// given relation to the provided values.
 //
 // The fixedIndices array should have size equal to the number of domain
```
asraa:
```suggestion
// The fixedValues array should have size equal to the number of domain
```

lib/Utils/Layout/Utils.h
```diff
@@ -61,19 +61,31 @@ presburger::IntegerRelation expandDimensions(
     const presburger::IntegerRelation& relation, RankedTensorType resultType,
     ArrayRef<ReassociationIndices> reassociation);
 
-// Returns a new relation produced by constraining the data index dimensions of
-// the given relation to the provided values.
+presburger::IntegerRelation fixVars(const presburger::IntegerRelation& relation,
```
asraa, line 78:
```suggestion
// Returns a new relation produced by constraining the index dimensions of type
// varKind to the given relation to the provided values.
// The fixedValues array size should equal the number of variables of type varKind.
presburger::IntegerRelation fixVars(const presburger::IntegerRelation& relation,
```

lib/Utils/Layout/Utils.h
```diff
@@ -90,6 +102,9 @@ struct PointCollector {
 void getRangePoints(const presburger::IntegerRelation& relation,
                     PointCollector& collector);
 
+// Get any point in the range of the relation.
+std::vector<int64_t> anyRangePoint(const presburger::IntegerRelation& relation);
```
asraa:
```suggestion
// Sample a point in the range of the relation.
std::vector<int64_t> anyRangePoint(const presburger::IntegerRelation& relation);
```

really nit worthy comment change, but I was curious about how this was sampling a range point until I saw it was using an ISL sample API that doesn't really have any docs, and mostly I wanted to know if it was a random choice, or enumerating the first point, etc. And figured sample would indicate that there is not guarantee on the choice

tests/Transforms/convert_to_ciphertext_semantics/tensor_extract.mlir
```diff
@@ -1,35 +1,27 @@
...
+// CHECK: [[C1:%.+]] = arith.constant 1 : i16
+// CHECK: [[CST:%.+]] = arith.constant dense<0> : tensor<1x16xi16>
+// CHECK: [[C3:%.+]] = arith.constant 3 : index
+// CHECK: [[RESULT:%.+]] = secret.generic([[ARG0:%.+]]: !secret.secret<tensor<1x16xi16>>)
+// CHECK: [[INSERTED:%.+]] = tensor.insert [[C1]] into [[CST]]{{\[}}[[C0]], [[C3]]] : tensor<1x16xi16>
+// CHECK: [[COLLAPSED:%.+]] = tensor.collapse_shape [[INSERTED]] {{\[}}[0, 1]] : tensor<1x16xi16> into tensor<16xi16>
+// CHECK: [[COLLAPSED_0:%.+]] = tensor.collapse_shape %input0 {{\[}}[0, 1]] : tensor<1x16xi16> into tensor<16xi16>
+// CHECK: [[MUL:%.+]] = arith.muli [[COLLAPSED]], [[COLLAPSED_0]] : tensor<16xi16>
+// CHECK: [[EXPANDED:%.+]] = tensor.expand_shape [[MUL]] {{\[}}[0, 1]] output_shape [1, 16] : tensor<16xi16> into tensor<1x16xi16>
+// CHECK: [[PERMUTED:%.+]] = tensor_ext.permute [[EXPANDED]] {permutation = #new_layout1} : tensor<1x16xi16>
```
asraa:
```suggestion
// CHECK: [[PERMUTED:%.+]] = tensor_ext.permute [[EXPANDED]] {permutation = {{.*}}} : tensor<1x16xi16>
```

I assume if we change the name of the new layout attr that the alias name will also change

could you also add an assert to validate the new layout is the scalar layout?

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -1234,6 +1275,222 @@ class ConvertTensorInsert
...
+
+    RankedTensorType ciphertextSemanticType =
+        cast<RankedTensorType>(adaptor.getTensor().getType());
+    arith::ConstantOp zeroTensorOp = arith::ConstantOp::create(
+        b, ciphertextSemanticType, b.getZeroAttr(ciphertextSemanticType));
+    arith::ConstantOp oneScalarOp = arith::ConstantOp::create(
+        b, ciphertextSemanticType.getElementType(),
+        b.getOneAttr(ciphertextSemanticType.getElementType()));
+
+    std::vector<int64_t> targetSlot = anyRangePoint(relWithFixedDomain);
```
asraa, line 1392: just a comment - it would be an very minor optimization to choose a range point that gives the most minimal rotation cost. e.g. if it was possible one of the range points was in the 0th position

lib/Utils/Layout/Utils.h
j2kun: For reference, it is a depth-first search described here: https://github.com/j2kun/isl/blob/58eaa613ffcd2b11c36a773c19531f75fbcfc66c/isl_sample.c#L326-L380

It appears to be deterministic, but without a guarantee on the choice, except in that it is the first value found according to some ordering of possible points.

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
j2kun: I will leave this as a followup.

Issue #2210, opened 2025-09-10
Migrate dialect conversion passes to `allowPatternRollback = false`


j2kun:
Most of them migrate smoothly. The following fail

   - lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
   - lib/Transforms/MemrefToArith/MemrefGlobalReplace.cpp

MemrefGlobalReplace will be deleted I think, so it can stay opted out.

ArithToCGGI has tests that... don't run secret-wrap-generic at the start? (e.g., tests/Examples/tfhe_rust_hl/cpu/fully_connected.mlir starts with --mlir-to-cggi so the {secret.secret} annotation is never touched). That test crashes at materialization because it's trying to materialize an unrealized conversion cast of a memref function argument, but with allowPatternRollback=true those materializations are never needed or run in the first place. So I suspect this pass could be fixed by instead properly converting the input IR to use secret.generic, secret types, and split-generic before converting to CGGI.

IR dump after failure:

```
// -----// IR Dump After CGGIToTfheRust Failed (cggi-to-tfhe-rust) //----- //
#constant_coefficient_encoding = #lwe.constant_coefficient_encoding<scaling_factor = 16777216>
#constant_coefficient_encoding1 = #lwe.constant_coefficient_encoding<scaling_factor = 1>
#key = #lwe.key<>
#ring_i32_1 = #polynomial.ring<coefficientType = i32, polynomialModulus = <1>>
#ring_i8_1 = #polynomial.ring<coefficientType = i8, polynomialModulus = <1>>
#ciphertext_space__D742 = #lwe.ciphertext_space<ring = #ring_i32_1, encryption_type = msb, size = 742>
!ct__D742 = !lwe.lwe_ciphertext<application_data = <message_type = i8, overflow = #lwe.preserve_overflow<>>, plaintext_space = <ring = #ring_i8_1, encoding = #constant_coefficient_encoding>, ciphertext_space = #ciphertext_space__D742, key = #key>
!ct__D742_1 = !lwe.lwe_ciphertext<application_data = <message_type = i32, overflow = #lwe.preserve_overflow<>>, plaintext_space = <ring = #ring_i32_1, encoding = #constant_coefficient_encoding1>, ciphertext_space = #ciphertext_space__D742, key = #key>
module {
  func.func @fn_under_test(%arg0: memref<1x1x!tfhe_rust.eui8> {secret.secret}) -> memref<1x1x!tfhe_rust.eui32> {
[...]

PR #2211, opened 2025-09-10
diable allowPatternRollback in non-context-aware dialect conversion drivers

j2kun:
Part of #2210

Sets `ConversionConfig config;  config.allowPatternRollback = false;` for most conversion passes. Two are failing when I do this, indicating a bug in their use of the dialect conversion API.


AlexanderViand:
I'm not sure a pattern rollback technically indicates a bug in the use of the (non-oneshot) dialect conversion API, but I get what you mean.
Shouldn't we also test with `MLIR_ENABLE_EXPENSIVE_PATTERN_API_CHECKS` before making the switch?

j2kun:
What I mean is: upstream I have been reviewing the dialect conversion changes around allowPatternRollback=false. When set, it materializes changes to the IR immediately rather than defer them to the end of the pass. Uses of the dialect conversion API may depend on previously-erased IR constructs still being visible in the IR, and these would cause hard errors when setting this flag. 
eebf4b1 gives an example where the pattern for TrivialEncryptOp was querying the (deleted) defining EncodeOp of its operand.

Not sure how to enable MLIR_ENABLE_EXPENSIVE_PATTERN_API_CHECKS, but I can try that. I don't think that is totally relevant here, since these are only "invalid" with backward-incompatible changes from allowPatternRollback=false.

j2kun:
Another illegal example (which would not be caught here) is how ElementwiseToAffine uses the DialectConversion driver with a RewritePattern instead of a ConversionPattern

j2kun:
The remaining opted-out conversion passes are

- lib/Dialect/Arith/Conversions/ArithToCGGI/ArithToCGGI.cpp
- lib/Transforms/MemrefToArith/MemrefGlobalReplace.cpp

MemrefGlobalReplace will be deleted I think, so it can stay opted out.

ArithToCGGI has tests that... don't run secret-wrap-generic at the start? (e.g., tests/Examples/tfhe_rust_hl/cpu/fully_connected.mlir starts with `--mlir-to-cggi` so the `{secret.secret}` annotation is never touched). That test crashes at materialization because it's trying to materialize an unrealized conversion cast of a memref function argument, but with allowPatternRollback=true those materializations are never needed or run in the first place. So I suspect this pass could be fixed by instead properly converting the input IR to use secret.generic, secret types, and split-generic before converting to CGGI.

lib/Dialect/CGGI/Conversions/CGGIToTfheRustBool/CGGIToTfheRustBool.cpp
```diff
@@ -260,25 +255,56 @@ class CGGIToTfheRustBool
...
     }
+
+    // Post-conversion cleanup: remove any remaining EncodeOps
+    // These would be EncodeOps that are not used by any TrivialEncrypt
+    // operations
+    SmallVector<lwe::EncodeOp> encodeOpsToErase;
+    op->walk([&](lwe::EncodeOp encodeOp) {
+      // Check if this EncodeOp has any TrivialEncrypt users
+      bool hasTrivalEncryptUser = false;
+      for (auto user : encodeOp.getResult().getUsers()) {
```
asraa: shouldn't it just be checking that there aren't any more users of the encode op (since the trivial encrypts would be gone) and erase if there aren't?

j2kun: Yes you're right. This can be simplified quite a bit.

PR #2219, opened 2025-09-12
add insert to new-layout-propagation

j2kun:
Also added an interface to determine which operands need layouts

### DIFF
--- a/lib/Transforms/LayoutPropagation/InterfaceImpl.cpp
+++ b/lib/Transforms/LayoutPropagation/InterfaceImpl.cpp
@@ -1,0 +1,60 @@
+#include "lib/Transforms/LayoutPropagation/InterfaceImpl.h"
+
+#include "lib/Dialect/HEIRInterfaces.h"
+#include "lib/Dialect/TensorExt/IR/TensorExtOps.h"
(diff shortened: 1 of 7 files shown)


lib/Transforms/LayoutPropagation/InterfaceImpl.cpp
```diff
@@ -0,0 +1,60 @@
...
+    return operandIndex == tensor::ExtractOp::odsIndex_tensor;
+  }
+};
+
+}  // namespace
+
+void registerOperandLayoutRequirementOpInterface(DialectRegistry& registry) {
+  registry.addExtension(
+      +[](MLIRContext* ctx, tensor_ext::TensorExtDialect* dialect) {
+        tensor_ext::RotateOp::attachInterface<
```
asraa, line 50: generally we're assuming here that the indices is plaintext? maybe this is an optimistic requirement? i think capturing the secretness state would be a bit weight. i like the approach of having an interface for this.
(supporting secret rotations is also totally unknown to me anyway)

j2kun: Yeah, IMO an insert/extract with secret indices is out of scope (I suspect such a situation would have first gone through `convert-secret-for-to-static-for` and the indices will become dynamic but cleartext).

PR #2225, opened 2025-09-16
try experimenting with macos ci

j2kun:


j2kun:
This seems to get the MacOS CI in line what happens on my local macbook (they both have the same failure). As such, it's an improvement over what was there before and I think we should merge it.

PR #2012, opened 2025-07-15
[Draft]: Initial Sbox

WoutLegiest:
CGGI Implementation based on 
[https://core.ac.uk/download/pdf/36694529.pdf](https://core.ac.uk/download/pdf/36694529.pdf)

Issue #1721, opened 2025-04-15
Yosys optimizer clean-ups and refactoring


asraa:
From office hours we discussed a few improvements to the yosys optimizer because it's very difficult to get working, also see #1712

- [ ] Avoid memref semantics and converge both pipelines on tensor semantics
- [ ] Allow for partial loop unrolling and add tests that work for that
- [ ] Cleanup the whole extract-body into functions situation

PR #2237, opened 2025-09-19
[Draft] Continue: cggi: remove memrefs from cggi pipeline #1910

WoutLegiest:
Continuation of [PR #1910](https://github.com/google/heir/pull/1910). Different PR, since original is submitted as internal CL. 
There is also a rebase of main and sbox (PR #2012). So #2012 should first be merged, before review. 

All of the test does not successes, unless some of the e2e tfhe-rs-rust test that are set to manual. Therefore we would need to write the tfhe-rs emitter a bit and would propose to do that in a later PR. Will add this to the issues list.


WoutLegiest:
@asraa 
Looks like the merge and squad is complete

j2kun:
I did a brief review of this PR, but not in super fine detail. My only concern is that if it breaks the e2e rust pipeline, whether there will be people who try to use it in between now and when the emitter is fixed for tensor.

@asraa I understand you worked on this as well, do you think it needs a more detailed review from me? We could patch it internally and fix the build and get it in.

WoutLegiest:
Fair point, I'm already working on the emitter updates. Can be found at: #2278

asraa:
Gotcha! Gimme a sec to finish the review. There's only one or two places that need a more detailed review and I'll call them out if I need it

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -553,6 +558,12 @@ void YosysOptimizer::runOnOperation() {
         ctx, std::vector<std::string>{"memref.store", "affine.store"});
   }
 
+  // Extracting tensor.inserts into their own generics avoids returning an
+  // entire tensor from Yosys.
```
asraa, line 562: were there any insert/extract_slices? since we use tensor to linalg affine loops i think it's just a scalar insert in the loops that appear from the high level ops

tests/Examples/tfhe_rust/BUILD
```diff
@@ -60,7 +60,10 @@ tfhe_rs_end_to_end_test(
         "--use-levels",
     ],
     mlir_src = "fully_connected.mlir",
-    tags = ["yosys"],
+    tags = [
+        "manual",
```
asraa, line 64: could you please add a TODO with a link to maybe https://github.com/google/heir/issues/1721 to make sure we add back the ones that we can tests e2e?

tests/Examples/tfhe_rust_hl/cpu/BUILD
```diff
@@ -43,5 +45,6 @@ tfhe_rs_end_to_end_test(
     ],
     heir_translate_flags = ["--emit-tfhe-rust-hl"],
     mlir_src = "hello_world_clean_xsmall.mlir",
+    tags = ["manual"],
```
asraa, line 48: same here to eventually remove manual https://github.com/google/heir/issues/1721

lib/Pipelines/PipelineRegistration.cpp
```diff
@@ -34,6 +38,15 @@ using mlir::func::FuncOp;
 
 namespace mlir::heir {
 
+void tosaToLinalg(OpPassManager& manager) {
```
asraa, line 41: i *think* we removed all the TOSA in e2e tests and replaced them w/ linalg - which ones broke that you needed code? I can fix in a follow-up

lib/Dialect/Secret/Conversions/SecretToCGGI/SecretToCGGI.cpp
```diff
@@ -673,95 +578,183 @@ struct ConvertSecretConcealOp
...
+              lwe::EncodeOp::create(b, b.getLoc(), ptTy, element,
+                                    b.getIndexAttr(plaintextBits),
+                                    overflowAttr),
+              b.getIndexAttr(ciphertextBits));
+
+          // b.create<lwe::TrivialEncryptOp>(
+          //      ctTy,
+          //      b.create<lwe::EncodeOp>(ptTy, element, ctTy.getEncoding()),
+          //      lwe::LWEParamsAttr())
+          //     .getResult();
```
asraa:
```suggestion
```

asraa (approved):
Yeah, the other guard we could do here to avoid e2e test failures in the meantime is to just disable mlir-to-cggi in the window between the fix PRs to just avoid the availability of the pipeline in this "tainted" region. But I'm OK with either

lib/Pipelines/PipelineRegistration.cpp
WoutLegiest: Normally, all the TOSA tests are converted to linalg and are now fixed. Also with the extension of PR #2278

tests/Examples/tfhe_rust/BUILD
WoutLegiest: Removed back in #2278

tests/Examples/tfhe_rust_hl/cpu/BUILD
WoutLegiest: Removed back in #2278

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
WoutLegiest: Mhh, don't know tbh. 
I also don't recognize this code, so think it was something from your original implementation

j2kun (approved):

j2kun:
@WoutLegiest can you remove or rebase the lockfile changes? It's blocking copybara

WoutLegiest:
> @WoutLegiest can you remove or rebase the lockfile changes? It's blocking copybara

Done!

WoutLegiest:
@j2kun @asraa looks like the CI builds fail. Cannot find a way to reproduce the failing, what to do?

asraa:
> @j2kun @asraa looks like the CI builds fail. Cannot find a way to reproduce the failing, what to do?

Looks like there are a couple of tensor operations that weren't handled in the verilog emitter (collapse and expand shape) and corner cases (0-D tensor), and then there's also another thing where sometimes arith ops are now operating on tensors instead of being load/stored from memrefs and so arith.extsi might operate on tensor values. 

I think that last issue is coming from a linalg canonicalization, which I might just disable for the tfhe-rs pipelines and/or try to get tensor-to-scalars to function properly

WoutLegiest:
Right, I changed the test to manual, because I indeed fix them in the next PR

asraa:
OH! Gotcha. They somehow were still running internally on presubmit.

j2kun:
Looks like this was merged as ff31b78d5a8afe3ecd503c4c8af690c5dc9fde1a

PR #2240, opened 2025-09-21
Generalize Vos-Vos-Erkin algorithm to support arbitrary mappings

j2kun:
Improves ImplementShiftNetworks to handle arbitrary mappings

- Uses ArithmetcDag for the implementation of a RotationGroup
- Extends ArithmeticDag slightly to account for simple tensor constants
- Improves the visitors to support multi-root dags

This is a port of https://github.com/j2kun/shift-networks/pull/1
which has a python implementation and a suite of randomized tests.


j2kun:
Still desired before it's totally ready:

- Port a handful of randomized tests from the python impl (short of https://github.com/google/heir/pull/2196)

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
```diff
@@ -14,51 +19,146 @@ namespace tensor_ext {
...
+inline ::llvm::hash_code hash_value(const CtSlot& obj) {
+  return llvm::hash_combine(obj.ct, obj.slot);
+}
+
+struct MappingEntry {
+  CtSlot source;
+  CtSlot target;
+};
+
+// An arbitrary mapping on the slots of a set of ciphertexts.
```
asraa, line 56: arbitrary meaning a source can go to many targets - but it should be that a target only has one source, right?

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
```diff
@@ -14,51 +20,145 @@ namespace tensor_ext {
...
+
+// The ShiftStrategy class applies power-of-two shifts to each set bit in some
+// order (default is LSB-to-MSB order, 1, 2, 4, 8, ...). Each shift amount is
+// considered a "round" in which a group of indices are shifted together. This
+// can be used both to identify conflicts for the graph coloring technique of
 // Vos-Vos-Erkin, and also to construct the concrete shift network after a
 // partition has been decided by Vos-Vos-Erkin.
 struct ShiftRound {
-  // Maps the index of the original input to its current position in the
+  // Maps the the original mapping source to its current position in the
```
asraa:
```suggestion
  // Maps the index of the original mapping source to its current position in the
```

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -47,279 +57,239 @@ inline int64_t normalizeShift(int64_t input, int64_t output,
...
-    }
-    llvm::dbgs() << "\n";
-  });
+void ShiftStrategy::evaluate(const Mapping& mapping) {
+  // First compute the virtual shifts needed for each source slot
+  SmallVector<SourceShift> sourceShifts;
+  sourceShifts.reserve(mapping.size());
+  for (const MappingEntry& entry : mapping) {
+    int64_t shift = getVirtualShift(entry.source, entry.target);
+    SmallVector<int64_t> neededShifts;
```
asraa:
```suggestion
```

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
```diff
@@ -68,17 +168,101 @@ class VosVosErkinShiftNetworks {
   // The returned ArrayRef is owned by this VosVosErkinShiftNetworks instance.
   // The resulting set of rotation groups are is cached, and the cache is used
   // on further calls to avoid recomputing the shift network.
-  ArrayRef<RotationGroup> computeShiftNetwork(const Permutation& permutation);
-
-  int64_t getCiphertextSize() const;
+  ShiftScheme findShiftScheme(const Mapping& mapping,
+                              ArrayRef<int64_t> shiftOrder = {});
```
asraa, line 174: the default is actually just a LSB to MSB power of two shift ordering? Probably worth a note

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -47,279 +57,239 @@ inline int64_t normalizeShift(int64_t input, int64_t output,
...
+    return schemeCache[cacheKey];
   }
 
-  ShiftStrategy strategy;
-  RotationGroup allIndices;
-  for (int64_t i = 0; i < ciphertextSize; i++) {
-    allIndices.insert(i);
-  }
-  strategy.evaluate(permutation, allIndices);
+  // FIXME: try many shift orders and pick the best
```
asraa: i see - the shift strategy can be selected to minimize the number of conflicts from what i understand. even power of two rotations are a choice in strategy right? (like you could use base 3 or something), but then you'd probably have more conflicts if there's less spots. but for the sake here you mean mostly shuffling the power of two rotations?

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -47,279 +57,239 @@ inline int64_t normalizeShift(int64_t input, int64_t output,
...
-    Value perGroupResult =
-        rotateGroup(op.getInput(), group, ciphertextSize, permKey, rewriter);
-    if (result.has_value())
-      result =
-          arith::AddIOp::create(rewriter, op.getLoc(), *result, perGroupResult);
-    else
-      result = perGroupResult;
+  b.setInsertionPointAfter(op);
+
+  // Could add a special case here if the numCipheretexts == 1, using
```
asraa:
```suggestion
  // Could add a special case here if the numCiphertexts == 1, using
```

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
```diff
@@ -47,279 +57,239 @@ inline int64_t normalizeShift(int64_t input, int64_t output,
...
+  // collapse_shape and expand_shape instead of insert/extract slice. No sure
+  // if that would be more efficient in the end...
```
asraa, line 275: one thing to note (and it's fine) is that convert-to-ciphertext-semantics drops unit dims around rotation ops / adds / muls but not permute ops. so after this runs i think you might need to add those patterns here too. and then honestly canonicalizations might take care of the dropped unit dims extract slice w/ collapse shape maybe?

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
asraa: ah nvm

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.h
j2kun: Done

lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
j2kun: Yes, I think we could experiment with non-power-of-two rotations, but the standard choice seems to be (e.g., it is this way in OpenFHE and Lattigo) to generate all power-of-two rotation keys.

But in general if you pick a different ordering of the power-of-two shifts, you can get very different results (one test case demonstrates this).

Will change the FIXME to a TODO

j2kun:
Rebased, I'm just going to add a few more tests before it's ready to go

j2kun:
OK So I fixed the bug in 924f462, and further, in a local branch I added 77 randomized tests output by the python reference code, all of which pass, but which produces some 12k lines of CPP source so I removed most of the tests once the bug was fixed and left it in https://github.com/j2kun/heir/pull/15/commits/8a65f40a311fe81b8367cc134939692ba34da1e3

We can probably reinstate some randomized tests once we add a fuzz testing framework. In the meantime, this PR is ready to go!

tests/Dialect/TensorExt/Transforms/implement_shift_network.mlir
```diff
@@ -22,168 +23,42 @@ func.func @test_no_conflicts(%0: tensor<64xi32>) -> tensor<64xi32> {
...
+// CHECK: %[[EMPTY:.*]] = tensor.empty() : tensor<1x64xi32>
+// CHECK: %[[INSERT:.*]] = tensor.insert_slice %[[SLICE]] into %[[EMPTY]][0, 0] [1, 64] [1, 1] : tensor<64xi32> into tensor<1x64xi32>
+// CHECK: return %[[INSERT]] : tensor<1x64xi32>
+#map3 = #tensor_ext.new_layout<"{ [ct1, slot1] -> [ct2, slot2] : ct1 = 0 and ct2 = 0 and slot1 = slot2 and slot1 >= 0 and 63 >= slot1 and slot2 >= 0 and 63 >= slot2 }">
+func.func @identity(%0: tensor<1x64xi32>) -> tensor<1x64xi32> {
+  %1 = tensor_ext.permute %0 {permutation = #map3} : tensor<1x64xi32>
+  return %1 : tensor<1x64xi32>
 }
+
+// FIXME: add some multi-ciphertext tests for IR materialization
```
asraa: maybe just add a small 2 ciphertext example? (but i see them in the randomized unit tests, so i'm confident but just a sanity check that the indices are materialized properly)

lib/Transforms/LayoutOptimization/LayoutConversionCost.cpp
```diff
@@ -3,100 +3,109 @@
...
+using kernel::SubtractNode;
+using kernel::SymbolicValue;
+using presburger::IntegerRelation;
+using tensor_ext::CtSlot;
+using tensor_ext::Mapping;
+using tensor_ext::NewLayoutAttr;
+using tensor_ext::ShiftScheme;
+
+// A visitor that counts the number of rotations in an ArithmeticDag.
+class RotationCountVisitor : public CachingVisitor<SymbolicValue, int64_t> {
```
asraa, line 34: this makes me SO happy.

lib/Dialect/TensorExt/Transforms/RotationGroupKernel.h
```diff
@@ -0,0 +1,403 @@
...
+                     int64_t ciphertextSize) {
+  using NodeTy = ArithmeticDagNode<T>;
+  using ValueTy = std::shared_ptr<NodeTy>;
+  int64_t numCiphertexts = input.size();
+
+  // We need to identify the (possibly two) target ciphertexts for each input
+  // ciphertext that was rotated.
+  //
+  // If there is only one target---i.e., if the rotation was exactly the power
+  // of two matching the ciphertext size---we can update the target with the
```
asraa:
```suggestion
  // of two matching a multiple of the ciphertext size---we can update the target with the
```

lib/Dialect/TensorExt/Transforms/RotationGroupKernel.h
```diff
@@ -0,0 +1,403 @@
...
+    std::vector<double> mask2(ciphertextSize, 0);
+    for (int64_t i = 0; i < ciphertextSize; i++) {
+      if (i < boundarySlot) {
+        mask1[i] = mask[i];
+      } else {
+        mask2[i] = mask[i];
+      }
+    }
+
+    // std::cout << "  mask1: ";
```
asraa, line 156: remove? code makes pretty good sense!

j2kun: I want to leave in some logging, but there is trouble when you do LLVM_DEBUG in a header that is included in other headers (the DEBUG_TYPE `#define` can be overridden), and I don't know how to enable LLVM_DEBUG from a unit test... Thoughts here?

tests/Dialect/TensorExt/Transforms/implement_shift_network.mlir
j2kun: Done (added two)

PR #2245, opened 2025-09-22
Call _mlir_ciface functions in plaintext tests

Time0o:
These prevent ABI issues, especially on macOS, see issue #2081.


tests/Examples/plaintext/add/add_test.c
j2kun, line 1: The only change I might suggest is to use the templated `StridedMemrefType` here instead of re-defining the struct type at each usage site.

https://github.com/google/heir/blob/ecaf7a3a307328f69019a4a0945d1ae9c4f5e73f/tests/llvm_runner/memref_types.h#L10

But I can always clean that up later. LGTM!

j2kun: The CI build errors suggest you may also need to wrap these declarations in `extern C { ... }`

```
undefined reference: _mlir_c_roberts_cross__encrypt__arg0
```

Cf. https://github.com/google/heir/blob/ecaf7a3a307328f69019a4a0945d1ae9c4f5e73f/tests/Dialect/ModArith/Conversions/mod_arith_to_arith/runner/lower_mac_test.cc#L8 for an example

j2kun:
Ah, sorry, you may also need to rebase over https://github.com/google/heir/pull/2250 since I broke the build at head :(

tests/Examples/plaintext/add/add_test.c
timo-lyceum: > The only change I might suggest is to use the templated `StridedMemrefType` here instead of re-defining the struct type at each usage site.
> 
> https://github.com/google/heir/blob/ecaf7a3a307328f69019a4a0945d1ae9c4f5e73f/tests/llvm_runner/memref_types.h#L10
> 
> But I can always clean that up later. LGTM!

That is a fair point, I kind of presumed these have to be C files. But I suppose not so I now went ahead and ported everything to Google test while I was at it to match the rest of the tests. Please let me know if I went too far here, especially with the added `StridedMemRefType` constructors.

timo-lyceum: > The CI build errors suggest you may also need to wrap these declarations in `extern C { ... }`
> 
> ```
> undefined reference: _mlir_c_roberts_cross__encrypt__arg0
> ```
> 
> Cf.
> 
> https://github.com/google/heir/blob/ecaf7a3a307328f69019a4a0945d1ae9c4f5e73f/tests/Dialect/ModArith/Conversions/mod_arith_to_arith/runner/lower_mac_test.cc#L8
> 
> for an example

This one is weird I think I still haven't fixed it. Sorry, it took me a while to get back to this, I'll most likely figure it out tomorrow.

Time0o: Argh, sorry forgot to switch accounts.

tests/llvm_runner/memref_types.h
```diff
@@ -1,18 +1,38 @@
...
   T* data;
   int64_t offset;
   int64_t sizes[N];
   int64_t strides[N];
+
+  StridedMemRefType() = default;
+
+  StridedMemRefType(T* basePtr, T* data, int64_t offset, int64_t size,
+                    int64_t stride)
+    requires(N == 1)
```
j2kun, line 23: I'm not sure this requirement should be here... what is the problem with allowing a multidimensional memref? I think in general we need to do multidimensional llvm runner tests.

tests/Examples/plaintext/add/add_test.cpp
j2kun, line 1: Considering this test does `tensor.extract` instead of adding, would you mind renaming the test files and functions?

j2kun (approved):

tests/llvm_runner/memref_types.h
Time0o: @j2kun This is just for convenience so that you can (but don't have to) write:

```
StridedMemRef foo = {base, data, 0, 10, 0};
```

Instead of:

```
StridedMemRef foo = {base, data, 0, {10}, {0}};
```

In the one-dimensional case. Benefit maybe debatable.

tests/Examples/plaintext/add/add_test.cpp
Time0o: @j2kun Done

tests/llvm_runner/memref_types.h
j2kun: Ah, yes, as an extra constructor for the special case, it's fine by me.

j2kun:
LGTM! I'll try to merge it in tomorrow.

j2kun:
Looks like asan/msan is reporting some memory leaks here, I am investigating this to see if they are legit.

j2kun:
Found the issues:

- Any memrefs created by LLVM-generated functions are heap allocated, so we need to free them before the test ends (the `.basePtr` member).
- The `debug` test can't use the gunit main because it needs to serve two purposes: running the actual test and creating a log file on disk. The custom main allows it to do both, and so I kept the StridedMemRef changes to that test and reverted the use of gunit and gunit_main

PR #2246, opened 2025-09-22
frontend: Emit ordered comparison predicates for floats

Time0o:
Floating point comparison predicates in the arith dialect differ from integer ones. This inserts the ordered variants fixing erroneous outputs such as `arith.cmpf slt` -> `arith.cmpf olt`.

This is another issue I had while experimenting with the Python frontend and here I'm a bit more confident that this is indeed a bug.


AlexanderViand (approved):
Great catch! If you ran into this "in the wild", you probably have examples that triggered this? If yes, it'd be great if you could add one as a unit test :)

PR #2248, opened 2025-09-22
fix lld/libomp misconfig in cibuildwheel

j2kun:
The forcing of the `lld` linker is breaking the cibuildwheel, and once I fixed that I found it was broken by a lack of openmp. I found a way to allow `libomp` and `libgomp` by using just the `-fopenmp` option, which allows the compiler to pick the exact linker flag based on what's available on the system. Also needed to install `libgomp` in the cibuildwheel container.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -23,9 +23,6 @@ build --action_env=CXX=clang++
 build --host_action_env=CC=clang
 build --host_action_env=CXX=clang++
 
-# linux-specific options
-build:linux --linkopt="-fuse-ld=lld"
-
 # macos specific options
 # openmp is not supported in apple clang
 build:macos --//:enable_openmp=0
--- a/.github/install_cibuildwheel_deps.sh
+++ b/.github/install_cibuildwheel_deps.sh
@@ -7,7 +7,7 @@ if ! bazel version; then
     arch="arm64"
   fi
   echo "Downloading $arch Bazel binary from GitHub releases."
-  curl -L -o $HOME/bin/bazel --create-dirs "https://github.com/bazelbuild/bazel/releases/download/8.1.0/bazel-8.1.0-linux-$arch"
+  curl -L -o $HOME/bin/bazel --create-dirs "https://github.com/bazelbuild/bazel/releases/download/8.4.0/bazel-8.4.0-linux-$arch"
   chmod +x $HOME/bin/bazel
 else
   # Bazel is installed for the correct architecture
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -34,6 +34,7 @@ bazel_dep(name = "googletest", version = "1.16.0")
 bazel_dep(name = "google_benchmark", version = "1.9.1")
 bazel_dep(name = "abseil-cpp", version = "20250127.0", repo_name = "com_google_absl")
 bazel_dep(name = "pybind11_bazel", version = "2.13.6")
+bazel_dep(name = "rules_shell", version = "0.6.1")
 
 # Yosys dependencies that are already on the bazel central registry
 # abc can't be used in place because it changes the names of the targets from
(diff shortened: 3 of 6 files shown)


asraa (approved):

j2kun:
Going to run cibuildwheel locally to test this before merging it, on top of Alex's changes to the python deps

j2kun:
ok so we actually have a slight problem here in that the manylinux specification forbids linking against openmp (linking against any shared lib not in this list implies noncompliance: https://peps.python.org/pep-0599/#the-manylinux2014-policy)

I think to get this to work now I have to bundle `libgomp`/`libomp` with the wheel, and then point to it in during the clang compile step. Or just ship `heir-py` with openmp disabled.

j2kun:
Specifically `cibuildwheel` builds and then fails at

```bash
Repairing wheel...

    + sh -c 'auditwheel repair -w /tmp/cibuildwheel/repaired_wheel /tmp/cibuildwheel/built_wheel/heir_py-0.0.2-cp311-abi3-linux_x86_64.whl'
INFO:auditwheel.main_repair:Repairing heir_py-0.0.2-cp311-abi3-linux_x86_64.whl
auditwheel: error: cannot repair "/tmp/cibuildwheel/built_wheel/heir_py-0.0.2-cp311-abi3-linux_x86_64.whl" to "manylinux_2_28_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.
```

Repro (from a venv with `cibuildwheel` installed):

```bash
CIBW_DEBUG_KEEP_CONTAINER=1 cibuildwheel --platform linux --archs x86_64
```

Then after it fails, `docker exec -it <container_id> /bin/bash`, and in the container:

```bash
$ /opt/python/cp311-cp311/bin/python -m pip install auditwheel-symbols
$ /opt/python/cp311-cp311/bin/auditwheel-symbols --manylinux 2_28  /tmp/cibuildwheel/built_wheel/heir_py-0.0.2-cp311-abi3-linux_x86_64.whl

heir/abc is manylinux_2_28 compliant.
heir/heir-opt is not manylinux_2_28 compliant because it links the following forbidden libraries:
libgomp.so.1
heir/heir-translate is manylinux_2_28 compliant.
heir/libOPENFHEbinfhe.so is not manylinux_2_28 compliant because it links the following forbidden libraries:
libc.so.6       offending versions: GLIBC_2.32
libstdc++.so.6  offending versions: GLIBCXX_3.4.29
libm.so.6       offending versions: GLIBC_2.29
heir/libOPENFHEcore.so is not manylinux_2_28 compliant because it links the following forbidden libraries:
libc.so.6       offending versions: GLIBC_2.32
libm.so.6       offending versions: GLIBC_2.29
libstdc++.so.6  offending versions: GLIBCXX_3.4.29
heir/libOPENFHEpke.so is not manylinux_2_28 compliant because it links the following forbidden libraries:
libm.so.6       offending versions: GLIBC_2.29
libstdc++.so.6  offending versions: GLIBCXX_3.4.29, GLIBCXX_3.4.26
libc.so.6       offending versions: GLIBC_2.32
heir/libopenfhe.so is not manylinux_2_28 compliant because it links the following forbidden libraries:
libgomp.so.1
heir_py-0.0.2.data/data/bin/heir-opt is not manylinux_2_28 compliant because it links the following forbidden libraries:
libgomp.so.1
heir_py-0.0.2.data/data/bin/heir-translate is manylinux_2_28 compliant.
```

j2kun:
The errors like `libc.so.6       offending versions: GLIBC_2.32` I can fix by bumping to a newer `manylinux` container version, but the libgomp error would not be fixed by this.

j2kun:
Just to unblock releases at this point, I'm going to disable openmp in the heir-py build and file an issue.

j2kun:
Another option: https://github.com/cerisier/toolchains_llvm_bootstrapped

Issue #2263, opened 2025-09-24
ImplementShiftNetworks: support rightward shifts as well


j2kun:
A layout change that reduces to a single rightward shift by 1 would require using all power-of-two rotation keys for a leftward shift, since it's evaluating -1 = 0b11111...1. Supporting rightward shifts as well as leftward shifts (perhaps in tandem with #2256) would fix this.

eymay:
NAF based representation could be useful here. Intuitively it can help to reduce the collisions as well since the digit set extends from `{0,1}` to `{-1,0,1}` and less digits are used for the same representation.

j2kun:
While having NAF support is not a bad idea (citation: https://eprint.iacr.org/2021/1161), I think the simplest path forward is to give negative powers of two as `shiftOrder` input to VosVosErkinShiftNetworks. The randomized approach of #2256 could do this, but it may also be useful to guarantee a standard set of negative shifts is deterministically attempted.

(cf. https://github.com/google/heir/pull/2240 which hasn't yet been merged)

j2kun:
Related: https://github.com/google/heir/issues/744

Time0o:
I tried to look into this after implementing random network construction but it seems to be a lot more complicated than it sounds.

It's easy enough to either all-left or all-right shifts, represent the latter by negative numbers and then adapt `ShiftStrategy::evaluate` such that it works with right shifts as well. However, the conflict graph that results is then not isomorphic to the one you'd get via left-shifting and I couldn't figure out how to adapt to that.

I'm pretty sure I read at least parts of the relevant paper somewhere but I can't locate an open access version anymore so I'm not if it talks about this detail.

j2kun:
[...]

Issue #2256, opened 2025-09-24
ImplementShiftNetworks: run Vos-Vos-Erkin method multiple times with randomly chosen orderings of power-of-two shifts


j2kun:
The only remaining feature of the Vos-Vos-Erkin paper that we haven't ported is the ability to randomly pick an order of power-of-two shifts and select the one that produces the smallest shift network. By "randomly," of course, I mean seeding the RNG in some deterministic way based on the input IR (can we "hash" an entire IR?).

Time0o:
@j2kun This sounded like something I could do so I gave it a shot here: https://github.com/Time0o/heir/commit/b05dab2af7f7e30934591c707aa451e75fbae101. It seems to work well but I have not settled on how to do seeding. MLIR has this https://mlir.llvm.org/doxygen/structmlir_1_1OperationEquivalence.html which could presumably be used to hash input IR (via an analysis pass?). But what would need to be hashed, The operation that the `LayoutOptimization` pass runs on, right?

PR #2271, opened 2025-09-28
Time0o vos vos erkin random shift orderings

Time0o:
This is my attempt at solving https://github.com/google/heir/issues/2256 as a step towards https://github.com/google/heir/issues/2263. While this does not break any tests it's a bit hard to tell whether the seeding mechanism I've implemented here is actually sensible.


lib/Dialect/TensorExt/Transforms/ImplementShiftNetworkTest.cpp
```diff
@@ -106,10 +148,8 @@ TEST(ImplementShiftNetworkTest, TestFig3) {
   mapping.add(CtSlot(0, 13), CtSlot(0, 2));
   mapping.add(CtSlot(0, 14), CtSlot(0, 9));
   mapping.add(CtSlot(0, 15), CtSlot(0, 1));
-  VosVosErkinShiftNetworks shiftNetworks;
-  auto scheme = shiftNetworks.findShiftScheme(mapping);
-  EXPECT_EQ(scheme.rotationGroups.size(), 3);
-  simulateShiftNetwork(mapping, scheme, numCts, ctSize);
+  checkMapping(mapping, numCts, ctSize);
```
Time0o: Remove this

lib/Transforms/LayoutOptimization/LayoutOptimization.td
```diff
@@ -42,6 +42,14 @@ def LayoutOptimization : Pass<"layout-optimization"> {
       "int",
       /*default=*/"1024",
       "Power of two length of the ciphertexts the data is packed in."
+    >,
+    Option<
+      "vveRandomTries",
+      "vve-random-tries",
+      "unsigned",
+      /*default=*/"100",
```
Time0o, line 59: I have no idea what a sensible value here is.

lib/Dialect/TensorExt/Transforms/ImplementShiftNetworkTest.cpp
```diff
@@ -74,16 +77,55 @@ void simulateShiftNetwork(const Mapping& mapping, const ShiftScheme& scheme,
...
+
+  // We try a large number of shift orders here such that we can be effectively
+  // certain that we will find a network that is at least as good as the "naive"
+  // one.
+  auto bestScheme = shiftNetworks.findBestShiftScheme(mapping, 1000);
+  unsigned bestNumRG = bestScheme.rotationGroups.size();
+  if (bestNumRG > naiveNumRG)
+    return ::testing::AssertionFailure()
+           << "Expected best found network with " << bestNumRG
+           << " rotation groups to not be worsei then naive network which has "
```
j2kun:
```suggestion
           << " rotation groups to not be worse then naive network which has "
```

lib/Transforms/LayoutOptimization/LayoutOptimization.td
j2kun: What is the rough time it takes to run a unit test with this value set to 100?

lib/Transforms/LayoutOptimization/LayoutOptimization.cpp
```diff
@@ -332,6 +305,24 @@ LayoutOptimization::OpHoistResult LayoutOptimization::hoistOp(
...
+  NewLayoutAttr fromLayoutAttr = dyn_cast<NewLayoutAttr>(fromLayout);
+  NewLayoutAttr toLayoutAttr = dyn_cast<NewLayoutAttr>(toLayout);
+
+  if (!fromLayoutAttr || !toLayoutAttr) {
+    return fromLayout == toLayout ? 0 : 1;
+  }
+
+  // We seed Vos-Vos-Erkin shift order randomization with a hash over the
+  // currently processed operation.
+  std::size_t vveRandomSeed = getAnalysis<OperationHashAnalysis>().hash;
```
j2kun: This is hashing the Value operand/results, and Values are essentially pointers. Wouldn't this vary from run to run (even on the same IR?).

Ideally we can have the seed be deterministic in relation to the input IR in some abstract sense. The options that occur to me:

- Hash pointers (potentially varies run to run or as upstream details of MLIR's implementation vary)
- Hash the textual representation of the IR (less brittle but still may vary as upstream details change), some [prior art here](https://github.com/openxla/xla/blob/f0c875ddaffd657489ecaef99cae6ff458a1e719/xla/hlo/ir/hlo_module.cc#L505-L542)
- Hash the string representation of the two layouts (say, their ISL strings).

The last option seems the simplest and probably least likely to change over time as ISL is quite mature and not actively developed. You can do this by just hashing `fromLayout.getLayout()` (a string) and same for `toLayout`, and `hash_combine`'ing them. I suppose it has the downside of, if the RNG doesn't hit a good shift order for one layout conversion op with a from/to layout, then it would fail to find a good shift order for all conversion ops with that same from/to layout. But at that point, how much hope can you put into randomness in a compiler pass...

j2kun (changes requested):
Generally looks good! Some minor comments

lib/Transforms/LayoutOptimization/LayoutOptimization.cpp
Time0o: @j2kun Yeah I think I really did not think this one through enough. My initial thought was also to hash the textual IR but then I skimped out and went with `computeHash`. I will try with `getLayout()`. Maybe it would make sense to allow passing a `vve-random-seed` parameter to the pass to be combined with this as well so that it is at least in theory possible to work around "bad RNG" in the case you describe?

j2kun: Passing in an RNG as a pass flag sounds good to me as well.

j2kun:
FYI you will need to rebase over main (e42ca74ebdc974ced23f80c6a0dd80573fbbce97) to fix some build failures.

lib/Transforms/LayoutOptimization/LayoutOptimization.td
Time0o: `bazel test //lib/Dialect/TensorExt/Transforms:ImplementShiftNetworkTest` still finishes in maybe a second or so even with this change.

lib/Dialect/TensorExt/Transforms/ImplementShiftNetworkTest.cpp
Time0o: Fixed

lib/Transforms/LayoutOptimization/LayoutOptimization.cpp
Time0o: Done

Time0o:
> FYI you will need to rebase over main ([e42ca74](https://github.com/google/heir/commit/e42ca74ebdc974ced23f80c6a0dd80573fbbce97)) to fix some build failures.

That actually seems to introduce a build failure for me, see: https://github.com/google/heir/issues/2282 (actually trying to compile again from scratch I get different errors but all in readline so same idea).

j2kun (approved):

j2kun:
Please run `pre-commit run --all-files` for the small change requested by CI, and then squash your commits into one and we'll merge it. I think we should be able to merge it despite https://github.com/google/heir/issues/2282 and we can investigate that separately.

Thanks for the contribution!

j2kun:
I can actually try bringing it into google and patching the rest.

Time0o:
@j2kun Thank you, I totally overlooked the pre-commit hooks

PR #2273, opened 2025-09-29
Add insert/extract slice to noise analysis

j2kun:
This is in support of the new layout system, which will use insert_slice and extract_slice for extracting a single row (one ciphertext) from a ciphertext-semantic tensor (multi-packed ciphertext).


asraa (approved):
easy enough for the case where it picks a certain ciphertext. remind me: at the stage that noise analysis has run, we've already lowered the non-trivial extract_slice operations into the appropriate masks / shifts?

j2kun:
Yes, this happens after convert-to-ciphertext-semantics.

PR #2277, opened 2025-09-30
Make BGV/CKKS ops elementwise-mappable, and add ElementwiseByOperandOpInterface

j2kun:
In removing the old layout system, I'm upgrading secret-to-scheme passes to support multi-packed ciphertexts, and as a result I find it's easiest to making all scheme ops elementwise mappable. However, `ckks.relinearize` cannot support this due to the trait verifier upstream that forces all tensor operands to be mapped over (its key-switching key argument is a different shape from the other operand).

This new interface (and upgrade to `convert-elementwise-to-affine`) allows operations like `relinearize` to be treated as "elementwise mappable" with respect to a subset of their operands. As future ops gain key material as arguments, they will need to be migrated from ElementwiseMappable to use this new interface.


lib/Dialect/LWE/IR/LWETypes.td
```diff
@@ -99,8 +99,10 @@ def LWECiphertext : LWE_Type<"LWECiphertext", "lwe_ciphertext", [MemRefElementTy
   let aliasSuffix = [{ getCiphertextSpace().getAliasSuffix(os); }];
 }
 
-def LWECiphertextLike : TypeOrValueSemanticsContainer<LWECiphertext, "lwe-ciphertext-like">;
 
 def LWEPlaintextOrCiphertext : AnyTypeOf<[LWECiphertext, LWEPlaintext]>;
 
+def LWECiphertextLike : TypeOrValueSemanticsContainer<LWECiphertext, "lwe-ciphertext-like">;
+def LWEPlaintextOrCiphertextLike : TypeOrValueSemanticsContainer<LWEPlaintextOrCiphertext, "lwe-ciphertext-like">;
```
asraa:
```suggestion
def LWEPlaintextOrCiphertextLike : TypeOrValueSemanticsContainer<LWEPlaintextOrCiphertext, "lwe-plaintext-or-ciphertext-like">;
```

asraa (approved):

AlexanderViand (approved):
❤️  

Do you think there's a chance to make this a thing upstream? If so, could ElementwiseMappable could be redefined to simply be ElementwiseByOperand on all operands (so that passes only need to handle the more generic case and not two different traits)?

j2kun:
> Do you think there's a chance to make this a thing upstream? If so, could ElementwiseMappable could be redefined to simply be ElementwiseByOperand on all operands (so that passes only need to handle the more generic case and not two different traits)?

Maybe. The scope of impacted stuff upstream is bigger: 4 traits involved, and passes involving linalg.elementwise. it's not clear that others have this issue of needing to replicate a tensor operand (rather than a scalar, which is already supported).

I don't think it's a high priority for us to have it upstream since we also own the pass that uses it. I'd rather spend my upstream dev time on higher priority stuff (I've started to poke around IRDL)

PR #2296, opened 2025-10-10
Add packing design doc to the HEIR website

### LINKED ISSUE #913: Unified notation for representing a packing of tensor data into a set of RLWE ciphertexts
Based on discussions in https://github.com/google/heir/issues/906 (e.g., https://github.com/google/heir/issues/906#issuecomment-2292793948), I think it would be prudent to have everyone agree on a notation for packing layouts. As I see it, the options are:

- TileTensor notation
- Fhelipe's index-bitstring (see https://github.com/google/heir/issues/800)
- A formalization of @edwjchen's comment https://github.com/google/heir/issues/906#issuecomment-2292793948

IMO we would ultimately formalize this as a custom attribute, similar to `affine_map`.

This would need to support:

- Packing multiple data items into one ciphertext (e.g., the example in Fhelipe of packing an entire matrix into a single ciphertext)
- Splitting one data item across multiple ciphertexts
- Packing multiple data across multiple ciphertexts?

### LINKED ISSUE #1888: Get a simple MNIST model e2e through a SIMD FHE pipeline

j2kun:
Fixes #1888 
Fixes #913

### DIFF
--- a/docs/content/en/docs/Design/layout.md
+++ b/docs/content/en/docs/Design/layout.md
@@ -1,0 +1,605 @@
+---
+title: Ciphertext Packing System
+weight: 9
+---
+
+This document describes HEIR's ciphertext packing system, including:
+
+- A notation and internal representation of a ciphertext packing, which we call
+  a _layout_.
+- An abstraction layer to associate SSA values with layouts and manipulate and
+  analyze them before a program is converted to concrete FHE operations.
+- A variety of layouts and kernels from the FHE literature.
+- A layout and kernel optimizer based on the
+  [Fhelipe compiler](https://github.com/fhelipe-compiler/fhelipe).
+- A layout conversion implementation of the
+  [Vos-Vos-Erkin graph coloring algorithm](https://link.springer.com/chapter/10.1007/978-3-031-17140-6_20).
+
+For background on what ciphertext packing is and its role in homomorphic
+encryption, see
+[this introductory blog post](https://www.jeremykun.com/2024/09/06/packing-matrix-vector-multiplication-in-fhe/).
+The short version of that blog post is that the SIMD-style HE computational
+model requires implementing linear-algebraic operations in terms of elementwise
+additions, multiplications, and cyclic rotations of large-dimensional vectors
+(with some exceptions like the
+[Park-Gentry matrix-multiplication kernel](https://eprint.iacr.org/2025/448)).
+
+Practical programs require many such operations, and the task of the compiler is
+to jointly choose ciphertext packings and operation kernels so as to minimize
+overall program latency. In this document we will call the joint process of
+optimizing layouts and kernels by the name "layout optimization." In FHE
+programs, runtime primarily comes from the quantity of rotation and bootstrap
+operations, the latter of which is in turn approximated by multiplicative depth.
+Metrics like memory requirements may also be constrained, but for most of this
+document latency is the primary concern.
+
+HEIR's design goal is to be an extensible HE compiler framework, we aim to
+support a variety of layout optimizers and multiple layout representations. As
+such, we separate the design of the layout representation from the details of
+the layout optimizer, and implement lowerings for certain ops that can be reused
+across optimizers.
+
+This document will begin by describing the layout representation, move on to the
+common, reusable components for working with that representation, and then
+finally describe one layout optimizer implemented in HEIR based on Fhelipe.
+
+## Layout representation
+
+A _layout_ is a description of how cleartext data is organized within a list of
+ciphertexts. In general, a layout is a partial function mapping from the index
+set of a list of ciphertext _slots_ to the index set of a cleartext tensor. The
+function describes which cleartext data value is stored at which ciphertext
+slot.
+
+A layout is _partial_ because not all ciphertext slots need to be used, and the
+function uses ciphertext slots as the domain and cleartext indices as the
+codomain because cleartext values may be replicated among multiple slots, but a
+slot can store at most one cleartext value.
+
+HEIR restricts the above definition of a layout as follows:
+
+- The partial function must be expressible as a _Presburger relation_, which
+  will be defined in detail below.
+- Unmapped ciphertext slots always contain zero.\[^zero\]
+
+We claim that this subset of possible layouts is a superset of all layouts that
+have been used in the FHE literature to date. For example, both the layout
+notation of Fhelipe and the TileTensors of HeLayers are defined in terms of
+specific parameterized quasi-affine formulas.
+
+Next we define a Presburger relation, then move on to examples.
+
+### Quasi-affine formulas and Presburger relations
+
+**Definition:** A _quasi-affine_ formula is a multivariate formula built from
+the following operations:
+
+- Integer literals
+- Integer-valued variables
+- addition and subtraction
+- multiplication by an integer constant
+- floor- and ceiling-rounded division by a nonzero integer constant
+- modulus by a nonzero integer constant
+
+Using the BNF grammar from the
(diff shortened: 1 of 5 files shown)


docs/content/en/docs/Design/layout.md
```diff
@@ -0,0 +1,608 @@
...
+**FIXME: kernels are currently considered free, cf. #1888**
+
+The cost of a layout conversion is estimated by simulating what the
+`implement-shift-network` would do if it ran on a layout conversion. And
+`layout-optimization` includes analyses that allow it to determine a folded cost
+for layout conversions that occur after other layout conversions, as well as the
+free cost of layout conversions that occur at function arguments, after
+`assign_layout` ops, or separated from these by ops that do not modify a layout.
+
+**FIXME: extend this pass to also pick the min kernel for an op even if it
```
j2kun: @asraa I am considering trying to fix this FIXME and the above before merging this design doc. Thoughts? LMK if you had already started on either of these.

j2kun: Moving it to a TODO

j2kun:
@asraa Ready for review!

docs/content/en/docs/Design/layout.md
```diff
@@ -0,0 +1,605 @@
...
+- For a block argument, the layout attribute is stored on the op owning the
+  block, using the `OperandAndResultAttrInterface` to give a consistent API for
+  accessing the attribute.
+
+These two differences are handled properly by a helper library,
+`lib/Utils/AttributeUtils.h`, which exposes setters and getters for layout
+attributes. As of 2025-10-01, the system does not provide a way to handle ops
+with multiple regions or multi-block regions.
+
+For example:
```
asraa:
```suggestion
For example, `#layout_attr` is associated with the SSA value `%1`:
```

docs/content/en/docs/Design/layout.md
```diff
@@ -0,0 +1,605 @@
...
+[Layout attributes](#layout-attributes) above), and data types are unchanged.
+
+In this regime, there are three special `tensor_ext` ops that are no-ops on
+data-semantic type, but are designed to manipulate the layout attributes. These
+ops are:
+
+- `tensor_ext.assign_layout`, which takes a data-semantic value and a layout
+  attribute, and produces the same data-semantic type. This is an "entry point"
+  into the layout system and lowers to a loop that packs the data according to
+  the layout.
```
asraa, line 203:
```suggestion
  attribute, and produces the same data-semantic type. This is an "entry point" into the layout system
  for cleartext values and lowers to a loop that packs the data according to the layout.
```

Actually I just wanted to understand this clearly just to double check. Cleartext values that are block args will also be assign_layout'd too right? (i.e. we won't annotate the block arg with a layout)

docs/content/en/docs/Design/layout.md
```diff
@@ -0,0 +1,605 @@
...
+  data-semantic value's current layout and a new layout. Typically this lowers
+  to a shift network.
+- `tensor_ext.unpack`, which clears the layout attribute on a data-semantic
+  value, and serves as an exit point from the layout system. This lowers to a
+  loop which extracts the packed cleartext data back into user data.
+
+A layout optimizer is expected to insert `assign_layout` ops for any server-side
+cleartexts that need to be packed at runtime.
+
+By contrast, in the ciphertext-semantic regime, all secret values are rank-2
```
asraa: "By contrast" - to what?
My thought was that cleartext would not always be rank-2 packed tensors, but I think that is in fact the case.

docs/content/en/docs/Design/layout.md
```diff
@@ -0,0 +1,605 @@
...
+sequences of `assign_layout` followed by `convert_layout` are folded together
+into combined `assign_layout` ops.
+
+### `layout-optimization`
+
+The [`layout-optimization`](/docs/passes/#-layout-optimization) pass has two
+main goals: to choose better kernels for ops, and to try to eliminate
+`convert_layout` ops. It does this by running a backward pass through the IR. If
+it encounters an op that is followed by a `convert_layout` op, it attempts to
+hoist the `convert_layout` through the op to its arguments.
```
asraa, line 299: I've also added an initial forward pass too! Ah! I guess I can update that after. But the forward pass would start by pushing the initial default layouts through the ops as much as possible (assuming the cost of the kernel does not increase in doing so).

For example, in the case of the slice insertion, we'll pick the best kernel on the forward pass because it'll be cheaper to choose the slice insertion result to minimize the layout conversion needed to insert the slice.

Honestly, now I'm thinking I should just make that be part of layout-propagation (and modify the assigned destination layout during the forward pass if it only has one use)

asraa (approved):

docs/content/en/docs/Design/layout.md
j2kun: "Blocks" don't have the ability to annotate by themselves, you can only annotate the owning op.

But, for example, cleartext function arguments will be assigned a layout, and it won't be until we add the preprocessing helper that we can move them to the client.

docs/content/en/docs/Design/layout.md
j2kun: Removed the clause. Indeed, cleartexts are packed as rank-2 tensors.

docs/content/en/docs/Design/layout.md
asraa: Ah I didn't mean annotate the block arg but annotate the arg on the owning op (the func or generic)

And it seems like the rest of the doc only annotated cleartext func args if they're used in a ctxt op

PR #2319, opened 2025-10-16
Remove manual cyclic repetition in LattigoEmitter

### LINKED ISSUE #1258: Move cyclic repetition code to earlier pipeline instead of in emitter

j2kun:
Fixes #1258

Still has some issues to iron out.

### DIFF
--- a/lib/Analysis/NoiseAnalysis/BGV/NoiseCanEmbModel.cpp
+++ b/lib/Analysis/NoiseAnalysis/BGV/NoiseCanEmbModel.cpp
@@ -148,9 +148,13 @@ typename Model::StateType Model::evalConstant(
   auto t = param.getSchemeParam()->getPlaintextModulus();
   auto phi = getPhi(param);
 
-  // noise part of the plaintext in a pt-ct multiplication
-  // v_const <= t * sqrt(phi(m) / 12)
-  return StateType::of(t * sqrt(phi / 12.0));
+  // Noise part of the plaintext in a pt-ct multiplication. Changed from
+  // MMLGA22 to account for ciphertext-plaintext muls with large but sparse
+  // plaintext masks. Cf.
+  // https://github.com/google/heir/pull/2319#issuecomment-3424387609
+  //
+  // v_const <= t * phi(m) / 2
+  return StateType::of(t * phi / 2.0);
 }
 
 typename Model::StateType Model::evalAdd(const StateType& lhs,
--- a/lib/Target/Lattigo/LattigoEmitter.cpp
+++ b/lib/Target/Lattigo/LattigoEmitter.cpp
@@ -1086,49 +1086,50 @@ LogicalResult LattigoEmitter::printOperation(BGVNewPlaintextOp op) {
 }
 
 LogicalResult LattigoEmitter::printOperation(BGVEncodeOp op) {
-  // cyclic repetition to mitigate openfhe zero-padding (#645)
-  // TODO(#1258): move cyclic repetition to earlier pipeline
-
-  // hack: access another op to get params then get MaxSlots
+  // hack: access another op to get params
   auto newPlaintextOp =
       mlir::dyn_cast<BGVNewPlaintextOp>(op.getPlaintext().getDefiningOp());
   if (!newPlaintextOp) {
     return failure();
   }
-  auto maxSlotsName = getName(newPlaintextOp.getParams()) + ".MaxSlots()";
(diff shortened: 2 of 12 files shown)


j2kun:
@ZenithalHourlyRate there is one remaining failing test, the dot_product_8_debug test, which after these changes actually hits a noise bound panic

```bash
$ bazel test --nocache_test_results //tests/Examples/lattigo/bgv/dot_product_8_debug:dotproduct8debug_test

...

lattigo.bgv.mul
  [240 0 0 0 ... 0 0]
  Noise: 48.43 Total: 52
  Noise Bound: 47.72 Gap: -0.71
panic: Noise Bound Exceeded [recovered]
        panic: Noise Bound Exceeded
```

Do you see anything in this PR that could impact noise growth? It seems the failure is not deterministic on my machine...

j2kun:
At HEAD I noticed the same test fails when you set `ciphertext-degree=8192`, and I noticed that the generated MLIR does indeed set `LogN: 14` so I think 8192 is the right choice... but then in my PR the failure is the same for smaller ciphertext sizes as well, I think because I switched from using max slots pulled from the params to using `len(value)` (the zero-initialized cleartext tensor). So my PR forces it to always use `ciphertext-size` instead of pulling MaxSlots from the params object.

What's the source of truth here?

ZenithalHourlyRate:
The root cause might be the expansion factor, i.e. the noise modeling for ct-pt mul. In our modified KPZ21 model, we use 2 sqrt(N) almost everywhere. Note that in OpenFHE they do not need to estimate the noise for ct-pt mul whereas we do.

The problematic mul is a ct-pt mul, with pt being [1, 0, 0, ....]. After encoding, pt will be something in Z_t^N, possibly with all coefficients being large.

We calculate the bound by ||ct * pt|| = 2 sqrt(N) * ||ct|| * t, but 2 sqrt expansion factor only works when both sides are random Gaussian polynomials (i.e. it has mostly small and a few big elements), but here we have a fixed big polynomial.

One way to fix it is to use expansion factor of N for ct-pt mul.

j2kun:
I patched in the noise fix from #2325 and now dot_product_8_debug is passing but the mono bound variant is failing (flakily, so the noise bound is very close to the limit)

```
$ bazel test --nocache_test_results //tests/Examples/lattigo/bgv/dot_product_8_debug_mono:dotproduct8debug_test

lattigo.bgv.mul
  [240 0 0 ... 0]
  Noise: 54.05 Total: 57
  Noise Bound: 52.18 Gap: -1.87
```

Same issue as before: I'm updating the `ciphertext-degree=8192`.

j2kun:
@ZenithalHourlyRate I noticed in the Mono bounds for BGV, the formula for evalConstant doesn't seem to take into account the sparsity of the constant. Would this be the place to make that noise model more intelligent?

ZenithalHourlyRate:
> @ZenithalHourlyRate I noticed in the Mono bounds for BGV, the formula for evalConstant doesn't seem to take into account the sparsity of the constant. Would this be the place to make that noise model more intelligent?

Indeed. The original paper says

> Here, we consider the constant α to be _uniformly_ distributed in Rt and the
bound on νconst comes from [15].

Practically we only know the coeffs of the constant polynomial to be in range [-t/2, t/2] and its canonical embedding norm is known to be bounded by Nt/2 instead of t * sqrt(N / 12). One worst-case example is to take the constant polynomial with all coefficients being t/2, and canonical embedding will make an N-sum of them (with roots of unity as shifts). There might be some some smaller constant from the norm of the Vandermonde matrix but the O(N) order is correct.

A way to change the formula is

```diff
  // noise part of the plaintext in a pt-ct multiplication
-  // v_const <= t * sqrt(phi(m) / 12)
-  return StateType::of(t * sqrt(phi / 12.0));
+  // v_const <= t * phi(m) / 2
+  return StateType::of(t * phi / 2.0);
```

j2kun:
Looks good! @ZenithalHourlyRate if you have a chance could you review the PR?

PR #2325, opened 2025-10-17
BGV: NoiseAnalysis KPZ21: more conservative on ct-pt mul

ZenithalHourlyRate:
See https://github.com/google/heir/pull/2319#issuecomment-3413879680 for backgrounds.

A 2sqrt(N) expansion factor for ct-pt mul is not enough as we do not know the distribution of pt. In previous testing for BGV, we only emphasized ct-ct cases where the v1 * v2 dominates but put less consideration for ct-pt case where v1 * v2 = 0.

This PR uses the expansion factor of N for ct-pt mul. I've experimentally observed an at-most 2 bit overflow so it suggests we could not adjust to some other Csqrt(N) (in this case this would mean 8sqrt(N)).

In BGV, ct-ct mul and ct-pt mul shares the same noise estimation function/formula and can be adjusted as in this PR. I would remark that ct-pt case is also relevant for BFV, but BFV ct-ct v.s. ct-pt use different multiplication structure (former uses multiplication technique, while later is much simpler), and the ct-ct formula is much more conservative when applied to ct-pt case so it should not cause overflow for BFV ct-pt mul. In other words, a new formula for ct-pt mul for BFV is an optimization but the current formula is good enough. 

@j2kun Could you please test 2319 again with this patch?

Cc @yspolyakov for what it's worth.

### DIFF
--- a/lib/Analysis/NoiseAnalysis/BGV/NoiseByBoundCoeffModel.cpp
+++ b/lib/Analysis/NoiseAnalysis/BGV/NoiseByBoundCoeffModel.cpp
@@ -156,16 +156,15 @@ typename NoiseByBoundCoeffModel::StateType NoiseByBoundCoeffModel::evalMul(
   auto ringDim = resultParam.getSchemeParam()->getRingDim();
 
   // (m_0 + tv_0) * (m_1 + tv_1) <=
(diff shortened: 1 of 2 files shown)


j2kun:
This causes the test to pass from #2319, however, the mono bounds (MMLGA22) of the same test fail as well. I will merge this one in and keep going on that other thread.

Issue #2328, opened 2025-10-17
Add an Orion entry dialect


j2kun:
Based on https://github.com/google/heir/issues/2326 and https://github.com/eymay/orion_heir_translator

the two examples from the translator repo produce these ops

```
rg -o '%[0-9]* = [^ ]*' mlp_fhe.mlir | grep -o '[^ ]*$' | sort | uniq
arith.constant
ckks.add
ckks.add_plain
ckks.linear_transform
ckks.mul
ckks.relinearize
ckks.rotate
lwe.rlwe_encode
```

```
rg -o '%[0-9]* = [^ ]*' resnet10_fhe.mlir | grep -o '[^ ]*$' | sort | uniq 
arith.constant
ckks.add
ckks.add_plain
ckks.bootstrap
ckks.chebyshev
ckks.linear_transform
ckks.mul
ckks.relinearize
ckks.rotate
lwe.rlwe_encode
```

The scheme params are populated

```
{ckks.schemeParam = #ckks.scheme_param<logN = 13, Q = [536903681, 67043329, 66994177, 67239937, 66961409, 66813953], P = [67108864], logDefaultScale = 26>}
```

and the LWE ciphertext type input to mlp has the following structure.

```
!lwe.new_lwe_ciphertext<
  plaintext_space = <ring = <coefficientType = f64, polynomialModulus = <1 + x**8192>>, encoding = #lwe.inverse_canonical_encoding<scaling_factor = 26>>,
  ciphertext_space = <ring = #polynomial.ring<coefficientType = !rns.rns<!mod_arith.int<536903681 : i64>, !mod_arith.int<67043329 : i64>, !mod_arith.int<66994177 : i64>, !mod_arith.int<67239937 : i64>, !mod_arith.int<66961409 : i64>, !mod_arith.int<66813953 : i64>>, polynomialModulus = <1 + x**8192>>, encryption_type = mix, size = 2>, 
  key = #lwe.key<>, 
  modulus_chain = <elements = <536903681 : i64, 67043329 : i64, 66994177 : i64, 67239937 : i64, 66961409 : i64, 66813953 : i64>, current = 5>>
```

Based on this I think `orion` dialect can reuse the LWE types, and will get its own ops for all of these. `chebyshev` and `linear_transformation` will need special lowerings and everything else will be 1-1 with `ckks`

PR #2330, opened 2025-10-17
Add orion entry dialect

j2kun:
Fixes https://github.com/google/heir/issues/2328

This corresponds to [3a09186eef73f046d4eff1f32e6ba58f4c8da349](https://github.com/j2kun/orion_heir_translator/commit/3a09186eef73f046d4eff1f32e6ba58f4c8da349) which runs orion and exports ckks + orion dialect MLIR.

From that project, after installing deps, I ran

```
python examples/mlp.py
python examples/resnet.py  # needs like 60GiB of RAM or lots of swap space, takes about 5m
```

Then ran the result through `heir-opt` with this PR (which just extracted type aliases), and persisted the result to the lit tests in this PR

### DIFF
--- a/lib/Dialect/Orion/IR/OrionOps.td
+++ b/lib/Dialect/Orion/IR/OrionOps.td
@@ -1,0 +1,37 @@
+#ifndef LIB_DIALECT_ORION_IR_ORIONOPS_TD_
+#define LIB_DIALECT_ORION_IR_ORIONOPS_TD_
+
+include "lib/Dialect/Orion/IR/OrionDialect.td"
+include "lib/Dialect/LWE/IR/LWETypes.td"
+
+include "mlir/IR/BuiltinAttributes.td"
+include "mlir/IR/CommonAttrConstraints.td"
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+class Orion_Op<string mnemonic, list<Trait> traits = []> :
+        Op<Orion_Dialect, mnemonic, traits> {
+  let cppNamespace = "::mlir::heir::orion";
+  let assemblyFormat = "operands attr-dict `:` functional-type(operands, results)" ;
+}
+
+def Orion_LinearTransformOp : Orion_Op<"linear_transform", [Pure]> {
+  // Use a cleartext for the diagonals since they need to be rotated in the
+  // BSGS implementation of linear transforms.
+  let arguments = (ins LWECiphertext:$input, FloatLike:$diagonals);
+  let results = (outs LWECiphertext:$result);
+  let summary = "Apply a linear transformation to the input ciphertext using pre-computed diagonals";
+}
+
+def Orion_ChebyshevOp : Orion_Op<"chebyshev", [Pure]> {
+  let arguments = (ins
+    LWECiphertext:$input,
+    ArrayAttr:$coefficients,
+    Builtin_FloatAttr:$domain_start,
+    Builtin_FloatAttr:$domain_end
+  );
+  let results = (outs LWECiphertext:$result);
+  let summary = "Evaluates a Chebyshev polynomial on a ciphertext using pre-computed coefficients";
+}
+
+#endif  // LIB_DIALECT_ORION_IR_ORIONOPS_TD_
--- a/tests/Dialect/Orion/IR/mlp.mlir
+++ b/tests/Dialect/Orion/IR/mlp.mlir
@@ -1,0 +1,64 @@
+// RUN: heir-opt %s
+
+!Z536903681_i64 = !mod_arith.int<536903681 : i64>
+!Z66813953_i64 = !mod_arith.int<66813953 : i64>
+!Z66961409_i64 = !mod_arith.int<66961409 : i64>
+!Z66994177_i64 = !mod_arith.int<66994177 : i64>
+!Z67043329_i64 = !mod_arith.int<67043329 : i64>
+!Z67239937_i64 = !mod_arith.int<67239937 : i64>
+#inverse_canonical_encoding = #lwe.inverse_canonical_encoding<scaling_factor = 26>
+#inverse_canonical_encoding1 = #lwe.inverse_canonical_encoding<scaling_factor = 104>
+#inverse_canonical_encoding2 = #lwe.inverse_canonical_encoding<scaling_factor = 52>
+#key = #lwe.key<>
+#modulus_chain_L5_C5 = #lwe.modulus_chain<elements = <536903681 : i64, 67043329 : i64, 66994177 : i64, 67239937 : i64, 66961409 : i64, 66813953 : i64>, current = 5>
+#ring_f64_1_x8192 = #polynomial.ring<coefficientType = f64, polynomialModulus = <1 + x**8192>>
+!rns_L5 = !rns.rns<!Z536903681_i64, !Z67043329_i64, !Z66994177_i64, !Z67239937_i64, !Z66961409_i64, !Z66813953_i64>
+!pt = !lwe.lwe_plaintext<application_data = <message_type = tensor<4096xf64>>, plaintext_space = <ring = #ring_f64_1_x8192, encoding = #inverse_canonical_encoding>>
+!pt1 = !lwe.lwe_plaintext<application_data = <message_type = tensor<4096xf64>>, plaintext_space = <ring = #ring_f64_1_x8192, encoding = #inverse_canonical_encoding2>>
+!pt2 = !lwe.lwe_plaintext<application_data = <message_type = tensor<4096xf64>>, plaintext_space = <ring = #ring_f64_1_x8192, encoding = #inverse_canonical_encoding1>>
+#ring_rns_L5_1_x8192 = #polynomial.ring<coefficientType = !rns_L5, polynomialModulus = <1 + x**8192>>
+#ciphertext_space_L5 = #lwe.ciphertext_space<ring = #ring_rns_L5_1_x8192, encryption_type = mix>
+#ciphertext_space_L5_D3 = #lwe.ciphertext_space<ring = #ring_rns_L5_1_x8192, encryption_type = mix, size = 3>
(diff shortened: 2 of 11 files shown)


j2kun:
CC @eymay

tests/Dialect/Orion/IR/mlp.mlir
```diff
@@ -0,0 +1,61 @@
...
+    %ct_7 = ckks.rotate %ct_6 {offset = 512 : i32} : !ct_L5
+    %ct_8 = ckks.add %ct_7, %ct_6 : (!ct_L5, !ct_L5) -> !ct_L5
+    %ct_9 = ckks.rotate %ct_8 {offset = 256 : i32} : !ct_L5
+    %ct_10 = ckks.add %ct_9, %ct_8 : (!ct_L5, !ct_L5) -> !ct_L5
+    %ct_11 = ckks.rotate %ct_10 {offset = 128 : i32} : !ct_L5
+    %ct_12 = ckks.add %ct_11, %ct_10 : (!ct_L5, !ct_L5) -> !ct_L5
+    %ct_13 = ckks.add_plain %ct_12, %pt : (!ct_L5, !pt) -> !ct_L5
+    %ct_14 = ckks.mul %ct_13, %ct_13 : (!ct_L5, !ct_L5) -> !ct_L5_D3
+    %ct_15 = ckks.relinearize %ct_14 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : (!ct_L5_D3) -> !ct_L5_2
+    %ct_16 = orion.linear_transform %ct_15, %arg1 {block_col = 0 : i32, block_row = 0 : i32, bsgs_ratio = 2.000000e+00 : f64, diagonal_count = 128 : i32, orion_level = 3 : i32, slots = 4096 : i32} : (!ct_L5_2, tensor<128x!pt>) -> !ct_L5_2
```
eymay: This is more related to the translator than this PR but the `linear_transform` having the same ciphertext input and return types can cause issues later during lowering. I fixed that with type propagation during lowering `linear_transform` but a cleaner approach could be ensuring that the types are correct during the construction. That is of course when we assume the lowering is fixed to a single kernel and we know what the return type will be.

eymay: Same should be true for ops like `chebyshev` or `bootstrap`. Bootstrap is tricky though because OpenFHE will bail out from a bootstrap that ends up with less levels in the end than more. So a misplaced bootstrap op here might actually not change the type of the ciphertext as it returns the same thing.

tests/Dialect/Orion/IR/resnet10.mlir
```diff
@@ -0,0 +1,147 @@
...
+    %ct_34 = ckks.mul %ct_33, %ct_33 : (!ct_L10, !ct_L10) -> !ct_L10_D3
+    %ct_35 = ckks.relinearize %ct_34 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : (!ct_L10_D3) -> !ct_L10_1
+    %ct_36 = ckks.mul %ct_35, %ct_35 : (!ct_L10_1, !ct_L10_1) -> !ct_L10_D3_1
+    %ct_37 = ckks.relinearize %ct_36 {from_basis = array<i32: 0, 1, 2>, to_basis = array<i32: 0, 1>} : (!ct_L10_D3_1) -> !ct_L10_2
+    %ct_38 = ckks.bootstrap %ct_37 : !ct_L10_2 -> !ct_L10
+    %ct_39 = orion.linear_transform %ct_38, %arg3 {block_col = 0 : i32, block_row = 0 : i32, bsgs_ratio = 2.000000e+00 : f64, diagonal_count = 256 : i32, orion_level = 2 : i32, slots = 32768 : i32} : (!ct_L10, tensor<256x!pt>) -> !ct_L10
+    %ct_40 = ckks.rotate %ct_39 {offset = 16384 : i32} : !ct_L10
+    %ct_41 = ckks.add %ct_40, %ct_39 : (!ct_L10, !ct_L10) -> !ct_L10
+    %ct_42 = ckks.bootstrap %ct_41 : !ct_L10 -> !ct_L10
+    %ct_43 = ckks.bootstrap %ct_42 : !ct_L10 -> !ct_L10
```
eymay, line 75: This is again about the translator but because the bootstrap placement of Orion is not tuned for OpenFHE, we try to be safe by emitting bootstrap before every chebyshev which is clearly not necessary here.

eymay (approved):
Thanks for adding Orion frontend support. The PR looks clean to me, I pointed out some limitations of the translator in comments.

tests/Dialect/Orion/IR/mlp.mlir
j2kun: Could you explain the meaning of the `'orion_level': 3` metadata attached here? I noticed it does change over `linear_transform` ops. (e.g., in the MLP example it is 5 for the first op, 3 for the second op).

Edit: is this the required level for encoding the plaintext input?

j2kun:
@eymay I did find it a bit strange that all the chebyshev ops emitted have a domain start/end as -1, 1 exactly, rather than some domain specific to the network/data distribution. Is that what you observed as well? Or does it indicate a bug on my part?

j2kun:
@eymay sorry to keep bugging you again, but I also noticed that the `bsgs_ratio` is always 2, and I can't find anywhere in the Orion codebase that sets this value to a non-hard-coded-2. Did you notice the same?

eymay:
> @eymay I did find it a bit strange that all the chebyshev ops emitted have a domain start/end as -1, 1 exactly, rather than some domain specific to the network/data distribution. Is that what you observed as well? Or does it indicate a bug on my part?

I think this is a desired result from the Orion compilation, this is to ensure a stable polynomial approximation. So Orion normalizes the data distribution for polynomial approximation, from the Orion [paper](https://arxiv.org/pdf/2311.03470#page=9):
> Range estimation: Both high-precision bootstrapping and
high-degree (Chebyshev) polynomial activation functions
require the inputs to be in the range [−1, 1].
Prior work chooses per network scaling factors by man-
ually inspecting and tracking the largest intermediate val-
ues seen [41, 52]. Orion handles this process automatically
through net.fit(), which accepts the entire training dataset
as input, calculates per layer scaling factors, and inserts scale-
down multiplications directly into the computational graph.

My default approach for the attributes in the Orion internal IR was to preserve them in the MLIR IR so they become explicit even if we don't end up using them.

eymay:
> @eymay sorry to keep bugging you again, but I also noticed that the `bsgs_ratio` is always 2, and I can't find anywhere in the Orion codebase that sets this value to a non-hard-coded-2. Did you notice the same?

Please I appreciate the discussion. Yes Orion seems to just keep `bsgs_ratio=2` for all the layers like `Linear`, `AvgPool2d` and `Conv2d`. On top of that the translator also emits the same number if there is no attribute found for the op. The ratio plugs into Lattigo's `LogBabyStepGiantStepRatio` parameter explained [here](https://github.com/tuneinsight/lattigo/blob/e455ddedadb2fa362317dd5dd1d2c4ce00ba58e1/circuits/common/lintrans/lintrans.go#L74). It seems like a heuristic number that stayed on but we don't need to take it into account while lowering.

tests/Dialect/Orion/IR/mlp.mlir
eymay: Yes I think so, Orion associates every `linear_transform` with a `level` attribute. This plugs into Lattigo's `LevelQ` parameter [here](https://github.com/tuneinsight/lattigo/blob/e455ddedadb2fa362317dd5dd1d2c4ce00ba58e1/circuits/common/lintrans/lintrans.go#L56). It looks like the level for encoding the diagonal plaintexts. Most attributes here don't seem necessary since we are compiling with full program context instead of calling an API with them. Maybe we can add a verbose mode for the translator emitting all the information like this that are not critical for compilation.

j2kun:
Ok, one change I want to make before this PR is submitted is to have the plaintext arguments be cleartext-typed, so that the function encodes them as plaintexts at runtime. This is not optimal, but we need the ability to rotate the plaintexts for the purpose of BSGS (to my knowledge none of these libraries support rotating an encoded plaintext; decoding just to rotate and re-encode seems silly). While we could pre-rotate the diagonals Orion produces in a manner specific to HEIR's implementation of BSGS, it seems quite brittle, and I suspect the runtime hit will not be that bad...

lib/Dialect/Orion/IR/OrionOps.cpp
```diff
@@ -0,0 +1,7 @@
+#include "lib/Dialect/Orion/IR/OrionOps.h"
```
asraa: nit: remove this?

lib/Dialect/Orion/IR/OrionOps.td
```diff
@@ -0,0 +1,38 @@
...
+
+def Orion_LinearTransformOp : Orion_Op<"linear_transform", [Pure]> {
+  // Use a cleartext for the diagonals since they need to be rotated in the
+  // BSGS implementation of linear transforms.
+  let arguments = (ins LWECiphertext:$input, FloatLike:$diagonals);
+  let results = (outs LWECiphertext:$result);
+  let summary = "Apply a linear transformation to the input ciphertext using pre-computed diagonals";
+}
+
+def Orion_ChebyshevOp : Orion_Op<"chebyshev", [Pure, AllTypesMatch<["input", "result"]>]> {
```
asraa: why would the input and result types match here? wouldn't the evaluation of the poly reduce the levels probably?

lib/Dialect/Orion/IR/OrionOps.cpp
j2kun: Removed the whole file.

lib/Dialect/Orion/IR/OrionOps.td
j2kun: Yes it should, but we don't know at the time this dialect is generated how many levels will be used. I can remove the constraint either way.

PR #2359, opened 2025-10-29
Add bicyclic layout and matmul kernel

j2kun:
This PR adds a new layout relation to represent the bicyclic matrix layout from https://eprint.iacr.org/2024/1762, along with the matmul kernel BMM-I from that paper, implemented in the ArithmeticDag framework.

This follows the python reference implementation at https://github.com/j2kun/fhe-packing/commit/87abe57193a6f2ee5d76bcd38888071d331dc814

However, it does not include the Baby-Step Giant-Step optimization as described in https://eprint.iacr.org/2025/1200. I wanted to get the initial kernel implemented correctly, then in a followup PR I can focus on the (independent) steps of:

- Integrating the layout/kernel into the pipeline
- Improving the kernel to use BSGS

Open question: the rotations used use this nontrivial structure

```
    auto rotA = NodeTy::leftRotate(packedADag, (-i * m));
    auto rotB = NodeTy::leftRotate(packedBDag, (i * (r * n - m)));
```

It was not clear to me if the existing `tensor_ext.rotate_and_reduce` would support this. Based on my understanding of section 6.2 of Lawrence's paper (https://eprint.iacr.org/2025/1200.pdf), BSGS only applies to the ciphertext-plaintext matmul kernel, so it's not immediately clear how to optimize the rotation schedule except in using the lower-level rotation hoisting.

Part of https://github.com/google/heir/issues/1376

### DIFF
--- a/lib/Kernel/KernelImplementation.h
+++ b/lib/Kernel/KernelImplementation.h
@@ -15,6 +15,7 @@
 #include "lib/Kernel/AbstractValue.h"
 #include "lib/Kernel/ArithmeticDag.h"
 #include "lib/Kernel/KernelName.h"
+#include "lib/Utils/APIntUtils.h"
 #include "lib/Utils/MathUtils.h"
 #include "mlir/include/mlir/Support/LLVM.h"  // from @llvm-project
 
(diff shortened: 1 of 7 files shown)


lib/Kernel/KernelImplementation.h
```diff
@@ -178,6 +178,37 @@ implementHaleviShoup(const T& vector, const T& matrix,
...
+std::enable_if_t<std::is_base_of<AbstractValue, T>::value,
+                 std::shared_ptr<ArithmeticDagNode<T>>>
+implementBicyclicMatmul(const T& packedA, const T& packedB, int64_t m,
+                        int64_t n, int64_t p) {
+  using NodeTy = ArithmeticDagNode<T>;
+  auto packedADag = NodeTy::leaf(packedA);
+  auto packedBDag = NodeTy::leaf(packedB);
+
+  // This implements the BMM-I algorithm from
+  // https://eprint.iacr.org/2024/1726
```
lawrencekhlim: Corrected Link:
// https://eprint.iacr.org/2024/1762

lib/Kernel/KernelImplementation.h
```diff
@@ -178,6 +178,37 @@ implementHaleviShoup(const T& vector, const T& matrix,
...
+// Returns an arithmetic DAG that implements the bicyclic matrix multiplication
+// algorithm.
+template <typename T>
+std::enable_if_t<std::is_base_of<AbstractValue, T>::value,
+                 std::shared_ptr<ArithmeticDagNode<T>>>
+implementBicyclicMatmul(const T& packedA, const T& packedB, int64_t m,
+                        int64_t n, int64_t p) {
+  using NodeTy = ArithmeticDagNode<T>;
+  auto packedADag = NodeTy::leaf(packedA);
+  auto packedBDag = NodeTy::leaf(packedB);
```
lawrencekhlim, line 197: A few quick questions as I'm not terribly familiar with the repository anymore: What does `NodeTy::leaf(packedA)` do? At this point, is the encoding of the matrix in the ciphertext repeated a number of times as is required by BMM-I?

lawrencekhlim: May want have a comment on what kind of packing is assumed at the beginning of this function? Zero-padded, no repeated, bicyclic encoding? Or is it zero-padded, but with repetition, bicyclic encoding?

j2kun: I will indeed add a comment explaining the assumptions. A different part of the pipeline ensures that the input to this function is properly packed.

For your other comment, the `ArithmeticDagNode` construction allows us to represent the kernel as a circuit in memory (in an easier way that representing it as MLIR) so that we can (1) evaluate it on test data in unit tests (2) analyze its structure and (3) generate IR from it. So `NodeTy::leaf` just asserts that the argument is an input to the circuit (but it could be a literal value, an SSA value, or a symbolic value)

j2kun:
> You could also adopt the formulation from my paper (batch matrix multiplications with h=1 in Section 5.2.1, Equation 9), mostly because I find it more intuitive, but this is fine too.

Ah, I recall struggling to get your formulation to work the same as the BMM-I when I first implemented it (in a python reference) but now when I tried again it was very straightforward. /shrug

> Lastly, BSGS does apply to ctxt-ctxt matrix multiplications (See section 6.2.2).

I will take a bit more to absorb this. Thanks for the precise reference.

j2kun:
OK the issue with our baby-step giant-step implementation is that the kernel assumes a structure of the form of equation 16-17 from 6.2.1 of https://eprint.iacr.org/2025/1200, and 6.2.2 requires us to apply an additional function to the rotation operand before applying the rotation.

This requires these changes, which I will defer to a followup PR:

1. Modify `implementRotateAndReduce` to take a callable to apply to the basic BSGS-decomposed shift before actually calling `leftRotate` with it. This is the `RotY` function in equation 21.
2. Modify the input to repeat the `y` ciphertext into a larger tensor shape to fit the API. Note this will produce IR that looks like a bunch of "tensor.insert_slice" to construct the tensor followed up "tensor.extract_slice" in the kernel to extract the individual ciphertexts. We need to ensure canonicalization is applied after this step so that these are folded back to just use the ciphertext in place. Either that or make a different, non-tensor-shaped kernel that doesn't require post-cleanup.
3. Rename the `plaintext` operand of `implementRotateAndReduce` to clarify it need not strictly be plaintext

j2kun:
@asraa Since Lawrence reviewed, marking pull_ready and you can stamp internally

PR #2360, opened 2025-10-29
Add bicyclic matmul to layout pipeline

j2kun:
This PR adds bicyclic matmul to layout pipeline

Minor fixes:
- ensure secretness lattice is updated if convert-layout is inserted
  mid-pass
- cleanup of Codegen.cpp and support for floordiv op

Left for next PR: adding e2e test of the full matmul op

Open question: is there a meaningful convert-layout hoister for bicyclic matmul? Otherwise I feel that layout conversions will never be naturally hoisted through it by the heuristic.

Rebased over https://github.com/google/heir/pull/2359
Part of https://github.com/google/heir/issues/1376

### DIFF
--- a/lib/Transforms/LayoutOptimization/InterfaceImpl.cpp
+++ b/lib/Transforms/LayoutOptimization/InterfaceImpl.cpp
@@ -76,6 +76,32 @@ struct MatvecHoistingImpl
   }
 };
 
+struct MatmulHoistingImpl
+    : public LayoutConversionHoistableOpInterface::ExternalModel<
+          MatmulHoistingImpl, linalg::MatmulOp> {
+  std::vector<Hoister> getHoisters(
+      Operation* op, tensor_ext::ConvertLayoutOp convertLayoutOp) const {
+    std::vector<Hoister> hoisters;
+
+    auto kernel = op->getAttrOfType<secret::KernelAttr>(
+        secret::SecretDialect::kKernelAttrName);
+    if (!kernel) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Kernel attribute not found on op " << *op << "\n");
+      return hoisters;
+    }
+
+    if (!op->hasAttr(tensor_ext::TensorExtDialect::kLayoutAttrName)) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Layout attribute not found on op " << *op << "\n");
+      return hoisters;
+    }
+
+    // TODO(#2385): try hoisting layout conversion through bicyclic matmul
+    return hoisters;
+  }
+};
+
 }  // namespace
 
 Hoister createTrivialHoister(Operation* op) {
@@ -144,6 +170,7 @@ void registerLayoutConversionHoistableInterface(DialectRegistry& registry) {
   });
   registry.addExtension(+[](MLIRContext* ctx, linalg::LinalgDialect* dialect) {
     linalg::MatvecOp::attachInterface<MatvecHoistingImpl>(*ctx);
+    linalg::MatmulOp::attachInterface<MatmulHoistingImpl>(*ctx);
   });
 }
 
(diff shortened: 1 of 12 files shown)


lib/Transforms/LayoutOptimization/InterfaceImpl.cpp
```diff
@@ -76,6 +76,32 @@ struct MatvecHoistingImpl
...
+      return hoisters;
+    }
+
+    if (!op->hasAttr(tensor_ext::TensorExtDialect::kLayoutAttrName)) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Layout attribute not found on op " << *op << "\n");
+      return hoisters;
+    }
+
+    // TODO(#FIXME): does bicyclic matmul have a meaningful hoister?
```
j2kun: Note to self: file this once the PR is otherwise approved

lawrencekhlim:
I get the general idea of what's going on, but since I'm not too familiar with the layout pipeline, it may be better if @asraa reviews this PR more closely.

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
```diff
@@ -639,55 +645,92 @@ LogicalResult LayoutPropagation::visitOperation(Conv2DOp op) {
...
+    setResultLayoutAttr(op);
+    debugAssignLayout(result, outputLayoutAttr);
+
+    auto kernelAttr = secret::KernelAttr::get(ctx, KernelName::MatmulBicyclic,
+                                              /*force=*/false);
+    op->setAttr(secret::SecretDialect::kKernelAttrName, kernelAttr);
+    return success();
+  }
+
+  // TODO(#1376): Implement bicyclic ciphertext-ciphertext matmul kernel.
```
asraa: nit: remove now?

lib/Utils/Layout/Codegen.cpp
```diff
@@ -43,11 +43,12 @@ namespace heir {
...
-    {isl_ast_op_mul, "arith.muli"},     {isl_ast_op_div, "arith.divsi"},
-    {isl_ast_op_pdiv_r, "arith.remsi"}, {isl_ast_op_pdiv_q, "arith.divsi"},
-    {isl_ast_op_max, "arith.maxsi"},    {isl_ast_op_min, "arith.minsi"},
-    {isl_ast_op_and, "arith.andi"},     {isl_ast_op_or, "arith.ori"},
+    {isl_ast_op_add, "arith.addi"},          {isl_ast_op_sub, "arith.subi"},
+    {isl_ast_op_mul, "arith.muli"},          {isl_ast_op_div, "arith.divsi"},
+    {isl_ast_op_pdiv_r, "arith.remsi"},      {isl_ast_op_pdiv_q, "arith.divsi"},
+    {isl_ast_op_max, "arith.maxsi"},         {isl_ast_op_min, "arith.minsi"},
+    {isl_ast_op_and, "arith.andi"},          {isl_ast_op_or, "arith.ori"},
+    {isl_ast_op_fdiv_q, "arith.floordivsi"},
```
asraa, line 51: hah, i needed this one too for the conv_2d_nchw_fchw!

lib/Transforms/LayoutOptimization/InterfaceImpl.cpp
asraa: For some layouts there should be a hoister. if there's a result layout that happens to be a rotation of the bicyclic encoding, then I think that the inputs could also be rotated. but agree to file and we could work it out on paper.
but i also don't think that would happen organically, and it's more likely that we would be hoisting a row-major layout through a bicyclic encoding.

asraa (approved):

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
j2kun: Fixed

PR #2379, opened 2025-11-03
[WIP] Support high-precision scale in HEIR

j2kun:
warning: first commit is raw LLM output, needs lots of cleanup.


j2kun:
@AlexanderViand I think you had mentioned you had tried this as well? Is your draft cleaner than this old one? :)

AlexanderViand:
> @AlexanderViand I think you had mentioned you had tried this as well? Is your draft cleaner than this old one? :)

Maybe? IIRC, my version is even based on/inspired by your attempt here, but we'll see -  I'll focus on getting the Cheddar backend PR cleaned up this week and then we can have a look at the noise management changes next week :)

Issue #2258, opened 2025-09-24
Start using `notifyMatchFailure` to improve debugging/error messages in passes


AlexanderViand:
Instead of simple returning failure, a Pattern can actually provide a diagnostic that explains WHY it didn't match:
https://mlir.llvm.org/docs/PatternRewriter/#pattern-rewriter

(This was a note on my phone from the "better debugging" session with Matthias at the MLIR School in Cambridge, but I figured it deserves to be an actual issue)

PR #2382, opened 2025-11-05
Replace some uses of `return failure()` with `notifyMatchFailure`

j2kun:
Fixes https://github.com/google/heir/issues/2258

### DIFF
--- a/lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
@@ -152,7 +152,7 @@ struct RemoveKeyArg : public OpRewritePattern<func::FuncOp> {
     }
 
     if (argsToErase.none()) {
-      return failure();
+      return rewriter.notifyMatchFailure(op, "no key arguments to erase");
     }
 
     rewriter.modifyOpInPlace(op, [&] { (void)op.eraseArguments(argsToErase); });
@@ -442,7 +442,7 @@ struct ConvertRlweDecodeOp : public OpConversionPattern<DecodeOp> {
 
     auto zeroAttr = rewriter.getZeroAttr(outputTensorType);
     if (!zeroAttr) {
-      return op.emitOpError() << "Unsupported type for lowering";
+      return rewriter.notifyMatchFailure(op, "Unsupported type for lowering");
     }
     auto alloc =
         AllocOp::create(rewriter, op.getLoc(), outputTensorType, zeroAttr);
--- a/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
@@ -123,7 +123,8 @@ struct ConvertRLWEDecrypt : public OpConversionPattern<RLWEDecryptOp> {
       // TODO (#882): For TFHE, which can support higher dimensional keys,
       // plaintexts, and ciphertexts, we need to add support for encrypt and
       // decrypt for those cases.
-      return failure();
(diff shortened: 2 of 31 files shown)


lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
```diff
@@ -442,7 +442,7 @@ struct ConvertRlweDecodeOp : public OpConversionPattern<DecodeOp> {
 
     auto zeroAttr = rewriter.getZeroAttr(outputTensorType);
     if (!zeroAttr) {
-      return op.emitOpError() << "Unsupported type for lowering";
+      return rewriter.notifyMatchFailure(op, "Unsupported type for lowering");
```
asraa, line 445: not your fault or anything but I don't think getZeroAttr even returns a logicalresult / failure...

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -431,12 +435,12 @@ struct ConvertRMul : public OpConversionPattern<RMulOp> {
...
       op.emitError() << "`lwe.rmul` expects ciphertext as two polynomials, got "
                      << xT.getNumElements() << " and " << yT.getNumElements();
-      return failure();
+      return rewriter.notifyMatchFailure(op, "ciphertext not two polynomials");
     }
 
     if (xT.getElementType() != yT.getElementType()) {
       op->emitOpError() << "`lwe.rmul` expects operands of the same type";
-      return failure();
+      return rewriter.notifyMatchFailure(op, "operands not of same type");
```
asraa, line 442: this one and the above also call `emitOpError` or `emitError` and I think it's probably more worth kicking that out and keeping the more verbose notifyMatchFailure instead

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -1170,21 +1192,22 @@ struct ConvertNTT : public OpConversionPattern<NTTOp> {
       op.emitError(
           "Can't directly lower for a tensor of polynomials. "
           "First run --convert-elementwise-to-affine.");
-      return failure();
+      return rewriter.notifyMatchFailure(
```
asraa, line 1192: same here with the emitError

lib/Dialect/TensorExt/Transforms/CollapseInsertionChains.cpp
```diff
@@ -70,14 +70,15 @@ struct FoldSequentialExtractInsertIntoRotate
...
+      return rewriter.notifyMatchFailure(
+          insertOp, "insert op does not source from an extract op");
     }
 
     auto shiftRes = calculateShift(insertOp, extractOp);
     if (failed(shiftRes)) {
       LLVM_DEBUG(llvm::dbgs()
                  << "Insert op " << insertOp << " has incalculable shift\n");
-      return failure();
+      return rewriter.notifyMatchFailure(insertOp, "failed to calculate shift");
```
asraa, line 76: likewise here i think LLVM_DEBUG would be redundant - if someone is already debugging they'd see the match failure

lib/Dialect/LWE/Conversions/LWEToLattigo/LWEToLattigo.cpp
j2kun: Docs on `getZeroAttr`

```
  // Returns a 0-valued attribute of the given `type`. This function only
  // supports boolean, integer, and 16-/32-/64-bit float types, and vector or
  // ranked tensor of them. Returns null attribute otherwise.
```

I think the implementation here matches that correctly.

j2kun:
Addressed comments and squashed

Issue #2384, opened 2025-11-05
SecretInsertMgmtCKKS: quadratic/exponential scaling on IRs


asraa:
I was struggling to a small CNN model to compile overnight, so I ran two experiments.

Here's a reproducer for the 70 second compilation
```
#map = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
#map1 = affine_map<(d0, d1) -> (d0, d1)>
#map2 = affine_map<(d0, d1) -> (d1)>
module {
  func.func @main(%arg0: tensor<1x1x32x32xf32> {secret.secret}) -> tensor<1x10xf32> {
    %cst = arith.constant dense_resource<torch_tensor_2_torch.float32> : tensor<1xf32>
    %cst_0 = arith.constant 0.000000e+00 : f32
    %cst_1 = arith.constant 1.000000e+00 : f32
    %cst_2 = arith.constant 4.000000e+00 : f32
    %cst_3 = arith.constant dense_resource<torch_tensor_2_1_5_5_torch.float32> : tensor<1x1x5x5xf32>
    %cst_4 = arith.constant dense_resource<torch_tensor_10_392_torch.float32> : tensor<10x196xf32>
    %cst_5 = arith.constant dense_resource<torch_tensor_10_torch.float32> : tensor<10xf32>
    %0 = tensor.empty() : tensor<1x1x28x28xf32>
    %broadcasted = linalg.broadcast ins(%cst : tensor<1xf32>) outs(%0 : tensor<1x1x28x28xf32>) dimensions = [0, 2, 3] 
    %1 = linalg.conv_2d_nchw_fchw {dilations = dense<1> : vector<2xi64>, strides = dense<1> : vector<2xi64>} ins(%arg0, %cst_3 : tensor<1x1x32x32xf32>, tensor<1x1x5x5xf32>) outs(%broadcasted : tensor<1x1x28x28xf32>) -> tensor<1x1x28x28xf32>
    %2 = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%1 : tensor<1x1x28x28xf32>) outs(%0 : tensor<1x1x28x28xf32>) {
    ^bb0(%in: f32, %out: f32):
      %13 = arith.negf %in : f32
      %14 = math.exp %13 : f32
      %15 = arith.addf %14, %cst_1 : f32
      %16 = arith.divf %cst_1, %15 : f32
[...]

PR #2391, opened 2025-11-06
Address performance of secret-insert-mgmt passes

j2kun:
Improves the runtime of `secret-insert-mgmt-*` passes by avoiding the need to re-run the data flow solver after each pattern.

Instead of clearing all solver analysis states at the end of the pattern and re-running, it manually sets the relevant solver state of the result of the op under consideration. This works because the patterns are all run via `walkAndApplyPatterns`, so the next operation visited will have to have all its operands handled by previous steps in the walk. Then, at the end of each walk, I re-run the data flow analysis so the next walk has fresh data for everything.

This should be a nonfunctional change with regards to the logic of `insert-mgmt`. But to keep the manual state-setting code consistent across all the variants of this pass and other uses of the same analyses, I refactored the passes into a "pipeline" helper, and refactored the analysis visitor logic into standalone functions that could be called outside of the DataFlow framework.

After the changes, the execution time report for the reproducer in #2384 is

```
===-------------------------------------------------------------------------===
                         ... Execution time report ...
===-------------------------------------------------------------------------===
  Total Execution Time: 4.9360 seconds

  ----Wall Time----  ----Name----
    0.0127 (  0.3%)  Parser
    ...
    0.7847 ( 15.9%)  LayoutPropagation
    1.5731 ( 31.9%)  LayoutOptimization
    0.1531 (  3.1%)  ConvertToCiphertextSemantics
    0.0092 (  0.2%)  ApplyFolders
    0.6836 ( 13.8%)  Canonicalizer
    ...
    0.2633 (  5.3%)  SecretInsertMgmtCKKS
    ...
    0.7725 ( 15.7%)  SecretDistributeGeneric
    4.9360 (100.0%)  Total
```

Note that the test is still failing due to requiring https://github.com/google/heir/pull/2390

Fixes https://github.com/google/heir/issues/2384

### DIFF
--- a/lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
+++ b/lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
@@ -24,6 +24,47 @@
 namespace mlir {
 namespace heir {
 
+LogicalResult updateResultLevelLattice(Operation* op, DataFlowSolver* solver) {
+  // Here we update the analysis state of the result of the original op This
+  // implies any downstream users now have an invalidated state in the data
+  // flow solver, so this is where we are requiring walkAndApplyPatterns for
+  // the use of any pattern that calls this function.
+  SmallVector<const LevelLattice*, 2> operandLattices;
+  for (auto operand : op->getOperands()) {
+    operandLattices.push_back(solver->lookupState<LevelLattice>(operand));
+  }
+
+  if (!op->getResults().empty()) {
+    for (auto result : op->getResults()) {
+      FailureOr<int64_t> resultLevel = deriveResultLevel(op, operandLattices);
+      auto* resultLattice = solver->getOrCreateState<LevelLattice>(result);
+      resultLattice->getValue().setLevel(resultLevel.value());
+    }
+  }
+
+  return success();
+}
+
+LogicalResult updateResultMulDepthLattice(Operation* op,
+                                          DataFlowSolver* solver) {
+  // Same warning as updateResultLevelLattice
+  SmallVector<const MulDepthLattice*, 2> operandLattices;
+  for (auto operand : op->getOperands()) {
+    operandLattices.push_back(solver->lookupState<MulDepthLattice>(operand));
+  }
+
+  if (!op->getResults().empty()) {
+    for (auto result : op->getResults()) {
+      FailureOr<int64_t> resultLevel =
+          deriveResultMulDepth(op, operandLattices);
+      auto* resultLattice = solver->getOrCreateState<MulDepthLattice>(result);
+      resultLattice->getValue().setMulDepth(resultLevel.value());
+    }
+  }
+
+  return success();
+}
+
 template <typename MulOp>
 LogicalResult MultRelinearize<MulOp>::matchAndRewrite(
     MulOp mulOp, PatternRewriter& rewriter) const {
@@ -45,9 +86,7 @@ LogicalResult MultRelinearize<MulOp>::matchAndRewrite(
   auto relinearized =
       mgmt::RelinearizeOp::create(rewriter, mulOp.getLoc(), result);
   result.replaceAllUsesExcept(relinearized, {relinearized});
-
-  solver->eraseAllStates();
-  return solver->initializeAndRun(top);
+  return success();
 }
 
 template <typename MulOp>
@@ -62,9 +101,7 @@ LogicalResult ModReduceAfterMult<MulOp>::matchAndRewrite(
   rewriter.setInsertionPointAfter(mulOp);
   auto modReduced = mgmt::ModReduceOp::create(rewriter, mulOp.getLoc(), result);
   result.replaceAllUsesExcept(modReduced, {modReduced});
-
-  solver->eraseAllStates();
-  return solver->initializeAndRun(top);
+  return success();
 }
 
 template <typename Op>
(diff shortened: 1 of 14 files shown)


lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
```diff
@@ -24,6 +24,47 @@
...
+  if (!op->getResults().empty()) {
+    for (auto result : op->getResults()) {
+      FailureOr<int64_t> resultLevel = deriveResultLevel(op, operandLattices);
+      auto* resultLattice = solver->getOrCreateState<LevelLattice>(result);
+      resultLattice->getValue().setLevel(resultLevel.value());
+    }
+  }
+
+  return success();
+}
```
ZenithalHourlyRate, line 46: OK I found the `rerunDataflow()`. I think it is this function that needs comment. It is not we are using greedy or walk, it is how we group the patterns. Let's say, "ModReduceBefore must be used inside a single PatternSet (with careful companion), and `rerunDataflow()` must be called after such PatternSet to correctly update the analysis state and further patterns that depends on such pattern can be applied."

<summary>Click to see reasoning</summary>
<details>
I am afraid this will results in ill analysis state

```mlir
%a = arith.muli %input1, %input2
%b = arith.muli %a, %a
%c = arith.addi %b, %b
%d = arith.muli %c, %input2
```
ModReduceBefore will first insert `mgmt.mod_reduce` before the second mul, and the MulDepth/LevelLattice analysis state for `%b` will be updated. However, state for `%c` will not be updated and when MatchCrossLevel visits the third mul, it will find invalid Level state.

For functionality, the ModReduceBefore code only cares about MulDepth being 0 or not so it is OK; but for MatchCrossLevel/MatchCrossMulDepth, this is not OK. I assume the code works now because some other solver re-running code are present.

The whole reason for re-run the solver is that there might be some intermediate op between two target op and the update-result mechanism wont propagate that far. This is also the bug I used to encounter with SecretnessAnalysis, where secret state is passed to results but not the grand-children and results in ill-behavior.

</details>

ZenithalHourlyRate: Thinking further, if we group patterns, these `updateResultLevelLattice` are not needed anyway in ModReduceBefore as it does not rely on updated states. It is those complex patterns like `BootstrapWaterline` that needs such complicated updating, and they happen infrequently.

MatchCrossLevel/MulDepth also does not change its result lattices, so they do not need update.

OK. Now I vote for deleting the updating analysis state logic in `ModReduceAfterMult`, `ModReduceBefore`, `MultRelinearize` and `MatchCrossLevel/MulDepth` completely, as they do not rely on updated state. Only leaving those `rerunDataflow` and the one inside `BootstrapWaterline`.

ZenithalHourlyRate (approved):
I am OK with this approach and I think it can be further simplified.

j2kun:
Ah, test is still failing now because of the need to lower `divf`. @asraa you may want to patch in the changes in the mean time?

j2kun:
I had originally thought there was a problem because the regression test times out in `dbg` mode, but on closer inspection, while the insert-mgmt pass still takes about 5s, LayoutOptimization takes 20s, Canonicalizer 12s, and others contribute more to the total. Most of the runtime of insert-mgmt appears to also be in the canonicalizer cleanup at the end, so I think it might be worthwhile to spend some time making the canonicalizer more efficient. Either way, it shouldn't block the submission of this PR.

```
$ bazel run -c dbg //tools:heir-opt -- --torch-linalg-to-ckks $PWD/tests/Regression/issue_2384.mlir --mlir-timing
===-------------------------------------------------------------------------===
                         ... Execution time report ...
===-------------------------------------------------------------------------===
  Total Execution Time: 77.2324 seconds

  ----Wall Time----  ----Name----
    0.0066 (  0.0%)  Parser
...
    6.4050 (  8.3%)  LayoutPropagation
   20.5750 ( 26.6%)  LayoutOptimization
    0.0171 (  0.0%)  CSE
    0.0000 (  0.0%)    (A) DominanceInfo
    1.0289 (  1.3%)  ConvertToCiphertextSemantics
    0.3306 (  0.4%)  ApplyFolders
   12.7498 ( 16.5%)  Canonicalizer
    0.1459 (  0.2%)  ImplementRotateAndReduce
    0.9955 (  1.3%)  ImplementShiftNetwork
    0.5405 (  0.7%)  Canonicalizer
    0.2382 (  0.3%)  CSE
...
    5.3456 (  6.9%)  SecretInsertMgmtCKKS
    7.5118 (  9.7%)  'builtin.module' Pipeline
    0.6535 (  0.8%)    Canonicalizer
    0.2746 (  0.4%)    CSE
    0.0000 (  0.0%)      (A) DominanceInfo
    1.1361 (  1.5%)    AnnotateMgmt
    2.3118 (  3.0%)    AnnotateMgmt
    0.5862 (  0.8%)    Canonicalizer
    0.0937 (  0.1%)    CSE
    0.0000 (  0.0%)      (A) DominanceInfo
    3.4163 (  4.4%)  OptimizeRelinearization
    0.7028 (  0.9%)  GenerateParamCKKS
    3.6221 (  4.7%)  PopulateScaleCKKS
    0.3589 (  0.5%)  Canonicalizer
    4.3232 (  5.6%)  SecretDistributeGeneric
    1.0501 (  1.4%)  Canonicalizer
    0.3271 (  0.4%)  SecretToCKKS
    2.1516 (  2.8%)  Rest
   77.2324 (100.0%)  Total
```

lib/Transforms/SecretInsertMgmt/SecretInsertMgmtPatterns.cpp
j2kun: The general pattern I'm aiming for here is: any pattern which queries the lattice for a particular state, must ensure that the op result lattice states are updated according to the data flow analysis.

In a case like `ModReduceBefore`, I see it querying the level state, so I assume conservatively that any later application of the same pattern in the same walk may need an updated level state for the result of that op.

However, now I notice that, actually, the LevelState is not used. It computes a maxLevel across operands and then never uses that variable. So I think I can indeed remove it.

asraa (approved):

Issue #2364, opened 2025-10-30
Support high-precision scale management


j2kun:
Some techniques for managing scale, for example Algorithm 2 of [BMPH20](https://eprint.iacr.org/2020/1203) which powers the scale management in Lattigo and is particularly important for their bootsrtapping algorithm, require full precision on the scaling factor being analyzed. In HEIR currently we only support analyzing scaling in the log2 domain.

I ran into this issue when trying to reproduce Lattigo's scale management for a particular lowering of chebyshev polynomial evaluation in https://github.com/google/heir/pull/2338, and also it showed up as a todo in #1640

This issue is to replace the log2-based scaling factors with APInt scaling factors. That will unlock a more precise scale management system for future work.

j2kun:
One thing I've determined, at least, is that we won't be able to reproduce lattigo's chebyshev polynomial evaluation exactly without high-precision scale analysis. This is because it requires setting the scale of plaintexts to quantities involving the particular primes used in the modulus chain (see section 3.2 of https://eprint.iacr.org/2020/1203).

Issue #1640, opened 2025-04-02
Implement parameter generation of "Approximate Homomorphic Encryption with Reduced Approximation Error"


ZenithalHourlyRate:
[Approximate Homomorphic Encryption with Reduced Approximation Error](https://ia.cr/2020/1118) introduced a way to reduce the rescaling error in CKKS via special parameter generation process.

This is somewhat related to #1169 but worth a separate issue for TODO tracking.

* [ ] Introduce the modulus selection algorithm in `generate-param-ckks`
* [ ] Modify scaling factor in LWE type to support high-precision scaling factor tracking, as we currently only tracks logScale (int) for CKKS for the ease of implementation in #1459

PR #2402, opened 2025-11-10
CKKS: Use prime selection logic from OpenFHE

ZenithalHourlyRate:
The prime selection logic from #1400 for CKKS just uses the same prime selection logic as BGV, and might cause some severe scaling factor mismatch issue / rescaling noise due to the nature of RNS-CKKS instantiation.

To better mitigate the issue, CKKS should have its own prime selection logic, and OpenFHE is the standard reference.

This also helps #2364 and #2379. I believe with this PR, the high-precision level-specific scaling factor in `populate-scale-ckks` can be safely instantiated after the "reduced error" prime selection logic in `generate-param-ckks`.

Relates to #1640 but that issue has lots of TODO related to high precision scaling factor so it can not be closed.

For @j2kun: Can the specific issue in #2364 be reproduced with these prime selection logic?

### DIFF
--- a/lib/Parameters/CKKS/Params.cpp
+++ b/lib/Parameters/CKKS/Params.cpp
@@ -5,20 +5,230 @@
 #include <vector>
 
 #include "lib/Parameters/RLWEParams.h"
+#include "lib/Parameters/RLWESecurityParams.h"
 #include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+#include "src/core/include/openfhecore.h"           // from @openfhe
 
 namespace mlir {
 namespace heir {
 namespace ckks {
 
+/// By original we mean the method in RNS-CKKS implementation
+/// Corresponds to FIXED* in OpenFHE
+static std::vector<int64_t> moduliQGenerationOpenFHEFixed(int logFirstMod,
+                                                          int logDefaultScale,
+                                                          int numLevel,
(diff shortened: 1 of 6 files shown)


lib/Parameters/CKKS/Params.cpp
```diff
@@ -5,20 +5,230 @@
...
 #include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
+#include "src/core/include/openfhecore.h"           // from @openfhe
 
 namespace mlir {
 namespace heir {
 namespace ckks {
 
+/// By original we mean the method in RNS-CKKS implementation
+/// Corresponds to FIXED* in OpenFHE
+static std::vector<int64_t> moduliQGenerationOriginal(int logFirstMod,
```
j2kun: maybe just call it `moduliQGenerationOpenFHEFixed`?

j2kun (approved):
LGTM!

j2kun:
> For @j2kun: Can the specific issue in #2364 be reproduced with these prime selection logic?

The problem in that situation was that the parameters were selected before HEIR, so I don't get to select parameters.

lib/Parameters/CKKS/Params.cpp
ZenithalHourlyRate: Name changed.

PR #2411, opened 2025-11-13
Improve Chebyshev Paterson-Stockmeyer evaluation

j2kun:
This PR improve the Chebyshev Paterson-Stockmeyer lowering by introducing some key changes present in the lattigo implementation that were missing from HEIR.

1. A more precise calculation of the BSGS splitting factor `k`, which results in 1 less mul depth when n is close to a power of two.
2. A more balanced recursive calculation of the baby step polynomials T_k(x)
3. A more balanced evaluation of the giant steps

### DIFF
--- a/lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
+++ b/lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
@@ -2,6 +2,7 @@
 #define LIB_UTILS_POLYNOMIAL_CHEBYSHEVPATERSONSTOCKMEYER_H_
 
 #include <algorithm>
+#include <bit>
 #include <cmath>
 #include <cstdint>
 #include <memory>
@@ -10,7 +11,6 @@
 
 #include "lib/Kernel/AbstractValue.h"
 #include "lib/Kernel/ArithmeticDag.h"
-#include "lib/Kernel/KernelImplementation.h"
 #include "lib/Utils/Polynomial/ChebyshevDecomposition.h"
 #include "llvm/include/llvm/ADT/ArrayRef.h"  // from @llvm-project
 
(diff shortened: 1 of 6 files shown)


j2kun:
@asraa as per our discussion: how much justification do you think I should provide for this PR? The reference repository is https://github.com/j2kun/bsgs and you can run `python compare_circuits.py` to show the examples that are equivalent to lattigo.

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
```diff
@@ -22,6 +22,65 @@ namespace polynomial {
...
+// A C++ port of Lattigo's `bignum.OptimalSplit` function.
+// This function determines the optimal split point for the Paterson-Stockmeyer
+// algorithm by calculating the number of multiplications for two possible
+// split points and choosing the better one.
+//
+// Args:
+//   logDegree: ceil(log2(degree)) of the polynomial.
+//
+// Returns:
+//   The optimal split parameter (log scale).
```
asraa:
```suggestion
//   The optimal split parameter (log split).
```
i'm just not sure what log scale meant

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
```diff
@@ -22,6 +22,65 @@ namespace polynomial {
...
+    // [Lee et al. 2020] : Maximize the number of odd terms of Chebyshev basis
+    // Find k = floor(log2(n-1))
+    int64_t k = 0;
+    int64_t temp = n - 1;
+    while (temp > 1) {
+      temp >>= 1;
+      k++;
+    }
+    int64_t a = (1LL << k) - 1;
+    int64_t b = n + 1 - (1LL << k);
```
asraa:
```suggestion
    int64_t b = n - a;
```

for slightly better readability in the choice for `b` is just the one that makes sense

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
```diff
@@ -22,6 +22,65 @@ namespace polynomial {
...
+//   logDegree: ceil(log2(degree)) of the polynomial.
+//
+// Returns:
+//   The optimal split parameter (log scale).
+inline int64_t optimalSplit(int64_t logDegree) {
+  int64_t logSplit = logDegree >> 1;
+  int64_t a = (1LL << logSplit) + (1LL << (logDegree - logSplit)) + logDegree -
+              logSplit - 3;
+  int64_t b = (1LL << (logSplit + 1)) + (1LL << (logDegree - logSplit - 1)) +
+              logDegree - logSplit - 4;
```
asraa, line 51: I see that this is just a direct port of lattigo, but is there any rationale? At least I don't really understand whats going on

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
```diff
@@ -46,6 +105,102 @@ std::vector<std::shared_ptr<kernel::ArithmeticDagNode<T>>> computePowers(
...
+//   Node representing T_n(x)
+template <typename T>
+std::shared_ptr<kernel::ArithmeticDagNode<T>> genChebyshevPowerRecursive(
+    std::shared_ptr<kernel::ArithmeticDagNode<T>> x, int64_t n,
+    std::map<int64_t, std::shared_ptr<kernel::ArithmeticDagNode<T>>>& cache) {
+  using NodeTy = kernel::ArithmeticDagNode<T>;
+
+  if (n == 0) {
+    // T_0(x) = 1
+    return NodeTy::constantScalar(1);
```
asraa: super minor but wouldn't grabbing this from the cache also minimize the # of constant 1's in the IR?

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
```diff
@@ -98,55 +253,87 @@ patersonStockmeyerChebyshevPolynomialEvaluation(
...
-      result = NodeTy::add(result, pol);
-    } else {
-      result = pol;
+    babySteps.push_back(pol);
+  }
+
+  auto y = chebPolynomialValuesMap[k];
+  auto yPower = y;
+
+  // Combine baby steps in tree-like manner
```
asraa: I think this is the only place that I'd want a little more explanation - maybe a closed form summation or something to reference since it's a bit hard to see what's going on here.

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
j2kun: Done

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
j2kun: I don't have a perfect explanation but let me investigate a bit. I think it's computing the size of the reduction tree for the giant steps as well as the computation tree for the baby steps, but I don't know why it's offsetting those by 3 and 4.

lib/Utils/Polynomial/ChebyshevPatersonStockmeyer.h
j2kun: Done, and simplified the code a bit.

Issue #2326, opened 2025-10-17
Infrastructure for comparisons with other compilers


j2kun:
We're starting up an effort to compare HEIR with other research compilers, and this ticket is meant to track infrastructure work related to that task, as well as be a place for discussion on the topic.

We want to compare with a few compilers in the literature, but for the sake of this ticket I will use Orion as the example.

There is an existing project thanks to @eymay that provides an initial Orion -> HEIR translation step, emitting CKKS dialect code (with some new ops needed by Orion but that the `ckks` dialect doesn't have). There are a few changes that need to be made to that project, which I will fork in https://github.com/j2kun/orion_heir_translator, but the general outline I want to establish for each compiler that wants to act as a frontend for HEIR:

1. Add a new entry dialect in HEIR that mirrors the output API of the frontend compiler. In our case it would be an `orion` dialect that includes a `linear_transform` operation and a `chebyshev` operation applied to ciphertexts.
2. Add a conversion pass that lowers from the entry dialect to the relevant HEIR dialect. For `orion` we'd lower to `ckks` dialect and would have to EITHER:
    1. Implement `linear_transform` and `chebyshev` using existing HEIR methods, extracting the relevant subset of code to helpers or using interfaces. The downside of this method is that `linear_transform` would only get single-hoisting optimizations, as double-hoisting requires the backend lib to expose it (Lattigo does, OpenFHE does not) and HEIR won't have it until we finish implementing the crypto in the compiler.
    2. Add separate ops in the chosen exit dialect (e.g., `lattigo.chebyshev`) and lower directly to those intsead of an intermediate `ckks` op. I am leaning against this option because (a) not all backends have APIs for these, and (b) this would preclude is from lowering from `orion` to `polynomial` when the crypto is implemented.
3. Lower from `ckks` further as normal.

So as the first part of this, I will add an `orion` dialect and lowerings.

PR #2412, opened 2025-11-14
Add type inference for LWE/CKKS ops

j2kun:
This came up in the conversion from Orion to CKKS (https://github.com/google/heir/issues/2326). That line of work requires me to reimplement chebyshev evaluation and the halevi-shoup kernel, but at the level of CKKS with parameters pre-selected.

https://github.com/google/heir/pull/2338 is addressing the finer points of that conversion, but in the mean time it has proven helpful to extend the result type inference capabilities of the LWE/CKKS dialects, because when I insert ciphertext mgmt ops at the CKKS level, I need to re-walk the IR and re-infer result types for all ops.

Yes, I know this is what @ZenithalHourlyRate warned against and why we have mgmt at the higher level. Normally I wouldn't want to do this either, but for comparing against another compiler that handles ciphertext management and parameter selection itself, I don't see another way :)

### DIFF
--- a/lib/Dialect/CKKS/IR/CKKSOps.cpp
+++ b/lib/Dialect/CKKS/IR/CKKSOps.cpp
@@ -79,6 +79,12 @@ LogicalResult RelinearizeOp::inferReturnTypes(
   return lwe::inferRelinearizeOpReturnTypes(ctx, adaptor, inferredReturnTypes);
 }
 
+LogicalResult LevelReduceOp::inferReturnTypes(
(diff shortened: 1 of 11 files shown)


j2kun:
For the reviewer: I'm looking for some extra scrutiny on the logic of the type inference, to make sure I'm not incorrectly dropping limbs or dividing by the wrong moduli.

ZenithalHourlyRate (approved):
LGTM!

PR #2425, opened 2025-11-19
add openfhe benchmark binary

j2kun:
Run with `bazel run -c opt //benchmark:openfhe_benchmark`


benchmark/openfhe_benchmark.cpp
```diff
@@ -0,0 +1,1074 @@
...
+//==================================================================================
+
+/*
+ * Main OpenFHE library benchmark that contains performance tests for standard
+ * operations in the following schemes: BFVrns, CKKSrns, BGVrns. It also
+ * contains several performance tests for NTT and INTT transformations.
+ */
+
+// HEIR note: this is ported from OpenFHE in order to ensure that our bazel
+// build produces similar timing numbers as the upstream OpenFHE cmake build.
```
asraa, line 42: Could you add a direct link to https://github.com/openfheorg/openfhe-development/blob/aa391988d354d4360f390f223a90e0d1b98839d7/benchmark/src/lib-benchmark.cpp#L33?

asraa (approved):
Looking at this, my guess is that in our compiled example our crypto params are just too different? I'm just curious about whether the primes are < 32 bits

benchmark/openfhe_benchmark.cpp
j2kun: Done

PR #2431, opened 2025-11-21
Convert orion to lattigo

j2kun:
This includes an e2e test of `linear_transform` and `chebyshev` ops in isolation, as well as an e2e execution (though not correctness test) of the mlp output of orion.

Note: the MLP example takes 2.5s in dbg mode, and 493ms in opt mode, which matches Orion's runtime when run manually. However, the one thing I couldn't figure out in this PR is how to get Lattigo to automatically select the right rotation keys based on the parameters of the linear transform struct. It seemed to still require me to manually generate all 127 rotation keys...


lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
```diff
@@ -274,4 +274,87 @@ def Lattigo_CKKSRotateOp : Lattigo_CKKSUnaryInplaceOp<"rotate"> {
...
+    4. Evaluate the transformation on the input ciphertext
+  }];
+  let arguments = (ins
+    Lattigo_CKKSEvaluator:$evaluator,
+    Lattigo_CKKSEncoder:$encoder,
+    Lattigo_RLWECiphertext:$input,
+    // Parameters corresponding to the lattigo lintrans.Parameters struct,
+    // and those that can be inferred from the IR are commented out.
+
+    RankedTensorOf<[AnyFloat]>:$diagonals,
```
asraa: You could also use 2DTensorOf<[AnyFloat]> to restrict to 2-D

lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
```diff
@@ -274,4 +274,87 @@ def Lattigo_CKKSRotateOp : Lattigo_CKKSUnaryInplaceOp<"rotate"> {
...
+    DenseI32ArrayAttr:$diagonal_indices,
+    Builtin_IntegerAttr:$levelQ,
+
+    // The same auxiliary prime used to generate evaluation keys
```
asraa, line 319: Was this left here to show what else is in this lattigo op but not needed for our op since we grab this from params?

j2kun: Yeah this is me just kind of documenting the process. Commented to clarify.

PR #2432, opened 2025-11-21
add openfhe ckks rotom chained_matmul test

edwjchen:
Adds an end-to-end test for chained matrix multiplication (a @ b @ c) where a is an encrypted input matrix and b, c are plaintext matrices. All matrices have shape 64x64 with n=4096.

The layout assignment was automatically generated by Rotom.


j2kun (changes requested):
Let's come up with a better approach for managing the test data.

(1) Is it possible to have one argument that stores all the packed plaintexts, e.g., a `tensor<400x4096>` instead of ~400 packed cleartext arguments? This would require adding `tensor.extract_slice` operations to the IR to get an individual packed cleartext.

(2) Would it be possible to use a compressed format for the input/output tensors? I have multiple ideas for what we could use, but I wonder first what data format you're using in rotom. If it's Python based (e.g., a torch tensor) you can follow [mnist_test](https://github.com/google/heir/blob/main/tests/Examples/openfhe/ckks/mnist/mnist_test.py) and do it all in python (but let me know what data format you'd like to use: npz? HDF5?). Otherwise—and this would be a good option in any case—you could define a very simple protocol buffer spec for the data and then compress it with zstd and check in the compressed binaryproto. I can help you with this if you're unfamiliar with protobufs. It'd be something like:

```proto
syntax = "proto3";

message Matrix {
  uint32 rows = 1;
  uint32 cols = 2;
  repeated double values = 3 [packed = true];  // row-major order, rows * cols elements
}
```

edwjchen:
@j2kun updated to use batched inputs! I also decided on using numpy's compressed format, since numpy is already supported dependency in Rotom. happy to switch to torch or protobuf if we want to keep the data serialization consistent with the mnist test.

edwjchen:
pre-committed, squashed, and moved the test description to be a docstring under `class ChainedMatmulTest`!

j2kun:
Ah, sorry, looks like it's still two commits. Can you squash it? Otherwise LGTM!

edwjchen:
ah oops, squashed!

j2kun:
Looks like everyone else is off for the holiday weekend, so this will probably not get merged until Monday.

PR #2445, opened 2025-12-01
OpenFHE interpreter overhead reduction (lenet example)

j2kun:
This PR improves the OpenFHE C++ interpreter by reducing the overhead of the interpreter. The main issues were:

- Copying ciphertexts and plaintexts a lot:
  - when reading from the environment or storing to the environment
  - when constructing a TypedCppValue or converting from a TypedCppValue to a specific variant
  - whenever handling a tensor op
- The overhead of having large variants with many options

This change introduces a few things to deal with this:

- Using `shared_ptr` to store values (and in TypedCppValue) which avoids copying
- Using std::move in the context of `insert_or_assign`
- Using a set of type-specific `DenseMap` for each type considered (e.g., `llvm::DenseMap<Value, std::shared_ptr<std::vector<int>>> intVectors` instead of having a single map that stores all values.
- Creating a static `operationDispatchTable` that avoids the overhead of type switching across all ops at visit time (in favor of hashing the TypeID associated with an Operation subclass)
- Checking if a tensor op's destination tensor is not used after, and does an in-place mutation of the tensor instead of creating a copy. This was a big part of the overhead of the packing loops, which run an ISL loop and insert a single element in the innermost body (creating n^2 or more copies to pack a single plaintext)

This also includes a patch of https://github.com/google/heir/pull/2421 in order to get the `lenet` example, but changes the `lenet_main.cpp` to take any file as input (so I could pass in truncated lenet openfhe IR for faster iteration).

To do the profiling:

```bash
# build in opt mode with debug symbols enabled
# (note for smaller examples I disable openmp to make the report cleaner,
# but the full lenet example takes a long time in this mode)
bazel build \
-c opt --copt=-g --linkopt=-lprofiler \
tests/Examples/openfhe/ckks/lenet:lenet_binary

# Run the binary and generate the profile
CPUPROFILE=prof.out bazel-bin/tests/Examples/openfhe/ckks/lenet/lenet_binary \
bazel-bin/tests/Examples/openfhe/ckks/lenet/lenet_heir_opt.mlir

# Convert the profile dump to a readable format
pprof --list=mlir::heir::openfhe::Interpreter::visit \
./bazel-bin/tests/Examples/openfhe/ckks/lenet/lenet_binary \
prof.out > profile.txt
```

This will produce a per-line runtime breakdown of the visit operations of the interpreter

Another view is

```bash
pprof --text --lines --focus=mlir::heir::openfhe::Interpreter \
./bazel-bin/tests/Examples/openfhe/ckks/lenet/lenet_binary \
prof.out > interpreter_focus.txt
```

Which gives a table of all runtime data for functions in the Interpreter namespace.

Running the above, this PR produces:

```
Interpreting function: lenet__generate_crypto_context ...
lenet__generate_crypto_context time: 0 seconds
Interpreting function: lenet__configure_crypto_context ...
Function returned 1 values
lenet__configure_crypto_context time: 18 seconds
Interpreting function: lenet__encrypt__arg0 ...
lenet__encrypt__arg0 time: 1 seconds
Interpreting function: lenet ...
lenet time: 91 seconds
Interpreting function: lenet__decrypt__result0 ...
lenet__decrypt__result0 time: 0 seconds
PROFILE: interrupts/evictions/bytes = 39553/8234/1086088
```

Profile: https://gist.github.com/j2kun/4ff4c2c9740b7f6cf3f3f5f370fda5b5
line-by-line annotation: https://gist.github.com/j2kun/7110641e3eaec18b3b8e6ec948e63c07

From the above profiles, it's clear that the overhead of dispatching and copying is eliminated, and the only remaining source of overhead I can see is the liveness checking, which is only 10ms.


lib/Target/OpenFhePke/Interpreter.cpp
```diff
@@ -137,9 +112,268 @@ using PrivateKeyT = PrivateKey<DCRTPoly>;
...
+        } else if constexpr (std::is_same_v<T, FastRotPrecompT>) {
+          fastRotPrecomps[v] = arg;
+        }
+      },
+      typedVal.value);
+}
+
+// Only used for type-agnostic block terminators (return, yield)
+TypedCppValue Interpreter::loadTypedValue(Value v) {
+  // Try each storage type
```
asraa: it seems a little rough to search through each storage type - and in `eraseValue` there is also a switch on MLIR type -> storage map. Could there by a map from MLIR type to a pointer to the map that you could use so you can do something like `mlirTypeToContainer[v.getType()]->at(v)` and then in `eraseValue` it can also to `mlirTypeToContainer[v.getType()]->erase(v)`?
Or is that indirection more overhead?

lib/Target/OpenFhePke/Interpreter.cpp
```diff
@@ -137,9 +112,268 @@ using PrivateKeyT = PrivateKey<DCRTPoly>;
...
+      .Case<openfhe::CryptoContextType>(
+          [&](auto ty) { cryptoContexts.erase(v); })
+      .Default([&](Type type) {
+        llvm::errs() << "Unsupported type " << type << " in eraseValue\n";
+      });
+}
+
+// Only used for type-agnostic block arguments (func args, iter args, etc.)
+void Interpreter::storeTypedValue(Value v, const TypedCppValue& typedVal) {
+  std::visit(
```
asraa, line 266: i know you mentioned type switching on the MLIR value is faster, so is it faster here to do an MLIR type switch to retrieve the right map and then assign it to the variant TypedCppValue?

lib/Target/OpenFhePke/Interpreter.cpp
```diff
@@ -845,86 +1127,88 @@ void Interpreter::visit(tensor::InsertSliceOp op) {
     return flatDestIndex;
   };
 
-  if (std::holds_alternative<std::vector<int>>(dest.value)) {
-    auto result = std::get<std::vector<int>>(dest.value);
-    const auto& srcVec = std::get<std::vector<int>>(source.value);
+  bool canModifyInPlace = liveness.isDeadAfter(op.getDest(), op);
```
asraa, line 1168: nice!!!

asraa (commented):
thank you!

lib/Target/OpenFhePke/Interpreter.cpp
j2kun: It only shows up as a cumulative 10ms in the profile (10ms for loadTypedValue, and 10ms for eraseValue, a total of 0.02%), which I think is sufficiently small to ignore.

j2kun: Eh, I'll just change it to a TypeSwitch. That should be straightforward

lib/Target/OpenFhePke/Interpreter.cpp
j2kun: The use of `constexpr` here means that the types are evaluated at compile time, so the only overhead comes from the variant (not the if statement). I'm not sure how this compares to a TypeSwitch...

PR #2458, opened 2025-12-09
Optimisation: Extend the Lut generation to LUT4 elements

WoutLegiest:
We can make the LUT CGGI pipeline more efficient by using 4 elements LUTs instead of 3 elements. Since each LUT only produces a single bit output, and the whole program only consits of 4-to-1 LUT, we cannot overflow a 4b ciphertext.
In the case we use TFHE-rs as backend, this will leads to a more efficient use of PBS (need to check the impact on the Jaxite backend). 

For now, would we like to dual support of lut3 and lut4 elements? For respectively, the jaxite and TFHE-rs backends?


j2kun:
Lut4 by default sounds great, but please make the lut width configurable on the yosys optimizer sub-pipeline via pass option. Maybe it makes sense to just extend the `mode` pass option to split `LUT` into `LUT3` and `LUT4`?

IIUC if we can limit the yosys optimizer pipeline to generate luts of a given size by configuration, then the rest of the pipeline doesn't need any special configuration and can just lower whatever luts it sees/supports at that stage. I.e., you don't need to touch the jaxite emitter and just let it fail on lut4, and secret-to-cggi can have both lut3 and lut4 patterns added without conflict.

I'm not too worried about having stuff in HEIR that `jaxite_bool` doesn't support (though adding support in jaxite_bool should be easy). But I know there are other groups (e.g., Cornami) that use the CGGI pipeline and may have limited support for certain LUT sizes.

j2kun:
Also, a long time ago when I was working on lut support in the yosys optimizer, I did some tests to see what was the best tradeoff of lut size vs runtime, and I recall lut5/lut6 being a net negative because you needed too large crypto parameters to support them, and the incremental reduction in circuit size didn't outweigh the added cost of the PBS. But lut3/lut4 were basically the same: lut4 took longer per PBS, but had a smaller circuit, and for the circuits I tried the runtime was similar.

At the time, my tfhe-rs CPU experiments put 3-bit PBS at 15.8ms, up to lut6 at 145ms, while circuit size reduction looked like this (for one example circuit), so 10x runtime for < 0.5x circuit reduction was pretty bad. I don't seem to have the exact lut4 numbers recorded from that experiment, though.

image

lib/Transforms/YosysOptimizer/yosys/techmap_lut4.v
j2kun, line 1: I'd recommend grepping the entire repo (including the .github/workflows) for instances of `techmap.v` to ensure that this new techmap file is included in all places.

For example, it's included in the release workflows like https://github.com/google/heir/blob/nightly/.github/workflows/nightly.yml

WoutLegiest: I've refactored the whole repo. Now, by default the LUT3 optimisation will be chosen.

WoutLegiest:
Ooh, very cool to see the past research! Is just thought about the ability to run on FPGAs, and would be a 'nice to have' as an appendix somewhere to write about the FPGA performance of the current CGGI pipelines (and mostly to show how much improvement we can do in the future).

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
```diff
@@ -81,18 +81,34 @@ using std::string;
...
+techmap -map {2}/techmap_lut3.v; opt_clean -purge;
+hierarchy -generate * o:Y i:*; opt; opt_clean -purge;
+clean;
+stat;
+)";
+constexpr std::string_view kYosysLut4Template = R"(
+read_verilog -sv {0};
+hierarchy -check -top \{1};
+proc; memory; stat;
+techmap -map {2}/techmap_lut4.v; stat;
```
asraa, line 104: i can't remember why but the old transpiler had all the lut cell definitions into this techmap.v and then later the abc -4 pass would remap and optimize with lut 4 cells. i think it may be possible to do that and still maintain just one techmap.v file, but i can experiment after as well

scripts/jupyter/heir_play/__init__.py
```diff
@@ -19,6 +19,6 @@ def load_ipython_extension(ipython):
       )
   )
   load_nightly("abc")
-  load_nightly("techmap.v")
+  load_nightly("techmap_lut3.v")
```
asraa, line 22: also include the lut4 - even if 3 is the default someone might pass lut4 into the optimizer pass

tools/heir-jaxite.bzl
```diff
@@ -30,7 +30,7 @@ def fhe_jaxite_lib(name, mlir_src, heir_opt_pass_flags = [], py_lib_target_name
             pass_flags = heir_opt_pass_flags,
             generated_filename = generated_heir_opt_name,
             HEIR_YOSYS = True,
-            data = ["@heir//lib/Transforms/YosysOptimizer/yosys:techmap.v"],
+            data = ["@heir//lib/Transforms/YosysOptimizer/yosys:techmap_lut3.v"],
```
asraa, line 34:
```suggestion
            # HEIR jaxite only supports lut 3 operations
            data = ["@heir//lib/Transforms/YosysOptimizer/yosys:techmap_lut3.v"],
```

lib/Transforms/YosysOptimizer/YosysOptimizer.cpp
WoutLegiest: Just did a small experiment and removed the `_tjc_lut3` from the techmap and now these is a small bug in the LUTImporter::getInputs function. Probably not so hard to fix, but then you shift the techmap set to the compiler i would think

scripts/jupyter/heir_play/__init__.py
WoutLegiest: Check, thanks

j2kun:
Not sure what happened to this PR. Going to try to re-import it to copybara

PR #2470, opened 2025-12-16
migrate pocketfft and cereal to MODULE.bazel

j2kun:
With this PR the last remaining thing to put on the BCR is openfhe (and wait for rules_hdl).

Note that one pesky change is that when the modules are on the BCR, if there are multiple versions in your local bazel cache, then bazel may not have the same path aliases, and will add a `+` at the end of the canonical name for the canonical path to the external dep. So I had to update some "local testing configs" to use `cereal+` and similar. This may come back as a new kind of error when we fix the python packaging workflow, if I didn't quite do it right here.


j2kun:
Needs one more step: https://github.com/bazelbuild/bazel-central-registry/pull/7047

asraa (approved):
thank you for documenting the weird `+` thing in case it comes up later!

PR #2495, opened 2025-12-22
Small update to the TfheRustEmitter to use Array

WoutLegiest:
Update of the TfheRustBoolEmitter to follow the same definition of the HL and normal TfheRustEmitter
Follow up after #2458


j2kun:
@WoutLegiest I lost track of this PR months ago. Is this still something you want to merge?

Issue #202, opened 2023-10-16
poly: determine if the poly_mod function name needs to include the input tensor type


j2kun:

ghost:
Hi all, Tomi here,I am a 2nd year Grad student at georgia tech and a software engineer based in the bay area. Looking to get a start contributing here and noticed this has not been assigned yet. can I take a stab at it?

j2kun:
Sounds good to me. I'd expect it should be a pretty small PR, good to start getting familiar with the build system and testing.

PR #2502, opened 2026-01-02
Issue #202 : Add tensor shape to polynomial mod function name

TomiAkeju:
This PR adds a tensor shape id to the function name generated in buildPolynomialModFunc to prevent potential name collisions. The function name now includes the input tensor dimensions (e.g., 2047 for a tensor<2047xi32>).

I want to note that I noticed <RankedTensorType> may have encodings, but they were not generated in parameters passed to buildPolynomialModFunc.  Very open to feedback. 

#202

### DIFF
--- a/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
+++ b/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
@@ -781,8 +781,8 @@ func::FuncOp PolynomialToModArith::buildPolynomialModFunc(FunctionType funcType,
         }
       },
       "x");
-  //Trailing identifier to match tensor syntax e.g tensor<2047x!coeff_ty>
-  os << "x"; 
+  // Trailing identifier to match tensor syntax e.g tensor<2047x!coeff_ty>
+  os << "x";
 
   auto coeffTy = ring.getCoefficientType();
   std::string coeffTyId;


lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -769,6 +769,23 @@ func::FuncOp PolynomialToModArith::buildPolynomialModFunc(FunctionType funcType,
   // TODO(#202): this function name probably also needs the input tensor type in
```
j2kun: nit: Remove the TODO

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -781,9 +798,9 @@ func::FuncOp PolynomialToModArith::buildPolynomialModFunc(FunctionType funcType,
     coeffTyId =
         llvm::formatv("{0}_i{1}", modulusStr, intTy.getIntOrFloatBitWidth());
   }
-  std::string funcName =
-      llvm::formatv("__heir_poly_mod_{0}_{1}", coeffTyId,
-                    ring.getPolynomialModulus().getPolynomial().toIdentifier());
+  std::string funcName = llvm::formatv(
```
j2kun, line 799: This name is matched on in lit tests, which will need to be updated. For example: https://github.com/google/heir/blob/73a4ca46958286c033f32db1120b46286efbf103/tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/lower_mul.mlir#L30

Please run `bazel test //tests/Dialect/Polynomial/...:all` to test the subset of tests that should be affected by this change. (Also, I believe this will cause CI to fail but it just started running, so that should point to the exact failing targets)

j2kun (changes requested):
LGTM, pending test updates

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
TomiAkeju: Thanks for the heads up. will take a look.

TomiAkeju: Added the new expected signatures to the lit tests and verified locally.  The bazel build took a lot more space than I anticipated but it was helpful to get more familiar with it.

PR #2510, opened 2026-01-07
Optimize chebyshev monomial

### LINKED ISSUE #2413: Detect if a chebyshev polynomial has a simpler monomial basis representation
In the lowering of a Chebyshev polynomial evaluation, we use this Paterson-Stockmeyer method along with baby-step giant-step to optimize the circuit. However, in some cases, even this method produces something that is suboptimal.

For example, the Chebyshev polynomial `[0.0, 0.75, 0.0, 0.25]` (ordered low-to-high degree) is equivalent in the monomial basis to x^3, and if you lower this via PS-Chebyshev (test case below), you get

```
module {
  func.func @chebyshev(%arg0: f32) -> f32 {
    %cst = arith.constant 5.000000e-01 : f32
    %0 = arith.mulf %cst, %arg0 : f32
    %cst_0 = arith.constant 5.000000e-01 : f32
    %1 = arith.mulf %cst_0, %arg0 : f32
    %cst_1 = arith.constant 2.000000e+00 : f32
    %2 = arith.mulf %arg0, %arg0 : f32
    %3 = arith.mulf %cst_1, %2 : f32
    %cst_2 = arith.constant 1.000000e+00 : f32
    %4 = arith.subf %3, %cst_2 : f32
    %5 = arith.mulf %1, %4 : f32
    %6 = arith.addf %0, %5 : f32
    return %6 : f32
  }
}
```

This is nontrivially more complex than the naive monomial evaluation of x^3, and since we know the chebyshev polynomial statically at compile time, it may provide some gains to identify when:

- The monomial basis does not have numerical stability issues with its coefficients (i.e., large degree coefficients that are very small)
- The monomial basis evaluation circuit would be simpler than the ps-cheb evaluation circuit.

In that case we should either (a) switch to the monomial basis and lower with a method like horner (or monomial-basis ps), or (b) analyze the circuit post-hoc for algebraic simplifications.

```
// RUN: heir-opt --lower-polynomial-eval="method=pscheb" %s | FileCheck %s

!poly_ty = !polynomial.polynomial<ring=<coefficientType=f32>>
#poly = #polynomial.typed_chebyshev_polynomial<[0.0, 0.75, 0.0, 0.25]> : !poly_ty

module {
  func.func @chebyshev(%ct: f32) -> f32 {
    %ct_0 = polynomial.eval #poly, %ct {coefficients = [0.0, 0.75, 0.0, 0.25], domain_lower = -1.000000e+00 : f64, domain_upper = 1.000000e+00 : f64} : f32
    return %ct_0 : f32
  }
}
```

Krishn1412:
- Fixes [issue #2413](https://github.com/google/heir/issues/2413).
- Uses a cost heuristic based on MultiplicativeDepthVisitorImpl to estimate polynomial evaluation complexity.
- Computes and compares the multiplicative depth of Chebyshev (PS) and monomial (Horner) evaluation DAGs.
- Selects the evaluation strategy with lower multiplicative depth.

### DIFF
--- a/lib/Transforms/LowerPolynomialEval/Patterns.cpp
+++ b/lib/Transforms/LowerPolynomialEval/Patterns.cpp
@@ -27,7 +27,7 @@
 #include "mlir/include/mlir/IR/Types.h"                  // from @llvm-project
 #include "mlir/include/mlir/IR/Value.h"                  // from @llvm-project
 #include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
-
+#include <bits/stdc++.h>
 #define DEBUG_TYPE "lower-polynomial-eval"
 
 namespace mlir {
@@ -40,16 +40,59 @@ using polynomial::EvalOp;
 using polynomial::FloatPolynomial;
 using polynomial::TypedFloatPolynomialAttr;
 
-static int estimateDagCost(ArithmeticDagNode<kernel::SSAValue>* node) {
-  int cost = 0;
-  node->walk([&](auto* n) {
-    if (n->isMul())
-      cost += 2;
-    else if (n->isAdd() || n->isSub())
-      cost += 1;
-  });
-  return cost;
-}
+using Dag = ArithmeticDagNode<kernel::SSAValue>;
+
+struct DagDebugPrinter {
+  std::unordered_set<const Dag*> visited;
+
+  void print(const std::shared_ptr<Dag>& node, int indent = 0) {
+    if (!node) {
+      indentPrint(indent);
+      std::cout << "<null>\n";
+      return;
+    }
+
+    const Dag* raw = node.get();
+    indentPrint(indent);
+
+    if (visited.count(raw)) {
+      std::cout << "<visited @" << raw << ">\n";
+      return;
+    }
+    visited.insert(raw);
+
+    node->visit([&](auto&& n) {
+      using N = std::decay_t<decltype(n)>;
+
+      if constexpr (std::is_same_v<N, mlir::heir::kernel::LeafNode<mlir::heir::kernel::SSAValue>>) {
+        std::cout << "Leaf(SSAValue)\n";
+      } else if constexpr (std::is_same_v<N, mlir::heir::kernel::ConstantScalarNode>) {
+        std::cout << "ConstScalar(" << n.value << ")\n";
+      } else if constexpr (std::is_same_v<N, mlir::heir::kernel::AddNode<mlir::heir::kernel::SSAValue>>) {
+        std::cout << "Add\n";
+        print(n.left, indent + 2);
+        print(n.right, indent + 2);
+      } else if constexpr (std::is_same_v<N, mlir::heir::kernel::SubtractNode<mlir::heir::kernel::SSAValue>>) {
+        std::cout << "Sub\n";
+        print(n.left, indent + 2);
+        print(n.right, indent + 2);
+      } else if constexpr (std::is_same_v<N, mlir::heir::kernel::MultiplyNode<mlir::heir::kernel::SSAValue>>) {
+        std::cout << "Mul\n";
+        print(n.left, indent + 2);
+        print(n.right, indent + 2);
+      } else if constexpr (std::is_same_v<N, mlir::heir::kernel::PowerNode<mlir::heir::kernel::SSAValue>>) {
+        std::cout << "Power(^" << n.exponent << ")\n";
+        print(n.base, indent + 2);
+      } else {
+        std::cout << "OtherNode\n";
+      }
+    });
+  }
+
+  void indentPrint(int n) {
+    for (int i = 0; i < n; ++i) std::cout << ' ';
+  }
+};
 
 LogicalResult LowerViaHorner::matchAndRewrite(EvalOp op,
                                               PatternRewriter& rewriter) const {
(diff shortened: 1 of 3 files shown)


lib/Transforms/LowerPolynomialEval/Patterns.cpp
```diff
@@ -12,6 +12,7 @@
 #include "lib/Utils/Polynomial/Horner.h"
 #include "lib/Utils/Polynomial/PatersonStockmeyer.h"
 #include "lib/Utils/Polynomial/Polynomial.h"
+#include "lib/Utils/Polynomial/PolynomialTestVisitors.h"
```
j2kun, line 15: nit: since this file was intended for unit testing only, and now it's used outside of a unit test, the MultiplicativeDepthVisitor should be extracted to a standalone library.

lib/Transforms/LowerPolynomialEval/Patterns.cpp
```diff
@@ -157,11 +157,29 @@ LogicalResult LowerViaPatersonStockmeyerChebyshev::matchAndRewrite(
...
+  auto xDag = kernel::ArithmeticDagNode<kernel::SSAValue>::leaf(xInput);
+  auto monoDag =
+    polynomial::hornerMonomialPolynomialEvaluation(
+        xDag, monomialCoeffs);
+  polynomial::test::MultiplicativeDepthVisitorImpl<kernel::SSAValue> depthVisitor;
+
+  double chebDepth = depthVisitor.process(chebDag);
+  double monoDepth = depthVisitor.process(monoDag);
+
+  bool useMonomial = chebDepth > monoDepth;
```
j2kun, line 179: This code is happening inside of a function that suggests the Paterson Stockmeyer Chebyshev method will be used without qualification.

In addition, if the pass option specifies that "pscheb" method must be used, I don't think the pass should second-guess it and use a different method, even if it's more efficient.

So what we should instead do here is: 

- Extract the construction of the DAG and the analysis into functions that are outside of the RewritePattern implementation.
- In `LowerPolynomialEval` [where the method is "auto"](https://github.com/google/heir/blob/b7f2f964cb9d201411a40317d6971c54a5f90910/lib/Transforms/LowerPolynomialEval/LowerPolynomialEval.cpp#L29), construct the dags, do the analysis on the depth, and then use that to determine which rewrite pattern to apply.

tests/Transforms/lower_polynomial_eval/chebyshev_to_monomial.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: heir-opt --lower-polynomial-eval="method=pscheb" %s | FileCheck %s
+
+!poly_ty = !polynomial.polynomial<ring=<coefficientType=f32>>
+#poly = #polynomial.typed_chebyshev_polynomial<[0.0, 0.75, 0.0, 0.25]> : !poly_ty
+
+module {
+  func.func @chebyshev(%ct: f32) -> f32 {
+    %ct_0 = polynomial.eval #poly, %ct {coefficients = [0.0, 0.75, 0.0, 0.25], domain_lower = -1.000000e+00 : f64, domain_upper = 1.000000e+00 : f64} : f32
```
j2kun, line 8: This file needs at least one `// CHECK: ...` statement to assert the output is correct.

lib/Transforms/LowerPolynomialEval/Patterns.cpp
j2kun, line 179: The second problem listed in #2413 which is not covered by this PR is numerical stability. The reason we use the Chebyshev basis instead of the monomial basis is that, for larger-degree polynomials, the monomial basis coefficients of high-degree terms will necessarily grow to be quite small (e.g., 1e-15) but cannot be ignored because of their large influence on the evaluated result, while Chebyshev basis coefficients remain relatively well normalized and small magnitude coefficients can be dropped without influencing the output.

So the other check that needs to occur to allow the monomial lowering is: will the monomial representation be unstable? While it may depend on which FHE scheme is being used and what precision is supported in that scheme, a good place to start would be to compute the condition number of polynomial evaluation for monomials, in https://epubs.siam.org/doi/10.1137/1.9780898718027.ch5 (Higham's _Accuracy and Stability of Numerical Algorithms_). However, that requires knowing the right value of `x` in advance, which may not be true in this case. We should have access to lower and upper bounds on the domain, so maybe sampling a few values in the domain would suffice.

j2kun (changes requested):

lib/Transforms/LowerPolynomialEval/Patterns.cpp
Krishn1412: As I understand from the code, `LowerPolynomialEval` only registers the rewrite patterns, the actual rewriting happens later when MLIR applies them. At registration time we don’t have access to the polynomials, so any analysis can’t be done there? Can we instead do the analysis inside matchAndRewrite, where the polynomial.eval op is available. In automatic mode, a pattern can return failure() if it decides it’s too expensive or unstable, allowing MLIR to try the other patterns. Apologies for the delay!

j2kun: Anything inside `runOnOperation` has access to the entire IR, but yes, you can move the code that actually mutates the IR into helper functions and then do the analysis inside a smaller number of patterns.

j2kun:
Closing. Please cf. https://github.com/google/heir/blob/main/CONTRIBUTING.md#L28 if you'd like to get more involved in this project.

PR #2528, opened 2026-01-13
bump rust dependencies

j2kun:
Fixes build errors related to `Error[E0658]: use of unstable library feature stdarch_x86_avx512`

Bytemuck is transitively depended on through the `tfhe` crate, and `tfhe` recently pushed 1.5.0 which pins to a newer, problematic `bytemuck` version. That's why we started seeing the issue yesterday.

I thought the crate versions were pinned in MODULE.bazel, but then with some further testing I discovered the version pins are ignored and it seems to always fetch the latest crate. Cf. https://github.com/bazelbuild/rules_rust/issues/3805


j2kun:
Edit: pinning the rustc version to the latest stable version seems to fix the issue!

PR #2534, opened 2026-01-15
Peel first iteration of loops with ciphertext iter-args (Halo Solution A-1)

j2kun:
Adds patterns to peel the first iteration of an `affine.for` or `scf.for` loop when the secretness of the init value and the yielded iter-arg disagree.

This is intended to be used at the `secret` + `arith` level, possibly with `mgmt` ops in the mix. I think this pattern will ultimately be used as part of the `insert-mgmt` pipeline, since that is where the bootstrap and level analysis occurs. But for now it's an independent pattern with a thin wrapper pass for testing.

### DIFF
--- a/lib/Transforms/Halo/BUILD
+++ b/lib/Transforms/Halo/BUILD
@@ -1,0 +1,65 @@
+load("@heir//lib/Transforms:transforms.bzl", "add_heir_transforms")
+
+package(
+    default_applicable_licenses = ["@heir//:license"],
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "Transforms",
+    hdrs = ["Passes.h"],
+    deps = [
+        ":ReconcileMixedSecretnessIterArgs",
+        ":pass_inc_gen",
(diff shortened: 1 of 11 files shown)


asraa:
if this pass is primarily for reconciling types (secret/plaintext), then why would it need to be part of the mgmt pipelines? those will change the level attrs and add new mgmt ops but I don't think at that point it would be changing types

j2kun:
> if this pass is primarily for reconciling types (secret/plaintext), then why would it need to be part of the mgmt pipelines? those will change the level attrs and add new mgmt ops but I don't think at that point it would be changing types

You're right about this pattern: it is a prerequisite to insert-mgmt because the level analysis would not converge otherwise, but not needed as "part" of insert-mgmt. But the later HALO passes definitely need to be part of insert-mgmt (or some alternative pipeline) because you need a level analysis to determine how much to unroll the loop.

j2kun:
I will squash this PR with joint authorship once Sara signs the CLA

PR #2536, opened 2026-01-16
Bootstrap loop iter args (Halo Solution A-2)

j2kun:
From 4.2 of https://dl.acm.org/doi/10.1145/3669940.3707275

> Solution A-2: Bootstrapping all the loop-carried variables at the beginning of each iteration. Regardless of consumed levels of ciphertexts, bootstrapping reinitializes their level. Therefore, inserting the bootstrap operation at the beginning of the loop iteration can resolve the level type mismatch problem. Figure 2: Solution A-2 depicts the level matching process of the input and output of the loop body. Before entering the loop body, HALO reduces each level of y and a to the minimum level using modswitch for live-in and loop-carried variables. Then, HALO recovers the levels of the loop inputs to 10 by bootstrapping all loop inputs at the beginning of the loop. After type matching process, the loop is transformed to a type-matched loop (repeatable). However, the type-matched loops are inefficient. For efficient loop execution, bootstrap-based loop restructuring is required.

For this PR I introduced a new `mgmt` op: `mgmt.level_reduce_min` which corresponds to "reducing the level to the minimum allowable level." I suspect we need this op in this situation because we will be inserting this `level_reduce_min` before knowing what the actual level budget is, and this ensures that our level analysis can reach a fixpoint when processing the loops that will be present after the HALO implementation is integrated.

Rebased over https://github.com/google/heir/pull/2534, and similar to that PR, I just created a pattern to handle the core operation, wrapped it in a dedicated pass for testing, and will figure out how to integrate it into the mgmt pipeline once all the remaining pieces of the Halo paper are implemented.

### DIFF
--- a/lib/Dialect/Mgmt/IR/MgmtOps.td
+++ b/lib/Dialect/Mgmt/IR/MgmtOps.td
@@ -62,6 +62,20 @@ def Mgmt_LevelReduceOp : Mgmt_Op<"level_reduce"> {
   let hasCanonicalizer = 1;
 }
 
+def Mgmt_LevelReduceMinOp : Mgmt_Op<"level_reduce_min"> {
+  let summary = "Reduce the level of input ciphertext to the minimum level";
+  let description = [{
+    This scheme-agonistic operation reduces the ciphertext level
+    to the minimum allowable level (typically 1 RNS limb). This is used in
+    HALO-style optimizations to ensure loop iter args and inits are invariant
+    across loop iterations.
+  }];
+  let arguments = (ins AnyType:$input);
+  let results = (outs AnyType:$output);
+  let assemblyFormat = "operands attr-dict `:` type($output)";
+}
+
+
 def Mgmt_RelinearizeOp : Mgmt_Op<"relinearize"> {
   let summary = "Relinearize the input ciphertext to be _linear_";
 
--- a/lib/Transforms/Halo/BootstrapLoopIterArgs.cpp
+++ b/lib/Transforms/Halo/BootstrapLoopIterArgs.cpp
@@ -1,0 +1,48 @@
+#include "lib/Transforms/Halo/BootstrapLoopIterArgs.h"
+
+#include "lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h"
+#include "lib/Transforms/Halo/Patterns.h"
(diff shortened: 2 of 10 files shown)


j2kun:
The CI failure was due to https://github.com/llvm/llvm-project/pull/175056#issuecomment-3750147952 which will be fixed in the next LLVM integrate.

lib/Transforms/Halo/BootstrapLoopIterArgs.cpp
```diff
@@ -0,0 +1,46 @@
...
+    dataflow::loadBaselineAnalyses(solver);
+    solver.load<SecretnessAnalysis>();
+
+    if (failed(solver.initializeAndRun(getOperation()))) {
+      getOperation()->emitOpError() << "Failed to run the analysis.\n";
+      signalPassFailure();
+      return;
+    }
+
+    patterns.add<PeelPlaintextAffineForInit, PeelPlaintextScfForInit>(context,
```
asraa: @j2kun is there a commit push missing?

j2kun: Yes! Fixed, sorry >_<

lib/Dialect/Mgmt/IR/MgmtOps.td
```diff
@@ -62,6 +62,19 @@ def Mgmt_LevelReduceOp : Mgmt_Op<"level_reduce"> {
   let hasCanonicalizer = 1;
 }
 
+def Mgmt_LevelReduceMinOp : Mgmt_Op<"level_reduce_min"> {
+  let summary = "Reduce the level of input ciphertext to the minimum level";
+  let description = [{
+    This scheme-agonistic operation reduces the ciphertext level
```
asraa, line 68: I'm guessing this is needed since the existing level reduce operation takes a number of levels to drop, and at this stage we don't know how many levels / limbs the input ciphertext has. I would probably clarify that if I'm right (that the minimum allowable level is always one limb?)

j2kun: Correct.

j2kun: Done!

PR #2542, opened 2026-01-20
Add a "scratchpad" pass for experimentation purposes

j2kun:


lib/Transforms/ScratchpadPass/ScratchpadPass.td
```diff
@@ -0,0 +1,18 @@
+#ifndef LIB_TRANSFORMS_SCRATCHPADPASS_SCRATCHPADPASS_TD_
+#define LIB_TRANSFORMS_SCRATCHPADPASS_SCRATCHPADPASS_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def ScratchpadPass : Pass<"scratchpad"> {
+  let summary = "An empty pass shell for experimentation purposes.";
```
asraa, line 7: Is this meant to be for local experimentation only (as a template?). I think for many patterns they should eventually end up in a pass, at least for the purposes of adding unit LIT tests for the pattern execution

j2kun: Yes, just for local experimentation. I made this PR mainly to show Eric Crockett the idea, and if he doesn't find it useful when ramping up on MLIR, we don't have to merge it.

PR #2547, opened 2026-01-21
pre-commit: use python3 instead of python

AlexanderViand:
I realized that on a freshly set-up macOS, vscode would fail to run the pre-commit checks because `python` isn't set up as an alias for `python3` by default. Actually, I think the same might be true on a fresh ubuntu machine (but I always use python-is-python3 so never tested 🤷)

EDIT 1: ignore the force-pushing, didn't have gpg, git and gh lined up properly so the signature didn't verify
EDIT 2: unrelated build failure means no CI, but it works on my machine (TM) ;)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,7 @@ repos:
   hooks:
     - id: disable-mdformat
       name: Ensure google-internal mdformat is disabled on md files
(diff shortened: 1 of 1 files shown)


j2kun:
Despite the CI failure, it should still merge properly. Thanks!

PR #2548, opened 2026-01-21
Migrate OpenFHE to bazelmod

j2kun:
Now that OpenFHE is on the BCR, we can move it to MODULE.bzl

Part of https://github.com/google/heir/issues/1446

Cherry-picking the LLVM integrate reversion


j2kun:
Looks like there's still some fixes needed for the frontend

j2kun:
Frontend tests should now be fixed, just had to update paths since bazelmod canonical repo names end in a plus

Issue #1181, opened 2024-12-12
Make LevelAnalysis/DimensionAnalysis aware of affine.for loop


ZenithalHourlyRate:
Check #1176.

> In tests/Dialect/Secret/Conversions/secret_to_ckks/matmul_loop.mlir, affine is not unrolled so LevelAnalysis can not work. It should not work as we do not have loop invariant kept here. For example, for the loop to carry out correctly in FHE, the invariant is the level is the same, which heavily depends on the body.
>
> More case analysis here. The body is a mul_plain + add, which in different mgmt op placement strategy there will be different level analysis result. For example, if we mod reduce immediately after mul_plain, we get the level reduced by one hence the invariant is not kept. For the current strategy we fortunately kept this invariant.

ZenithalHourlyRate:
Decided to give up on this by #1569. Also #1459 provides a way to skip these analyses.

(do not close this issue due to TODO-backlinks)

PR #2551, opened 2026-01-21
Rewrite level analysis for loop support

j2kun:
Extracted just the level analysis rewrite from https://github.com/google/heir/pull/2538

This PR includes a rewrite of LevelAnalysis to make it aware of loops, and, provided the patterns from https://github.com/google/heir/pull/2534 and https://github.com/google/heir/pull/2536 are applied, the analysis will quickly reach a fixpoint.

As a byproduct, this PR rewrites the LevelAnalysis code from scratch to be a bit more ergonomic.

- Transfer logic is isolated into dedicated functions
- The LevelState is extended to include explicit sentinels for "uninitialized", "invalid", and "max level"
- The "max level" sentinel is intended to represent the maximum level required, even before that maximum level is known, so that ops like `level_reduce_min` can be sensibly analyzed.
- Additional documentation has been added to clarify how the analysis works.
- Includes a new interface to show which operands of an op need back-propagation for level/dimension analysis.
- Moved the hoisting interface to its own file to break dependency cycles

*Warning*: If the loop does not have relinearization ops inserted properly, then the DimensionAnalysis may loop infinitely. And since most uses of the LevelAnalysis simultaneously use the DimensionAnalysis, this may cause issues during integration. I was thinking I could also rewrite the DimensionAnalysis to have an "invalid" state, and call any dimension over 5 (say) "invalid", which would ensure a fixpoint and avoid infinite loops (at the expense of a possibly mysterious error later). Either way, this PR isolates the loop-based tests into individual passes so I can handle integration in a followup PR.

Part of https://github.com/google/heir/issues/1181


lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h
```diff
@@ -199,19 +213,25 @@ class SecretnessAnalysisDependent {
...
-   * those that are not secret to the provided vector.
+   * The input operation must either have no legal plaintext operands, or else
+   * implement PlaintextOperandInterface to identify which operands may be
+   * plaintext. Those are then filtered by secretness to populate the outparam.
    *
    * @param op The operation to analyze.
    * @param nonSecretOperands A vector to store the non-secret operands.
    */
-  void getNonSecretOperands(Operation* op,
+  void getPlaintextOperands(Operation* op,
```
asraa, line 225: nit: if you want to, update the name of the `nonSecretOperands` to reflect the restriction to just plaintexts and not just nonsecret

lib/Analysis/LevelAnalysis/LevelAnalysis.h
```diff
@@ -66,33 +140,34 @@ class LevelState {
...
 ///
-/// This forward analysis roots from user input as 0, and
-/// after each modulus switching operation, the level will
-/// increase by 1, until the output.
+/// This forward analysis roots from user input as 0, and after each modulus
+/// switching operation, the level will increase by 1, until the output.
 ///
-/// Special case is the bootstrapping operation, where the level
-/// will be set back to 0 (input level).
+/// Special cases:
```
asraa: mod_reduce is also in here right?

lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h
j2kun: Done

lib/Analysis/LevelAnalysis/LevelAnalysis.h
j2kun: Done

PR #2546, opened 2026-01-21
frontend: use signed extension when changing bitwidth

AlexanderViand:
This PR fixes an issue in the python frontend related to bit width handling: when the frontend upcasts, it now uses signed extension instead of unsigned (fixes #2496)

PR #2552, opened 2026-01-21
Frontend literal upcasting

### LINKED ISSUE #2525: A strange bug related to plaintext computation
Hi! I meet a very strange issue with HEIR's python frontend (heir-py 0.0.2) 😂

If I write my code like this:

```
from heir import compile
from heir.mlir import I32, Secret

@compile()
def func(x: Secret[I32]):
    v2 = x + 10
    v3 = 15 * 3
    v4 = 10 * v3
    v5 = v4 + v4
    return v5 + v2

if __name__ == "__main__":
    x = 2
    func.setup()
    enc_x = func.encrypt_x(x)
    result_enc = func.eval(enc_x)
    result = func.decrypt_result(result_enc)
    print(f"Expected result: {func.original(x)}")
    print(f"FHE result: {result}")
```

It will output:

```
Expected result: 912
FHE result: 144
```

However, if I reduce this program and write it as:

```
from heir import compile
from heir.mlir import I32, Secret

@compile()
def func(x: Secret[I32]):
    v2 = x + 10
    v3 = 10 * 15 * 3
    v5 = v3 + v3
    return v5 + v2

if __name__ == "__main__":
    x = 2
    func.setup()
    enc_x = func.encrypt_x(x)
    result_enc = func.eval(enc_x)
    result = func.decrypt_result(result_enc)
    print(f"Expected result: {func.original(x)}")
    print(f"FHE result: {result}")
```

It works well:
```
Expected result: 912
FHE result: 912
```

And, if I use this program:
```
from heir import compile
from heir.mlir import I32, Secret

@compile()
def func(x: Secret[I32]):
    v2 = x + 10
    v3 = 15 * 3
    v4 = 10 * v3
    v5 = v4 + v4
    return v5 + v3

if __name__ == "__main__":
    x = 2
    func.setup()
    enc_x = func.encrypt_x(x)
    result_enc = func.eval(enc_x)
    result = func.decrypt_result(result_enc)
    print(f"Expected result: {func.original(x)}")
    print(f"FHE result: {result}")
```

It will crash with red error message.

AlexanderViand:
While we still prevent numba from doing its default upcasting of literals to int32, there is now some logic that ensures that the inputs to binary operations on literals/constants do get upcast to prevent overflow (fixes #2525)

This is necessary, despite Python's eager constant folding pre-bytecode, because of examples like this:
```python
a = 45 # or a = 3 * 15, which will be constant-folded to 45 in the byte code already
b = 10 * a
```
This will be translated to a binop (mul) between two literals (each of type int8, because we overwrite numba to pick the smallest numba type that matches), and since binop rules in numba don't promote, computed as int8 * int8 = int8 which will overflow and give -62 instead of 450. I.e., we would get:

```llvm
%0 = arith.constant 45 : i8
%1 = arith.constant 10 : i8
%2 = arith.muli %0, %1 : i8
```

The fix works in two parts:
* The binop emitter now takes a result type, and if that's wider than the input types, emits extensions for the input operands. This by itself would be a no-op change, as the result type would be the same as the input types by numba's rules.

* Therefore, our numba typing override now introduces a special exception for binary operations where both inputs are literal integers, and assigns the output type based on the result, which will then cause the emitter code above to actually emit the extensions:

```llvm
%0 = arith.constant 45 : i8
%1 = arith.constant 10 : i8
%2 = arith.extsi %1 : i8 to i16
%3 = arith.extsi %0 : i8 to i16
%4 = arith.muli %2, %3 : i16
```

Note: the second part is pretty much entirely LLM generated as I don't remember much from my deep dive into numba's typing system last year. ~~Marking as draft until I can look at that code a second/third time~~ Had a look and it, imho, checks out, also added a few more tests. I'm not certain this is the last time we'll run into bitwidth issues, but at least for this issue, this seems like a solid fix.

### DIFF
--- a/frontend/heir/mlir_emitter.py
+++ b/frontend/heir/mlir_emitter.py
@@ -43,7 +43,7 @@ def isIntegerLike(typ: NumbaType | MLIRType) -> bool:
   match typ:
     case I1() | I8() | I16() | I32() | I64() | nt.Integer() | nt.Boolean():
       return True
-    case MLIRType() | nt.Type:
+    case MLIRType() | nt.Type():
       return False
     case _:
       raise InternalCompilerError(
@@ -55,7 +55,7 @@ def isFloatLike(typ: NumbaType | MLIRType) -> bool:
   match typ:
     case nt.F32 | nt.F64 | nt.Float():
       return True
-    case MLIRType() | nt.Type:
+    case MLIRType() | nt.Type():
       return False
     case _:
       raise InternalCompilerError(
@@ -508,7 +508,10 @@ class TextualMlirEmitter:
         )
         return ""
       case ir.Expr(op="binop"):
-        emitted_expr, ext, ty = self.emit_binop(assign.value)
+        result_type = self.typemap.get(assign.target.name)
+        emitted_expr, ext, ty = self.emit_binop(
+            assign.value, result_type=result_type
+        )
         expr = f"{emitted_expr} : {mlirType(ty)} {mlirLoc(assign.loc)}"
         # if the var is being reassigned, then create a new SSA var
         assign_str = self.reassign_and_forward_name(assign.target, expr)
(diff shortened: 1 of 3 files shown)


j2kun:
Will wait for https://github.com/google/heir/pull/2546 to merge before marking this as pull_ready.

AlexanderViand:
> Will wait for #2546 to merge before marking this as pull_ready.

rebased :)

j2kun (approved):

PR #2555, opened 2026-01-21
Migrate yosys and ABC to bazelmod

j2kun:
Fixes https://github.com/google/heir/issues/1446

Depends on https://github.com/bazelbuild/bazel-central-registry/pull/7258
Rebased over https://github.com/google/heir/pull/2548

This will require some nontrivial internal patches for copybara


j2kun:
OK, one more build error, looks like the external kernel module is not pulling in all the cpp files it needs

```
ld.lld: error: undefined reference: Yosys::RTLIL_BACKEND::dump_sigspec(std::ostream&, Yosys::RTLIL::SigSpec const&, bool)
>>> referenced by bazel-out/k8-opt/bin/_solib_k8/libexternal_Syosys+_Slibkernel.so (disallowed by --no-allow-shlib-undefined)
```

j2kun:
Update: I got it building and tests passing locally after https://github.com/bazelbuild/bazel-central-registry/pull/7507

j2kun:
OK it's ready!

asraa (approved):

PR #2561, opened 2026-01-22
Integrate Halo loop patterns into secret-insert-mgmt

j2kun:
This PR integrates the HALO loop support patterns into the secret-insert-mgmt pipeline.

The vast majorty of this PR is debug statements added to the insert-mgmt pipeline, guarded by the `experimental-disable-loop-unroll`, and the various analyses so that I could figure out the small number of issues that needed fixing. Those issues are:

1. When an analysis depends on another analysis, the dependence relationship between the two analyses is not established until the first time that one analysis queries a lattice element from another analysis. Moreover, the order in which analyses are run is deterministic, but appears to be arbitrary, so if analysis A depends on analysis B, then A can run before B. When it queries B, the analysis state is not present. You're supposed to default to a pessimistic lattice value, and then due to the change to the dependence structure, the analysis framework will trigger another run of analysis B after analysis A is run on that program point.
2. The default bootstrap params is way too slow, resulting in like a 3 minute bootstrap latency. I overrode the crypto-context params/configure step in the final tests to use faster params so we could at least test correctness. But I also created a benchmark file to experiment with this outside of the context of HEIR's compilation pipeline.
3. We need some way to seed the number of available levels for loop unrolling, otherwise HEIR will be default think that we don't need very many levels at all (i.e., if there are no mul's in the loop, then it thinks we need zero levels). So I added a pipeline flag `level-budget` that gives a total number of levels (excluding what is needed for bootstrap). This is mainly a hack that allows me to get the compilation to work properly.

### DIFF
--- a/benchmark/bootstrap_example_main.cpp
+++ b/benchmark/bootstrap_example_main.cpp
@@ -1,0 +1,47 @@
+#include <iostream>
+#include <vector>
+
+#include "benchmark/bootstrap_example.h"
+
+// A simple test for loop support.
+//
+// Computes the function
+//
+//     def f(x):
+//       sum = 1.0
+//       for i in range(8):
+//         sum = sum * x - 1.0
+//       return sum
+//
+int main(int argc, char** argv) {
+  std::cout << "Bootstrap Example Main\n";
+  std::cout << "Generating crypto context\n";
+  auto cryptoContext = loop__generate_crypto_context();
+  std::cout << "Generating keys\n";
+  auto keyPair = cryptoContext->KeyGen();
+  auto publicKey = keyPair.publicKey;
+  auto secretKey = keyPair.secretKey;
+  std::cout << "Configuring crypto context\n";
+  cryptoContext = loop__configure_crypto_context(cryptoContext, secretKey);
+
+  std::vector<float> arg0 = {0.,         0.14285714, 0.28571429, 0.42857143,
+                             0.57142857, 0.71428571, 0.85714286, 1.};
+  std::vector<float> expected = {-1.,         -1.16666629, -1.39989342,
+                                 -1.74687019, -2.29543899, -3.19507837,
+                                 -4.66914279, -7.};
+
+  std::cout << "Encrypting arg\n";
+  auto arg0Encrypted = loop__encrypt__arg0(cryptoContext, arg0, publicKey);
+  std::cout << "Running loop\n";
+  auto outputEncrypted = loop(cryptoContext, arg0Encrypted);
+  std::cout << "Decrypting result\n";
+  auto actual =
+      loop__decrypt__result0(cryptoContext, outputEncrypted, secretKey);
+
+  for (size_t i = 0; i < actual.size(); i++) {
+    std::cout << "actual[" << i << "] = " << actual[i] << ", expected[" << i
+              << "] = " << expected[i] << "\n";
+  }
+
+  return 0;
+}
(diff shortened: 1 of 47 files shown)


j2kun:
@asraa this PR is finally ready for a review.

benchmark/bootstrap_example_main.cpp
```diff
@@ -0,0 +1,37 @@
+#include <iostream>
+#include <vector>
+
+#include "benchmark/bootstrap_example.h"
+
+int main(int argc, char** argv) {
```
asraa, line 16: for the sake of making this easier for us to understand later and so people know what "loop" is trying to compute, could you put a link to the IR that this was modeled after (that simple loop with the mul inside)

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
```diff
@@ -489,7 +489,24 @@ LogicalResult OpenFhePkeEmitter::printOperation(affine::AffineForOp op) {
 }
 
 LogicalResult OpenFhePkeEmitter::printOperation(affine::AffineYieldOp op) {
-  // Assume all yielded loop values have already been assigned.
+  // Assign each operand to the corresponding iter arg. This may be strictly
+  // necessary in such cases as the iter_arg being the result of inserting
+  // a value into an empty tensor (defined outside the loop) which is then
```
asraa, line 494: how come the tensor.insert operations don't already handle the assignment?

tests/Examples/openfhe/ckks/loop_support/debug_helper.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "src/core/include/lattice/hal/lat-backend.h"  // from @openfhe
+#include "src/core/include/utils/inttypes.h"           // from @openfhe
+#include "src/pke/include/encoding/plaintext-fwd.h"    // from @openfhe
+
+using lbcrypto::DCRTPoly;
+using PlaintextT = lbcrypto::Plaintext;
+
+// DecryptCore not accessible from CryptoContext
+// so copy from @openfhe//src/pke/lib/schemerns/rns-pke.cpp
+DCRTPoly decryptCore(const std::vector<DCRTPoly>& cv,
```
asraa: I'm a little suspicious of this helper since it's not named "DecryptCore" like the docstring..

asraa (approved):

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
j2kun: Ah, this was a significant bug to fix, that suggests we should probably lower to memref before C++ codegen, tbh.

The loop looked like:

```
%0 = tensor.empty()
%1 = tensor.insert %init into %0[0]
for %i from 1...10 iter_args(%iter = %1)
  ...
  %next = tensor.insert %computed_val into %iter[0]
  affine.yield %next
```

This code-genned to a situation where, due to the reference following in the OpenFHE codegen, the final tensor.insert referred to `%0` (!). So we got a line at the end of the code-genned loop that assigned to that `%0`. On top of that, the loop init itself code-genned a copy of the init so that the loop could mutate the init/iter_arg without interfering with other uses of the init. Code gen was like:

```
std::vector<CiphertextTy> foo(1);
foo[0] = init;
auto iter_arg = foo;

for (int i = 0; i < 10; ++i) {
   ...
   foo[0] = computed_val;
   // iter_arg ignored
}
```

So that insert op was not actually inserting into the iter_arg, and the next iteration of the loop started back from the original init val again.

I think this would all kind of be avoided if:

1. We lowered tensor to memref as the last step of the openfhe pipeline
2. We used emitC instead of hand-writing this stuff

tests/Examples/openfhe/ckks/loop_support/debug_helper.cpp
j2kun: I don't actually need that function at all, it was copied from the other debug helper. Removing.

benchmark/bootstrap_example_main.cpp
j2kun: Done

Issue #2519, opened 2026-01-09
Try EvalFastRotationExt over EvalFastRotation in OpenFHE backend


j2kun:
We use `EvalFastRotation` in our OpenFHE backend, but there is another method `EvalFastRotationExt` https://github.com/openfheorg/openfhe-development/blob/aa391988d354d4360f390f223a90e0d1b98839d7/src/pke/include/cryptocontext.h#L2438

It says that one is supported for hybrid key switching, which we support in HEIR. However, I don't recall if we use hybrid key switching by default, or how to determine when to use it versus other options.

Search the codebase for `SetKeySwitchTechnique(HYBRID)` for references.

Main goal would be to understand the performance implications of using the `Ext` variant, and use it in HEIR when appropriate.

toprakmurat:
Hi! I’d be happy to work on this issue. I’m an undergraduate student at ITU and have recently started doing research in FHE. Could you please assign this issue to me?

asraa:
I was just looking at this because I found that FastRotation was taking a lot of time in my program - yes, HEIR enables hybrid key switching by default. OpenFHE's default settings enable rotations with HYBRID and HEIR will only SetKeySwitchTechnique in codegen/interpreter when the GenParams op enables BV key switching.

Ext mode enables the fast rotation to be computed in the extended modulus (P*Q). After calling EvalFastRotationExt, you need to explicitly mod down back to the modulus Q with KeySwitchDown. In the meantime, you can perform computations in the extended modulus like EvalAddExt/EvalAddExtInPlace or EvalMultExt/EvalMultExtInPlace

It makes the most sense to do this when we can add/mul_plain many rotation results in the extended modulus and call KeySwitchDown once after taking the sum.

j2kun:
This sounds like it should then be a larger transformation pass to take advantage of this opporuntity at the OpenFHE level.

asraa:
Totally - the openfhe EvalLinearTransform functions use this internally so we are losing out on some performance by not using that API when we are computing linear transforms for matvecs/etc

j2kun:
OK so for @toprakmurat the scope of the transformation will be:

- At the `openfhe` dialect level (since we don't have a notion of "extended basis operations" at the LWE level)
- Add new ops for `openfhe.fast_rotation_ext` as well as other ops that can happen at the `_ext` level.
- Add a new op for the KeySwitchDown (reverting from the extended basis to the normal basis).
[...]

PR #2574, opened 2026-01-24
Add extended basis rotation operations for OpenFHE dialect

toprakmurat:
## Summary

This PR adds support for extended basis operations in the OpenFHE dialect, allowing the use of an optimization technique where multiple rotations are performed in a larger (P*Q) CRT basis and combined before a single `key_switch_down` converts the result back to the normal (Q) basis.

This is groundwork for addressing #2519.

## New Operations

| Operation | Description |
|-----------|-------------|
| `openfhe.fast_rotation_ext` | Fast rotation in extended P*Q basis (defers key switching) |
| `openfhe.key_switch_down` | Converts ciphertext from extended to normal basis |
| `openfhe.add_ext` | Addition of ciphertexts in extended basis |
| `openfhe.add_ext_inplace` | In-place addition in extended basis |
| `openfhe.mul_ext` | Multiplication in extended basis (uses `EvalMultNoRelin`) |

### DIFF
--- a/lib/Dialect/Openfhe/IR/OpenfheOps.cpp
+++ b/lib/Dialect/Openfhe/IR/OpenfheOps.cpp
@@ -23,6 +23,93 @@ namespace mlir {
 namespace heir {
 namespace openfhe {
 
+namespace {
+
+// Trait to get the operation creator for each rotation type
+template <typename RotationOpType>
+struct RotationOpCreator {
+  static_assert(sizeof(RotationOpType) == 0,
+                "RotationOpCreator not specialized for this type");
+};
+
+template <>
+struct RotationOpCreator<FastRotationOp> {
+  static Operation* create(OpBuilder& b, Location loc, Type ctType,
+                           FastRotationOp sourceOp, Value idx) {
+    return openfhe::FastRotationOp::create(
+        b, loc, ctType, sourceOp.getCryptoContext(), sourceOp.getInput(), idx,
+        sourceOp.getCyclotomicOrder(), sourceOp.getPrecomputedDigitDecomp());
+  }
+};
+
+template <>
+struct RotationOpCreator<FastRotationExtOp> {
+  static Operation* create(OpBuilder& b, Location loc, Type ctType,
+                           FastRotationExtOp sourceOp, Value idx) {
+    return openfhe::FastRotationExtOp::create(
+        b, loc, ctType, sourceOp.getCryptoContext(), sourceOp.getInput(), idx,
+        sourceOp.getPrecomputedDigitDecomp(), sourceOp.getAddFirst());
+  }
+};
+
+template <typename RotationOpType>
+FailureOr<Operation*> buildBatchedRotationOperation(
+    RotationOpType sourceOp, MLIRContext* context, OpBuilder& builder,
+    const SmallVector<Operation*>& batchedOperations) {
+  SmallVector<Value> indexVals;
+  indexVals.reserve(batchedOperations.size());
+
+  for (auto* op : batchedOperations) {
+    auto rotateOp = cast<RotationOpType>(op);
+    indexVals.push_back(rotateOp.getIndex());
+  }
+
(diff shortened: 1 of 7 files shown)


toprakmurat:
I'd appreciate feedback on the operation signatures in OpenfheOps.td (are the arguments and attributes appropriate?), whether we need a distinct type for extended-basis ciphertexts or can reuse Openfhe_Ciphertext, and any concerns with the overall approach before I proceed with implementing the conversion pass and optimization patterns to push `key_switch_down` later in the IR.

lib/Dialect/Openfhe/IR/OpenfheOps.td
```diff
@@ -469,6 +492,26 @@ def FastRotationOp : Openfhe_Op<"fast_rotation", [Pure,
...
+def FastRotationExtOp : Openfhe_Op<"fast_rotation_ext", [Pure,
+  AllTypesMatch<["input", "output"]>,
+  BatchVectorizableOpInterface,
+  DeclareOpInterfaceMethods<ElementwiseByOperandOpInterface>,
+  DeclareOpInterfaceMethods<
+      BatchVectorizableOpInterface,
+      ["isBatchCompatible", "buildBatchedOperation"]>
+]> {
+  let summary = "Fast (hoisted) rotation in extended (P*Q) CRT basis.";
+  let arguments = (ins
```
asraa, line 492: It looks like from https://github.com/openfheorg/openfhe-development/blob/aa391988d354d4360f390f223a90e0d1b98839d7/src/pke/include/cryptocontext.h#L2438

that the arguments needed are (cryptoContext, input, index, precomputedDigitDecomp, addFirst) - in other words i don't think you neeed cyclotomicOrder anymore.

lib/Dialect/Openfhe/IR/OpenfheOps.cpp
```diff
@@ -99,6 +99,81 @@ FailureOr<Operation*> FastRotationOp::buildBatchedOperation(
...
+              rhsRotate.getPrecomputedDigitDecomp() &&
+          getCyclotomicOrder() == rhsRotate.getCyclotomicOrder() &&
+          getAddFirst() == rhsRotate.getAddFirst());
+}
+
+FailureOr<Operation*> FastRotationExtOp::buildBatchedOperation(
+    MLIRContext* context, OpBuilder& builder,
+    SmallVector<Value> vectorizedOperands,
+    SmallVector<Operation*> batchedOperations) {
+  SmallVector<Value> indexVals;
```
asraa: I wonder if you can share a function with EvalFastRotation? The code is very very similar.

lib/Dialect/Openfhe/IR/OpenfheOps.td
j2kun: +1, I submitted the patch to remove that argument, though we should double check it's in the OpenFHE version we're depending on

asraa: your patch was only for the regular evalfastrotation, this API was unchanged and i guess never had that m param

lib/Target/OpenFhePke/Interpreter.cpp
```diff
@@ -1937,6 +1942,46 @@ void Interpreter::visit(FastRotationPrecomputeOp op) {
...
+                 cc->EvalFastRotationExt(ct, index, digits, addFirst));
+}
+
+void Interpreter::visit(KeySwitchDownOp op) {
+  auto cc = cryptoContexts.at(op.getCryptoContext());
+  auto ct = ciphertexts.at(op.getCiphertext());
+  TIME_OPERATION("KeySwitchDown", op.getResult(), cc->KeySwitchDown(ct));
+}
+
+void Interpreter::visit(AddExtOp op) {
```
asraa: ah! sorry for not noticing this earlier, but OpenFHE API doesn't have a specific Add/Mul operation for extended basis - so is it worth mirroring that in the openfhe dialect at all? Maybe they can just be omitted

PR #2593, opened 2026-01-30
RelinearizeOp lowering to transform in CKKS Dialect

crockeea:
This commit reorganizes CKKS lowerings a bit in preparation for a full implementation of CKKS in HEIR.

As a warm up (for me; this is my first PR) I change the RelinearizeOp lowering (CKKS->Polynomial) into a CKKS transform. This required the addition of a few new CKKSOps. KeySwitchInner was raised from Polynomial dialect to the CKKS dialect. 

I also introduced a new LWE type: LWERingElt representing an (abstract) ring element. This allows ops to deal with ring elements without worrying about their *representation* (e.g., as a polynomial or in evaluation form).

Since my next PR will focus on a transformation for KeySwitchInner, I didn't bother fixing up what we had written before; I just deleted it.

As this is my first PR, I strongly encourage nits. Pay special attention to build files, as I have no idea what I'm doing.

### DIFF
--- a/lib/Dialect/CKKS/IR/CKKSOps.cpp
+++ b/lib/Dialect/CKKS/IR/CKKSOps.cpp
@@ -102,6 +102,67 @@ LogicalResult LevelReduceOp::inferReturnTypes(
   return lwe::inferLevelReduceOpReturnTypes(ctx, adaptor, inferredReturnTypes);
 }
 
+LogicalResult KeySwitchInnerOp::inferReturnTypes(
+    MLIRContext* ctx, std::optional<Location>, ValueRange operands,
+    DictionaryAttr attrs, mlir::OpaqueProperties properties,
+    mlir::RegionRange regions, SmallVectorImpl<Type>& results) {
+  KeySwitchInnerOpAdaptor op(operands, attrs, properties, regions);
+  lwe::LWERingEltType ringEltType =
+      cast<lwe::LWERingEltType>(op.getValue().getType());
+  results.push_back(ringEltType);
+  results.push_back(ringEltType);
+  return success();
+}
+
+LogicalResult ExtractCoeffOp::inferReturnTypes(
+    MLIRContext* ctx, std::optional<Location>, ValueRange operands,
+    DictionaryAttr attrs, mlir::OpaqueProperties properties,
+    mlir::RegionRange regions, SmallVectorImpl<Type>& results) {
+  ExtractCoeffOpAdaptor op(operands, attrs, properties, regions);
+
+  lwe::LWECiphertextType ctType =
+      cast<lwe::LWECiphertextType>(op.getValue().getType());
+  polynomial::RingAttr ringAttr = ctType.getCiphertextSpace().getRing();
+  lwe::LWERingEltType outputType =
+      lwe::LWERingEltType::get(ctx, ringAttr, ctType.getModulusChain());
+
+  results.push_back(outputType);
+  return success();
+}
+
+LogicalResult KeySwitchInnerOp::verify() {
+  // TODO(#2157): check the ksk's RNS chain extends the value's RNS chain.
+  return success();
+}
+
+LogicalResult ExtractCoeffOp::verify() {
+  int numCTCoeffs = this->getValue().getType().getCiphertextSpace().getSize();
+  int idx = this->getIndex().getZExtValue();
+
+  if (idx < 0) {
+    return emitOpError() << "index " << idx << " cannot be negative";
+  }
+
+  if (idx >= numCTCoeffs) {
+    return emitOpError()
+           << "index " << idx
+           << " must be smaller than the number of ciphertext components "
+           << numCTCoeffs;
+  }
+
+  return success();
+}
+
+LogicalResult FromCoeffsOp::verify() {
+  int numCoeffs = this->getCoeffs().size();
+  if (numCoeffs < 1) {
+    return emitOpError()
+           << "Ciphertexts must have at least two components; got "
+           << numCoeffs;
+  }
+  return success();
+}
+
 void MulPlainOp::getCanonicalizationPatterns(RewritePatternSet& results,
                                              MLIRContext* context) {
   results.add<lwe::PutCiphertextInFirstOperand<MulPlainOp>>(context);
--- a/lib/Dialect/CKKS/IR/CKKSOps.td
+++ b/lib/Dialect/CKKS/IR/CKKSOps.td
@@ -226,4 +226,69 @@ def CKKS_BootstrapOp : CKKS_Op<"bootstrap", [ElementwiseMappable]> {
   let assemblyFormat = "operands attr-dict `:` qualified(type($input)) `->` qualified(type($output))" ;
 }
 
+def CKKS_KeySwitchInnerOp : CKKS_Op<"key_switch_inner", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Fundamental key-switching kernel.";
+  let description = [{
+    The key-switching key is the encryption of a key `s_{in}` under the key `s_{out}`.
+    The input `value` is a ring element of a ciphertext that gets multiplied with
+    `s_{in}`. KeySwitchInner outputs a linear ciphertext `ct'` such that decrypting
+    `ct'` under `s_{out}` is `value * s_{in}`. By adding `ct'` to the components
+    of the original ciphertext that do are *not* multiplied by `s_{in}`, we obtain
+    a ciphertext that encrypts the same message as the original ciphertext, but under
+    `s_{out}`.
+
+    Concretely, for relinearization, a ciphertext `[c_0, c_1, c_2]` decrypts as
+    `c_0 + c_1*s + c_2*s^2`, and the key-switch key is an encryption of `s^2` under `s`.
+    Then we apply `KeySwitchInner` to `c_2`, which produces `[c_0', c_1']`, where
+    `c_0'+c_1'*s = c_2 * s^2`. Then relinearization outputs `[c_0+c_0', c_1+c_1']`.
+
+    This operation is intended to be an internal implementation detail of
+    higher-level ciphertext operations such as `ckks.relinearize`, isolated
+    here for reuse among multiple op lowerings.
+  }];
+  let arguments = (ins
+      LWERingElt:$value,
+      RankedTensorOf<[LWECiphertext]>: $keySwitchingKey
+  );
+  let results = (outs
+    LWERingElt:$constTerm,
+    LWERingElt:$linearTerm
+  );
+  let hasVerifier = 1;
+}
+
+def CKKS_ExtractCoeffOp : CKKS_Op<"extract_coeff", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Extract a ciphertext component";
+  let description = [{
+    A ciphertext can be viewed as a polynomial in the indeterminant `S`,
+    the abstract secret key. For example, a ciphertext with two components
+    `[c0, c1]` represents the polynomial `c0 + c1 * S`. This op returns `c_i`.
+  }];
+  let arguments = (ins
+      LWECiphertext:$value,
+      IndexAttr:$index
+  );
+  let results = (outs
+    LWERingElt:$output
+  );
+  let hasVerifier = 1;
+}
+
+def CKKS_FromCoeffsOp : CKKS_Op<"from_coeffs", [Pure]> {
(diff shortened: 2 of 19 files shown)


crockeea:
I'm working on the CLA. I copied the structure of the transform (using Passes.\*, Patterns.\*, and \<MyTransform\>.\* from TensorExt/Transforms, but I can't say I entirely understand the purpose of splitting it up this way. In particular, it's not clear that Patterns.\* adds anything, and maybe that code should be incorporated directly into \<MyTransform\>.\*. Thoughts?

AlexanderViand:
Woohoo, nice to see your first PR 🎉 Since you’re asking for nits: I noticed that the pre-commit checks are failing, so you probably need to do https://heir.dev/docs/development/#pre-commit :)

> I'm working on the CLA. I copied the structure of the transform (using Passes.*, Patterns.*, and <MyTransform>.* from TensorExt/Transforms, but I can't say I entirely understand the purpose of splitting it up this way. In particular, it's not clear that Patterns.* adds anything, and maybe that code should be incorporated directly into <MyTransform>.*. Thoughts?

The separate files are mostly an artifact of the TableGen (*.td) to C++ conversion, where it’s usually easier to have every type of TableGen definable thing (Dialect(s), Operations, Types, Passes, Patterns, etc) in their own file, since they all need to be passed through the generator slightly differently and will each generate a .h.inc/.cpp.inc file. Even then it’s not technically necessary, and in your case where you don’t use Patterns.td since everything is defined directly in C++, it really is a bit redundant.

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -226,4 +226,59 @@ def CKKS_BootstrapOp : CKKS_Op<"bootstrap", [ElementwiseMappable]> {
   let assemblyFormat = "operands attr-dict `:` qualified(type($input)) `->` qualified(type($output))" ;
 }
 
+def CKKS_KeySwitchInnerOp : CKKS_Op<"key_switch_inner", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Key switch on an RNS polynomial component";
```
j2kun:
```suggestion
  let summary = "Key switch on a CKKS ciphertext component";
```

nit: since the input is an abstract ring element, describing it as a polynomial mixes abstractions. Perhaps you could say in the description: for a standard RNS CKKS implementation, the operand `value` is a single RNS polynomial component.

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -226,4 +226,59 @@ def CKKS_BootstrapOp : CKKS_Op<"bootstrap", [ElementwiseMappable]> {
...
+  );
+  let hasVerifier = 1;
+}
+
+def CKKS_ExtractCoeffOp : CKKS_Op<"extract_coeff", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Extract a ciphertext component";
+  let description = [{
+    A ciphertext can be viewed as a polynomial in the indeterminant S,
+    the abstract secret key. For example, a ciphertext with two components
+    [c0, c1] represents the polynomial c0 + c1 * S. This op returns c_i.
```
j2kun:
```suggestion
    `[c0, c1]` represents the polynomial `c0 + c1 * S`. This op returns `c_i`.
```

Note: when these docs are rendered on the website, they have MathJAX available, so if you like you can use TeX mathmode syntax here instead.

Example: https://github.com/google/heir/blob/8116223c2d9fbf276d92a494901809d15d91e092/lib/Dialect/LWE/IR/LWEAttributes.td#L148-L151
And rendered: https://heir.dev/docs/dialects/lwe/#inversecanonicalencodingattr

lib/Dialect/CKKS/IR/CKKSOps.cpp
```diff
@@ -102,6 +102,55 @@ LogicalResult LevelReduceOp::inferReturnTypes(
...
+  results.push_back(outputType);
+  return success();
+}
+
+LogicalResult KeySwitchInnerOp::verify() {
+  // TODO(#2157): check the ksk's RNS chain extends the value's RNS chain.
+  return success();
+}
+
+LogicalResult ExtractCoeffOp::verify() {
```
j2kun, line 138: Do you expect these verifiers to do anything? If not, you can remove the `let hasVerifier = 1;` from the op definition and delete them. Mainly I think the verifier would check that the result type agrees with the component types of the input ciphertext, and vice versa for `CTFromCoeffsOp::verify`

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -226,4 +226,59 @@ def CKKS_BootstrapOp : CKKS_Op<"bootstrap", [ElementwiseMappable]> {
...
+      LWECiphertext:$value,
+      IndexAttr:$index
+  );
+  let results = (outs
+    LWERingElt:$output
+  );
+  let hasVerifier = 1;
+}
+
+def CKKS_CTFromCoeffsOp : CKKS_Op<"ct_from_coeffs", [Pure]> {
```
j2kun:
```suggestion
def CKKS_FromCoeffsOp : CKKS_Op<"from_coeffs", [Pure]> {
```

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -226,4 +226,59 @@ def CKKS_BootstrapOp : CKKS_Op<"bootstrap", [ElementwiseMappable]> {
...
+  let summary = "Create a ciphertext from a vector of components";
+  let description = [{
+    A ciphertext can be viewed as a polynomial in the indeterminant S,
+    the abstract secret key. For example, a ciphertext with two components
+    [c0, c1] represents the polynomial c0 + c1 * S. Given [c0, c1], this op
+    return the corresponding ciphertext.
```
j2kun:
```suggestion
    returns the corresponding ciphertext.
```

lib/Dialect/CKKS/IR/CKKSOps.cpp
```diff
@@ -102,6 +102,55 @@ LogicalResult LevelReduceOp::inferReturnTypes(
...
+    mlir::RegionRange regions, SmallVectorImpl<Type>& results) {
+
+  KeySwitchInnerOpAdaptor op(operands, attrs, properties, regions);
+
+  RankedTensorType tensorType = cast<RankedTensorType>(op.getKeySwitchingKey().getType());
+  lwe::LWECiphertextType ctType = cast<lwe::LWECiphertextType>(tensorType.getElementType());
+  lwe::LWERingEltType ringEltType = cast<lwe::LWERingEltType>(op.getValue().getType());
+  polynomial::RingAttr ringAttr = ringEltType.getRing();
+  lwe::CiphertextSpaceAttr ctAttr = lwe::CiphertextSpaceAttr::get(ctx, ringAttr, ctType.getCiphertextSpace().getEncryptionType(), 2);
+  // Using ctType.getModulusChain is potentially suspect here
```
j2kun: I agree but in this case I think it's fine. The key-switching key may have a different RNS type, depending on what flavor of slicing we are going to assert it should have as input to this op, but I think the modulus chain will be the same as the original ciphertext that the input RingElt came from.

(If this ends up not working out, the alternative would be to not infer the return type, and instead just run a verifier.)

lib/Dialect/CKKS/Transforms/Passes.h
```diff
@@ -0,0 +1,29 @@
...
+#include "lib/Dialect/CKKS/Transforms/Passes.h.inc"
+
+struct DecomposeRelinearizePattern
+    : public OpRewritePattern<RelinearizeOp> {
+  using OpRewritePattern<RelinearizeOp>::OpRewritePattern;
+
+ public:
+  LogicalResult matchAndRewrite(RelinearizeOp op,
+                                PatternRewriter& rewriter) const override;
+};
```
j2kun:
```suggestion
#define GEN_PASS_REGISTRATION
#include "lib/Dialect/CKKS/Transforms/Passes.h.inc"
```

Shouldn't need the pattern declaration here since it's already in Patterns.h

lib/Dialect/CKKS/Transforms/Passes.td
```diff
@@ -0,0 +1,11 @@
+#ifndef LIB_DIALECT_CKKS_TRANSFORMS_PASSES_TD_
+#define LIB_DIALECT_CKKS_TRANSFORMS_PASSES_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def DecomposeRelinearize : Pass<"ckks-decompose-relinearize"> {
+  let summary = "Decomposes ckks:relinearize into ckks:keyswitchinner and addition";
+  let dependentDialects = ["mlir::heir::ckks::CKKSDialect"];
```
j2kun, line 11: One recent practice I want to enforce is to have each pass mark a designated lit test as a "doctest." The tooling is set up so that, if you include a line like this in the `description` field of a pass:

https://github.com/google/heir/blob/8116223c2d9fbf276d92a494901809d15d91e092/lib/Transforms/PopulateScale/PopulateScale.td#L44

Then the documentation building step will build heir-opt, run the pass, and then show the command/input/output in the resulting documentation entry. For the example linked above, this is what's rendered: https://heir.dev/docs/passes/#example-28

You should be able to use `tests/Dialect/CKKS/Transforms/relin.mlir` for this. The only real restriction is that the doctest should only have one test in the file.

```
   (* example filepath=tests/Dialect/CKKS/Transforms/relin.mlir *)
```

lib/Dialect/CKKS/Transforms/Patterns.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace ckks {
+
+LogicalResult DecomposeRelinearizePattern::matchAndRewrite(
+    RelinearizeOp op, PatternRewriter& rewriter) const {
+  ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+  // TODO: should these go in the verifier instead?
```
j2kun:
```suggestion
```

I think not yet, since the operand is truly marked as optional. Eventually I'd expect it becomes non-optional and then we have to update whatever passes that insert the relinearize op to manage the ksk input, at which point this clause would be removed.

tests/Dialect/CKKS/Transforms/decompose_relinearize.mlir
j2kun, line 1: Rename this file decompose_relinearize.mlir after the pass it tests.

ai-mannamalai:
> I'm working on the CLA. I copied the structure of the transform (using Passes.*, Patterns.*, and <MyTransform>.* from TensorExt/Transforms, but I can't say I entirely understand the purpose of splitting it up this way. In particular, it's not clear that Patterns.* adds anything, and maybe that code should be incorporated directly into <MyTransform>.*. Thoughts?

I'm finding out on Cornami side. I should have an update by EOD.

lib/Dialect/CKKS/IR/CKKSOps.cpp
crockeea: I don't have much intuition here. I would think that a reasonable verifier for `ExtractCoeffOp` might be something like "ensure that the index is valid". And for `FromCoeffsOp` that "there are at least two components". Do those sound reasonable?

j2kun: Those sound reasonable to me.

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -251,7 +251,8 @@ def CKKS_KeySwitchInnerOp : CKKS_Op<"key_switch_inner", [Pure, DeclareOpInterfac
       RankedTensorOf<[LWECiphertext]>: $keySwitchingKey
   );
   let results = (outs
-    LWECiphertext:$output
+    LWERingElt:$constTerm,
+    LWERingElt:$linearTerm
```
j2kun, line 255: Good call. I missed that change from how the original op was.

crockeea: Well, I intentionally changed it until I got well into writing KeySwitch and realized it was a poor decision :)

j2kun: And if I had done my job reviewing it, I would have noticed and asked about the change :)

j2kun (approved):
I'm happy with it.

The last step is for you to squash all the commits into one, and then I can mark it as `pull_ready`

crockeea:
Done!

j2kun (approved):

PR #2594, opened 2026-01-31
Cornami MX2 systolic array backend

ai-mannamalai:
Cornami MX2 systolic array backend for TFHE and CKKS schemes lowering  into SCIFR Bool/Ckks dialects

- map the said dialect into SCIFR Concrete and SCIFR Liberate topologies suitable for hardware

cornami backend: move tests into common area

move scifr-opt and scifr-translate into the parent HEIR tools to avoid rebuild etc.

- update heir-translate to include HEIR_BACKEND_CORNAMI and Target SCIFRBoolEmitter

Cornami: HEIR dialect SCIFRBool emitter and optimizer merged into heir-opt or heir-translate
```bash
$ bazelisk test //tests/Dialect/SCIFRBool/... //tests/Dialect/SCIFRCkks/...
```

### DIFF
--- a/lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
+++ b/lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
@@ -1,0 +1,166 @@
+#include "lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.h"
+
+#include <cstdint>
+#include <ratio>
+#include <utility>
+
+#include "lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.h.inc"
+#include "lib/Dialect/CGGI/IR/CGGIDialect.h"
+#include "lib/Dialect/CGGI/IR/CGGIOps.h"
+#include "lib/Dialect/LWE/IR/LWEDialect.h"
+#include "lib/Dialect/LWE/IR/LWEOps.h"
+#include "lib/Dialect/LWE/IR/LWETypes.h"
+#include "lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.h"
+#include "lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.h"
+#include "lib/Dialect/SCIFRBool/IR/SCIFRBoolTypes.h"
+#include "lib/Utils/ConversionUtils.h"
+#include "lib/Utils/Utils.h"
+#include "llvm/include/llvm/ADT/SmallVector.h"           // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"   // from @llvm-project
+#include "mlir/include/mlir/Dialect/MemRef/IR/MemRef.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypeInterfaces.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypes.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"   // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/Visitors.h"               // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"              // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"     // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
+#define DEBUG_TYPE "cggi-to-scifrbool"
+
+using namespace mlir;
+using namespace heir;
+
+namespace mlir {
+namespace cornami {
+namespace scifrbool {
+
+#define GEN_PASS_DEF_CGGITOSCIFRBOOL
+#include "lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.h.inc"
+
+class CGGIToSCIFRBoolTypeConverter : public TypeConverter {
+ public:
+  CGGIToSCIFRBoolTypeConverter(MLIRContext *ctx) {
+    addConversion([](Type type) { return type; });
+    addConversion([this](ShapedType type) -> Type {
+      return type.cloneWith(type.getShape(),
+                            this->convertType(type.getElementType()));
+    });
+  }
+};
+
+/// Convert a func by adding key arguments
+struct AddKeyArgs : public OpConversionPattern<func::FuncOp> {
+  AddKeyArgs(mlir::MLIRContext *context)
+      : OpConversionPattern<func::FuncOp>(context, /* benefit= */ 2) {}
+
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      func::FuncOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    if (!containsDialects<lwe::LWEDialect, cggi::CGGIDialect>(op)) {
+      return failure();
+    }
+
+    Type bskType = scifrbool::SCIFRBoolBootstrapKeyType::get(getContext());
+    Type kskType = scifrbool::SCIFRBoolKeySwitchKeyType::get(getContext());
+    Type serverParamsType =
+        scifrbool::SCIFRBoolServerParametersType::get(getContext());
+
+    rewriter.modifyOpInPlace(op, [&] {
+      (void)op.insertArgument(0, bskType, nullptr, op.getLoc());
+      (void)op.insertArgument(1, kskType, nullptr, op.getLoc());
+      (void)op.insertArgument(2, serverParamsType, nullptr, op.getLoc());
+    });
+    return success();
+  }
+};
+
+template <typename BinOp, typename SCIFRBoolBinOp>
+struct ConvertCGGITRBBinOp : public OpConversionPattern<BinOp> {
+  using OpConversionPattern<BinOp>::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      BinOp op, typename BinOp::Adaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op->getLoc(), rewriter);
+
+    rewriter.replaceOp(
+        op, SCIFRBoolBinOp::create(b, adaptor.getLhs(), adaptor.getRhs()));
+    return success();
+  }
+};
+
+struct ConvertBoolNotOp : public OpConversionPattern<cggi::NotOp> {
+  ConvertBoolNotOp(mlir::MLIRContext *context)
+      : OpConversionPattern<cggi::NotOp>(context) {}
+
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      cggi::NotOp op, OpAdaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op->getLoc(), rewriter);
+
+    rewriter.replaceOp(op, b.create<scifrbool::NotOp>(adaptor.getInput()));
+    return success();
+  }
+};
+
+struct CGGIToSCIFRBool : impl::CGGIToSCIFRBoolBase<CGGIToSCIFRBool> {
+  using CGGIToSCIFRBoolBase::CGGIToSCIFRBoolBase;
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    auto *op = getOperation();
+
+    CGGIToSCIFRBoolTypeConverter typeConverter(context);
+    RewritePatternSet patterns(context);
+    ConversionTarget target(*context);
+    addStructuralConversionPatterns(typeConverter, patterns, target);
+
+    target.addLegalDialect<scifrbool::SCIFRBoolDialect>();
+    target.addIllegalDialect<cggi::CGGIDialect>();
+    target.addIllegalDialect<lwe::LWEDialect>();
+
+    target.addDynamicallyLegalOp<memref::AllocOp, memref::DeallocOp,
+                                 memref::StoreOp, memref::LoadOp,
+                                 memref::SubViewOp, memref::CopyOp,
+                                 tensor::FromElementsOp, tensor::ExtractOp>(
+        [&](Operation *op) {
+          return typeConverter.isLegal(op->getOperandTypes()) &&
+                 typeConverter.isLegal(op->getResultTypes());
+        });
+    target.addDynamicallyLegalOp<func::FuncOp>([&](func::FuncOp op) {
+      bool hasKeyArg = op.getFunctionType().getNumInputs() > 0 &&
+                       mlir::isa<scifrbool::SCIFRBoolBootstrapKeyType>(
+                           *op.getFunctionType().getInputs().begin());
+      return (!containsDialects<lwe::LWEDialect, cggi::CGGIDialect>(op) ||
+              hasKeyArg);
+    });
+
+    target.addLegalOp<mlir::arith::ConstantOp>();
+
+    patterns.add<AddKeyArgs, ConvertCGGITRBBinOp<cggi::AndOp, scifrbool::AndOp>,
+                 ConvertCGGITRBBinOp<cggi::NandOp, scifrbool::NandOp>,
+                 ConvertCGGITRBBinOp<cggi::OrOp, scifrbool::OrOp>,
+                 ConvertCGGITRBBinOp<cggi::NorOp, scifrbool::NorOp>,
+                 ConvertCGGITRBBinOp<cggi::XorOp, scifrbool::XorOp>,
+                 ConvertCGGITRBBinOp<cggi::XNorOp, scifrbool::XNorOp>,
+                 ConvertBoolNotOp, ConvertAny<memref::AllocOp>,
+                 ConvertAny<memref::DeallocOp>, ConvertAny<memref::StoreOp>,
+                 ConvertAny<memref::LoadOp>, ConvertAny<memref::SubViewOp>,
+                 ConvertAny<memref::CopyOp>, ConvertAny<tensor::FromElementsOp>,
+                 ConvertAny<tensor::ExtractOp> >(typeConverter, context);
+
+    if (failed(applyPartialConversion(op, target, std::move(patterns)))) {
+      return signalPassFailure();
+    }
+  }
+};
+}  // namespace scifrbool
+}  // namespace cornami
+}  // namespace mlir
--- a/lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.td
+++ b/lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.td
@@ -1,0 +1,27 @@
+//===- SCIFRBoolDialect.td - Cornami SCIFR Boolean Dialect -*- tablegen -*-===//
+//     This dialect is an exit dialect for targeting Cornami's FracTLCore(R) hardware accelerator for CGGI programs.
+//===----------------------------------------------------------------------===//
+
+// Ref: Resource Estimation of CGGI and CKKS scheme workloads on
+// FracTLcore Computing Fabric, ArXiV: https://arxiv.org/abs/2510.16025, 2025.
+
+#ifndef SCIFRBOOL_DIALECT
+#define SCIFRBOOL_DIALECT
+
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/OpBase.td"
+include "mlir/Pass/PassBase.td"
+
+//===----------------------------------------------------------------------===//
+// Cornami SCIFR Boolean dialect definition.
+//===----------------------------------------------------------------------===//
+
+def SCIFRBool_Dialect : Dialect {
+    let name = "scifrbool";
+    let summary = "Cornami SCIFR Boolean dialect for FHE Applications.";
+    let cppNamespace = "::mlir::cornami::scifrbool";
+    let useDefaultTypePrinterParser = 1;
+}
+
+
+#endif // SCIFRBOOL_DIALECT
--- a/lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
+++ b/lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
@@ -1,0 +1,86 @@
+//===- SCIFRBoolOps.td - SCIFRBool Dialect ops -----------*- tablegen -*-===//
+
+#ifndef SCIFRBool_OPS
+#define SCIFRBool_OPS
+
+include "SCIFRBoolDialect.td"
+include "mlir/IR/AttrTypeBase.td"
+
+include "mlir/IR/OpBase.td"
+include "mlir/IR/BuiltinAttributes.td"
+include "mlir/IR/CommonAttrConstraints.td"
+include "mlir/Interfaces/InferTypeOpInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "mlir/Interfaces/CallInterfaces.td"
+include "mlir/Interfaces/CastInterfaces.td"
+
+include "lib/Dialect/HEIRInterfaces.td"
+
+include "lib/Dialect/LWE/IR/LWETypes.td"
+
+class SCIFRBool_Op<string mnemonic, list<Trait> traits = []> :
+        Op<SCIFRBool_Dialect, mnemonic, traits> {
+  let assemblyFormat = [{
+    operands attr-dict `:` functional-type(operands, results)
+  }];
+  let cppNamespace = "::mlir::cornami::scifrbool";
+}
+
+// --- Operations for a gate-bootstrapping API of a SCIFRBool library ---
+class SCIFRBool_BinaryGateOp<string mnemonic>
+  : SCIFRBool_Op<mnemonic, [
+    Pure,
+    Commutative,
+    SameOperandsAndResultType
+]> {
+  let arguments = (ins LWECiphertext:$lhs, LWECiphertext:$rhs);
+  let results = (outs LWECiphertext:$output);
+  let assemblyFormat = "operands attr-dict `:` qualified(type($output))" ;
+}
+
+def SCIFRBool_AndOp : SCIFRBool_BinaryGateOp<"and"> { let summary = "Logical AND of two ciphertexts."; }
+def SCIFRBool_NandOp : SCIFRBool_BinaryGateOp<"nand"> { let summary = "Logical NAND of two ciphertexts."; }
+def SCIFRBool_NorOp  : SCIFRBool_BinaryGateOp<"nor">  { let summary = "Logical NOR of two ciphertexts."; }
+def SCIFRBool_OrOp  : SCIFRBool_BinaryGateOp<"or">  { let summary = "Logical OR of two ciphertexts."; }
+def SCIFRBool_XorOp : SCIFRBool_BinaryGateOp<"xor"> { let summary = "Logical XOR of two ciphertexts."; }
+def SCIFRBool_XNorOp : SCIFRBool_BinaryGateOp<"xnor"> { let summary = "Logical XNOR of two ciphertexts."; }
+
+def SCIFRBool_NotOp : SCIFRBool_Op<"not", [
+    Pure,
+    Involution,
+    SameOperandsAndResultType,
+]> {
+  let arguments = (ins LWECiphertext:$input);
+  let results = (outs LWECiphertext:$output);
+  let assemblyFormat = "operands attr-dict `:` qualified(type($output))";
+  let summary = "Logical NOT of two ciphertexts";
+}
+
+//===----------------------------------------------------------------------===//
+def SCIFRBOOL_AnyNumber : AnyTypeOf<[AnyFloat],
+                               "number">;
+
+
+// Operator for kernel outlining pass; this pass will contain all the ops
+// that are able to be scheduled on the currently available hardware resources
+def SCIFRBOOL_SectionOp : SCIFRBool_Op<"section",[NoTerminator]> {
+  let summary = "Wraps a single operation";
+  let description = [{
+      Section op is intended to handle kernel outlining, move kernels of instructions. Section contains the code scheduled to Fractal Array.
+  }];
+
+  let arguments = (ins
+    Variadic<AnyType>:$input
+  );
+
+  let results = (outs
+    Variadic<AnyType>:$output
+  );
+
+  let regions = (region AnyRegion:$body);
+
+  let assemblyFormat = "`(` operands `)` $body attr-dict `:` functional-type(operands, results)";
+  let hasCustomAssemblyFormat = 1;
+}
+
+#endif // SCIFRBool_OpS
--- a/lib/Dialect/SCIFRBool/IR/SCIFRBoolTypes.td
+++ b/lib/Dialect/SCIFRBool/IR/SCIFRBoolTypes.td
@@ -1,0 +1,32 @@
+//===- SCIFRBoolTypes.td - SCIFRBool Dialect Types -----------*- tablegen -*-===//
+
+#ifndef SCIFRBool_TYPES
+#define SCIFRBool_TYPES
+
+include "SCIFRBoolDialect.td"
+
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/BuiltinTypeInterfaces.td"
+include "mlir/IR/CommonTypeConstraints.td"
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/InferTypeOpInterface.td"
+
+class SCIFRBool_Type<string name, string typeMnemonic, list<Trait> traits = []>
+    : TypeDef<SCIFRBool_Dialect, name, traits> {
+  let mnemonic = typeMnemonic;
+}
+
+def SCIFRBool_Bootstrap_Key : SCIFRBool_Type<"SCIFRBoolBootstrapKey", "bootstrap_key"> {
+  let summary = "The key required to perform Bootstrap operation in SCIFRBool.";
+}
+
+def SCIFRBool_Key_Switch_Key : SCIFRBool_Type<"SCIFRBoolKeySwitchKey", "key_switch_key"> {
+  let summary = "The key required to perform Keyswitch operation in SCIFRBool.";
+}
+
+def SCIFRBool_Server_Parameters : SCIFRBool_Type<"SCIFRBoolServerParameters", "server_parameters"> {
+  let summary = "The server parameters required to map to cornami hardware in SCIFRBool.";
+}
+
+#endif // SCIFRBool_Types
--- a/lib/Dialect/SCIFRBool/Transforms/DialectResourceMap.cpp
+++ b/lib/Dialect/SCIFRBool/Transforms/DialectResourceMap.cpp
@@ -1,0 +1,27 @@
+#include "lib/Dialect/SCIFRBool/Transforms/DialectResourceMap.h"
+
+namespace mlir {
+namespace cornami {
+
+void dump(heir::graph::Graph<Operation*>& graph) {
+  size_t n_vertices = graph.getVertices().size();
+  size_t n_out_edges = 0;
+  size_t n_in_edges = 0;
+  for (auto vertexptr : graph.getVertices()) {
+    n_out_edges += graph.getOutDegree(vertexptr);
+  }
+  for (auto vertexptr : graph.getVertices()) {
+    n_in_edges += graph.getInDegree(vertexptr);
+  }
+  auto n_sources = graph.getSources().size();
+  auto n_sinks = graph.getSinks().size();
+  llvm::outs() << "====== graph @ display =====\n";
+  llvm::outs() << "vertices = " << n_vertices << ", in-edges = " << n_in_edges
+               << ", out-edges = " << n_out_edges << ", sinks = " << n_sinks
+               << ", source = " << n_sources << "\n";
+  llvm::outs() << "====== xxxxxxxxxxxxxxx =====\n";
+  return;
+}
+
+}  // namespace cornami
+}  // namespace mlir
--- a/lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
+++ b/lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
@@ -1,0 +1,99 @@
+#include "lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.h"
+
+#include <cstdint>
+#include <unordered_set>
+
+#include "lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.h"
+#include "lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.h"
+#include "llvm/include/llvm/Support/Debug.h"            // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/ValueRange.h"            // from @llvm-project
+#include "mlir/include/mlir/IR/Visitors.h"              // from @llvm-project
+#include "mlir/include/mlir/Transforms/Passes.h"        // from @llvm-project
+
+// clang-format off
+#include "lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.h.inc"
+// clang-format on
+
+namespace mlir {
+namespace cornami {
+namespace scifrbool {
+
+#define GEN_PASS_DEF_REPLACEOPWITHSECTION
+#include "lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.h.inc"
+
+struct ReplaceOpWithSection
+    : impl::ReplaceOpWithSectionBase<ReplaceOpWithSection> {
+  using ReplaceOpWithSectionBase::ReplaceOpWithSectionBase;
+
+  // TODO: Fuse Not op to the previous section greedily
+  void runOnOperation() override {
+    Operation* op = getOperation();
+    MLIRContext& context = getContext();
+
+    op->walk([&](func::FuncOp op) { convertFunc(op, context); });
+  }
+  std::string getValueName(mlir::Value val) {
+    std::string opName;
+    llvm::raw_string_ostream ss(opName);
+    mlir::OpPrintingFlags pFlags;
+
+    val.printAsOperand(ss, pFlags);
+    return opName;
+  }
+
+  void convertFunc(Operation* funcOp, MLIRContext& context) {
+    uint32_t nChipsPerOp = 2;
+    uint32_t nOpsPerSection = nChips / nChipsPerOp;
+    std::vector<Operation*> opList;
+    funcOp->walk([&](Operation* op) {
+      if (isa<SCIFRBoolDialect>(op->getDialect()) && !isa<SectionOp>(op)) {
+        opList.push_back(op);
+      }
+    });
+    OpBuilder builder(&context);
+
+    for (uint32_t i = 0; i < (uint32_t)opList.size(); i += nOpsPerSection) {
+      auto startInsertionPoint = opList[i];
+      // Collect the inputs and input types of the original operation
+      SmallVector<Type, 4> inputTypes;
+      SmallVector<Value, 4> inputTensors;
+      std::unordered_set<std::string> inputNames;
+      for (uint32_t j = 0; j < nOpsPerSection && i + j < opList.size(); j++) {
+        Operation* scifrOp = opList[i + j];
+        // Assuming the operation has operands (inputs)
+        for (Value operand : scifrOp->getOperands()) {
+          std::string opName = getValueName(operand);
+          if (inputNames.find(opName) == inputNames.end()) {
+            inputNames.insert(opName);
+            inputTypes.push_back(operand.getType());
+            // TODO: Do not add operands that will be created inside this
+            // section as arguments
+            inputTensors.push_back(operand);
+          }
+        }
+      }
+      builder.setInsertionPoint(startInsertionPoint);
+      // Create a new SectionOp to wrap the original operator
+      Operation* sectionOp = SectionOp::create(
+          builder, startInsertionPoint->getLoc(), inputTypes[0], inputTensors);
+
+      Region& region = sectionOp->getRegion(0);
+      builder.createBlock(&region);
+      Block* sectionOpBlock = &region.front();
+
+      builder.setInsertionPointToStart(sectionOpBlock);
+      for (uint32_t j = 0; j < nOpsPerSection && i + j < opList.size(); j++) {
+        Operation* scifrOp = opList[i + j];
+        Operation* scifrOpCopy = builder.clone(*scifrOp);
+        scifrOpCopy->getBlock()->moveBefore(sectionOpBlock);
+        scifrOp->replaceAllUsesWith(sectionOp->getResults());
+        scifrOp->erase();
+      }
+    }
+  }
+};  // struct ReplaceOpWithSection
+
+}  // namespace scifrbool
+}  // namespace cornami
+}  // namespace mlir
--- a/lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.td
+++ b/lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.td
@@ -1,0 +1,22 @@
+#ifndef LIB_DIALECT_SCIFRBOOL_CONVERSIONS_REPLACEOPWITHSECTION_TD_
+#define LIB_DIALECT_SCIFRBOOL_CONVERSIONS_REPLACEOPWITHSECTION_TD_
+
+include "mlir/Pass/PassBase.td"
+
+def ReplaceOpWithSection : Pass<"replace-op-with-section"> {
+  let summary = "Move each CGGI/SCIFRBool Operation inside a Section Operation";
+  let description = [{
+      Move each CGGI/SCIFRBool Operation inside a Section Operation.
+      We generate mlir with sections for larger boolean circuits.
+  }];
+
+  let dependentDialects = [
+    "mlir::heir::cggi::CGGIDialect",
+    "mlir::cornami::scifrbool::SCIFRBoolDialect"
+  ];
+  let options = [
+    Option<"nChips", "nChips", "uint32_t", "2", "number of available chips">
+  ];
+}
+
+#endif  // LIB_DIALECT_SCIFRBOOL_CONVERSIONS_REPLACEOPWITHSECTION_TD_
--- a/lib/Dialect/SCIFRCkks/IR/SCIFRCkksDialect.h
+++ b/lib/Dialect/SCIFRCkks/IR/SCIFRCkksDialect.h
@@ -1,0 +1,23 @@
+#ifndef SCIFRCKKS_SCIFRCKKSDIALECT_H
+#define SCIFRCKKS_SCIFRCKKSDIALECT_H
+
+#include "llvm/include/llvm/Support/CommandLine.h"     // from @llvm-project
+#include "llvm/include/llvm/Support/InitLLVM.h"        // from @llvm-project
+#include "llvm/include/llvm/Support/SourceMgr.h"       // from @llvm-project
+#include "llvm/include/llvm/Support/ToolOutputFile.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/Builders.h"             // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypes.h"         // from @llvm-project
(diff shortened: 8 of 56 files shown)


ai-mannamalai:
Okay

ai-mannamalai:
I will make the changes but I prefer to have one big commit.

ai-mannamalai:
@j2kun  - many of the items you have mentioned are re-worked. Please take a look

BUILD.bazel
```diff
@@ -83,3 +83,23 @@ config_setting(
     },
     visibility = ["//visibility:public"],
 )
+
+# Cornami Backend
+string_flag(
```
j2kun: I don't see a particular need for this flag. We have it for yosys because yosys often has build issues on different platforms. But the Cornami dialects and pipelines introduce no additional dependencies.

OK to remove?

tests/Dialect/SCIFRBool/Transforms/and_gate.mlir
j2kun, line 1: As far as I can tell, none of the tests in this directory are being run because of the lack of a BUILD file in this directory. The simplest build file would be:

```
load("//bazel:lit.bzl", "glob_lit_tests")

package(default_applicable_licenses = ["@heir//:license"])

glob_lit_tests(
    name = "all_tests",
    data = ["@heir//tests:test_utilities"],
    driver = "@heir//tests:run_lit.sh",
    test_file_exts = ["mlir"],
)
```

Also: 

-  I'm not sure how the cpp files in this directory as intended to be tested. Maybe turn them into CHECK directives on the relevant MLIR test files instead? 
- Some lit files in here with no FileCheck directives (this would cause a test failure if the tests were run)
- Some of the output mlir files would be treated as test files and cause a failure because they have no `// RUN` directive. Maybe these should also become CHECK lines in the corresponding input MLIR files?

scripts/cornami/README_cornami.md
j2kun, line 1: Maybe for a followup PR: if you want to build a pipeline that corresponds to the `cggi-to-cornami` pipeline, then it would make sense to put this in docs/content/en/docs/pipelines.md so the docs show up on the website.

scripts/cornami/scifr.bats
j2kun, line 1: Help me understand the purpose of this test file. It won't be run in CI as-is. Do you want help converting it to a test that combines heir-opt and heir-translate? In the simplest case, you can make a lit test that pipes heir-opt to heir-translate. Otherwise we can do something more complicated with bazel rules/macros.

lib/Dialect/SCIFRCkks/docs/SCIFRCkks.md
j2kun, line 1: This would be auto-generated on https://heir.dev, so this file can probably be deleted.

lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.td
```diff
@@ -0,0 +1,41 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// SCIFRBool Attributes.
+//===----------------------------------------------------------------------===//
+
+class SCIFRBool_Attr<string attrName, string attrMnemonic, list<Trait> traits = []>
+    : AttrDef<SCIFRBool_Dialect, attrName, traits> {
+  let mnemonic = attrMnemonic;
+}
```
j2kun:
```suggestion
```

This appears to be unused. Remove?

lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.td
```diff
@@ -0,0 +1,41 @@
...
+    let summary = "Cornami SCIFR Boolean dialect for FHE Applications.";
+    let description = [{
+MLIR Dialect for Cornami SCIFR Boolean Dialect | SCIFRBool
+    Cornami SCIFR Boolean dialect
+    Concepts
+        Section
+            Contains Operators and HBM Regions and memory transfers operations
+            Section can be reduced to FA (for executable) or default (which is not executable)
+        Operator
+            Operators are from AppStream FHE or FractlsBase only
```
j2kun: IMO this documentation section could be improved. I would expect this is the first place someone working on HEIR would look to understand what scifrbool is. Perhaps a link to the arxiv paper would help.

Maybe also lead with a high level description like "This dialect is an exit dialect for targeting Cornami's FractlBase hardware accelerator for CGGI programs." before going into details about the concepts.

lib/Dialect/SCIFRBool/IR/SCIFRBoolTypes.td
```diff
@@ -0,0 +1,36 @@
...
+class SCIFRBool_Type<string name, string typeMnemonic, list<Trait> traits = []>
+    : TypeDef<SCIFRBool_Dialect, name, traits> {
+  let mnemonic = typeMnemonic;
+}
+
+def SCIFRBool_Ciphertext : SCIFRBool_Type<"SCIFRBoolCiphertext", "ciphertext"> {
+  let summary = "A type for SCIFRBool Ciphertext";
+}
+
+def SCIFRBool_Bootstrap_Key_Standard : SCIFRBool_Type<"SCIFRBoolBootstrapKeyStandard", "bootstrap_key_standard"> {
```
j2kun: Is there a plan to add "nonstandard" bootstrap keys? Otherwise maybe drop `Standard` from the name and mnemonic.

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// --- Operations for a gate-bootstrapping API of a SCIFRBool library ---
+
+class SCIFRBool_BinaryGateOp<string mnemonic>
+  : SCIFRBool_Op<mnemonic, [
+    Pure,
+    Commutative,
+    SameOperandsAndResultType,
+    ElementwiseMappable,
```
j2kun: If it is ElementwiseMappable and Scalarizable, then the types need to be generalized to support tensors of ciphertexts (we have `LWECiphertextLike` for this but it also requires adding extra code paths to check for tensor types in your lowerings). That said, I suspect you aren't using these traits and the simplest fix is to remove them.

Same for NotOp below

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+    let results = (outs AnyTensor:$res);
+
+    let assemblyFormat = [{
+        $input attr-dict `:` type($input)
+    }];
+}
+
+//===----------------------------------------------------------------------===//
+// Operator: sigmoid
+//===----------------------------------------------------------------------===//
```
j2kun:
```suggestion
```

(it's a key switch op, and anyway the ascii art is mostly noise)

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+def SCIFRBOOL_AnyNumber : AnyTypeOf<[AnyFloat],
+                               "number">;
+
+
+def SCIFRBOOL_PBSOp : SCIFRBool_Op<"PBS", [SameOperandsAndResultType]> {
+    let summary = "Programmatic Bootstrapping (PBS) operation.";
+    let description = [{
+        The `SCIFRBOOL.PBS` creates a PBS for all inputs.
+
+        This operation takes an tensor argument and returns a tensor.
```
j2kun: How does this op specify what LUT is applied during the PBS? That should be documented here, especially if it is not part of the op itself.

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// Operator: sigmoid
+//===----------------------------------------------------------------------===//
+def SCIFRBOOL_KSOp : SCIFRBool_Op<"KS",[SameOperandsAndResultType]> {
+  let summary = "Computes Key Switching of input.";
+
+  let description = [{
+    KS : Key Switching of the operation
```
j2kun: How does this op specify the key switching key if it's not an operand?

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+
+class SCIFRBool_BinaryGateOp<string mnemonic>
+  : SCIFRBool_Op<mnemonic, [
+    Pure,
+    Commutative,
+    SameOperandsAndResultType,
+    ElementwiseMappable,
+    Scalarizable
+]> {
+  let arguments = (ins LWECiphertext:$lhs, LWECiphertext:$rhs);
```
j2kun, line 36: Should these ops be using SCIFRBool_Ciphertext instead of LWECiphertext? (Otherwise it appears SCIFRBool_Ciphertext is unused)

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+
+def SCIFRBOOL_PBSOp : SCIFRBool_Op<"PBS", [SameOperandsAndResultType]> {
+    let summary = "Programmatic Bootstrapping (PBS) operation.";
+    let description = [{
+        The `SCIFRBOOL.PBS` creates a PBS for all inputs.
+
+        This operation takes an tensor argument and returns a tensor.
+    }];
+
+    let arguments = (ins AnyTensor:$input);
```
j2kun: Can you specify more specifically what kind of tensor is allowed here? To me, a PBS operation takes as input a ciphertext, not a tensor. So is this intended to be a tensor of `SCIFRBool_Ciphertext`? If so, you can express that directly in the type constraint rather than use AnyTensor.

Same question for key switch below.

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// Operator: LinearOp
+//===----------------------------------------------------------------------===//
+def SCIFRBOOL_LinearOp : SCIFRBool_Op<"Linear",[SameOperandsAndResultType]> {
+  let summary = "Computes linear combination of the input with fixed coefficients.";
+
+  let description = [{
+    compute the linear combination of the input FHE vectors using fixed coefficients
```
j2kun: Similar to above, the "fixed coefficients" are not specified on the op, so how are they determined?

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+  );
+
+  let results = (outs
+    AnyTensor:$output
+  );
+}
+/// boolean gates in ciphertext ///
+
+def SCIFRBOOL_SectionOp : SCIFRBool_Op<"section",[NoTerminator]> {
+  let summary = "Wraps a single operation";
```
j2kun, line 67: I think this op needs a bit more explanation. I'm guessing this is related to the dialect description:

> Contains Operators and HBM Regions and memory transfers operations

I guess part of what is not clear from this spec is what the purpose of the op is. Does a section roughly correspond to dispatching set of instructions?

lib/Dialect/SCIFRCkks/docs/SCIFRCkks.md
ai-mannamalai: Done.

scripts/cornami/README_cornami.md
ai-mannamalai: sure we can do that

scripts/cornami/scifr.bats
ai-mannamalai: these can be done as simple lit tests; we just had them as shell scripts

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
ai-mannamalai: Okay done.

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
ai-mannamalai: SectionOp is container for the kernel outlining pass;

ai-mannamalai:
@j2kun - I think most of the items are addressed for a good first pass commit IMHO; lmk if this is okay from your end

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+
+    let arguments = (ins AnyTensor:$input);
+    let results = (outs AnyTensor:$res);
+
+    let assemblyFormat = [{
+        $input attr-dict `:` type($input)
+    }];
+}
+
+//WIP: low level implementation of SCIFR Bool API not exposed yet
```
j2kun: Instead of commenting it out, remove the ops from the PR. Feel free to file an issue or post a followup draft PR.

lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.td
j2kun: This comment was marked as resolved but nothing appears to have changed. Same with some other comments, such as the question about the semantics of the section op.

lib/Dialect/SCIFRCkks/IR/SCIFRCkksOps.td
```diff
@@ -0,0 +1,126 @@
...
+  let assemblyFormat = "operands attr-dict `:` qualified(type($output))";
+  let summary = "Logical NOT of two ciphertexts";
+}
+
+//===----------------------------------------------------------------------===//
+def SCIFRCkks_AnyNumber : AnyTypeOf<[AnyFloat],
+                               "number">;
+
+
+def SCIFRCkks_PBSOp : SCIFRCkks_Op<"PBS", [SameOperandsAndResultType]> {
```
j2kun: I'm seeing a lot of the same issues with the CKKS ops as the Bool ops: PBS has no lookup table operand/attribute (does Cornami's CKKS implementation support PBS?), linear transform op has no coefficients operand, operands are listed as "AnyTensor" when it should probably be restricted to a ciphertext or tensor of ciphertexts.

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,140 @@
...
+//  let results = (outs
+//    AnyTensor:$output
+//  );
+//}
+/// boolean gates in ciphertext ///
+
+// Operator for kernel outlining pass
+def SCIFRBOOL_SectionOp : SCIFRBool_Op<"section",[NoTerminator]> {
+  let summary = "Wraps a single operation";
+  let arguments = (ins
```
j2kun, line 72: Add to the description field: the section op is intended to handle kernel outlining, move kernels of instructions.

scripts/cornami/README_cornami.md
j2kun, line 1: Please remove this file for now.

lib/Backend/cornami/scifr.bats
j2kun, line 1: As mentioned before, this doesn't fit with the testing infrastructure in this project. Please remove. Let me know if you're trying to do something here that you can't figure out how to do with the existing infrastructure and we can work on a solution.

lib/Backend/cornami/Dialect/SCIFRBool/docs/Cornami-HEIR.png
j2kun, line 1: Please remove this docs/ directory for now

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,88 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+def SCIFRBOOL_AnyNumber : AnyTypeOf<[AnyFloat],
+                               "number">;
+
+
+//WIP: low level implementation of SCIFR Bool API not exposed yet
+// PBSOp, KSOp, and LinearOp.
+// e.g. all Boolean ops can be  written in LinearOps etc.
```
j2kun:
```suggestion
```

Please remove "WIP" and "TODO" style comments

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
```diff
@@ -0,0 +1,88 @@
...
+        Op<SCIFRBool_Dialect, mnemonic, traits> {
+  let assemblyFormat = [{
+    operands attr-dict `:` functional-type(operands, results)
+  }];
+  let cppNamespace = "::mlir::scifrbool";
+}
+
+// --- Operations for a gate-bootstrapping API of a SCIFRBool library ---
+// TODO: Traits ElementwiseMappable, Scalarizable to be supported fully in
+//       transforms and conversion passes after adding to SCIFRBool_BinaryGateOp
```
j2kun:
```suggestion
```

lib/Dialect/SCIFRBool/IR/SCIFRBoolDialect.td
```diff
@@ -0,0 +1,37 @@
...
+    }];
```
j2kun:
```suggestion
```

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.td
j2kun, line 1: nit: ReplaceOpWithSection should live in Transforms/, not Conversions/, because it does not perform dialect conversion.

lib/Dialect/SCIFRBool/Conversions/ReplaceOpWithSection.cpp
```diff
@@ -0,0 +1,109 @@
...
+      return llvm::TypeSwitch<Operation&, WalkResult>(*op)
+          .Case<func::FuncOp>([&](auto op) {
+            convertFunc(op, context);
+            return WalkResult::advance();
+          })
+          .Default([&](Operation& op) { return WalkResult::advance(); });
+    });
+    if (walkResult.wasInterrupted()) {
+      signalPassFailure();
+    }
```
j2kun:
```suggestion
    op->walk([&]func::FuncOp op) {
        convertFunc(op, context);
    });
```

nit: this can be simplified quite a bit:

1. No path produces an interrupted walk result
2. walk supports more specific template arguments, so you can "merge" it with the type switch when there's only one op type you want to walk over.

lib/Dialect/SCIFRBool/Conversions/ReplaceOpWithSection.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return opName;
+  }
+
+  void convertFunc(Operation* funcOp, MLIRContext& context) {
+    uint32_t nChipsPerOp = 2;
+    uint32_t nOpsPerSection = nChips / nChipsPerOp;
+    std::vector<Operation*> opList;
+    funcOp->walk([&](Operation* op) {
+      llvm::TypeSwitch<Operation&>(*op)
+          .Case<scifrbool::AndOp, scifrbool::OrOp, scifrbool::NotOp,
```
j2kun: Rather than opt-in here, it will be more future-proof (and slightly more efficient) to check the dialect name and otherwise opt out of unsupported ops.

```
if (isa<SCIFRBoolDialect>(op->getDialect()) && !isa<SectionOp>(op)) {
  opList.push_back(op);
}
```

That way when a new op is added, unless it's special like section, it doesn't need to manually be added to this list.

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "mlir/include/mlir/IR/ValueRange.h"            // from @llvm-project
+#include "mlir/include/mlir/IR/Visitors.h"              // from @llvm-project
+#include "mlir/include/mlir/Transforms/Passes.h"        // from @llvm-project
+
+// clang-format off
+#include "lib/Dialect/SCIFRBool/Conversions/ReplaceOpWithSection.h.inc"
+// clang-format on
+
+namespace mlir {
+namespace cornami {
```
j2kun, line 19: nit: I would also add `namespace scifrbool {` based on this pass's home in the SCIFRBool directory. This will also reduce code size in the rest of the file.

lib/Dialect/SCIFRBool/Conversions/ReplaceOpWithSection.cpp
```diff
@@ -0,0 +1,109 @@
...
+      Region& region = sectionOp->getRegion(0);
+      builder.createBlock(&region);
+      Block* sectionOpBlock = &region.front();
+
+      builder.setInsertionPointToStart(sectionOpBlock);
+      for (uint32_t j = 0; j < nOpsPerSection && i + j < opList.size(); j++) {
+        Operation* scifrOp = opList[i + j];
+        Operation* scifrOpCopy = builder.clone(*scifrOp);
+        scifrOpCopy->getBlock()->moveBefore(sectionOpBlock);
+        // TODO: Replace sectionOp with multiple results
```
j2kun:
```suggestion
```

Looks like this is already done on the line below

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
```diff
@@ -0,0 +1,109 @@
...
+          }
+        }
+      }
+      builder.setInsertionPoint(startInsertionPoint);
+      // Create a new SectionOp to wrap the original operator
+      Operation* sectionOp = builder.create<scifrbool::SectionOp>(
+          startInsertionPoint->getLoc(), inputTypes[0], inputTensors);
+
+      Region& region = sectionOp->getRegion(0);
+      builder.createBlock(&region);
```
j2kun, line 82: I see a few issues with this section op construction. Typically an op like

```
%results = op.foo %operand1, %operand2 {
^bb0(%blockarg1, %blockarg2):
   ...
}
```

has three components:

- op operands
- region block arguments
- control flow from the block back to the op

Op operands are what you expect. Block arguments usually have some semantic distinction from the operands, and the body region usually has a terminator that describes how the work done inside the region flows back out to the results of the op. 

In this pass, `section` has op operands copied from the inner op, but its region has no block arguments and no terminator.
Without a terminator, the SSA value produced by the inner op will be viewed as dead code and eliminated. Moreover, scoping rules prevent any code outside the region from accessing SSA values defined in the region.

I suspect if you add a lit test that runs this pass, and then you run `-cse -canonicalize`, you'll see that behavior.

So I think you may need to define a terminator op (`scifrbool.return`?) to consume the results of the last op in a section.

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
j2kun, line 1: This pass should have lit tests.

lib/Dialect/SCIFRBool/Transforms/DialectResourceMap.cpp
```diff
@@ -0,0 +1,55 @@
+#include "lib/Dialect/SCIFRBool/Transforms/DialectResourceMap.h"
+
+namespace mlir {
+namespace cornami {
+
+cornami::DialectResourceMap make_HELayers_resource_map() {
```
j2kun: nit: don't call it `HELayers`; seems like it was accidentally copied from a different project, and IBM does not have permissive licensing so I want to be sure we're not running afoul of any legal issues here.

lib/Dialect/SCIFRBool/Transforms/SCIFRBoolEmitter.td
j2kun, line 1: Based on the files in Target/SCIFRBool, these three files can be deleted from this directory:

- SCIFRBoolEmitter.td
- SCIFRBoolEmitter.h
- SCIFRBoolEmitter.cpp

lib/Dialect/SCIFRCkks/IR/SCIFRCkksDialect.td
```diff
@@ -0,0 +1,41 @@
...
+    Concepts
+        Section
+            Contains Operators and HBM Regions and memory transfers operations
+            Section can be reduced to FA (for executable) or default (which is not executable)
+        Operator
+            Operators are from AppStream FHE or FractlsBase only
+    }];
+    let cppNamespace = "::mlir::scifrckks";
+    let useDefaultTypePrinterParser = 1;
+}
```
j2kun, line 30:
```suggestion
```

lib/Dialect/SCIFRCkks/IR/SCIFRCkksOps.cpp
```diff
@@ -0,0 +1,12 @@
+//===- CFAIROps.cpp - CFAIR dialect ops ---------------*- C++ -*-===//
```
j2kun:
```suggestion
//===- SCIFRCkksOps.cpp - CFAIR dialect ops ---------------*- C++ -*-===//
```

lib/Dialect/SCIFRCkks/IR/SCIFRCkksDialect.h
```diff
@@ -0,0 +1,23 @@
+#ifndef SCIFRCkks_SCIFRCkksDialect_H
```
j2kun: nit: capitalize ifdef guards. Here and in SCIFRCkksOps.h and SCIFRCkksOps.td

lib/Dialect/SCIFRCkks/Transforms/CKKSEstimator.cpp
```diff
@@ -0,0 +1,352 @@
...
+      // "PARAM": return {0, 0};
+      // "BOOTSTRAPREAL": return {0, 0};
+      // "SET": return {0, 0};
+      // "SET_CI": return {0, 0};
+      // "SELECTOR": return {0, 0};
+      // "ADD_SCALAR_INT": return {0, 0};
+      // "ADD_SCALAR_DOUBLE": return {2, 2};
+      // "MUL_SCALAR_DOUBLE": return {2, 2};
+      // "OUTPUT": return {0, 0};
+      // "None": return {0, 0};
```
j2kun: Remove or add as cases. Looks like most of these have already been added

lib/Dialect/SCIFRCkks/Transforms/CKKSEstimator.cpp
```diff
@@ -0,0 +1,352 @@
...
+    } else if (llvm::isa<ckks::SubPlainOp>(op)) {
+      opcode = CKKSOpcode::SUBPLAIN;
+    } else if (llvm::isa<ckks::MulOp>(op)) {
+      opcode = CKKSOpcode::MUL;
+    } else if (llvm::isa<ckks::MulPlainOp>(op)) {
+      opcode = CKKSOpcode::MULPLAIN;
+    } else if (llvm::isa<ckks::RotateOp>(op)) {
+      opcode = CKKSOpcode::ROTATE;
+      //} else if  ( llvm::isa<ckks::ExtractOp>(op) ) {
+      // opcode = CKKSOpcode::EXTRACT;
```
j2kun:
```suggestion
```

IIRC we removed this op

lib/Dialect/SCIFRCkks/Transforms/CKKSEstimator.cpp
```diff
@@ -0,0 +1,352 @@
...
+    } else if (llvm::isa<ckks::NegateOp>(op)) {
+      opcode = CKKSOpcode::NEGATE;
+    } else if (llvm::isa<ckks::RelinearizeOp>(op)) {
+      opcode = CKKSOpcode::RELINEARIZE;
+    } else if (llvm::isa<ckks::RescaleOp>(op)) {
+      // m_ckksops.n_rescaleop++;
```
j2kun:
```suggestion
```

j2kun (changes requested):
I still have to do:

- CGGIToSCIFRBool.cpp
- SCIFRBoolEmitter.cpp

Will get back to it later today.

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+using namespace heir;
+using namespace mlir::scifrbool;
+
+namespace mlir {
+namespace cornami {
+
+#define GEN_PASS_DEF_CGGITOSCIFRBOOL
+#include "lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.h.inc"
+
+class CGGIToSCIFRBoolTypeConverter : public TypeConverter {
```
j2kun, line 43: It seems like this type converter can simply be removed. It's not actually converting anything. You should be able to run the dialect conversion framework without passing it a type converter.

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
+#define DEBUG_TYPE "cggi-to-scifrbool"
+
+using namespace mlir;
+using namespace heir;
+using namespace mlir::scifrbool;
+
+namespace mlir {
+namespace cornami {
```
j2kun, line 37:
```suggestion
namespace mlir {
namespace cornami {
namespace scifrbool {
```

(and add the trailing `}  // namespace scifrbool` needed at the end of the file)

With this, you do not need to add `scifrbool::` elsewhere in this file.

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+    Type serverParamsType =
+        scifrbool::SCIFRBoolServerParametersType::get(getContext());
+
+    rewriter.modifyOpInPlace(
+        op, [&] { (void)op.insertArgument(0, bskType, nullptr, op.getLoc()); });
+    rewriter.modifyOpInPlace(
+        op, [&] { (void)op.insertArgument(1, kskType, nullptr, op.getLoc()); });
+    rewriter.modifyOpInPlace(op, [&] {
+      (void)op.insertArgument(2, serverParamsType, nullptr, op.getLoc());
+    });
```
j2kun:
```suggestion
    rewriter.modifyOpInPlace(
        op, [&] { 
          (void)op.insertArgument(0, bskType, nullptr, op.getLoc());  
          (void)op.insertArgument(1, kskType, nullptr, op.getLoc()); 
          (void)op.insertArgument(2, serverParamsType, nullptr, op.getLoc());
        });
```

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+struct ConvertCGGITRBBinOp : public OpConversionPattern<BinOp> {
+  using OpConversionPattern<BinOp>::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(
+      BinOp op, typename BinOp::Adaptor adaptor,
+      ConversionPatternRewriter &rewriter) const override {
+    ImplicitLocOpBuilder b(op->getLoc(), rewriter);
+
+    rewriter.replaceOp(
+        op, b.create<SCIFRBoolBinOp>(adaptor.getLhs(), adaptor.getRhs()));
```
j2kun:
```suggestion
        op, SCIFRBoolBinOp::create(b, adaptor.getLhs(), adaptor.getRhs()));
```

The `b.create` form is deprecated. Use `OpType::create(b, ...)`. You will also get better IDE autocompletions with that form.

Here and below.

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+    auto *op = getOperation();
+
+    CGGIToSCIFRBoolTypeConverter typeConverter(context);
+    RewritePatternSet patterns(context);
+    ConversionTarget target(*context);
+    addStructuralConversionPatterns(typeConverter, patterns, target);
+
+    target.addLegalDialect<scifrbool::SCIFRBoolDialect>();
+    target.addIllegalDialect<cggi::CGGIDialect>();
+    target.addIllegalDialect<lwe::LWEDialect>();
```
j2kun, line 127: I'm a bit confused by this line. Wouldn't this make `lwe` types also illegal? But you are leaving the LWE types in the IR and not converting them. Maybe this line should be removed (I would expect it to cause the pass to fail...)

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    target.addDynamicallyLegalOp<memref::AllocOp, memref::DeallocOp,
+                                 memref::StoreOp, memref::LoadOp,
+                                 memref::SubViewOp, memref::CopyOp,
+                                 tensor::FromElementsOp, tensor::ExtractOp>(
+        [&](Operation *op) {
+          return typeConverter.isLegal(op->getOperandTypes()) &&
+                 typeConverter.isLegal(op->getResultTypes());
+        });
```
j2kun, line 136: Because the type converter is not converting types, all types are legal and this has no effect.

```suggestion
```

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+    target.addDynamicallyLegalOp<func::FuncOp>([&](func::FuncOp op) {
+      bool hasKeyArg = op.getFunctionType().getNumInputs() > 0 &&
+                       mlir::isa<scifrbool::SCIFRBoolBootstrapKeyType>(
+                           *op.getFunctionType().getInputs().begin());
+      return typeConverter.isSignatureLegal(op.getFunctionType()) &&
+             typeConverter.isLegal(&op.getBody()) &&
+             (!containsDialects<lwe::LWEDialect, cggi::CGGIDialect>(op) ||
+              hasKeyArg);
```
j2kun:
```suggestion
      return (!containsDialects<lwe::LWEDialect, cggi::CGGIDialect>(op) ||
              hasKeyArg);
```

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
```diff
@@ -0,0 +1,169 @@
...
+    });
+
+    target.addLegalOp<mlir::arith::ConstantOp>();
```
j2kun, line 145:
```suggestion
```

I suspect this line also has no effect, because `arith` was never declared illegal

lib/Target/SCIFRBool/SCIFRBoolEmitter.h
```diff
@@ -0,0 +1,158 @@
...
+  LogicalResult printOperation(AndOp op);
+  LogicalResult printOperation(NandOp op);
+  LogicalResult printOperation(OrOp op);
+  LogicalResult printOperation(NorOp op);
+  LogicalResult printOperation(NotOp op);
+  LogicalResult printOperation(XorOp op);
+  LogicalResult printOperation(XNorOp op);
+  // LogicalResult printOperation(KSOp op);
+  // LogicalResult printOperation(LinearOp op);
+  // LogicalResult printOperation(PBSOp op);
```
j2kun:
```suggestion
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.h
```diff
@@ -0,0 +1,158 @@
...
+        m_variableNames.try_emplace(result, m_variableCount++);
+      }
+      for (Region &region : opr->getRegions()) {
+        for (Block &block : region) {
+          for (Value arg : block.getArguments()) {
+            m_variableNames.try_emplace(arg, m_variableCount++);
+          }
+        }
+      }
+      return WalkResult::advance();
```
j2kun:
```suggestion
```

No return value necessary

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+                                       "code against the Concrete Engine "
+                                       "API")};
+};
+
+static llvm::ManagedStatic<TranslateOptions> translateOptions;
+
+void registerTranslateOptions() {
+  // Forces initialization of options.
+  *translateOptions;
+}
```
j2kun, line 46: These appear to be unused and can be removed. (If you want to keep it, the `packageName` option should have the name `package-name` and an appropriate description.)

```suggestion
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+                using cornami::fhe::concrete::ConcreteEngine;
+                using cornami::fhe::concrete::KeySwitchKey;
+                using cornami::fhe::concrete::StreamLwe;
+
+                using cornami::fhe::concrete::BinaryBooleanOp;
+                using cornami::fhe::concrete::BooleanEncodingType;
+                using cornami::fhe::concrete::ConcreteEngineBoolean;
+                using cornami::fhe::concrete::UnaryBooleanOp;
+
+                using SCIFRBoolCiphertext = std::vector<uint64_t>;
```
j2kun, line 466:
```suggestion
using cornami::fhe::ServerParameters;
using cornami::fhe::concrete::BootstrapKeyStandard;
using cornami::fhe::concrete::ConcreteEngine;
using cornami::fhe::concrete::KeySwitchKey;
using cornami::fhe::concrete::StreamLwe;

using cornami::fhe::concrete::BinaryBooleanOp;
using cornami::fhe::concrete::BooleanEncodingType;
using cornami::fhe::concrete::ConcreteEngineBoolean;
using cornami::fhe::concrete::UnaryBooleanOp;

using SCIFRBoolCiphertext = std::vector<uint64_t>;
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+
+LogicalResult translateToSCIFRBool(Operation *op, llvm::raw_ostream &os) {
+  SCIFRBoolEmitter emitter(os);
+  emitter.autoGenComment();
+  emitter.beginCode();
+  LogicalResult result = emitter.translate(*op);
+  return result;
+}
+
+// TODO: check if more operators need to be added
```
j2kun:
```suggestion
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+          .Case<ModuleOp>([&](auto op) { return success(); })
+          // Func ops
+          .Case<func::FuncOp, func::ReturnOp>(
+              [&](auto op) { return success(); })
+          // Arith ops
+          .Case<arith::ConstantOp>([&](auto op) { return printOperation(op); })
+          // SCIFRBool ops
+          .Case<AndOp, NandOp, OrOp, NorOp, NotOp, XorOp, XNorOp,
+                /* KSOp,
+                   LinearOp, PBSOp, */
```
j2kun:
```suggestion
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+  }
+  m_returnStr = "";
+  std::string returnVar = getNameForValue(op.getOperands()[0]);
+  m_returnStr += m_fromElementsStr;
+  m_returnStr += "engineBoolean.m_engine->profiler_report();\n";
+  m_returnStr += "return " + returnVar + ";\n";
+  return success();
+}
+
+LogicalResult SCIFRBoolEmitter::printSksMethod(::mlir::ValueRange operands,
```
j2kun, line 208: nit: the name "sks" was specific to the server key argument of the TfheRs emitters. Might want to change it to be more descriptive to Cornami. Since the "server key" is encapsulated in the engineBoolean which is not passed as an argument, maybe just call this `printBinaryBooleanMethod` or similar.

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+      return failure();
+    }
+    m_os << intAttr.getValue() << ";\n";
+  } else {
+    return op.emitError() << "Unknown constant type " << valueAttr.getType();
+  }
+  return success();
+}
+
+void SCIFRBoolEmitter::createStream(std::string opName) {
```
j2kun, line 412: Fine for now, but the stream concept and management seems like it should be part of a dialect with passes to manage the cloning or creation process.

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+                using cornami::fhe::concrete::UnaryBooleanOp;
+
+                using SCIFRBoolCiphertext = std::vector<uint64_t>;
+  )cpp";
+}
+
+FailureOr<std::string> SCIFRBoolEmitter::convertType(Type type) {
+  if (auto shapedType = dyn_cast<ShapedType>(type)) {
+    // A lambda in a type switch statement can't return multiple types.
+    // FIXME: why can't both types be FailureOr<std::string>?
```
j2kun:
```suggestion
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+    }
+    m_os << getNameForValue(op.getMemRef()) << "["
+         << getNameForValue(op.getIndices()[0]) << "];\n";
+  }
+  return success();
+}
+
+LogicalResult SCIFRBoolEmitter::printOperation(memref::StoreOp op) {
+  m_os << getNameForValue(op.getMemref());
+  m_os << ".push_back(" << getNameForValue(op.getValueToStore()) << ");\n";
```
j2kun, line 394: `memref.store` stores a value in a specific index (given as an operand), while this ignores the index and stores the value at the end. Surely this will cause major bugs.

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+}
+
+LogicalResult SCIFRBoolEmitter::printOperation(memref::AllocOp op) {
+  m_os << "std::vector<";
+  if (failed(emitType(op.getMemref().getType().getElementType()))) {
+    return op.emitOpError() << "Failed to get memref element type";
+  }
+  std::string resultVar = getNameForValue(op.getMemref());
+  m_os << "> " << resultVar << ";\n";
+
```
j2kun, line 367: You may want to add a resize here to ensure the vector has allocated enough space.

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+return printSksMethod({op.getInput()}, op.getResult(), "KS");
+}
+
+LogicalResult SCIFRBoolEmitter::printOperation(LinearOp op) {
+return printSksMethod({op.getInput()}, op.getResult(), "Linear");
+}
+
+LogicalResult SCIFRBoolEmitter::printOperation(PBSOp op) {
+return printSksMethod({op.getInput()}, op.getResult(), "PBS");
+}*/
```
j2kun:
```suggestion
```

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+
+LogicalResult SCIFRBoolEmitter::printOperation(SectionOp op) {
+  m_os << "// =================== start of section " << m_sectionCount + 1
+       << " ===================\n\n";
+  if (m_sectionCount > 0)
+    m_os << "engineBoolean.m_engine->resetTopology(true);\n";
+  m_os << "engineBoolean.m_engine->SetDefaultDirTopology(\"/tmp/viz/"
```
j2kun: `/tmp/viz/` seems like debug code that was left in by mistake. Maybe that should be a command line option?

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
j2kun: Or else, as below, make options that you actually use in the emitter (SetDefaultDirTopology seems like a good candidate)

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
```diff
@@ -0,0 +1,542 @@
...
+          std::string resultName = getNameForValue(operation.getResult());
+          m_oprToStrm[sectionOpResultName].push_back(
+              m_oprToStrm[resultName].back());
+        })
+        .Default([&](Operation &) {
+          // Do nothing
+        });
+  });
+  m_os << "engineBoolean.m_engine->routeToHost("
+       << m_oprToStrm[sectionOpResultName].back() << ");\n";
```
j2kun, line 296: The lack of a terminator on `section`, plus this code, makes the implicit assumption that only the results of the very last op in a `section` are the output of the entire `section` op. If you want to be able to have a section produce multiple results, you would probably need a terminator at that point.

If these are the intended semantics of `section`, this should be added to the tablegen description for the `section` op.

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
j2kun, line 1: The emitter should have a lit test, especially to exercise some of the more detailed aspects such as the handling of `section`

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: makes sense me fold it in

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: there is a basic --emit-scifrbool based test; nothing for section op per-se and given section-op pass needs more work I'll have to do this at a later time

lib/Dialect/SCIFRCkks/Transforms/CKKSEstimator.cpp
ai-mannamalai: sure let me clean it

lib/Dialect/SCIFRCkks/IR/SCIFRCkksDialect.h
ai-mannamalai: sure

lib/Dialect/SCIFRBool/Transforms/DialectResourceMap.cpp
ai-mannamalai: turns out this is dead code; unused 🫨

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
ai-mannamalai: that was not my observed behavior; not sure if something has changed

lib/Dialect/SCIFRBool/Conversions/ReplaceOpWithSection.cpp
ai-mannamalai: OK. I guess mileage may vary but totally agree from maintainer POV

lib/Target/SCIFRBool/SCIFRBoolEmitter.h
ai-mannamalai: OK

lib/Dialect/SCIFRBool/Transforms/SCIFRBoolEmitter.td
ai-mannamalai: ahh! indeed - thanks for catching it

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.td
ai-mannamalai: mm okay

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
ai-mannamalai: OK

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
ai-mannamalai: not so sure when I tried the code

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
ai-mannamalai: skip for now

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: Test exists at file tests/Dialect/SCIFRBool/Transforms/and_gate.mlir

lib/Dialect/SCIFRBool/IR/SCIFRBoolOps.td
ai-mannamalai: removed PBS, KS and LinearOp for now.

lib/Dialect/SCIFRBool/Conversions/ReplaceOpWithSection.cpp
ai-mannamalai: accepted your change.

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
ai-mannamalai: done

lib/Dialect/SCIFRBool/Transforms/ReplaceOpWithSection.cpp
ai-mannamalai: added in file: **tests/Dialect/SCIFRBool/Transforms/scifr_sectionop.mlir**

lib/Dialect/SCIFRBool/Transforms/SCIFRBoolEmitter.td
ai-mannamalai: done

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: OK

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: done

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: Got it

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: done

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: good point

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
ai-mannamalai: I propose to work on section op pass in a later PR

lib/Dialect/CGGI/Conversions/CGGIToSCIFRBool/CGGIToSCIFRBool.cpp
ai-mannamalai: its used in the code few lines below

lib/Target/SCIFRBool/SCIFRBoolEmitter.cpp
j2kun: Those tests run the pass, but do not assert anything about the correctness of the output. I think we can merge this in, but if I were you I would follow up this PR with some smaller PRs that enhance the testing to ensure the passes and code generation is doing what you expect it to do.

In particular, without assertions, routine maintenance may make backwards-incompatible changes to the passes and dialects.

j2kun:
Please run `pre-commit run --all-files` on this PR to fix the linter errors.

ai-mannamalai:
> Please run `pre-commit run --all-files` on this PR to fix the linter errors.

done; thanks

Issue #2157, opened 2025-08-27
Lower ckks.relinearize to polynomial dialect


j2kun:
First step was done in https://github.com/google/heir/pull/2013

Still needs doing:

- lowering `key_switch_inner`
- getting the key switching key added to the IR at a higher level
- getting parameter-selected params forwarded to this lowering

crockeea:
I'm actually not sure what the remaining items in the issue are referring to:
- getting the key switching key added to the IR at a higher level
- getting parameter-selected params forwarded to this lowering

j2kun:
I will close this an open a new issue for the more concrete task of getting an end-to-end LLVM test of relinearization

PR #2602, opened 2026-02-04
Transform KeySwitchInner

crockeea:
This PR builds on https://github.com/google/heir/pull/2593 to decompose `KeySwitchInner` into lower-level operations. This PR replaces #2157. 

This required a few changes:
- Added `lwe:ExtractSliceOp`, `lwe:ConvertBasisOp`, and `lwe:RRingMulOp`
- Moved `ExtractCoeffOp` and `FromCoeffsOp` from CKKS dialect to LWE dialect.
- A new pass to decompose `KeySwitchInnerOp`. Due to the changes in #2593, this is now a relatively simple operation (relative to #2157) that avoids a lot of the complexity and tensor manipulation that happened when this transform was a lowering from `Polynomial` to `ModArith`; a big win!
- A change to `LWECiphertext::verify` that skips the `modulusChain` check if a `modulusChain` isn't present, which is true for KeySwitchKeys; see below.
- I removed the optional `modulusChain` attribute from `LWERingElt`. I determined it was causing lots of problems, and wasn't solving any. Moreover, I don't think it makes sense for an isolated ring element.
- I hoisted `rns:ExtractSlice::verify` to a reusable function since it is very similar to the new `ckks:ExtractSliceOp`.
- #2604 attempted to fix a circular import problem with `SchemeParamAttr` and `getSchemeParamFromAttr`, but this did not resolve the problem in this PR. After discussion, we decided to move ``getSchemeParamFromAttr` from Parameters/CKKS/Utils to lib/Dialect/CKKS/CKKSAttributes.
- Updated relinearize test now that the KeySwitchInner verifier is more strict.
- Refactored relinearize transform to remove Patterns.h/cpp.

A word on the `ModulusChain` attribute of ciphertexts: it's not clear (to me) what the intended usage/meaning of this attribute is. For CKKS, does it only include the "normal" primes, or should it also include key-switch primes? If it's the former, this is an issue for key-switch keys (which are currently ciphertexts INCLUDING special primes), and if it's the latter, then the `current` attribute of the `ModulusChain ` doesn't always correspond to level (or there's an implicit maximum value of `current` that is smaller than the length of the `ModulusChain`). For now I opted to make KeySwitchKeys be a tensor of `LWECiphertext`, but *without* a `ModulusChain`.


lib/Dialect/CKKS/IR/CKKSOps.cpp
```diff
@@ -128,15 +132,122 @@ LogicalResult ExtractCoeffOp::inferReturnTypes(
...
+  results.push_back(resultType);
+  return success();
+}
+
+LogicalResult ConvertBasisOp::inferReturnTypes(
+    MLIRContext* ctx, std::optional<Location>, ValueRange operands,
+    DictionaryAttr attrs, mlir::OpaqueProperties properties,
+    mlir::RegionRange regions, SmallVectorImpl<Type>& results) {
+  ConvertBasisOpAdaptor op(operands, attrs, properties, regions);
+  lwe::LWERingEltType inputType =
```
j2kun: Stylistically speaking, it's acceptable to use `auto` for the type declaration when you're explicitly casting, because the type is duplicated and so having it twice provides no extra information

```
auto inputType = cast<lwe::LWERingEltType>(op.getValue().getType());
```

lib/Dialect/CKKS/IR/CKKSOps.cpp
```diff
@@ -128,15 +132,122 @@ LogicalResult ExtractCoeffOp::inferReturnTypes(
...
+  }
+
+  int kskRank = keyTensorType.getRank();
+  if (kskRank != 1) {
+    return emitOpError()
+           << "KeySwitchingKey must be a rank-1 tensor, but it has rank  "
+           << kskRank;
+  }
+
+  // TODO: I'm probably doing something wrong, but when I try to link against
```
j2kun: Cf. #2604 to unblock this

lib/Dialect/CKKS/IR/CKKSOps.td
```diff
@@ -291,4 +292,31 @@ def CKKS_FromCoeffsOp : CKKS_Op<"from_coeffs", [Pure]> {
   let hasVerifier = 1;
 }
 
+// Should this be a CKKS op or an LWE op?
+def CKKS_ExtractSliceOp : CKKS_Op<"extract_slice", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
```
j2kun: These seem like LWE ops to me, and I don't see any obvious obstacles to moving them there.

lib/Dialect/CKKS/Transforms/BUILD
```diff
@@ -45,6 +50,22 @@ cc_library(
...
 
+cc_library(
+    name = "DecomposeKeySwitch",
+    srcs = ["DecomposeKeySwitch.cpp"],
+    hdrs = ["DecomposeKeySwitch.h"],
+    deps = [
+        ":Patterns",
+        ":pass_inc_gen",
+        "@heir//lib/Dialect/CKKS/IR:Dialect",
+        "@heir//lib/Parameters/CKKS:Params",
```
j2kun:
```suggestion
```

Looks like this one is not used (that said, we would clean it up automatically during import to google's monorepo)

lib/Dialect/CKKS/Transforms/Patterns.cpp
```diff
@@ -47,6 +54,128 @@ LogicalResult DecomposeRelinearizePattern::matchAndRewrite(
   return success();
 }
 
+LogicalResult DecomposeKeySwitchPattern::matchAndRewrite(
+    KeySwitchInnerOp op, PatternRewriter& rewriter) const {
+  ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+  // TODO(#2157): enable this for more than just CKKS
```
j2kun: This TODO seems obsolete now that it's a CKKS-specific transform. I feel like we can return to this if/when we get to bfv/bgv implementations (which we may never do, tbh)

```suggestion
```

lib/Dialect/CKKS/Transforms/Patterns.cpp
```diff
@@ -47,6 +54,128 @@ LogicalResult DecomposeRelinearizePattern::matchAndRewrite(
...
+  SchemeParamAttr schemeParamAttr =
+      op->getParentOfType<ModuleOp>()->getAttrOfType<SchemeParamAttr>(
+          CKKSDialect::kSchemeParamAttrName);
+  if (!schemeParamAttr) {
+    return rewriter.notifyMatchFailure(
+        op, "Cannot find scheme param attribute on parent module");
+  }
+  auto schemeParam = SchemeParam::getSchemeParamFromAttr(schemeParamAttr);
```
j2kun: Will need an update and a new import after #2604

lib/Dialect/LWE/IR/LWEOps.cpp
```diff
@@ -42,6 +42,38 @@ LogicalResult RMulOp::verify() { return lwe::verifyMulOp(this); }
 
 LogicalResult RMulPlainOp::verify() { return lwe::verifyMulPlainOp(this); }
 
+LogicalResult RRingMulOp::verify() {
+  lwe::LWECiphertextType ct;
+  lwe::LWERingEltType pt;
+  // TODO: Can the verifier be run before OR after canonicalization, or does
```
j2kun: Generally, canonicalization and verification are independent, and the verifier may not depend on the operation being in a canonical form.

One rule of thumb is that verification is run before and after each pass, and canonicalization is itself a pass.

lib/Dialect/LWE/IR/LWEOps.cpp
```diff
@@ -42,6 +42,38 @@ LogicalResult RMulOp::verify() { return lwe::verifyMulOp(this); }
...
+  }
+  // verify ring and modulusChain matches
+  auto ctCiphertext = ct.getCiphertextSpace();
+  if (ctCiphertext.getRing() != pt.getRing()) {
+    return emitOpError() << "Input rings do not match";
+  }
+  // Note: KeySwitch keys don't have a ModulusChain
+  // if (ct.getModulusChain() != pt.getModulusChain()) {
+  //   return emitOpError() << "Input modulus chains do not match";
+  // }
```
j2kun:
```suggestion
```

lib/Dialect/LWE/IR/LWEOps.td
```diff
@@ -168,13 +168,26 @@ def LWE_RMulOp : LWE_BinOp<"rmul", [SameOperandsAndResultRings, InferTypeOpAdapt
...
   let hasVerifier = 1;
 }
-
 def LWE_RMulPlainOp : LWE_CiphertextPlaintextOp<"rmul_plain", [Commutative, IncreasesMulDepthOpInterface]> {
   let summary = "Multiplication between RLWE ciphertext-plaintext";
   let hasVerifier = 1;
   let hasCanonicalizer = 1;
 }
 
+// TODO: Should this have IncreasesMulDepthOpInterface? For the purposes of keyswitching, no. However, if the ring element is an encoded plaintext, then "yes".
```
j2kun: This interface is used by `MulDepthAnalysis`, so we would only need to put it on this op if we expect this op to show up at a time when we need to compute the multiplicative depth of the IR. My expectation is that (a) we need to compute mul depth when doing parameter selection and ciphertext management insertion, (b) by the time we get to decomposing key switch, that work is done and (c) we won't be re-computing that at an intermediate stage like when we insert these ring mul ops.

lib/Dialect/LWE/IR/LWEOps.td
```diff
@@ -168,13 +168,26 @@ def LWE_RMulOp : LWE_BinOp<"rmul", [SameOperandsAndResultRings, InferTypeOpAdapt
...
+// I would say that "IncreasesMulDepth" should be a property of the non-ciphertext input rather than the op itself.
+//
+// TODO: What about IsCiphertextPlaintextOp?
+//
+// TODO: I see a lot of the types in other ops are more generic than the strict ones I use; not sure how to generalize (or why I would want to).
+def LWE_RRingMulOp : LWE_Op<"ring_mul", [Commutative, InferTypeOpAdaptor, ElementwiseMappable]> {
```
j2kun:
```suggestion
def LWE_RRingMulOp : LWE_Op<"ring_mul", [InferTypeOpAdaptor, AllTypesMatch<["rhs", "output"]>]> {
```

A few things with the traits:

1. ElementwiseMappable implies that the inputs can be both type T and tensors with element type T, for some T. The op definition you have here doesn't allow that, so ElementwiseMappable is incorrect to use here. I also don't think it's necessary to add.
2. Commutative implies that the tablegen allows more flexible types for the lhs and rhs operands, but here you have the operands being rigid types (lhs can only be a RingElt, and rhs can only be a ciphertext). I think this is fine, and I don't see a reason to allow them to be commutative. The purpose of the Commutative trait is to reduce the number of canonicalization patterns you need to write to simplify expressions. In particular, if you are working with `arith` ops, then the canonicalization patterns use Commutative to put constant-like operands always on the rhs, and then subsequent canonicalization patterns that look for constants (e.g., mul by 2 simplifying to a bit shift) only need to match on one operand. Since we don't have constant-like operands for LWERingElt or LWECiphertext, Commutative doesn't do anything (in particular, it does not magically relax the type constraints specified in the tblgen definition). However, if you wanted to allow this, what you'd have to do is define a type constraint like `def LWE_RingEltOrCt : AnyTypeOf<[LWERingElt, LWECiphertext]>;` and use that as the argument "type" (`let arguments = (ins LWE_RingEltOrCt:$lhs, LWE_RingEltOrCt:$rhs);`). Then the rest of the codebase would have to `dyn_cast` to see which type each operand happens to be. As I said earlier, we probably don't need this, and we can just force any code path that creates these ops to put the RingElt in the LHS. This means you can also remove `hasCanonicalizer` and the corresponding `getCanonicalizationPatterns` method.
3. In the verifier you check equality of the rhs type and the output type. The suggested `AllTypesMatch<["rhs", "output"]>` auto-generates verification code that checks the same thing, and you can remove the manual check from your verifier.

j2kun: Any reason why we can't simplify the name to `RingMulOp` instead of `RRingMulOp`?

tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir
```diff
@@ -0,0 +1,71 @@
...
+    // CHECK-DAG: return [[const_ext]], [[linear_ext]] : !ringelt, !ringelt
+    %constTerm, %linearTerm = ckks.key_switch_inner %x, %arg0 : (!ringelt_L1, tensor<2x!ct_L2>) -> (!ringelt_L1, !ringelt_L1)
+    return %constTerm, %linearTerm: !ringelt_L1, !ringelt_L1
+  }
+}
+
+// TODO: Ideally there would be at least two more tests here:
+//  - One to test multiple partitions
+//  - One to test partial partitions
+// However, Jeremy said I couldn't do that in the same file for doctest purposes; where should they go?
```
j2kun: Just make a second mlir file in this same directory.

tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir
```diff
@@ -0,0 +1,71 @@
...
+#ciphertext_space_L2 = #lwe.ciphertext_space<ring = #ring_L2x1024, encryption_type = lsb, size = 2>
+// encoding probably doesn't make sense for KSKs
+#inverse_canonical_encoding = #lwe.inverse_canonical_encoding<scaling_factor = 0>
+#key = #lwe.key<>
+// ModulusChain probably isn't appropriate for keyswitch keys. The problem is that in order to be a valid ciphertext, the modulus chain needs to include
+// the key-switch primes, but these don't correspond to available "levels".
+!ct_L2 = !lwe.lwe_ciphertext<application_data = <message_type = i3>,
+                             plaintext_space = <ring = #ring_L2x1024, encoding = #inverse_canonical_encoding>,
+                             ciphertext_space = #ciphertext_space_L2,
+                             key = #key>
```
j2kun, line 24: TBH with this mismatch it seems like we need a new type for the key switching key, which can have exactly the necessary information attached to it. If you agree, is the key switching key specification tied to CKKS, or applicable to all RLWE schemes?

j2kun (changes requested):
I think this is closer to good that it seems.

>  Is there a way to run a broad set of tests without running all of the examples?

You can run `bazel test tests/path/to/dir/...:all` to glob for tests in any subdirectory. Generally speaking the slow tests are in `tests/Examples`, and this pass should not impact them at all. So if you pass tests in `tests/Dialect/CKKS/...:all` then you can push it and let the whole test suite run in CI for the final check.

Also generally `bazel test -c opt` will be faster than `bazel test` because by default we do a debug build for developers, and the end-to-end tests are much slower in debug mode (mainly due to OpenFHE relying on `-O3`). However, switching between `-c opt` and `-c dbg` discards the bazel cache and rebuilds, so typically I stick to one or the other during most of my development, depending on how intense my tests are.

lib/Dialect/CKKS/IR/CKKSOps.td
crockeea: Should I make _all_ of the new ops LWEOps? 
- `KeySwitchInnerOp`
- `ExtractCoeffOp`
- `FromCoeffsOp`
- `ExtractSliceOp`
- `ConvertBasisOp`

or only a subset?

j2kun: The only one that gives me pause is key switch inner, because I'm not sure how specific the key switch implementation is to CKKS. I'll leave it to you to decide.

lib/Dialect/CKKS/IR/CKKSOps.cpp
crockeea: I'm still getting a circular dependency:
```
SchemeParamAttr schemeParamAttr =
      getParentOfType<ModuleOp>()->getAttrOfType<SchemeParamAttr>(
          CKKSDialect::kSchemeParamAttrName);
  if (!schemeParamAttr) {
    return emitOpError() << "Cannot find scheme param attribute on parent module";
  }
  auto schemeParam = SchemeParam::getSchemeParamFromAttr(schemeParamAttr);
  ```
  
  `getSchemeParamFromAttr` is in `lib/Parameters/CKKS/Utils.h`, which requires the build dependency `"@heir//lib/Parameters/CKKS:Utils"`, which results in a circular import.

Comment (j2kun) [reply lib/Dialect/CKKS/IR/CKKSOps.cpp]: Ah this is because you're using that helper. I suppose the way to fix this is then to move the utility function instead to CKKSAttributes.h, since anyone depending on that utility also depends on the attribute it takes as input.

File: lib/Dialect/CKKS/IR/CKKSOps.td (hunk 3 above)
Comment (crockeea) [reply lib/Dialect/CKKS/IR/CKKSOps.td]: I'm not familiar enough with modern implementations of other schemes to say. However, if I keep it as a CKKSOp, I still need a parallel LWE op that it can be lowered to, right?

File: lib/Dialect/LWE/IR/LWEOps.td (hunk 10 above)
Comment (crockeea) [reply lib/Dialect/LWE/IR/LWEOps.td]: There's an `LWE_AddOp` and an `LWE_RAddOp`. Since the other ops had an `R` in front for ring ops (vs pure LWE), I figured this should to. Open to your suggestion though.

File: lib/Dialect/CKKS/IR/CKKSOps.td (hunk 3 above)
Comment (j2kun) [reply lib/Dialect/CKKS/IR/CKKSOps.td]: > I still need a parallel LWE op that it can be lowered to, right?

Not necessarily, but ultimately we will lower to the polynomial dialect. So if we have to duplicate that conversion pattern because the lowering from `bgv.relinearize` is identical, then it would make more sense to have the lowering pattern anchor on an `lwe` op, and then patterns that lower `relinearize` to produce `key_switch_inner` would produce the `lwe` op directly instead of having parallel ops at the `ckks` and `bgv` level.

(In particular, if the lowering for `relinearize` is the same in BFV/BGV as it is in CKKS, then it might even make sense to move the `relinearize` op to the `lwe` dialect for the same reason).

I think today we handle this in a manner that has some duplication because we weren't sure about differences in implementation, and that will be clarified by this work and then I can come in and do the refactoring to make the flow clearer.

File: lib/Dialect/LWE/IR/LWEOps.td (hunk 10 above)
Comment (j2kun) [reply lib/Dialect/LWE/IR/LWEOps.td]: what about `LWE_RMulRingElt : LWE_op<"rmul_ring_elt", ...>`   ?

In a bunch of other places in the codebase we call "specialized" mul ops `mul_plain`, `mul_scalar`, `mul_const`, etc., so I think this fits better with that convention.

File: lib/Dialect/CKKS/IR/CKKSOps.td (hunk 3 above)
Comment (crockeea) [reply lib/Dialect/CKKS/IR/CKKSOps.td]: I moved everything but `KeySwitchInner` to LWEOps. 

> (In particular, if the lowering for relinearize is the same in BFV/BGV as it is in CKKS, then it might even make sense to move the relinearize op to the lwe dialect for the same reason).

>I think today we handle this in a manner that has some duplication because we weren't sure about differences in implementation, and that will be clarified by this work and then I can come in and do the refactoring to make the flow clearer.

I did ask Craig [Gentry] about this. He's also not very familiar with modern key-switching for BGV, but he thinks it should be the same, except with a different mod_down step (which shouldn't change the keyswitch lowering).

That said, given that I don't have expertise in BGV right now, I think we're going to have to make some educated guesses and fix things later if needed. I'm indifferent to the location of `KeySwitchInner`; if I had to choose I'd say put it in LWE just so the CKKS dialect stays a nice high-level API.

File: tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir (hunk 12 above)
Comment (crockeea) [reply tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir:24]: I thought about that too, and even got part-way through the implementation before turning back. The natural choice would be `LWEPublicKey`, since a KSK is indeed precisely a list of PublicKeys. The problem is that (as far as I can tell), LWEPublicKey won't play nice with other ops.

For example, I'd need to make RMulRingEltOp support multiplication of a ring elt and a public key, not just a ciphertext. We also would need to be able to add PublicKeys and extract their coeffs. I suppose that's not terrible [to add support for LWEPublicKey to those ops, in addition to CT]; I can do that if you think it sounds reasonable.

Comment (crockeea) [reply tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir:24]: The structure of BinOp could make things challenging.

File: lib/Dialect/CKKS/IR/CKKSOps.td (hunk 3 above)
Comment (crockeea) [reply lib/Dialect/CKKS/IR/CKKSOps.td]: Moving KSI to LWEOps might resolve the circular import problem by moving the verification code to LWEOps.cpp.

File: lib/Dialect/CKKS/IR/CKKSOps.cpp (hunk 2 above)
Comment (crockeea) [reply lib/Dialect/CKKS/IR/CKKSOps.cpp]: Office hours decision: move `getSchemeParamFromAttr` to CKKSAttributes.

File: lib/Dialect/CKKS/IR/CKKSOps.td (hunk 3 above)
Comment (crockeea) [reply lib/Dialect/CKKS/IR/CKKSOps.td]: Office hours summary:
Relinearize transform would be the same for BGV and CKKS.
KeySwitchInner transform would mostly be the same for BGV and CKKS, except that you would need to call <scheme>.ModDown instead of LWE.ModDown, since those would be different ops.

Decision: keep KeySwitchInner in CKKS.

File: tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir (hunk 12 above)
Comment (crockeea) [reply tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir:24]: Decision: Leave my hack (ignore ModulusChain for KeySwitchKeys, whcih are LWECiphertexts) for now, but fix this situation in #2628.

File: tests/Dialect/CKKS/Transforms/decompose_relinearize.mlir [hunk 13]
```diff
@@ -1,34 +1,68 @@
-// RUN: heir-opt --ckks-decompose-relinearize %s | FileCheck %s
+// RUN: heir-opt --ckks-decompose-relinearize %s --mlir-print-ir-after-failure | FileCheck %s
```
j2kun:
```suggestion
// RUN: heir-opt --ckks-decompose-relinearize %s | FileCheck %s
```

tests/Dialect/CKKS/Transforms/decompose_keyswitch.mlir
```diff
@@ -0,0 +1,54 @@
...
+
+module attributes {
+    ckks.schemeParam = #ckks.scheme_param<
+      logN = 10,
+      // TODO: make a ticket to convert these to integerAttrs
```
j2kun:
```suggestion
```

tests/Dialect/CKKS/Transforms/decompose_relinearize.mlir
```diff
@@ -1,34 +1,68 @@
...
+#ciphertext_space_L2 = #lwe.ciphertext_space<ring = #ring_L2x1024, encryption_type = lsb, size = 2>
+!ct_L2 = !lwe.lwe_ciphertext<application_data = <message_type = i3>,
+                             plaintext_space = <ring = #ring_L2x1024, encoding = #inverse_canonical_encoding>,
+                             ciphertext_space = #ciphertext_space_L2,
+                             key = #key>
+
+module attributes {
+    ckks.schemeParam = #ckks.scheme_param<
+      logN = 10,
+      // TODO: make a ticket to convert these to integerAttrs
```
j2kun:
```suggestion
```

j2kun (approved):
Just a minor cleanup. LGTM!

j2kun (approved):

Issue #2591, opened 2026-01-30
OMP: Error #15 (Duplicate OpenMP runtime) when running heir-opt tests via Bazel


crockeea:
I am encountering an OpenMP runtime error when running tests for heir-opt using Bazel. The error indicates that multiple copies of the OpenMP runtime (libomp) have been linked into the binary, causing the program to abort during initialization.

#### Steps to Reproduce (intermittent, may need a failing test?)
Run the lwe-to-polynomial tests (or potentially any test invoking heir-opt with complex lowerings). E.g.,
`bazel test //tests/Dialect/LWE/Conversions/lwe_to_polynomial:decrypt_ops.mlir.test`

`OMP: Error #15: Initializing libomp.so.5, but found unknown library already initialized. OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://openmp.llvm.org/
error: command failed with exit status: -6`

I have reproduced this error with the tip of main.

#### According to AI
It appears that heir-opt is linking against the OpenMP runtime twice. This often happens if:

The toolchain provides -fopenmp implicitly.

A dependency (possibly in @llvm-project) explicitly links a static archive of OpenMP.

#### Expected Behavior
The tests should run without conflicting OpenMP runtimes. This error makes debugging [even] harder.

j2kun:
Could you let me know what platform you're running this on?

crockeea:
WSL with Ubuntu 24.04

j2kun:
I think this is a mistake in my migration of OpenFHE to the bazel central registry.

https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/openfhe/1.4.2.bcr.1/overlay/BUILD

There you can see `linkstatic=True` in each of the OpenFHE build targets. I had the same setting in our HEIR-internal build files, but I suspect this is the culprit. I can try to remove them, but I won't have access to my own WSL machine to test it for a while (my basement where my office usually sits is currently under construction).

j2kun:
[...]

PR #2619, opened 2026-02-06
Workaround for #2591

crockeea:
This PR is a workaround for #2591. It does not address the root cause, which I was unable to identify, but it does appear to prevent test failures due to OMP collisions.


j2kun:
Since this appears to be only happening on your system, what you can do is put these lines into your `~/.bazelrc`, which will layer on top of the `.bazelrc` in HEIR.

PR #2624, opened 2026-02-08
Add extended basis rotation passes for OpenFHE (ConvertToExtendedBasis + HoistKeySwitchDown)

toprakmurat:
## Summary
Continuation of #2574, related to #2519.
Adds two new passes for optimizing rotation-heavy FHE programs by deferring expensive key-switch operations:
1. `--openfhe-convert-to-extended-basis` - Converts `fast_rotation` → `fast_rotation_ext` + `key_switch_down`
2. `--openfhe-hoist-key-switch-down`

### Pass 1: ConvertToExtendedBasis
- Converts each `openfhe.fast_rotation` to `openfhe.fast_rotation_ext` + `openfhe.key_switch_down`
- Semantically equivalent transformation that enables further optimization
### Pass 2: HoistKeySwitchDown
- Pushes `key_switch_down` operations later in the IR to reduce key-switch count.
- Works on `Add`, `AddInPlace`, and `Mul` operations.
- When **at least one** operand is `key_switch_down` (with single use), the pass swaps:

```mlir
// Case 1: Both operands are key_switch_down
%a = key_switch_down(%a_ext)
%b = key_switch_down(%b_ext)
%sum = add(%a, %b)
```
to:
```mlir
%sum_ext = add(%a_ext, %b_ext)
%sum = key_switch_down(%sum_ext)
```
and,
```mlir
// Case 2: One operand is key_switch_down (enables chained optimization)
%a = key_switch_down(%a_ext)
%sum = add(%a, %ct)  // %ct is not key_switch_down
```
to:
```mlir
%sum_ext = add(%a_ext, %ct)
%sum = key_switch_down(%sum_ext)
```

Through greedy iteration, this reduces multiple key-switches in chained operations to a single final key-switch.

### DIFF
--- a/lib/Dialect/Openfhe/Transforms/ConvertToExtendedBasis.cpp
+++ b/lib/Dialect/Openfhe/Transforms/ConvertToExtendedBasis.cpp
@@ -1,0 +1,48 @@
+#include "lib/Dialect/Openfhe/Transforms/ConvertToExtendedBasis.h"
+
+#include "lib/Dialect/Openfhe/IR/OpenfheOps.h"
+#include "mlir/include/mlir/IR/PatternMatch.h"  // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"     // from @llvm-project
+#include "mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace openfhe {
+
+#define GEN_PASS_DEF_CONVERTTOEXTENDEDBASIS
+#include "lib/Dialect/Openfhe/Transforms/Passes.h.inc"
+
+struct ConvertFastRotationOp : public OpRewritePattern<FastRotationOp> {
+  using OpRewritePattern<FastRotationOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(FastRotationOp op,
+                                PatternRewriter& rewriter) const override {
+    auto fastRotExt = FastRotationExtOp::create(
+        rewriter, op->getLoc(), op.getType(), op.getCryptoContext(),
+        op.getInput(), op.getIndex(), op.getPrecomputedDigitDecomp(),
+        /*addFirst=*/true);
+
+    auto keySwitchDown =
+        KeySwitchDownOp::create(rewriter, op->getLoc(), op.getType(),
+                                op.getCryptoContext(), fastRotExt.getResult());
+
+    rewriter.replaceOp(op, keySwitchDown.getResult());
+    return success();
+  }
+};
+
+struct ConvertToExtendedBasis
+    : impl::ConvertToExtendedBasisBase<ConvertToExtendedBasis> {
+  using ConvertToExtendedBasisBase::ConvertToExtendedBasisBase;
+
+  void runOnOperation() override {
+    MLIRContext* context = &getContext();
+    RewritePatternSet patterns(context);
+    patterns.add<ConvertFastRotationOp>(context);
+    walkAndApplyPatterns(getOperation(), std::move(patterns));
+  }
+};
+
+}  // namespace openfhe
+}  // namespace heir
+}  // namespace mlir
--- a/lib/Dialect/Openfhe/Transforms/HoistKeySwitchDown.cpp
+++ b/lib/Dialect/Openfhe/Transforms/HoistKeySwitchDown.cpp
@@ -1,0 +1,65 @@
+#include "lib/Dialect/Openfhe/Transforms/HoistKeySwitchDown.h"
+
+#include "lib/Dialect/Openfhe/IR/OpenfheOps.h"
+#include "mlir/include/mlir/IR/PatternMatch.h"  // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"     // from @llvm-project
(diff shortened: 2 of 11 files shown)


lib/Dialect/Openfhe/Transforms/ConvertToExtendedBasis.cpp
```diff
@@ -0,0 +1,47 @@
...
+    SmallVector<FastRotationOp> fastRotOps;
+    getOperation()->walk([&](FastRotationOp op) { fastRotOps.push_back(op); });
+
+    for (FastRotationOp op : fastRotOps) {
+      builder.setInsertionPoint(op);
+
+      auto fastRotExt = FastRotationExtOp::create(
+          builder, op->getLoc(), op.getType(), op.getCryptoContext(),
+          op.getInput(), op.getIndex(), op.getPrecomputedDigitDecomp(),
+          /*addFirst=*/true);
```
toprakmurat: I set `addFirst=true` for all rotations, assuming hoisting pass will optimize redundant key-switches. Does this cause any problems?

lib/Dialect/Openfhe/Transforms/HoistKeySwitchDown.cpp
```diff
@@ -0,0 +1,82 @@
...
+};
+
+struct HoistKeySwitchDown : impl::HoistKeySwitchDownBase<HoistKeySwitchDown> {
+  using HoistKeySwitchDownBase::HoistKeySwitchDownBase;
+
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    patterns.add<HoistKeySwitchThrough<AddOp>>(&getContext());
+    patterns.add<HoistKeySwitchThrough<AddInPlaceOp>>(&getContext());
+    patterns.add<HoistKeySwitchThrough<MulOp>>(&getContext());
```
toprakmurat, line 55: Are `Add`, `AddInPlace`, and `Mul` the complete set of operations where hoisting key_switch_down is valid?
@j2kun mentioned those ones in [related issue comment](https://github.com/google/heir/issues/2519#issuecomment-3741079771)

toprakmurat:
Hi @asraa, I'd appreciate your review of these changes when you have time. Should I also add interpreter benchmarks comparing performance with and without these passes?

j2kun:
@AlexanderViand Asra and I are a bit strapped for time these days. Do you think you'd have time to review this PR?

lib/Dialect/Openfhe/Transforms/HoistKeySwitchDown.cpp
```diff
@@ -0,0 +1,82 @@
...
+    rewriter.replaceOp(op, newKs.getResult());
+    if (lhsKs) rewriter.eraseOp(lhsKs);
+    if (rhsKs) rewriter.eraseOp(rhsKs);
+
+    return success();
+  }
+};
+
+// Transforms: key_switch_down(key_switch_down(a)) => key_switch_down(a)
+struct EliminateRedundantKeySwitchDown
```
j2kun: If you add the Idempotent trait to the KeySwitchDown op's tablegen definition, then you can remove this pattern an the greedy pattern rewrite driver should fold it automatically.
https://github.com/llvm/llvm-project/blob/e6f30334de720b0473f1d797c008ccf74fb300b8/mlir/lib/IR/Operation.cpp#L870

Example trait usage upstream: https://github.com/llvm/llvm-project/blob/77b31b90d7bcc86692a9dda98409768686546cd0/mlir/include/mlir/Dialect/Arith/IR/ArithOps.td#L733

I believe the trait is defined in OpBase.td so you shouldn't need any extra imports.

lib/Dialect/Openfhe/Transforms/ConvertToExtendedBasis.cpp
j2kun: I actually don't understand what `addFirst` is doing in OpenFHE. The docs say `If true, the first element c0 is also computed`. Not clear to me what that means. I suspect this may come up again when these passes are tested end-to-end.

lib/Dialect/Openfhe/Transforms/ConvertToExtendedBasis.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+struct ConvertToExtendedBasis
+    : impl::ConvertToExtendedBasisBase<ConvertToExtendedBasis> {
+  using ConvertToExtendedBasisBase::ConvertToExtendedBasisBase;
+
+  void runOnOperation() override {
+    IRRewriter builder(getOperation()->getContext());
+
+    SmallVector<FastRotationOp> fastRotOps;
+    getOperation()->walk([&](FastRotationOp op) { fastRotOps.push_back(op); });
```
j2kun: nit: you can also use `mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h` which provides a `walkAndApplyPatterns` for this. Not necessary since this pass is relatively simple, but doing this would (a) handle the insertion point management and op filtering and (b) give you access to the `PatternRewriter` which makes debugging easier IMO.

j2kun:
Also when you're ready, please squash your commits into one, as per https://heir.dev/docs/development/#pull-request-review-flow

j2kun:
Thank you @toprakmurat!

PR #2632, opened 2026-02-10
Update LWEToPolynomial lowering

crockeea:
The primary goal of this PR is to update the LWE->Polynomial lowering in light of the new LWE ops introduced in #2602. This required the addition of two new `Polynomial` ops: `ExtractSliceOp` and `ConvertBasisOp`.

I did some light housekeeping along the way:
- Converted several `cast`s to `dyn_casts` with a check. I found that without this, errors were impenetrable.
- Extracted the core of `ExtractSliceOp` type inference logic and moved it to a template in RNSOps.h. This is now used in `rns::ExtractSliceOp`, `lwe::ExtractSliceOp`, and `polynomial::ExtractSliceOp`.
- Made `ExtractSliceOp` consistent across RNS, LWE, and Polynomial dialects
- Added several `assemblyFormat`s to new ops for consistency and test readability
- Added missing op descriptions

The main impetus for this PR was the ability to lower from `ckks.relinearize` to `Polynomial`. I added a "test", but I haven't really done a test like this before. I think the goal of the test should be that it doesn't crash, rather than specifying any specific output (since each step is verified already). I'm not sure where it should go, or what it should look like.

### DIFF
--- a/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
@@ -12,6 +12,7 @@
 #include "lib/Dialect/Polynomial/IR/PolynomialAttributes.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialOps.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialTypes.h"
+#include "lib/Dialect/RNS/IR/RNSOps.h"
 #include "lib/Dialect/Random/IR/RandomEnums.h"
 #include "lib/Dialect/Random/IR/RandomOps.h"
 #include "lib/Dialect/Random/IR/RandomTypes.h"


lib/Dialect/LWE/IR/LWEOps.td
```diff
@@ -306,6 +306,7 @@ def LWE_ExtractCoeffOp : LWE_Op<"extract_coeff", [Pure, DeclareOpInterfaceMethod
     LWERingElt:$output
   );
   let hasVerifier = 1;
+  let assemblyFormat = "$value attr-dict `:` type($value) `->` type($output)";
```
j2kun:
```suggestion
  let assemblyFormat = "$value attr-dict `:` type($value)";
```

Optional: I believe ops that have `InferReturnTypeOpInterface` can actually leave the return type out of the assembly format. So you could omit this if you prefer brevity over consistency (and if you want to override the default `operands attr-dict `:` functional-type(operands, results)` then maybe go as brief as possible?)

lib/Dialect/Polynomial/IR/PolynomialOps.td
```diff
@@ -396,5 +396,37 @@ def Polynomial_EvalOp : Polynomial_Op<"eval", [AllTypesMatch<["value", "output"]
...
+  let hasVerifier = 1;
+}
+
+def Polynomial_ConvertBasisOp : Polynomial_Op<"convert_basis", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Change a polynomial's RNS basis";
+  let description = [{
+    Given a polynomial with RNS coefficients, output a polynomial where the coefficients
+    have the target output basis. The output value of each coefficient is the RNS
+    representation of the input coefficient, lifted to its centered representative over
+    the integers, and reduced by the output modulus.
```
j2kun:
```suggestion
    representation of the input coefficient, equivalent to the result of
    lifting the input RNS limbs to its centered representative over
    the integers, and reducing by the output modulus.
```

nit: I try to have the op semantics not describe a particular lowering one has in mind, since optimizations may make that incorrect. In particular, we don't enforce a particular representative for an RNS-typed value at any given point in the IR because, IMO, the whole point of using equivalence classes is that we can be flexible with when we reduce in order to get performance benefits.

tests/Dialect/LWE/Conversions/lwe_to_polynomial/lower_relinearize_to_polynomial.mlir
j2kun, line 1: I think it would be fine to put this smoke test in the same directory as the lwe-to-polynomial pass tests. Eventually we'd have a pipeline that encapsulates a set of passes and we have test directories for testing those pipelines in isolation. E.g., https://github.com/google/heir/tree/main/tests/Transforms/mlir_to_openfhe_ckks

(probably they should live in like `tests/Pipelines` or similar, but that's where they are now)

lib/Dialect/LWE/IR/LWEOps.td
crockeea: I don't understand everything you said there, but I updated several places with type inference to exclude the `-> type($output)`.

lib/Dialect/Polynomial/IR/PolynomialOps.td
crockeea: Fix here and in LWEOps.td

tests/Dialect/LWE/Conversions/lwe_to_polynomial/lower_relinearize_to_polynomial.mlir
crockeea: Moved to tests/Dialect/LEW/Conversions/lwe_to_polynomial/lower_relinearize_to_polynomial.mlir

j2kun (approved):

crockeea:
NOTE: I just made a questionable change by removing `ConvertAny` from `addTensorOfTensorConversionPatterns` as a hack to fix the now-failing LWE/Conversions/lwe_to_polynomial/ops_dimension_error test. 

I added `addTensorOfTensorConversionPatterns` to the LWEToPolynomial lowering since we now lower tensors of ciphertexts (e.g., KeySwitch keys), which get lowered to tensors of tensors of Polynomials. However, `addTensorOfTensorConversionPatterns` registers the `ConvertAny` pattern, which, for reasons I don't really understand, make ops_dimension_error fail. 

Make sure this is reasonable!

j2kun:
> NOTE: I just made a questionable change by removing `ConvertAny` from `addTensorOfTensorConversionPatterns` as a hack to fix the now-failing LWE/Conversions/lwe_to_polynomial/ops_dimension_error test.
> 
> I added `addTensorOfTensorConversionPatterns` to the LWEToPolynomial lowering since we now lower tensors of ciphertexts (e.g., KeySwitch keys), which get lowered to tensors of tensors of Polynomials. However, `addTensorOfTensorConversionPatterns` registers the `ConvertAny` pattern, which, for reasons I don't really understand, make ops_dimension_error fail.
> 
> Make sure this is reasonable!

I checked this and all tests are still passing; only PolynomialToModArith uses it and those tests are quite exhaustive. So if someone liked it then they shoulda put a test on it :ring:

j2kun:
Once the pull_ready label goes on it and I start applying internal patches, additional changes you make to the branch may not be updated. I recommend rebasing over main if you have additional changes you'd like to make on this branch, but easiest would be to close the PR and open a new one (as the main changes were merged as 342da339d7ca80f95243f3728fa59d49c8444612)

crockeea:
I think that's all the changes needed. I didn't run the full test suite, and it failed after your initial approval. I'm assuming it wouldn't be auto-merged if CI failed, so I think this should have merged all of my changes.

crockeea:
I'm confused: this appears to now be merged into main.

j2kun:
Right, I did some internal patches and made the same fixes as you. No worries.

j2kun:
> I'm confused: this appears to now be merged into main.

Cf. https://heir.dev/docs/development/#internal-review-details

crockeea:
This PR was merged, but the diff is wrong. The real diff is [here](https://github.com/google/heir/commit/342da339d7ca80f95243f3728fa59d49c8444612).

PR #2647, opened 2026-02-12
Refactor of NTT/iNTT to make inputs and outputs both Polynomials

crockeea:
The primary goal of this PR is to make inputs and outputs of NTT/INTT both Polynomial, and to support the creation of NTT nodes with an RNS root of unity. To that end:
- Created ModArithAttr for ModArith roots of unity (single modulus)
- Created RNSAttr for RNS roots of unity
- Added a "form" attribute to polynomials indicating whether the polynomial is in coefficient or evaluation (NTT) form.
- Changed NTTOp and INTTOp to take polynomials and produce polynomials (with a different form)
- Added type inference to NTTOp/INTTOp
- Updated corresponding verifiers
- Updated NTT lowering in PolynomialToModArith. Note that the lowering does not support RNS-NTT ops yet; that will be in another PR.
- Updated tests

### DIFF
--- a/lib/Dialect/ModArith/IR/ModArithAttributes.td
+++ b/lib/Dialect/ModArith/IR/ModArithAttributes.td
@@ -1,0 +1,37 @@
+#ifndef LIB_DIALECT_MODARITH_IR_MODARITHATTRIBUTES_TD_
+#define LIB_DIALECT_MODARITH_IR_MODARITHATTRIBUTES_TD_
+
+include "lib/Dialect/ModArith/IR/ModArithDialect.td"
+include "lib/Dialect/ModArith/IR/ModArithTypes.td"
+
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
+
+class ModArith_Attr<string name, string attrMnemonic, list<Trait> traits = []>
+    : AttrDef<ModArith_Dialect, name, traits> {
+  let mnemonic = attrMnemonic;
+}
+
+def ModArith_ModArithAttr : ModArith_Attr<"ModArith", "value"> {
+  let summary = "a typed modular arithmetic value";
+  let description = [{
+    A typed modular arithmetic value.
+
+    The `type` parameter is expected to be a modular arithmetic coefficient
+    type, and `value` is the corresponding integer representative.
+
+    Example:
+
+    ```mlir
+    #v = #mod_arith.value<17 : !mod_arith.int<...>>
+    ```
+  }];
+  let parameters = (ins
+    "::mlir::heir::mod_arith::ModArithType":$type,
+    "::mlir::IntegerAttr":$value
+  );
+  let hasCustomAssemblyFormat = 1;
+}
+
+#endif  // LIB_DIALECT_MODARITH_IR_MODARITHATTRIBUTES_TD_
--- a/lib/Dialect/Polynomial/IR/PolynomialOps.cpp
+++ b/lib/Dialect/Polynomial/IR/PolynomialOps.cpp
@@ -186,69 +186,100 @@ bool isPrimitiveNthRootOfUnity(const APInt& root, const APInt& n,
 
 /// Verify that the types involved in an NTT or INTT operation are
 /// compatible.
-static LogicalResult verifyNTTOp(Operation* op, PolynomialType poly,
-                                 RankedTensorType tensorType,
-                                 std::optional<PrimitiveRootAttr> root) {
-  Attribute encoding = tensorType.getEncoding();
-  if (!encoding) {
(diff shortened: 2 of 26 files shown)


lib/Dialect/ModArith/IR/ModArithAttributes.td
```diff
@@ -0,0 +1,37 @@
...
+  let description = [{
+    A typed modular arithmetic value.
+
+    The `type` parameter is expected to be a modular arithmetic coefficient
+    type, and `value` is the corresponding integer representative.
+
+    Example:
+
+    ```mlir
+    #v = #mod_arith.value<type=!mod_arith.int<17 : i32>, value=3 : i32>
```
j2kun: I'm fine with it, but if you feel this is too verbose it can be simplified with a custom parser/printer. An example syntax might be:

`#mod_arith.value<3 mod 17 : i32>`

At least, this would make it simpler to enforce that the value has the same storage type as the modulus.

I can also do this in a followup if desired.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
```diff
@@ -186,68 +186,77 @@ bool isPrimitiveNthRootOfUnity(const APInt& root, const APInt& n,
...
-    if (!isPrimitiveNthRootOfUnity(rootValue, rootDegree, cmod)) {
-      return op->emitOpError()
-             << "provided root " << rootValue.getZExtValue()
-             << " is not a primitive root " << "of unity mod "
-             << cmod.getZExtValue() << ", with the specified degree "
-             << rootDegree.getZExtValue();
+    auto maCoeffType =
+        dyn_cast<mod_arith::ModArithType>(inputRing.getCoefficientType());
+    auto rnsCoeffType = dyn_cast<rns::RNSType>(inputRing.getCoefficientType());
+    if (maCoeffType) {
```
j2kun: This is fine, but stylistically you may appreciate using `llvm::TypeSwitch` for this. Example: https://github.com/google/heir/blob/5b2512584cc4fcab3209c65523a10fbd90bb779f/lib/Dialect/LWE/IR/LWEAttributes.cpp#L30-L36

```cpp
return llvm::TypeSwitch<Type, LogicalResult>(inputRing.getCoefficientType())
      .Case<mod_arith::ModArithType>(
          [&](auto ty) { ... })
      .Case<rns::RNSType>(
          [&](auto ty) { ... })
      .Default([&](Type ty) { 
           return op->emitOpError() 
              << "Ring has unsupported coefficient type " << ty; });
```

lib/Dialect/RNS/IR/RNSAttributes.td
```diff
@@ -0,0 +1,39 @@
...
+  let description = [{
+    A typed RNS value with one integer per basis limb.
+
+    The `type` parameter is expected to be an RNS coefficient type, and
+    `values` stores one residue for each limb.
+
+    Example:
+
+    ```mlir
+    #v = #rns.value<type=!rns.rns<...>, values = [3 : i32, 9 : i32]>
```
j2kun:
```suggestion
    #v = #rns.value<[3 , 9] : !rns.rns<...>>
```

To match the `assemblyFormat` below

lib/Dialect/ModArith/IR/ModArithAttributes.td
crockeea: Thanks for the suggestion; I didn't know such a thing was possible! One thing I'm trying to do is prevent/reduce the duplication of integers in the mlir files, so I opted for this: `#root_val = #mod_arith.value<1925 : !coeff_ty>`

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: Nifty; I'll keep this in mind!

lib/Dialect/RNS/IR/RNSAttributes.td
crockeea: I think this is updated correctly now.

crockeea:
@j2kun Per https://github.com/google/heir/issues/2649#issuecomment-3900273801, check my RNS dependency on ModArith.

j2kun:
> @j2kun Per [#2649 (comment)](https://github.com/google/heir/issues/2649#issuecomment-3900273801), check my RNS dependency on ModArith.

Try this patch

https://gist.github.com/j2kun/986af859eefa3431c49dfbe4ab5035aa

It makes the syntax slightly different, but has the benefit of inferring the RNS type from the values directly. (Maybe, as you pointed out before, this is a benefit because then the value attributes provided to the RNS attr are themselves attributes and can be referred to via aliases and deduped)

crockeea:
Thanks Jeremy, I was struggling to figure that out :)

crockeea:
Squashed commits; ready for merge.

PR #2653, opened 2026-02-13
Adds ElementAttr to tensor_ext.layout

VedantParanjape:
TensorExt.Layout could only be assigned a LayoutAttr, which is an quasi-affine relation. This adds support to handle arbitrary permutations of ctxt to vector slots.

### DIFF
--- a/lib/Dialect/TensorExt/IR/TensorExtOps.td
+++ b/lib/Dialect/TensorExt/IR/TensorExtOps.td
@@ -92,6 +92,9 @@ def TensorExt_RemapOp : TensorExt_Op<"remap", [Pure, AllTypesMatch<["input", "ou
 // Forces ops to use a general Attribute and dyn_cast to the specific kind of
 // layout they support.
 def LayoutLike : AnyAttrOf<[
+  // A list of tuples [a, b, c, d] representing an explicit map (ct, slot) ->
+  // (ct, slot) defined by f(a, b) = (c, d).
+  AnyI64ElementsAttr,
   TensorExt_LayoutAttr,
 ]>;
 
--- a/lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
+++ b/lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
@@ -195,6 +195,51 @@ static FailureOr<Value> implementUnpackOpNew(
   return loop.value().getResults()[0];
 }
 
+static FailureOr<Value> implementAssignLayoutPermutation(
+    Value input, DenseIntElementsAttr permutation, int64_t ciphertextSize,
+    ImplicitLocOpBuilder& builder,
+    const std::function<void(Operation*)>& createdOpCallback) {
+  auto elementType = getElementTypeOrSelf(input.getType());
+
+  // TODO(#2666): This logic assumes ctxt = 0, this can be extended to handle a
+  // more general case. permutation is <N x 4 x i64>: each row is [src_ct,
+  // src_slot, dst_ct, dst_slot]. src_ct and dst_ct are always 0 (single
+  // ciphertext), so treat as 1D: extract from input[0][src_slot] and insert
+  // into result[0][dst_slot].
+  auto ciphertextType = RankedTensorType::get({1, ciphertextSize}, elementType);
+  auto zeroCtxt = arith::ConstantOp::create(
+      builder, ciphertextType, builder.getZeroAttr(ciphertextType));
+  createdOpCallback(zeroCtxt);
+  Value result = zeroCtxt.getResult();
+  auto ctIdxOp = arith::ConstantIndexOp::create(builder, 0);
+  createdOpCallback(ctIdxOp);
+  Value ctIdx = ctIdxOp.getResult();
+
+  for (auto it = permutation.value_begin<APInt>();
+       it != permutation.value_end<APInt>();) {
+    // skip src_ct (always 0)
+    ++it;
+    int64_t srcSlot = (*it++).getSExtValue();
+    // skip dst_ct (always 0)
+    ++it;
+    int64_t dstSlot = (*it++).getSExtValue();
+
+    auto srcSlotIdx = arith::ConstantIndexOp::create(builder, srcSlot);
+    createdOpCallback(srcSlotIdx);
+    auto extracted = tensor::ExtractOp::create(builder, input,
+                                               ValueRange{ctIdx, srcSlotIdx});
+    createdOpCallback(extracted);
+    auto dstSlotIdx = arith::ConstantIndexOp::create(builder, dstSlot);
+    createdOpCallback(dstSlotIdx);
+    auto insertOp = tensor::InsertOp::create(
+        builder, extracted.getResult(), result, ValueRange{ctIdx, dstSlotIdx});
+    createdOpCallback(insertOp);
+    result = insertOp.getResult();
+  }
+
+  return result;
+}
+
 FailureOr<Value> implementAssignLayout(
     Value input, Attribute layout, int64_t ciphertextSize,
     ImplicitLocOpBuilder& builder,
@@ -203,6 +248,10 @@ FailureOr<Value> implementAssignLayout(
   if (LayoutAttr layoutAttr = dyn_cast<LayoutAttr>(layout)) {
     return implementAssignLayoutNew(input, layoutAttr, ciphertextSize, builder,
                                     createdOpCallback);
+  } else if (DenseIntElementsAttr elementAttr =
+                 dyn_cast<DenseIntElementsAttr>(layout)) {
+    return implementAssignLayoutPermutation(input, elementAttr, ciphertextSize,
+                                            builder, createdOpCallback);
   }
   return builder.emitError() << "Unsupported layout attribute type: " << layout;
 };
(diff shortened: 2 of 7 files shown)


lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -195,6 +195,43 @@ static FailureOr<Value> implementUnpackOpNew(
...
 }
 
+static FailureOr<Value> implementAssignLayoutPermutation(
+    Value input, ElementsAttr permutation, int64_t ciphertextSize,
+    ImplicitLocOpBuilder& builder,
+    const std::function<void(Operation*)>& createdOpCallback) {
+  auto inputType = dyn_cast<RankedTensorType>(input.getType());
+  auto elementType = getElementTypeOrSelf(input.getType());
+
+  auto permValues = SmallVector<int64_t>(permutation.getValues<int64_t>());
```
asraa: since there's no layout attribute verifier and the code below only checks if the attribute is an ElementsAttr, you might need an additional check somewhere that the elements are integer/i64 and not floats since I think otherwise getValues will assert failure

j2kun:
@VedantParanjape In some of the examples we discussed, the packing was not purely a permutation (there were replicated values among slots). I worry that a strict "permutation" does not suffice, when actually you want a mapping from slots to data values.

In the case where we do remapping from (ct, slot) -> (ct, slot), an AnyI64ElementsAttr suffices because we can group the values in groups of four (a, b, c, d) being interpreted as mapping f(a, b) = (c, d). Can you also do that in your case? (maybe yes because you're supporting flattened input tensors?)

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
VedantParanjape: I added a simple check, does that suffice?

VedantParanjape:
> Hey @VedantParanjape ! Code looks really great, thanks for the contribution.
> 
> I was just wondering if you could add a few small tests with a permutation layout in tests/Transforms/convert_to_ciphertext_semantics

I have added a test case, but I need to fix the verifier before this testcase actually passes.

VedantParanjape:
> In the case where we do remapping from (ct, slot) -> (ct, slot), an AnyI64ElementsAttr suffices because we can group the values in groups of four (a, b, c, d) being interpreted as mapping f(a, b) = (c, d). Can you also do that in your case? (maybe yes because you're supporting flattened input tensors?)

yeah, since I am just supporting a flattened input tensor, the ct (a, b) will really just be 0 in all the cases, so we can just eliminate it? 

Overall your design makes more sense, I will make the changes. I need to just duplicate MappingLike code.

lib/Dialect/TensorExt/IR/TensorExtOps.td
```diff
@@ -92,6 +92,11 @@ def TensorExt_RemapOp : TensorExt_Op<"remap", [Pure, AllTypesMatch<["input", "ou
 // Forces ops to use a general Attribute and dyn_cast to the specific kind of
 // layout they support.
 def LayoutLike : AnyAttrOf<[
+  // An explicit permutation from logical data indices to ciphertext slot
```
asraa: nit: needs an update now

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -195,6 +195,84 @@ static FailureOr<Value> implementUnpackOpNew(
...
 
+static FailureOr<Value> implementAssignLayoutPermutation(
+    Value input, DenseIntElementsAttr permutation, int64_t ciphertextSize,
+    ImplicitLocOpBuilder& builder,
+    const std::function<void(Operation*)>& createdOpCallback) {
+  auto elementType = getElementTypeOrSelf(input.getType());
+
+  // permutation is <N x 4 x i64>: each row is [src_ct, src_slot, dst_ct, dst_slot].
+  // src_ct and dst_ct are always 0 (single ciphertext), so treat as 1D:
+  // extract from input[src_slot] and insert into result[0][dst_slot].
```
asraa:
```suggestion
  // src_ct and dst_ct are always 0 (single ciphertext), so treat as 1D:
  // extract from input[0][src_slot] and insert into result[0][dst_slot].
```

FWIW in your work the ciphertexts are always single, but the implementation here can still be general enough to support that (otherwise why not just omit the src_ct / dst_ct?) so maybe i would support all general permutations since it's not too much work to do that. But it's only your input creation that would happen to have ct = 0 in all cases

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -195,6 +195,84 @@ static FailureOr<Value> implementUnpackOpNew(
...
+
+  for (auto it = permutation.value_begin<APInt>();
+       it != permutation.value_end<APInt>();) {
+    ++it;  // skip src_ct (always 0)
+    int64_t srcSlot = (*it++).getSExtValue();
+    ++it;  // skip dst_ct (always 0)
+    int64_t dstSlot = (*it++).getSExtValue();
+
+    Value srcSlotIdx = arith::ConstantIndexOp::create(builder, srcSlot);
+    Value extracted  = tensor::ExtractOp::create(builder, input, ValueRange{srcSlotIdx});
```
asraa: since the tensor is still 2 dimensional, i think you'll still need two slot indices to access

lib/Transforms/ConvertToCiphertextSemantics/TypeConversion.cpp
```diff
@@ -41,5 +41,12 @@ Type materializeScalarLayout(Type type, LayoutAttr attr, int ciphertextSize) {
   return RankedTensorType::get({1, ciphertextSize}, type);
 }
 
+Type materializePermutationLayout(Type elementType,
+                                  DenseIntElementsAttr permutation,
+                                  int ciphertextSize) {
+  // src_ct and dst_ct are always 0; output is always a single ciphertext.
```
asraa, line 48: if you decide to generalize, then this would also need an update to the max output ct value

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
VedantParanjape: I want to implement it for a more general case, but I am struggling with making the pipeline work as is. Is it okay if I submit another PR to extend this specialized case to a more general one?

lib/Transforms/ConvertToCiphertextSemantics/TypeConversion.cpp
VedantParanjape: I'll add TODO blocks at these call sites that need more work.

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
asraa: Oh yeah! Make an issue and leave a TODO comment to generalize it :)

VedantParanjape: perfect! ty

VedantParanjape:
@j2kun @asraa The current version passes the new test case locally. Could you please trigger the workflow job?

tests/Transforms/convert_to_ciphertext_semantics/assign_layout.mlir
```diff
@@ -64,3 +64,75 @@ module {
...
+// CHECK-DAG: %[[c4:.*]] = arith.constant 4 : i32
+// CHECK-DAG: %[[c3:.*]] = arith.constant 3 : i32
+// CHECK-DAG: %[[c5:.*]] = arith.constant 5 : i32
+// CHECK-DAG: %[[c2:.*]] = arith.constant 2 : i32
+// CHECK-DAG: %[[c6:.*]] = arith.constant 6 : i32
+// CHECK-DAG: %[[c1:.*]] = arith.constant 1 : i32
+// CHECK-DAG: %[[c7:.*]] = arith.constant 7 : i32
+// CHECK-DAG: %[[cst:.*]] = arith.constant dense<0> : tensor<1x32xi16>
+// CHECK-DAG: %[[c0:.*]] = arith.constant 0 : i32
+// CHECK-DAG: %[[EXT0:.*]] = tensor.extract %arg0[%[[c0]], %[[c0]]] : tensor<1x32xi16>
```
j2kun: nit: in general I think it's safe to leave out things like the type suffix when it's otherwise clear from the context. Less noise in the test files to update later.

lib/Transforms/ConvertToCiphertextSemantics/TypeConversion.cpp
```diff
@@ -41,5 +41,13 @@ Type materializeScalarLayout(Type type, LayoutAttr attr, int ciphertextSize) {
   return RankedTensorType::get({1, ciphertextSize}, type);
 }
 
+Type materializePermutationLayout(Type elementType,
+                                  DenseIntElementsAttr permutation,
+                                  int ciphertextSize) {
+  // TODO: Extend to a more general case where src_ct and dst_ct != 0
```
j2kun: File an issue and use the notation

```
// TODO(#<issue_number>): foo
```

And [this action](https://github.com/j2kun/todo-backlinks) will trigger to put a link to the source line on the GH issue directly. (And notify if the issue is closed while the TODO is still in the codebase)

Same for the other TODO above (they can both link to the same issue)

j2kun:
Ah, you may also need to run `pre-commit run --all-files` to ensure the formatters are applied.

j2kun (approved):

PR #2656, opened 2026-02-15
Move CKKSToLWE from Conversions to Transforms

### LINKED ISSUE #2627: Move CKKS->LWE Conversion to Transform
On 2/9 office hours, we discussed when an (abstract) transform should be a lowering, and when it should be a Transform. Decision was: if it changes types, it's a lowering/Conversion, if not, it belongs in Transforms.

Currently, CKKS->LWE is a Conversion that doesn't change types, so it should be a Transform for consistency.

toprakmurat:
Fixes #2627.

This PR moves the pass from `lib/Dialect/CKKS/Conversions/CKKSToLWE/` to 
`lib/Dialect/CKKS/Transforms/` and updates all references accordingly. 
The `--ckks-to-lwe` flag name remains unchanged.


asraa:
It looks like we have some merge conflicts above ^

PR #2660, opened 2026-02-17
[Interpreter] Implement thread-safe parallelization for scf.forall

### LINKED ISSUE #2544: Support parallelization in scf.forall in the openfhe interpreter
Right now the loop will just execute in sequence, since if you enable the loop parallelization with
`#pragma omp parallel for`

then there's a race and the ciphertextVector result won't necessarily store the resulting elements after the loop iterations are executed in parallel.
see https://github.com/google/heir/pull/2535/files

akashmadhu4:
Previously, scf.forall executed sequentially. Naive OpenMP parallelization caused data races on shared maps.
This PR fixes #2544 by cloning the interpreter per thread.

Improvements / Fixes
Interpreter now holds read-only liveness analysis via std::shared_ptr (no deep copy).
visit(scf::ForallOp) updated with #pragma omp parallel for.
TestOpenfheParallelFastRotation passes under parallel execution.

Fixes #2544

### DIFF
--- a/lib/Target/OpenFhePke/BUILD
+++ b/lib/Target/OpenFhePke/BUILD
@@ -1,5 +1,6 @@
 # OpenFhePke Emitter
 
+load("@heir//bazel/openfhe:copts.bzl", "OPENMP_COPTS", "OPENMP_LINKOPTS")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 load("@rules_cc//cc:cc_test.bzl", "cc_test")
 
@@ -124,7 +125,8 @@ cc_library(
     copts = select({
         "@heir//:config_openfhe_enable_timing": ["-DOPENFHE_ENABLE_TIMING"],
         "//conditions:default": [],
-    }),
+    }) + OPENMP_COPTS,
+    linkopts = OPENMP_LINKOPTS,
     deps = [
         "@heir//lib/Dialect/Mgmt/IR:Dialect",
         "@heir//lib/Dialect/ModArith/IR:Dialect",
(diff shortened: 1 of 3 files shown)


lib/Target/OpenFhePke/BUILD
```diff
@@ -124,7 +124,9 @@ cc_library(
     copts = select({
         "@heir//:config_openfhe_enable_timing": ["-DOPENFHE_ENABLE_TIMING"],
         "//conditions:default": [],
-    }),
+    }) + ["-fopenmp"],
```
asraa: you may want to use https://github.com/google/heir/blob/1210ad37dc9531d6e60d8ddbce81dbd79f7626a6/bazel/openfhe/copts.bzl#L13-L19 instead, so it would look like

```suggestion
    }) + OPENMP_COPTS,
```

lib/Target/OpenFhePke/Interpreter.h
```diff
@@ -246,7 +246,8 @@ class Interpreter {
   llvm::DenseMap<Value, EvalKeyT> evalKeys;
   llvm::DenseMap<Value, FastRotPrecompT> fastRotPrecomps;
 
-  Liveness liveness;
+  // liveness as shared_ptr to avoid expensive copying
```
asraa, line 249: nice, thank you!

asraa (commented):
@akashmadhu4 wonderful thank you so so much for fixing this issue!

just a quick fix-me on the BUILD files to make sure that openmp is only linked and added to the copts when parallelism is enabled using the bazel configuration

lib/Target/OpenFhePke/BUILD
akashmadhu4: Thanks you for the suggestion,@asraa !. I’ve updated the BUILD files to use OPENMP_COPTS and OPENMP_LINKOPTS instead of hardcoding -fopenmp, and also fixed some formatting issues

Issue #2598, opened 2026-02-02
Upstream KeyMemRT to HEIR


j2kun:
KeyMemRT is a memory management runtime built on top of HEIR as a research project. It reduces memory requirements of a program by storing key material on disk and loading it in advance of it being needed in the program.

I have discussed this work at length with the authors. They plan to open source the code soon, and then it will be fair game to start upstreaming. There are a few caveats with their work vis-a-vis loop support, but the core dialects and transforms should be viable to upstream with minimal changes. It may also require adding a slight patch to our OpenFHE dependency to provide the appropriate API and/or bug fixes for serialization and deserialization.

https://arxiv.org/abs/2601.18445

> KeyMemRT Compiler and Runtime: Unlocking Memory-Scalable FHE
> Eymen Ünay, Björn Franke, Jackson Woodruff
> Fully Homomorphic Encryption (FHE) enables privacy preserving computation but it suffers from high latency and memory consumption. The computations are secured with special keys called rotation keys which often take up the majority of memory. In complex FHE applications, these rotation keys can cause a large memory bottleneck limiting program throughput. Existing compilers make little effort to solve this problem, instead relying on systems with massive memory availability. This resource requirement is a barrier to FHE uptake because optimizing FHE programs by hand is challenging due to their scale, complexity and expertise required.
> In this work, we present KeyMemRT; an MLIR based compiler and runtime framework that individually manages rotation key lifetimes to lower memory utilization and to allow arbitrary number of rotation indices to be supported without memory bloating. KeyMemRT relies on dataflow analysis to determine key lifetimes and is the first FHE compiler to provide automatic key management, handle fine-grained key-mangement and manage boostrap keys. We implement frontends for Orion and HEIR and show improvements over state-of-the-art FHE compilers. KeyMemRT achieves memory reduction of 1.74x and a speedup of 1.20x over ANT-ACE, and memory reduction of 1.16x and a speedup of 1.73x over memory-optimized compiler Fhelipe. We provide KeyMemRT as a post-optimizing compiler that can be targeted by any FHE compiler.

j2kun:
CC @eymay, could you drop a link to the open source code when it's ready, no rush.

eymay:
Hey @j2kun, thanks for considering upstreaming KeyMemRT. Here is the open source version of the compiler part: https://github.com/eymay/KeyMemRT-Compiler. I will follow up on the full system including runtime and the pipelines.

PR #2661, opened 2026-02-17
kmrt: add KeyMemRuntime dialect

j2kun:
Part of #2598

Ports https://github.com/eymay/KeyMemRT-Compiler/tree/main/lib/Dialect/KMRT/IR with minor naming/documentation changes.


j2kun:
@eymay Would you be open to reviewing the pieces of my port? I also have one question about the `clear_key` op: is the semantics of this lazy? (e.g., marking a key for eviction without doing a memory op at that moment?)

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeDialect.td
```diff
@@ -0,0 +1,30 @@
...
+include "mlir/IR/OpBase.td"
+
+def KeyMemRuntime_Dialect : Dialect {
+  let name = "kmrt";
+  let summary = "A dialect for KeyMemRuntime operations";
+  let description = [{
+    KeyMemRuntime individually manages rotation key lifetimes to lower memory
+    utilization and allow an arbitrary number of rotation indices to be
+    supported without memory bloating. It treats rotation keys as dataflow
+    elements and provides automatic, fine-grained key management.
```
eymay: This describes more of the KeyMemRT compiler in the paper than the dialect I suppose. It can be better to rename or make it clear to what it is referring.

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeDialect.td
```diff
@@ -0,0 +1,30 @@
...
+
+    The KeyMemRuntime dialect provides:
+
+    - A new type for rotation keys (`!kmrt.rot_key`).
+    - Operations for loading and clearing keys (`kmrt.load_key`, `kmrt.clear_key`).
+    - Prefetching hints to overlap computation with key movement (`kmrt.prefetch_key`).
+    - Support for rotation decomposition and bootstrap key management.
```
eymay: Last point is more of a system feature than the KMRT dialect in the paper so can be removed

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeOps.td
```diff
@@ -0,0 +1,185 @@
...
+    ```
+  }];
+
+  let arguments = (ins AnyTypeOf<[Index, I64]>:$index);
+  let results = (outs KeyMemRuntime_RotKeyType:$rot_key);
+  let assemblyFormat = "$index attr-dict `:` type($index) `->` type($rot_key)";
+}
+
+def KeyMemRuntime_UseKeyOp : Op<KeyMemRuntime_Dialect, "use_key", [Pure]> {
+  let summary = "Reference an already-loaded rotation key";
```
eymay: `kmrt.use_key` and `kmrt.assume_loaded` are quick fix ops to make loop nesting work. While `use_key` is more harmless, `assume_loaded` can cause key management bugs and is not safe. The tension is between having the keys globally and the IR attempting to treat them locally. I would appreciate some improvement in how to ensure type safety while being able to make use of loop nestings.

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeOps.td
```diff
@@ -0,0 +1,185 @@
...
+    This operation should typically be preceded by a `load_key` operation
+    that produced the rotation key being cleared.
+  }];
+
+  let arguments = (ins KeyMemRuntime_RotKeyType:$rot_key);
+  let results = (outs);
+  let assemblyFormat = "$rot_key attr-dict `:` type($rot_key)";
+}
+
+def KeyMemRuntime_RotationOp : Op<KeyMemRuntime_Dialect, "rotation", [Pure, AllTypesMatch<["ciphertext", "result"]>]> {
```
eymay: This op can be improved, it is not actively used in KeyMemRT but aims to be library agnostic rotation operation using a key. Instead, KeyMemRT simply extends `openfhe.rot` with a key argument and uses that https://github.com/eymay/KeyMemRT-Compiler/blob/main/lib/Dialect/Openfhe/IR/OpenfheOps.td#L306

Extending the definition of `rotation` operation in KeyMemRT with a key argument is useful to ensure every rotation operation can use a key that is alive. Since a `load_key` op creates the key value and `clear_key` consumes it, the rotation operation needs to be between the ops to make use of the key. However, this is applicable to `!kmrt.rot_key` types with static indices since we can check the static index.

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeTypes.td
```diff
@@ -0,0 +1,79 @@
...
+}
+
+def KeyMemRuntime_RotKeyType : KeyMemRuntime_Type<"RotKey", "rot_key", [MemRefElementTypeInterface]> {
+  let summary = "A rotation key type with optional static index parameter";
+  let description = [{
+    The RotKeyType represents rotation keys used in the KeyMemRuntime system.
+
+    - If rotation_index is present: Static constant rotation key
+    - If rotation_index is absent: Dynamic rotation key (determined at runtime,
+      e.g., from affine loop induction variables)
```
eymay, line 35: This is strongly tied to the loop representation and the nesting model but this can be improved to only accept tighter ranges such as affine maps. Having a dynamic type breaks the type safety in the IR to a great extent.

eymay (commented):
Hi @j2kun thanks for upstreaming KeyMemRT, I'm happy to review this. I think the naming `KeyMemRuntime` can cause confusion with the actual runtime. We named it `KMRT` because it was a core dialect of KeyMemRT compiler in the paper. I think for HEIR it can be a good idea to make it clear by calling it something like `KeyManagement`, `KeyMgmt` or `KeyMemMgmt` to show it is about the key memory management. But it is still true that for some features (like prefetching) there needs to be a runtime so might as well keep runtime in the name.

An important thing is to ensure the dialect is library agnostic as possible. We focused on OpenFHE so we modified some ops (e.g. `openfhe.rot` [here](https://github.com/eymay/KeyMemRT-Compiler/blob/main/lib/Dialect/Openfhe/IR/OpenfheOps.td#L306)) in its dialect but that might not be necessary.

eymay:
> @eymay Would you be open to reviewing the pieces of my port? I also have one question about the `clear_key` op: is the semantics of this lazy? (e.g., marking a key for eviction without doing a memory op at that moment?)

Sure! The compiler marks it for clearing but the runtime can react to it differently depending on its modes of execution. I figured it's best to keep `clear_key` eagerly executed to reduce the memory usage immediately regardless of whether the runtime is prefetching (KeyMemRT - Balanced) or eagerly following the ops (KeyMemRT - Low Memory). The compiler should do analysis to remove immature clears by merging with loads beforehand so the runtime does not have much prediction to do. In OpenFHE, that becomes finding the key in the key dictionary and removing its value.

j2kun:
> I think the naming `KeyMemRuntime` can cause confusion with the actual runtime. We named it `KMRT` because it was a core dialect of KeyMemRT compiler in the paper. I think for HEIR it can be a good idea to make it clear by calling it something like `KeyManagement`, `KeyMgmt` or `KeyMemMgmt` to show it is about the key memory management. But it is still true that for some features (like prefetching) there needs to be a runtime so might as well keep runtime in the name.

There is a prior art of MLIR dialects with "runtime" in the name, for example [tensorflow runtime (tfrt)](https://github.com/tensorflow/runtime/blob/master/documents/tfrt_host_runtime_design.md) and [AIR runtime (airrt)](https://xilinx.github.io/mlir-air/AIRRtDialect.html). So I'm not too worried, but I'll think on it.

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeOps.td
j2kun: yes I think I will remove this op, and rely on other dialects to support the key type from this dialect.

lib/Dialect/KeyMemRuntime/IR/KeyMemRuntimeTypes.td
j2kun: So we actually did some static analysis of this since this PR was first pushed (cf. RotationAnalysis), and while it currently only annotates the overall module with the total set of rotation indices required (dynamic or static), we could go further and annotate the individual rotation ops with the indices that it uses.

I would be interested to see a bit more details about the static analysis you did, since tracing the def-use chain and analyzing the affine context is what I _wanted_ to do, and I only had time for "brute force simulation" of the loop in cleartext 🙈 

I'll leave this as-is and leave this comment here for context :)

tools/heir-opt.cpp
```diff
@@ -191,6 +192,7 @@ int main(int argc, char** argv) {
   registry.insert<debug::DebugDialect>();
   registry.insert<jaxite::JaxiteDialect>();
   registry.insert<jaxiteword::JaxiteWordDialect>();
+  registry.insert<key_mgmt::KeyMgmtDialect>();
```
asraa, line 195: nit: also add to heir-lsp

asraa (approved):

Issue #2662, opened 2026-02-18
Analyze the needed set of rotation keys in the presence of loops and computed rotation shifts


j2kun:
Context: to support rolled linalg kernels (#2655), we will need to introduce loops that have rotation operations that depend on SSA values computed dynamically. Currently all lower-level rotation ops have static attributes for the offset.

This will ultimately require an analysis that can determine a static set of rotation shifts used by a program, that supports the combination of loops (i.e., analyzing the iteration space of an affine/scf for loop) and computed expressions (because the shift used may be an arithmetic expression of the induction variables).

This would be easiest with affine, since `affine.apply` makes static expressions easier to write, but I believe scf.for can be supported as well with constant folding.

PR #2655, opened 2026-02-13
Add option to keep rotate-and-reduce kernel rolled

j2kun:
This PR adds an option to the `convert-to-ciphertext-semantics` pass to keep kernels rolled. So far it only supports the matvec kernel (though not much is left to support the others; I just wanted to get an end-to-end test in first—that's coming next!) and I had to make some small tweaks to the kernel and the IRMaterializingVisitor from the last PR that only did unit testing.

The new test is an exact copy of the input MLIR from the `matvec.mlir` test, with the unroll flag flipped and the CHECK statements changed. For example, here is the output of the `square` matvec kernel:

```mlir
  func.func @square(
      %arg0: !secret.secret<tensor<1x1024xf32>> {tensor_ext.original_type = #original_type}
  ) -> (!secret.secret<tensor<1x1024xf32>> {tensor_ext.original_type = #original_type}) {
    %c-4 = arith.constant -4 : index
    %c1 = arith.constant 1 : index
    %c4 = arith.constant 4 : index
    %c0 = arith.constant 0 : index
    %cst = arith.constant dense<0.000000e+00> : tensor<1x1024xf32>
    %cst_0 = arith.constant dense<0.000000e+00> : tensor<16xf32>
    %cst_1 = arith.constant dense<"0x5036CB..."> : tensor<16x16xf32>
    %0 = call @_assign_layout_3137146596124505264(%cst_1) : (tensor<16x16xf32>) -> tensor<16x1024xf32>
    %1 = call @_assign_layout_6118843033991999771(%cst_0) : (tensor<16xf32>) -> tensor<1x1024xf32>
    %2 = secret.generic(%arg0: !secret.secret<tensor<1x1024xf32>>) {
    ^body(%input0: tensor<1x1024xf32>):
      %3 = scf.for %arg1 = %c0 to %c4 step %c1 iter_args(%arg2 = %cst) -> (tensor<1x1024xf32>) {
        %5 = scf.for %arg3 = %c0 to %c4 step %c1 iter_args(%arg4 = %cst) -> (tensor<1x1024xf32>) {
          %9 = arith.muli %arg1, %c4 : index
          %10 = arith.addi %arg3, %9 : index
          %extracted_slice = tensor.extract_slice %0[%10, 0] [1, 1024] [1, 1] : tensor<16x1024xf32> to tensor<1x1024xf32>
          %11 = arith.muli %arg1, %c-4 : index
[...]

PR #2663, opened 2026-02-18
Support dynamic and static rotation shifts (naively)

j2kun:
First step of #2662

Context: to support rolled linalg kernels (#2655), we will need to introduce loops that have rotation operations that depend on SSA values computed dynamically. Currently all lower-level rotation ops have static attributes for the offset.

This change is the first step toward supporting dynamic rotation indices in the lower level parts of the compiler. It makes all scheme-level and library-level rotation ops have optional `dynamic_shift` (Value) and `static_shift` (Attribute) operands, and modifies the rest of the codebase to support both at a very naive level.

In particular, dynamic shifts are only supported insofar as the defining op of the `dynamic_shift` operand is an `arith.constant`. This is to maintain compatibility as the larger feature is split across multiple PRs. It preserves the ability for the lower-levels of the pipeline to use the current analysis methods to find all the rotation keys that need to be generated by the client, since that does not yet support loops.

The next step will be to add an analysis that supports loops and computed expressions, and use that in passes like FastRotationPrecompute and ConfigureCryptoContext.

MOST of this PR is renaming operands (e.g., index -> static_shift) and updating tests to match.


tests/Examples/dacapo/mlp.dacapo.mlir
```diff
@@ -54,510 +54,510 @@
 module attributes {ckks.schemeParam = #ckks.scheme_param<logN = 17, Q = [1106058412451299513, 1056763241666817029, 957769724367225479, 919081519653443687, 1030837924888066153, 1084354410096143723, 1135846243351935917, 1087115004561311021, 997960547764032911, 892538949448853293, 1002528331340998513, 1100798419621231379, 981696679688787961, 1061922508412786269], P = [1152921504606846976], logDefaultScale = 60>, scheme.ckks} {
   func.func @_hecate_MLP(%ct: !ct_L13) -> !ct_L1 {
     %ct_0 = ckks.level_reduce %ct {levelToDrop = 10 : i64} : !ct_L13 -> !ct_L3
-    %ct_1 = ckks.rotate %ct_0 {offset = 0 : i32} : !ct_L3
+    %ct_1 = ckks.rotate %ct_0 {static_shift = 0 : i32} : !ct_L3
```
asraa, line 57: i can't file an issue in that repo, but i'll need to migrate the translator

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
```diff
@@ -828,8 +828,17 @@ LogicalResult OpenFhePkeEmitter::printOperation(RotOp op) {
...
   os << variableNames->getNameForValue(op.getCryptoContext()) << "->"
      << "EvalRotate" << "("
-     << variableNames->getNameForValue(op.getCiphertext()) << ", "
-     << op.getIndex().getValue() << ");\n";
+     << variableNames->getNameForValue(op.getCiphertext()) << ", ";
+
+  if (op.getStaticShiftAttr()) {
+    os << op.getStaticShift()->getValue();
+  } else if (op.getDynamicShift()) {
+    os << variableNames->getNameForValue(op.getDynamicShift());
```
asraa: this file also uses a function to get a string for a constant

  auto getConstantOrValue = [&](Value value) -> std::string {
    return getStringForConstant(value).value_or(
        variableNames->getNameForValue(value));
  };

lib/Dialect/Openfhe/IR/OpenfheOps.td
```diff
@@ -390,7 +390,8 @@ def RotOp : Openfhe_Op<"rot", [
   let arguments = (ins
     Openfhe_CryptoContext:$cryptoContext,
     Openfhe_Ciphertext:$ciphertext,
-    Builtin_IntegerAttr:$index
+    Optional<AnySignlessIntegerOrIndex>:$dynamic_shift,
```
asraa, line 393: for these and others, should we verify that it's one or the other?

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -108,7 +109,17 @@ struct ConfigureCryptoContext
     std::set<int64_t> distinctRotIndices;
     walkFuncAndCallees(op, [&](Operation* op) {
       if (auto rotOp = dyn_cast<openfhe::RotOp>(op)) {
-        distinctRotIndices.insert(rotOp.getIndex().getInt());
+        if (rotOp.getStaticShift().has_value()) {
+          distinctRotIndices.insert(rotOp.getStaticShift()->getInt());
+        } else if (rotOp.getDynamicShift()) {
+          // Try to extract constant from dynamic_shift SSA value
```
asraa, line 115: i *think* we need to fail if we have a dynamic rotation, since we won't be able to generate the key for it (unless openfhe ever ends up supporting that by having a complete set of rotation keys and doing the necessary rotation calculations on its own, or we do something like that in the IR itself)

def a follow up or a low priority

tests/Examples/dacapo/mlp.dacapo.mlir
j2kun: I can do that

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
j2kun: In the next PR, I intend to add a dataflow analysis that can determine this for the purpose of identifying the total set of rotations required.

lib/Target/OpenFhePke/OpenFhePkeEmitter.cpp
j2kun: Done

lib/Dialect/Openfhe/IR/OpenfheOps.td
j2kun: Done (added a verifier)

PR #2664, opened 2026-02-18
Improves the Form attribute syntax of Polynomial

crockeea:
Inspired by Jeremy's suggestions to #2647, I have updated the syntax of the Polynomial "form" attribute. In addition to just being much more pleasing to the eye and much simpler to use, this is a very useful change for the upcoming NTT insertion pass. This change really happened as a *result* of that work, but it got a bit out of hand and is quite large in itself, so I decided to split it off into its own PR.

The main issue is that this introduces EnumAttrs, which I had a hard time compiling. Ideally they would go into PolynomialAttributes, but I guess there's something different about the way TableGen works with Enums vs other attributes. If anyone can simplify the build logic, that'd be great!


lib/Dialect/Polynomial/IR/BUILD
```diff
@@ -72,6 +74,17 @@ add_heir_dialect_library(
     ],
 )
 
+# I couldn't figure out how to put this in the Attributes file where they belong
```
j2kun: I think this is actually the correct way to do it, though I believe some people alternatively declare the enum in C++, and make the parameter to the attr use the C++ class directly (naming it as a string, like `"bool"`), ensure the includes line up properly, and then maybe customize the parser to print/parse the enum name.

```suggestion
```

lib/Dialect/Polynomial/IR/PolynomialEnums.td
```diff
@@ -0,0 +1,19 @@
+#ifndef LIB_DIALECT_POLYNOMIAL_IR_POLYNOMIALENUMS_TD_
+#define LIB_DIALECT_POLYNOMIAL_IR_POLYNOMIALENUMS_TD_
+
+include "lib/Dialect/Polynomial/IR/PolynomialDialect.td"
+// include "mlir/IR/BuiltinAttributeInterfaces.td"
+// include "mlir/IR/OpBase.td"
+// include "mlir/IR/OpAsmInterface.td"
```
j2kun:
```suggestion
```

PR #2667, opened 2026-02-18
Add ILP bootstrap placement pass

edwjchen:
This PR is an initial attempt at implementing a naive ILP-based bootstrap placement pass for more efficient bootstrap and secret-mgmt operation placement. The pass does not rely on the current secret-mgmt-insert-pass, as the pass initially solves for the level of each op, thereby determining the bootstrap and other secret-mgmt operation placement. 

- New pass --ilp-bootstrap-placement with option bootstrap-waterline
- LIT tests: bootstrap_placement_comparison.mlir, ilp_relin_modreduce_placement.mlir.

At a high-level, the formulation is as follows:
- Variables
  - the input level of each op, the output level of each op, and whether a bootstrap is placed on an op. 
- Constraints 
  - Level bounds: 0 <= op.level <= bootstrapWaterline 
  - Operand matching: operand levels must be equal
  - Multiplication: assumes multiplications consume 1 level 
  - Non-mult: all other operations forward the operand level 
  - Bootstrap (big-M): if bootstrap, then the output level is the bootstrapWaterline, otherwise the input level is forwarded to the output level. 
- Objective: minimize number of bootstrap operations

Note: This is still a naive ILP implementation and there are concerns yet to be addressed. 
- This ILP formulation relies on the assumption that every multiplication consumes a level.  This can be relaxed with more fine-grain scale analysis, (e.g., that delays multiple multiplications before a rescale instruction). 
- The ILP formulation only considers the input and output level of each vertex. However, this misses potential level changes along edges (e.g., when two independent data paths converge at an op with a level mismatch). The assumption assumes a modReduce operation after every multiplication, and all operands levels are equal without additional modReduce operations. 

These limitations will be addressed in the future.

### DIFF
--- a/lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
+++ b/lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
@@ -1,0 +1,372 @@
+#include "lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.h"
+
+#include <algorithm>
+#include <cassert>
+#include <sstream>
+#include <string>
+#include <unordered_set>
+#include <utility>
+
+#include "absl/status/statusor.h"  // from @com_google_absl
+#include "absl/time/time.h"        // from @com_google_absl
+#include "lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h"
+#include "lib/Dialect/Secret/IR/SecretOps.h"
+#include "llvm/include/llvm/ADT/DenseMap.h"            // from @llvm-project
+#include "llvm/include/llvm/ADT/TypeSwitch.h"          // from @llvm-project
+#include "llvm/include/llvm/Support/Casting.h"         // from @llvm-project
+#include "llvm/include/llvm/Support/Debug.h"           // from @llvm-project
+#include "llvm/include/llvm/Support/raw_ostream.h"     // from @llvm-project
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinOps.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/Operation.h"            // from @llvm-project
+#include "mlir/include/mlir/IR/Value.h"                // from @llvm-project
+#include "mlir/include/mlir/IR/ValueRange.h"           // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"            // from @llvm-project
+#include "mlir/include/mlir/Support/LogicalResult.h"   // from @llvm-project
+#include "ortools/math_opt/cpp/math_opt.h"  // from @com_google_ortools
+
+// The level describes the remaining multiplicative depth of a ciphertext.
+// Bootstrap operations reset the level to a maximum value (or a target level
+// if specified), allowing further operations to be performed.
+//
+// This implementation tracks the level state through the computation
+// and uses ILP to determine optimal bootstrap placement to minimize costs
+// while ensuring level constraints are satisfied.
+//
+// ILP Formulation:
+// - Variables: level[value] for each SSA value, input_level[op] for each op,
+//   bootstrap[op] for each operation
+// - Constraints:
+//   * Level bounds: 0 <= level[value] <= bootstrapWaterline (levels
+//   0..bootstrapWaterline)
+//   * Operand matching: for each op, operand_i_level == input_level[op]
+//   * Multiplication: output_level = input_level - 1
+//   * Non-mult: output_level = input_level
+//   * Bootstrap (big-M): if bootstrap[op] = 1, level_after = bootstrapWaterline
+//                        else level_after = level_before
+// - Objective: minimize sum of bootstrap decisions
+
+namespace math_opt = ::operations_research::math_opt;
+
+#define DEBUG_TYPE "ilp-bootstrap-placement"
+
+namespace mlir {
+namespace heir {
+
+// Helper to get secret operands
+static void getSecretOperands(Operation* op, DataFlowSolver* solver,
+                              SmallVector<Value>& secretOperands) {
+  for (OpOperand& operand : op->getOpOperands()) {
+    if (isSecret(operand.get(), solver)) {
+      secretOperands.push_back(operand.get());
+    }
+  }
+}
+
+LogicalResult ILPBootstrapPlacementAnalysis::solve() {
+  math_opt::Model model("ILPBootstrapPlacementAnalysis");
+
+  // Get the secret.generic operation
+  auto genericOp = dyn_cast<secret::GenericOp>(opToRunOn);
+  if (!genericOp) {
+    return failure();
+  }
+
+  Block* body = genericOp.getBody();
+
+  int nextOpaqueId = 0;
+  llvm::DenseMap<Operation*, int> opaqueIds;
+  auto uniqueName = [&](Operation* op) {
+    std::string varName;
+    llvm::raw_string_ostream ss(varName);
+    ss << op->getName().getStringRef() << "_";
+    if (opaqueIds.count(op) == 0)
+      opaqueIds.insert(std::make_pair(op, nextOpaqueId++));
+    ss << opaqueIds.lookup(op);
+    return ss.str();
+  };
+
+  // Map operations to bootstrap decision variables
+  llvm::DenseMap<Operation*, math_opt::Variable> decisionVariables;
+  // Map SSA values to level variables (after bootstrap decision)
+  llvm::DenseMap<Value, math_opt::Variable> levelVars;
+  // Map SSA values to level variables before bootstrap
+  llvm::DenseMap<Value, math_opt::Variable> beforeBootstrapVars;
+  // Input level: level at which operands are consumed
+  llvm::DenseMap<Operation*, math_opt::Variable> inputLevelVars;
+
+  // Big-M constant for big-M method
+  // NOTE: This BIG_M does not account for "freshly encrypted" ciphertexts
+  // starting at a higher level than the bootstrap waterline. This should
+  // be addressed in future work.
+  const int BIG_M = bootstrapWaterline;
+
+  // Create variables for all SSA values in the body
+  // First, handle block arguments (inputs)
+  for (BlockArgument arg : body->getArguments()) {
+    if (!isSecret(arg, solver)) continue;
+
+    std::stringstream ss;
+    ss << "levelArg" << arg.getArgNumber();
+    auto levelVar =
+        model.AddContinuousVariable(0, bootstrapWaterline, ss.str());
+    levelVars.insert(std::make_pair(arg, levelVar));
+    // Inputs start at maximum level (bootstrapWaterline)
+    model.AddLinearConstraint(
+        levelVar == bootstrapWaterline,
+        "initLevelArg" + std::to_string(arg.getArgNumber()));
+  }
+
+  // Walk operations in the body
+  for (Operation& op : body->getOperations()) {
+    // Check if this operation produces secret results
+    if (llvm::none_of(op.getResults(), [&](OpResult result) {
+          return isSecret(result, solver);
+        })) {
+      continue;
+    }
+    if (isa<secret::YieldOp>(op)) continue;
+
+    std::string opName = uniqueName(&op);
+
+    // Create bootstrap decision variable for this operation
+    auto bootstrapVar = model.AddBinaryVariable("bootstrap" + opName);
+    decisionVariables.insert(std::make_pair(&op, bootstrapVar));
+
+    // Create input level variable: level at which operands are consumed
+    auto inputLevelVar = model.AddContinuousVariable(0, bootstrapWaterline,
+                                                     "inputLevel" + opName);
+    inputLevelVars.insert(std::make_pair(&op, inputLevelVar));
+
+    // Create level variables for results
+    for (OpResult result : op.getResults()) {
+      if (!isSecret(result, solver)) continue;
+
+      std::stringstream ss;
+      ss << "level" << opName << result.getResultNumber();
+      auto levelVar =
+          model.AddContinuousVariable(0, bootstrapWaterline, ss.str());
+      levelVars.insert(std::make_pair(result, levelVar));
+
+      std::stringstream ss2;
+      ss2 << "levelBefore" << opName << result.getResultNumber();
+      auto beforeVar =
+          model.AddContinuousVariable(0, bootstrapWaterline, ss2.str());
+      beforeBootstrapVars.insert(std::make_pair(result, beforeVar));
+    }
+  }
+
+  // Add constraints for operations
+  for (auto& [op, _] : opaqueIds) {
+    std::string opName = uniqueName(op);
+
+    // Get secret operands
+    SmallVector<Value> secretOperands;
+    getSecretOperands(op, solver, secretOperands);
+
+    // Operand matching: all operands must be at the same level when consumed.
+    // inputLevel = level at which operands are consumed.
+    auto inputLevelVar = inputLevelVars.at(op);
+    for (size_t i = 0; i < secretOperands.size(); ++i) {
+      Value operand = secretOperands[i];
+      if (!levelVars.contains(operand)) continue;
+
+      std::stringstream ss;
+      ss << "operandMatch" << opName << "Op" << i;
+      model.AddLinearConstraint(levelVars.at(operand) == inputLevelVar,
+                                ss.str());
+    }
+
+    // Output level: for multiplication, output = input - 1; else output = input
+    if (isa<arith::MulFOp>(op) || isa<arith::MulIOp>(op)) {
+      for (OpResult result : op->getResults()) {
+        if (!isSecret(result, solver)) continue;
+        if (!beforeBootstrapVars.contains(result)) continue;
+
+        auto resultBeforeVar = beforeBootstrapVars.at(result);
+        // outputLevel = inputLevel - 1
+        std::stringstream ss;
+        ss << "mulOutput" << opName << result.getResultNumber();
+        model.AddLinearConstraint(resultBeforeVar == inputLevelVar - 1,
+                                  ss.str());
+        std::stringstream ssMin;
+        ssMin << "mulLevelMin" << opName << result.getResultNumber();
+        model.AddLinearConstraint(resultBeforeVar >= 0, ssMin.str());
+      }
+    } else {
+      for (OpResult result : op->getResults()) {
+        if (!isSecret(result, solver)) continue;
+        if (!beforeBootstrapVars.contains(result)) continue;
+
+        auto resultBeforeVar = beforeBootstrapVars.at(result);
+        // outputLevel = inputLevel (level flows through unchanged)
+        std::stringstream ss;
+        ss << "flowOutput" << opName << result.getResultNumber();
+        model.AddLinearConstraint(resultBeforeVar == inputLevelVar, ss.str());
+      }
+    }
+
+    // Add bootstrap constraints using big-M method
+    for (OpResult result : op->getResults()) {
+      if (!isSecret(result, solver)) continue;
+      if (!levelVars.contains(result) || !beforeBootstrapVars.contains(result))
+        continue;
+
+      auto resultLevelVar = levelVars.at(result);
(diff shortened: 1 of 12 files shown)


lib/Pipelines/BooleanPipelineRegistration.cpp
j2kun, line 1: nit: your changes don't impact any pipelines at all, so best to revert this file change.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.td
```diff
@@ -0,0 +1,35 @@
...
+
+include "mlir/Pass/PassBase.td"
+
+def ILPBootstrapPlacement : Pass <"ilp-bootstrap-placement"> {
+    let summary = "Optimize placement of bootstrap ops using ILP";
+    let description = [{
+        This pass uses an integer linear program to determine the optimal level
+        of each term in the MLIR, and thus the placement of bootstrap and
+        modreduce operations. It does not have any dependencies on the existing
+        secret-insert-mgmt-ckks pass.
```
j2kun:
```suggestion
        This pass uses an integer linear program to determine the optimal level
        of each term in the MLIR, and thus the placement of bootstrap and
        modreduce operations.
```

nit: phrases like "the existing" get out of date quickly.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.td
```diff
@@ -0,0 +1,35 @@
...
+
+        The pass runs on raw IR (secret.generic with arith ops etc.). It
```
j2kun:
```suggestion
        The pass runs on [ciphertext-semantic](https://heir.dev/docs/design/layout/#data-semantic-and-ciphertext-semantic-tensors)
        IR (secret.generic with arith ops operating on pre-packed tensors). It
```

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.td
```diff
@@ -0,0 +1,35 @@
...
+    }];
+
+    let dependentDialects = ["mlir::heir::mgmt::MgmtDialect"];
+
+  let options = [
+    Option<"bootstrapWaterline",
+           "bootstrap-waterline",
+           "int",
+           /*default=*/"3",
+           "Bootstrap waterline (max level). Levels are 0..bootstrap-waterline (inclusive); inputs start at bootstrap-waterline.">,
```
j2kun, line 34: IIUC a freshly encrypted ciphertext is usually encrypted at a higher level than the bootstrap waterline. Is that something to incorporate later? Generally we're assuming there is a "main entry function" whose secret inputs are freshly encrypted, though we don't formalize that well...

lib/Transforms/SecretInsertMgmt/Passes.td
j2kun, line 1: nit: No need to modify this file.

tests/Transforms/ilp_bootstrap_placement/ilp_bootstrap_placement_level3.mlir
```diff
@@ -0,0 +1,53 @@
...
+
+// Test ILP bootstrap placement with bootstrap-waterline=3 on the same computation.
+// ILP places 2 bootstraps (optimal for this DAG with levels 0..3):
+//   - after L3 (input3*input4), and after L4 (l1*l2).
+//
+// Computation: L1..L10 as in ilp_bootstrap_placement.mlir.
+
+// CHECK: func.func @bootstrap_placement_test
+// CHECK: secret.generic
+// First bootstrap: after L3 (arith.mulf %input3, %input4)
```
j2kun: The optimal solution may be unique for this test, but what you're really testing here is that the ILP produces exactly two bootstraps. This is more robustly asserted by using `// CHECK-COUNT-2: mgmt.bootstrap` followed by `// CHECK-NOT: mgmt.bootstrap` instead of relying on a particular order of operations. In theory, if the solution is not unique, then some superficial change to the ILP could cause this test to fail by placing the same two bootstraps in slightly different locations in the IR, even though that would still be correct.
Similarly, in the greedy test above you could use `// CHECK-COUNT-3: mgmt.bootstrap` to assert at least three bootstraps.
Moreover, you can combine these two tests into one lit file by having two RUN lines and using `FileCheck --check-prefix=UNIQUE_NAME`

For example

```
// RUN: heir-opt --ilp-bootstrap-placement=bootstrap-waterline=3 %s | FileCheck %s --check-prefix=CHECK-ILP
// RUN: heir-opt --secret-insert-mgmt-ckks=bootstrap-waterline=3 %s | FileCheck %s --check-preifx=CHECK-GREEDY

// CHECK-ILP-COUNT-2: mgmt.bootstrap
// CHECK-ILP-NOT: mgmt.bootstrap

// CHECK-GREEDY-COUNT-3: mgmt.bootstrap
```

This way makes it immediately clear that you're doing a direct comparison of the ILP method to the greedy method.
That said, it is probably also valuable to have at least one test that asserts correct insertion of relin/modreduce after muls, and doing what I said would remove the ability to do that in the same test. However, (1) keeping those two tests separate is better since they're testing different functional behaviors and (2) you're reusing existing tested patterns so really this would just ensure that those patterns are accidentally deleted from the pass.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+struct ILPBootstrapPlacement
+    : impl::ILPBootstrapPlacementBase<ILPBootstrapPlacement> {
+  using ILPBootstrapPlacementBase::ILPBootstrapPlacementBase;
+
+  void processSecretGenericOp(secret::GenericOp genericOp,
+                              DataFlowSolver* solver) {
+    // Remove all bootstrap ops. This makes the IR invalid, because the level
+    // states are incorrect. However, the correctness of the ILP ensures the
+    // level states are made correct at the end.
```
j2kun: This is technically not true, because the level is not part of IR validation at this stage of the pipeline. I think the corresponding comment in https://github.com/google/heir/blob/3ed0da33f81984b32a32f8490e2de1f07ed14c03/lib/Transforms/OptimizeRelinearization/OptimizeRelinearization.cpp#L37 is also out of date because the key basis operand was removed from that op at some point.

```suggestion
```

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    ILPBootstrapPlacementAnalysis analysis(genericOp, solver,
+                                           bootstrapWaterline);
+    if (failed(analysis.solve())) {
+      genericOp->emitError(
+          "Failed to solve the bootstrap placement optimization problem");
+      return signalPassFailure();
+    }
+
+    // Copy bootstrap decisions before any IR changes that might invalidate
```
j2kun: I suspect the logic in this section can be made quite a bit simpler than it's done here. My first instinct would be to organize it like this:

1. Have the ILP return a set of `Value`s, each of which the ILP decided needs a bootstrap after.
2. Looping over these `Value`s, insert a bootstrap and replace all uses (as done below)
3. Use a similar helper to relinearize below for modreduce as well, outside of this function. (insertModReduceBeforeOrAfterMult should work just as well)

You seem to have hit an obstacle with the analysis, but I'm not sure I understand the problem. I don't think inserting a modreduce op can "reallocate" the block, and the analysis should not be invalidated because you insert new ops, so long as the old Values are unchanged (the DataFlowSolver is ultimately a mapping from Value to Lattice states and is unchanged by IR modifications). What would not be accessible are Lattices of newly inserted Values, but if you need to you can make Lattice states for new Values manually by using `GetOrCreateState` and then mutating the returned LatticeState's internals in-place.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  void runOnOperation() override {
+    Operation* module = getOperation();
+
+    DataFlowSolver solver;
+    dataflow::loadBaselineAnalyses(solver);
+    solver.load<SecretnessAnalysis>();
+
+    if (failed(solver.initializeAndRun(getOperation()))) {
```
j2kun, line 85: If you are having troubles with the dataflow solver becoming invalid, moving the solver instantiation and initializeAndRun to occur just before you query the analysis may help. It doesn't need to live here, and can instead live inside `processSecretGenericOp` and you can feel free to re-run it to refresh it, so long as you don't re-run it in a loop.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+  math_opt::Model model("ILPBootstrapPlacementAnalysis");
+
+  // Get the secret.generic operation
+  auto genericOp = dyn_cast<secret::GenericOp>(opToRunOn);
+  if (!genericOp) {
+    return failure();
+  }
+
+  Block* body = genericOp.getBody();
+  if (!body) {
```
j2kun: This is unreachable (the input op would not verify). Can safely remove.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+  const int BIG_M = bootstrapWaterline * 100;
+
+  // Create variables for all SSA values in the body
+  // First, handle block arguments (inputs)
+  for (BlockArgument arg : body->getArguments()) {
+    if (!isSecret(arg, solver)) continue;
+
+    std::stringstream ss;
+    ss << "levelArg" << arg.getArgNumber();
+    auto levelVar = model.AddIntegerVariable(0, bootstrapWaterline, ss.str());
```
j2kun: I believe these variables do not need to be integer. They will be forced to have integer values in any feasible solution by the rest of the model. Making them continuous will give a performance boost.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+    for (OpResult result : op.getResults()) {
+      if (isSecret(result, solver)) {
+        hasSecretResults = true;
+        break;
+      }
+    }
+
+    if (!hasSecretResults) {
+      continue;
+    }
```
```suggestion
    if (llvm::none_of(op.getResults(), 
         [&](OpResult result) { return isSecret(result, solver); })) {
      continue;
    }
```

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+        levelVar == bootstrapWaterline,
+        "initLevelArg" + std::to_string(arg.getArgNumber()));
+  }
+
+  // Walk operations in the body
+  for (Operation& op : body->getOperations()) {
+    if (isa<secret::YieldOp>(op)) {
+      // Handle yield: the result level should match the yielded value
+      continue;
+    }
```
```suggestion
```

Subsumed by the check below for "has no secret results"

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    // Create bootstrap decision variable for this operation
+    auto bootstrapVar = model.AddBinaryVariable("bootstrap" + opName);
+    decisionVariables.insert(std::make_pair(&op, bootstrapVar));
+
+    // Track only "core" ops (exclude mgmt ops that may have been inserted by
+    // a prior pass) so index matches the pass when it skips those.
+    if (!isa<mgmt::ModReduceOp>(op) && !isa<mgmt::RelinearizeOp>(op) &&
+        !isa<mgmt::BootstrapOp>(op)) {
```
j2kun: bootstrap ops were deleted, remove? Also, why not remove the other two ops as well in pre-processing? They are inserted at the end of the pass, and the hope is for this ILP to eventually decide those as well. May as well add that logic now.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+      orderedOps.push_back(&op);
+      orderedDecisionVars.push_back(bootstrapVar);
+    }
+
+    // Create input level variable: level at which operands are consumed
+    auto inputLevelVar =
+        model.AddIntegerVariable(0, bootstrapWaterline, "inputLevel" + opName);
```
j2kun: Ditto as above: none of these variables need to be integer.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+      std::stringstream ss;
+      ss << "operandMatch" << opName << "Op" << i;
+      model.AddLinearConstraint(levelVars.at(operand) == inputLevelVar,
+                                ss.str());
+    }
+
+    // Output level: for multiplication, output = input - 1; else output = input
+    if (isa<arith::MulFOp>(op) || isa<arith::MulIOp>(op)) {
+      for (OpResult result : op.getResults()) {
```
j2kun: MulF and MulI always have one op result, so you can just use `op->getResults()[0]` or similar. With the neighboring comment about iterating over known-good operations, you can also remove checks for `isSecret(result, solver)` and `beforeBootstrapVars.contains(result)`

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+      std::stringstream ss2;
+      ss2 << "levelBefore" << opName << result.getResultNumber();
+      auto beforeVar =
+          model.AddIntegerVariable(0, bootstrapWaterline, ss2.str());
+      beforeBootstrapVars.insert(std::make_pair(result, beforeVar));
+    }
+  }
+
+  // Add constraints for operations
+  for (Operation& op : body->getOperations()) {
```
j2kun: Why not loop just over, say, the operations in `opaqueIds`? That would allow you to avoid re-filtering on things like "is it a yield" or "does it have secret operands/results"

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+  llvm::SmallVector<math_opt::Variable, 32> orderedDecisionVars;
+  // Map SSA values to level variables (after bootstrap decision)
+  llvm::DenseMap<Value, math_opt::Variable> levelVars;
+  // Map SSA values to level variables before bootstrap
+  llvm::DenseMap<Value, math_opt::Variable> beforeBootstrapVars;
+  // Input level: level at which operands are consumed
+  llvm::DenseMap<Operation*, math_opt::Variable> inputLevelVars;
+
+  // Big-M constant for big-M method
+  const int BIG_M = bootstrapWaterline * 100;
```
j2kun: If you have a bigger input example to test on, could you try tightening BIG_M to be equal to bootstrapWaterline? Based on the formulation here, that is the minimal value of BIG_M that still makes the model valid. My reasoning:

```
// levelAfter <= bootstrapWaterline + bootstrapWaterline * (1 - bootstrap)
// levelAfter >= bootstrapWaterline - bootstrapWaterline * (1 - bootstrap)
// levelAfter <= levelBefore + bootstrapWaterline * bootstrap
// levelAfter >= levelBefore - bootstrapWaterline * bootstrap
```

If bootstrap = 0, the first two constraints are `0 <= levelAfter <= 2 * bootstrapWaterline` which is still trivially satisfied. The second two constraints are an equality.

If bootstrap = 1, the first two make an equality and the last two become `before - bootstrapWaterline <= 0 <= levelAfter <= bootstrapWaterline <= before + bootstrapWaterline` is trivially satisfied.

This may become slightly incorrect if we allow inputs to the IR to have a "fresh encryption" level that exceeds the bootstrap waterline, but in that case big-M can be the total number of levels.

In any case, making big-M tight should always improve solver performance by making the LP relaxation tighter.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+    }
+  }
+
+  // Objective: minimize number of bootstraps
+  math_opt::LinearExpression obj;
+  for (auto& [op, decisionVar] : decisionVariables) {
+    obj += decisionVar;
+  }
+  model.Minimize(obj);
+
```
j2kun, line 256: Could you add a pass option that dumps the built model and solution to a file for debugging purposes?

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -0,0 +1,344 @@
...
+  // Extract solution
+  auto varMap = result.variable_values();
+  for (auto& [op, decisionVar] : decisionVariables) {
+    solution.insert(std::make_pair(op, varMap[decisionVar] > 0.5));
+  }
+
+  // Build bootstrapByOpIndex from ordered decision variables (index-based),
+  // not from solution.lookup(orderedOps[i]), so we don't rely on Operation*
+  // pointer identity across environments (e.g. sandbox vs local).
+  bootstrapByOpIndex.clear();
```
j2kun: I suspect with the suggested change to the pass organization mentioned earlier, this extra work is unnecessary.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.td
edwjchen: You're right in that a freshly encrypted ciphertext is usually higher than the bootstrap waterline (e.g., in DaCapo, the initial level of a ciphertext is 29 where was the bootstrap waterline is 16). This is something that we should incorporate later. 

I'll make a note of this in the description!

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
edwjchen: Added a printSolution helper function that is called when the debug flag is set. 

for example: 
```bazel run //tools:heir-opt -- -debug-only=ilp-bootstrap-placement --ilp-bootstrap-placement=bootstrap-waterline=3 "$(pwd)/tests/Transforms/ilp_bootstrap_placement/ilp_relin_modreduce_placement.mlir"```

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
edwjchen: This check was used in the OptimizeRelinearization pass. I personally haven't had any troubles initializing the solver. If it's unnecessary, we can remove it!

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
edwjchen: Tightening the big-M here works for the test cases! I'll leave a note when we want to support freshly encrypted ciphertexts.

j2kun:
Please run `pre-commit run --all-files` to fix the linter errors. You can set it up to run automatically as a git hook via https://heir.dev/docs/development/#pre-commit

edwjchen:
> Please run `pre-commit run --all-files` to fix the linter errors. You can set it up to run automatically as a git hook via https://heir.dev/docs/development/#pre-commit

Updated!

PR #2671, opened 2026-02-19
Add a pass to insert NTT/INTT into Polynomial

### LINKED ISSUE #1095: convert-polynomial-mul-to-ntt: migrate to new mod arith type/poly type
mod arith dialect chooses to use a mod arith type to carry out all operations, see #1084. As the `convert-polynomial-mul-to-ntt` pass needs mod arith, it needs to migrate to new type.

Also, polynomial type is also undergoing migration (#1085). So it might be migrated after both type being migrated.

crockeea:
This PR introduces a pass to (optimally) insert NTTs and INTTs into a Polynomial IR.

In short, the pass creates a [CP-SAT instance](https://developers.google.com/optimization/cp/cp_solver) based on the input AST, solves the instance, and uses the solution to insert NTTs/INTTs.

Along the way:
- Several changes in PolynomialOps to support tensor-of-polynomial inputs
- Changes to correctly get not-yet-parsed attributes during type inference
- The pass uses the CP-SAT problem. MLIR doesn't appear to have any advanced solvers for problems like max-flow/min-cut or ILP, so I introduced a dependency on Google's ORTools.
- NTTSolver and CPSATSolution are translate between CP-SAT concepts and NTT-insertion/MLIR IR concepts.

Open items:
- ~~Unclear what the ntt_rewrites.mlir test does now; it passes (I think), but the code for the corresponding pass is now gone. Related to #1095~~
- ~~I suspect this PR may resolve #1095, but I haven't had time to investigate.~~
- The pass is missing the insertion of real roots of unity; I think the work to generate those roots of unity has been done elsewhere (though it may need to be updated to support RNS RUs), but I didn't have time to link them together.

Unfortunately, I don't know that I'm going to have any more time to resolve the remaining issues in the immediate future. I encourage someone else to get this over the finish line; feel free to contact me if you have questions. If I do find time to work on this, I will leave comments.


crockeea:
Test failures resolved.

crockeea:
Addressing the open items: I think #1095  can be closed, and ntt_rewrites.mlir can be deleted.

The one piece it tests that is not in the tests introduced in this PR is a “bad ring” test, and that’s because I’m not inserting real roots of unity (which is where a bad ring world be detected).

One path to getting this PR merged is to punt on adding real roots of unity (all existing tests pass without them), and do that when we fix up the lowering from polynomial to modarith.

It all of this sounds good, I will do the cleanup of #1095, and then we can get this under review.

tools/BUILD
```diff
@@ -60,6 +60,7 @@ cc_binary(
         "@heir//lib/Dialect/LWE/Conversions/LWEToOpenfhe",
         "@heir//lib/Dialect/LWE/Conversions/LWEToPolynomial",
         "@heir//lib/Dialect/LWE/IR:Dialect",
+        # Why do we sometimes use `Transforms` alone (like LWE), and other times use specific transforms (like Lattigo)?
```
j2kun:
```suggestion
```

`@heir//lib/Dialect/LWE/Transforms` is shorthand for `@heir//lib/Dialect/LWE/Transforms:Transforms`. Generally speaking, internal google tooling requires that any symbol used has a corresponding `#include` that directly declares that symbol, and any `#include` has a corresponding build dependency.  So there may be "duplicates" in this list to satisfy that requirement when the `:Transforms` target provides a registration function for all defined passes while the more specific libraries provide individual pass registration hooks.

j2kun:
> Addressing the open items: I think #1095 can be closed, and ntt_rewrites.mlir can be deleted.
> 
> The one piece it tests that is not in the tests introduced in this PR is a “bad ring” test, and that’s because I’m not inserting real roots of unity (which is where a bad ring world be detected).
> 
> One path to getting this PR merged is to punt on adding real roots of unity (all existing tests pass without them), and do that when we fix up the lowering from polynomial to modarith.
> 
> It all of this sounds good, I will do the cleanup of #1095, and then we can get this under review.

This sounds good to me. A quick scan of that issue and this PR makes me agree. I can put the roots of unity together in a followup PR.

lib/Dialect/Polynomial/Transforms/Passes.td
```diff
@@ -3,18 +3,21 @@
 
 include "mlir/Pass/PassBase.td"
 
-def PolyMulToNTT : Pass<"convert-polynomial-mul-to-ntt"> {
-  let summary = "Rewrites polynomial operations to their NTT equivalents";
+// TODO(#1095): migrate to mod arith type
```
j2kun:
```suggestion
```

tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt_optimal_placement.mlir
```diff
@@ -0,0 +1,38 @@
...
+!poly_ty_2 = !polynomial.polynomial<ring=#ring_2, form=coeff>
+
+module {
+  // Covers: a shared-dataflow circuit where optimal placement must avoid
+  // redundant conversions on mul chains while still satisfying coeff-only
+  // consumers. The optimal weighted conversion count here is 3:
+  // 1 NTT on %x, and 2 INTTs (on %diff and %prod).
+  // CHECK: func.func @optimal_ntt_placement(
+  // CHECK-COUNT-1: polynomial.ntt
+  // CHECK-COUNT-2: polynomial.intt
```
j2kun:
```suggestion
  // CHECK-COUNT-1: polynomial.ntt
  // CHECK-NOT: polynomial.ntt
  // CHECK-COUNT-2: polynomial.intt
  // CHECK-NOT: polynomial.intt
```

This will ensure the count is exactly 1 (resp, 2) and not at least 1 (resp, 2).

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+  // Consistency check on return values
+  func.walk([&](func::ReturnOp ret) {
+    for (auto [i, output] : llvm::enumerate(ret->getOperands())) {
+      if (output.getType() != newResultTypes[i]) {
+        ret->emitOpError("Function return types are inconsistent");
+        signalPassFailure();
+        return;
+      }
+    }
+  });
```
j2kun, line 613: nit: no need to walk since there's always just one Return op

```suggestion
  Operation *returnOp = func.front().getTerminator();
  for (auto [i, output] : llvm::enumerate(returnOp->getOperands())) {
    if (output.getType() != newResultTypes[i]) {
      returnOp->emitOpError("Function return types are inconsistent");
      signalPassFailure();
      return;
    }
  }
```

lib/Dialect/Polynomial/Transforms/NTTSolver.cpp
```diff
@@ -0,0 +1,171 @@
...
+  return 0;
+}
+
+const NTTSolver::RepVars& NTTSolver::getVars(Value& v) const {
+  auto it = vars.find(v);
+  if (it != vars.end()) {
+    return it->second;
+  } else {
+    llvm_unreachable("Var does not exist");
+  }
```
j2kun: I think it would suffice to `return vars.at(v);` which will raise an exception if the key is not present.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+  // op constraints. We optimize at the function level, which means there should
+  // be no NTTs/INTTs on inputs (unless both forms are needed) and no NTTs/INTTs
+  // on outputs. Instead, we allow inputs and outputs to be in either form, and
+  // choose whatever is "naturally" falls out.
+  // This is optimal because the function we are optimizing does no
+  // unnecessary transformations. Because of this constraint in particular,
+  // max-flow/min-cut is not an option: it cannot express, e.g., "accept this
+  // input in whatever form is best". Instead, we set up a (binary) constraint
+  // satisfaction problem (essentially a subset of a generic integer linear
+  // program) and find an optimal solution that miniimzes the number of
```
j2kun:
```suggestion
  // program) and find an optimal solution that minimizes the number of
```

lib/Dialect/Polynomial/Transforms/NTTSolver.h
j2kun, line 1: One thing you can do here, if you so chose, is write plain-old unit tests of the core solver logic without involving any MLIR at all. It may be worth coming back to this if we run into any nontrivial bugs while integrating this pass.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+    }
+    // These ops have coeff-form inputs and no (poly) outputs
+    // We therefore require their operands to be available in coeff form.
+    else if (isa<ToTensorOp, LeadingTermOp, EvalOp>(op)) {
+      for (Value v : polyOperands) {
+        solver.fixForm(v, Form::COEFF);
+      }
+    }
+    // These ops have coeff-form outputs (and possibly coeff-form inputs)
+    else if (isa<ConvertBasisOp, MonicMonomialMulOp, FromTensorOp>(op)) {
```
j2kun: Not necessary for this PR, but this definitely seems like the kind of situation where we will want this mapping of ops to "form cases" in another setting, and that is the time to use an OpInterface. That op interface can also provide methods to return the particular operands and results that are polynomials, which would avoid the filtering above.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+    }
+    func.emitError()
+        << "polyMulToNTT:typeToForm expected polynomial-like type, got " << ty;
+    return Type();
+  };
+
+  // Convert the value v to the given form. This adds an op to the AST.
+  // Do NOT convert if the input is a tensor.
+  auto addConversion = [&](Value& v, Form outputForm) -> Value {
+    // TODO need to insert PrimitiveRootAttrs here
```
j2kun: File an issue and link it here

```
// TODO(#123): insert PrimitiveRootAttrs
```

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+          PolynomialType::get(elem.getContext(), elem.getRing(), form);
+      return RankedTensorType::get(rt.getShape(), newElem, rt.getEncoding());
+    }
+    func.emitError()
+        << "polyMulToNTT:typeToForm expected polynomial-like type, got " << ty;
+    return Type();
+  };
+
+  // Convert the value v to the given form. This adds an op to the AST.
+  // Do NOT convert if the input is a tensor.
```
j2kun: This comment seems ignored or stale.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+      return NTTOp::create(b, v, PrimitiveRootAttr()).getOutput();
+    } else {
+      return INTTOp::create(b, v, PrimitiveRootAttr()).getOutput();
+    }
+  };
+
+  // First, deal with function arguments. We save the argument types
+  // for use in step 5.
+  SmallVector<Type> newInputTypes = llvm::to_vector(func.getArgumentTypes());
+  if (!func.isDeclaration()) {
```
j2kun: I would quit at the very beginning of `runOnOperation` if the func is a declaration. Nothing you can do in that case, unless you want to handle `func.call` in this pass (and in that case you'd need to insert extra NTTs for whatever form is required by the declaration's argument types).

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+        }
+        return WalkResult::advance();
+      });
+  if (wr.wasInterrupted()) return;
+
+  /***************************************************
+   ********** Step 1: Build CP-SAT instance **********
+   **************************************************/
+
+  for (Operation* op : rewriteOrder) {
```
j2kun, line 174: Might be worth noting somewhere (maybe in pass description) that this does not support ops with regions (e.g., a loop), and file an issue to add loop support.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+      // This includes MonicMonomialMulOp, FromTensorOp, MonomialOp, ConstantOp
+      continue;
+    }
+
+    b.setInsertionPoint(op);
+
+    if (isa<func::ReturnOp>(op)) {
+      // We have to iterate over *all* operands here, because some might not be
+      // polynomials. If you only iterate over polyResults, the indexing will
+      // be wrong.
```
j2kun: nit: if you wanted to avoid this, you can filter for polynomials on `op->getOpOperands()` which produces an `OpOperand` that has a `getOperandNumber()` as well as the underlying `Value` (instead of filtering on `op->getOperands()`.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -0,0 +1,583 @@
...
+  // have been created and are in the AST. In this step, we point op
+  // inputs to the correct value/form.
+
+  // Given an input-AST value and a target form, output the corresponding
+  // AST value with that form
+  auto formToValue = [&](Value& v, Form form) -> Value {
+    if (form == Form::COEFF) {
+      return coeffFormCache[v];
+    } else {
+      return evalFormCache[v];
```
j2kun:
```suggestion
      return coeffFormCache.at(v);
    } else {
      return evalFormCache.at(v);
```

nit: err on missing keys

tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt.mlir
```diff
@@ -0,0 +1,246 @@
...
+!poly_ty_4 = !polynomial.polynomial<ring=#ring_4, form=coeff>
+!ntt_poly_ty_4 = !polynomial.polynomial<ring=#ring_4, form=eval>
+
+module {
+  // Covers: baseline MulOp rewrite from coeff-form input to eval-form signature/result.
+  // CHECK: func.func @test_ntt_insertion1([[x:%.+]]: [[T1:![^ ]+]]) -> [[T1]] {
+  func.func @test_ntt_insertion1(%x: !poly_ty_1) -> !poly_ty_1 {
+    // CHECK-DAG: [[xsq:%.+]] = polynomial.mul [[x]], [[x]] : [[T1]]
+    %xsq = polynomial.mul %x, %x : !poly_ty_1
+    // CHECK-DAG: return [[xsq]] : [[T1]]
```
j2kun:
```suggestion
    // CHECK: [[xsq:%.+]] = polynomial.mul [[x]], [[x]] : [[T1]]
    %xsq = polynomial.mul %x, %x : !poly_ty_1
    // CHECK: return [[xsq]] : [[T1]]
```

nit: CHECK-DAG allows the statements to occur in any order.

tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt.mlir
```diff
@@ -0,0 +1,246 @@
...
+!poly_ty_3 = !polynomial.polynomial<ring=#ring_3, form=coeff>
+!ntt_poly_ty_3 = !polynomial.polynomial<ring=#ring_3, form=eval>
+
+#ring_4 = #polynomial.ring<coefficientType = !Zq1, polynomialModulus = <1 + x**8>>
+!poly_ty_4 = !polynomial.polynomial<ring=#ring_4, form=coeff>
+!ntt_poly_ty_4 = !polynomial.polynomial<ring=#ring_4, form=eval>
+
+module {
+  // Covers: baseline MulOp rewrite from coeff-form input to eval-form signature/result.
+  // CHECK: func.func @test_ntt_insertion1([[x:%.+]]: [[T1:![^ ]+]]) -> [[T1]] {
```
j2kun: This test does not (as the comment on line 22 states) assert any particular form is chosen. It just asserts that the same type is used in the input/output, and for the op. Perhaps instead you want to add something like `form=eval` to the regex capturing `T1`? You may need to add `--mlir-print-local-scope` to the `RUN` line for this to inline the type aliases. Then you can simplify the assertion to something like

```
// CHECK: @test_ntt_insertion1
// CHECK-SAME: form=eval
// CHECK-NOT: polynomial.ntt
// CHECK: return
// CHECK-SAME: form=eval
```

Or else, if you keep the types aliased, you can add `CHECK` statements at the top of the file that capture the name of the alias on the line that has the expected type declaration. E.g.,

```
// CHECK: [[ring_3:#[^ ]*]] = #polynomial.ring<coefficientType = {{.*}}, polynomialModulus = <1 + x**8>>
// CHECK: [[poly_ty3:![^ ]*]] = !polynomial.polynomial<ring=[[ring_3]], form=coeff>
```

This may change how you approach the other tests in this file as well, though most examples seem to have explicit ntt/intt that implicitly assert the same thing.

crockeea:
> I can see room for improvement on using OpInterfaces to reduce some of the duplicate logic

Can you tell me what you had in mind? I'm very interested in making this HEIR as a whole as clean as possible. 

I certainly noticed some crufty code in PolynomialOps.cpp, particularly in the type inference functions.
- I was getting crashes due to attributes not being fully parsed when the type-inference func is called. This seems like a weird MLIR design choice, but not much I can do about it right now. AI proposed a very complex work-around, and I wonder if there's a simpler solution.
- Similarly, hoisting type inference to handle tensors involved a lot of boilerplate. Any way to simplify that?

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
crockeea: Why is there always just one return op? AI gave me the impression that you could (e.g.) return in an if/else block, meaning there could be multiple ReturnOps.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
crockeea: Sounds like this could be very beneficial to readability!

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
j2kun: I did some experiments and stand corrected. In particular, `func.func` can have multiple `return` ops, as in this example

```mlir
module {
  func.func @select(%arg0: i32, %arg1: i32, %arg2: i1) -> i32 {
    cf.cond_br %arg2, ^bb1(%arg0 : i32), ^bb2(%arg1 : i32)
  ^bb1(%0: i32):  // pred: ^bb0
    return %0 : i32
  ^bb2(%1: i32):  // pred: ^bb0
    return %1 : i32
  }
}
```

However, this is not possible with `scf.if` because `scf.if` and similar ops in the `scf` dialect restrict their regions to have `scf.yield` as the (sole) terminator. I thought this was also true of `func.func` but it is not true.

j2kun:
> I certainly noticed some crufty code in PolynomialOps.cpp, particularly in the type inference functions.

I noticed indeed that `getInherentAttrFromAttrsOrProperties` seems complicated, but I'm not sure what the right way to do that is.

One possible way to clean things up is to use the overloaded pattern, which lets you type-match on multiple values at the same time

```cpp
// Helper for the "overloaded" pattern
template <class... Ts>
struct Overloaded : Ts... {
  using Ts::operator()...;
};
template <class... Ts>
Overloaded(Ts...) -> Overloaded<Ts...>;

...

return std::visit(
    Overloaded{
        [&](PolynomialType polyTy1, PolynomialType polyTy2) { ... },
        [&](ShapedType shapedTy1, ShapedType ShapedTy2) { ... },
        [&](auto, auto) { /*default case*/ },
    },
    ty1, ty2);
```

> using OpInterfaces

My idea is, in PolyMulToNTT.cpp, instead of checking on specific op names for each step before/after the solver construction (e.g., below)

```cpp
    else if (isa<ToTensorOp, LeadingTermOp, EvalOp>(op)) {
      for (Value v : polyOperands) {
        solver.fixForm(v, Form::COEFF);
      }
    }
```

Instead you would define an OpInterface like

```
def NttFormOpInterface : OpInterface<"NttFormOpInterface"> {
  let cppNamespace = "::mlir::heir";
  let description = [{
    An interface that describes requirements on the polynomial form of operands and results.
  }];

  let methods = [
    InterfaceMethod<
      "Describes which polynomial-like operands must be in coefficient form",
      "std::vector<OpOperand>", "operandsMustBeCoeffForm"
    >,
    InterfaceMethod<
      "Describes which polynomial-like operands must be evaluation form",
      "std::vector<OpOperand>", "operandsMustBeEvalForm"
    >,
    InterfaceMethod<
      "Return true if the input and output must have the same form",
      "bool", "operandsAndResultRequireSameForm"
    >,
    InterfaceMethod<
      "Returns true if the op converts form",
      "bool", "convertsFormWithCost"
    >,
    ...
  ];
}
```

You can implement these on each polynomial op, and then you'd cast to this interface and do something like

```cpp
    NttFormOpInterface nttFormInterface = dyn_cast<NttFormOpInterface>(op);
    if (!nttFormInterface) continue; // or 
      for (OpOperand v : nttFormInterface.operandsMustBeCoeffForm()) {
        solver.fixForm(v.get(), Form::COEFF);
      }
      for (OpOperand v : nttFormInterface.operandsMustBeEvalForm()) {
        solver.fixForm(v.get(), Form::EVAL);
      }
      ...
    }
```

I'm not sure it is easy to make this interface clear enough while also providing what you need, but the basic idea is that the interface exposes the minimal thing that is needed for the transformation, and then the transformation doesn't need to depend on the details of the op definitions, filter for polynomials, etc.

crockeea:
I really like your suggestion, because I think it makes it much clearer how to add new ops. One question I have though: what do you do about ops like ReturnOp or tensor::ExtractOp where [I presume] I don't have control over what interfaces they implement? Would those just have to be special cases? It would dilute the elegance of this approach significantly.

j2kun:
> I really like your suggestion, because I think it makes it much clearer how to add new ops. One question I have though: what do you do about ops like ReturnOp or tensor::ExtractOp where [I presume] I don't have control over what interfaces they implement? Would those just have to be special cases? It would dilute the elegance of this approach significantly.

Your opinion of MLIR is about to be increased slightly, because they thought of that and you can register interfaces on upstream ops. See for example here:

https://github.com/google/heir/blob/main/lib/Transforms/LayoutOptimization/InterfaceImpl.cpp#L169

Roughly the steps are:

1. Define a class that subclasses `YourOpInterface::ExternalModel`

```cpp
struct MatvecHoistingImpl
    : public LayoutConversionHoistableOpInterface::ExternalModel<
          MatvecHoistingImpl, MatvecOp> 
```

2. Create a helper function that calls `resgistry.addExtension` (you need one such call for each dialect involved, due to the lambda capturing the specific dialect pointer).
 
```cpp
registry.addExtension(+[](MLIRContext* ctx, linalg::LinalgDialect* dialect) {
    linalg::MatvecOp::attachInterface<MatvecHoistingImpl>(*ctx);
    // .. more ops here...
  });
```

3. Call the helper from `heir-opt.cpp`

https://github.com/google/heir/blob/5a9c463a8ccaead640d6480bbd2b937a6e9d5e27/tools/heir-opt.cpp#L387

tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt.mlir
crockeea: I made several tests stricter

tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt.mlir
crockeea:
```
// CHECK: @test_ntt_insertion1
// CHECK-SAME: form=eval
// CHECK-NOT: polynomial.ntt
// CHECK: return
// CHECK-SAME: form=eval
```
I was getting aliases in my types, so `form=eval` wasn't appearing.

I opted for the second approach, even though it ended up being a bit of a pain to define FileCheck aliases for all the types. Make sure I did it correctly!

crockeea:
I spent some time understanding the true nature of the problem regarding `getInherentAttrFromAttrsOrProperties`. The problem arises precisely when all of the following are true:
1. The op has an attribute
2. The op has type inference
3. Type inference requires the attribute
4. The op does *not* explicitly require the type of `output` in its IR, i.e., it is something like:
    ```
    let arguments = (ins
        Value:$value,
        Attr:$attr
    );
    let results = (outs
      Value:$output
    );
    let assemblyFormat = "$value attr-dict `:` type($value)"; // no `-> ($output)`
      ```
  
Under these conditions, the parser runs type inference before it fully parses the attribute (which seems to be integral to how MLIR works), so trying to write the following fails:
```
LogicalResult MyOp::inferReturnTypes(
    MLIRContext* context, std::optional<Location> /*loc*/, ValueRange operands,
    DictionaryAttr attrs, mlir::OpaqueProperties properties,
    mlir::RegionRange regions, SmallVectorImpl<Type>& results) {
  MyOpAdaptor op(operands, attrs, properties, regions);
  Attr attr = op.getAttr(); // crash (not a graceful exit) here
```
https://github.com/llvm/llvm-project/issues/88506 seems relevant, but it was closed, so I'm not sure. At any rate, this is more an MLIR issue than a HEIR issue. Let me know if you think it's worth opening a ticket there.

The options are as follows:
1. Use a complicated attribute getter like `getInherentAttrFromAttrsOrProperties`. This is pretty ugly in the type inference function.
2. Make the return type explicit in the IR. This seems to defeat the purpose of type inference (since it doesn't compress the IR), but there are still some benefits, namely that you can construct the op in C++ without providing an explicit return type.

Ultimately, I opted for option 2 and removed `getInherentAttrFromAttrsOrProperties`. My reason is that given mutually exclusive options, I'd rather have simpler human-written/read code like `inferReturnTypes` than simpler largely-machine-read IR.

I did not really intend to open this can of worms when starting this PR, I just ran into accidentally. I will avoid the issue for now by making the return type explicit in the IR for relevant ops, and someone else can revisit this down the road if the redundancy bothers them (it certainly bothers me!)

j2kun:
> 1. Use a complicated attribute getter like `getInherentAttrFromAttrsOrProperties`. This is pretty ugly in the type inference function.
> 2. Make the return type explicit in the IR. This seems to defeat the purpose of type inference (since it doesn't compress the IR), but there are still some benefits, namely that you can construct the op in C++ without providing an explicit return type.
> 
> Ultimately, I opted for option 2 and removed `getInherentAttrFromAttrsOrProperties`. My reason is that given mutually exclusive options, I'd rather have simpler human-written/read code like `inferReturnTypes` than simpler largely-machine-read IR.

I think this is a good choice. Thanks for looking into it.

j2kun:
@crockeea let me know when this is ready for a second review. It looks like it's nearly there?

crockeea:
Getting close, but I want to try using the OpInterfaces to make the pass easier to read, understand, modify, and extend.

crockeea:
I ended up taking a simpler approach by defining an enum inside the pass, classifying each op once, and using the enums for branching in the walks of the AST. I think this got most of the benefit without much boilerplate, but let me know if you think otherwise.

Once the tests pass, I think this is ready to merge!

j2kun (approved):
Looks great! Let's get it in.

asraa:
I think you might need to rebase this one

crockeea:
@asraa I'm assuming that comment was for @j2kun ? I learned my lesson last time, and I'm not touching anything since the `pull_ready` tag is attached.

j2kun:
@crockeea  yes please rebase. The import failed and a rebase is required for it to proceed.

j2kun:
Very strangely, when run on google-internal infra, `@test_ntt_insertion5` fails. I will comment it out so we can get this in, and then do a followup investigation.

crockeea:
That is interesting. If I had to guess, we're seeing some non-determinism in the SAT solver. An equally-valid solution to ntt_insertion_5 is to have the inputs and outputs be in NTT form.

j2kun:
> That is interesting. If I had to guess, we're seeing some non-determinism in the SAT solver. An equally-valid solution to ntt_insertion_5 is to have the inputs and outputs be in NTT form.

I think it's something more sinister. The failing output is actually inserting one ntt op when none are needed. I will dig into it, it's probably a very minor issue.

crockeea:
If you figure out how to reproduce it, I'm interested in debugging.

PR #2676, opened 2026-02-20
Migrate HEIR to use a hermetic LLVM toolchain (clang, lld, etc.)

### LINKED ISSUE #2675: Move HEIR to a hermetic C++ toolchain
Users (myself included) are often having trouble compiling HEIR due to differences of the host system. E.g., dealing with openmp, libstdc++/libc++ versions, gcc vs clang and their version differences (and Apple Clang/XCode), etc.

This is also a problem with building the Python frontend (e.g., https://github.com/google/heir/issues/2559).

It seems like the best path forward is to finally configure HEIR to use a hermetic C++ toolchain, so that there are no dependencies on the host system. After doing some research it seems https://github.com/cerisier/toolchains_llvm_bootstrapped is the best option. The main downside is that this adds longer build time because it builds the C++ standard libraries (at least) from source. The main upside is that all major targets seem to be supported, so no more issues with MacOS/XCode, and the Python frontend build issues should be naturally resolved.

In my initial experiments, so far it seems the blockers are:

- OpenMP support, filed https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/322
- Some of our dependencies don't build with clang-21, which appears hard-coded in the toolchain
  - Filed https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/324 to as if you can configure an older version
  - Filed https://github.com/bazelbuild/bazel-central-registry/issues/7642 to see if we can fix the underlying compilation errors in the respective BCR entries

j2kun:
Fixes https://github.com/google/heir/issues/2675

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,9 +1,12 @@
 module(name = "heir")
 
 bazel_dep(name = "bazel_skylib", version = "1.9.0")
+bazel_dep(name = "rules_cc", version = "0.2.18")
 
-# This must come before apple_support cf. https://github.com/bazelbuild/bazel/issues/25728
-bazel_dep(name = "rules_cc", version = "0.2.13")
+# hermetic LLVM toolchain (for C++ compiler and linker)
+bazel_dep(name = "llvm", version = "0.7.8")
+
+register_toolchains("@llvm//:all")
 
 # Apple compilation support.
 # Note this requires full xcode, not just commandline tools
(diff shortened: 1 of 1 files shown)


j2kun:
Slight progress here as sed and gawk were patched in BCR to work with clang 21.

The remaining builds that need fixing are

- bison
- fmt
- openfhe (needs omp.h)

I included the build logs in a (temporary) commit

j2kun:
Bumping to clang 22 and... IT WORKS!!!!

PR #2690, opened 2026-02-24
layout: Improve error handling and support for permutation layout

VedantParanjape:
Add support for <NxType> as input vector in the layout permutation operation. This also adds bound checking for the layout indices.


VedantParanjape:
This is a follow-up to clean up the code and handle more edge cases before I take a stab at #2666

PR #2692, opened 2026-02-25
[NFC] ArithmeticDag visitor returns vector

j2kun:
This commit changes the return type of all ArithmeticDag visitor operator() methods to `std::vector<T>` from T. This is in preparation for supporting ForLoop DAG nodes that have multiple iter args, toward keeping our FHE kernels rolled.

- Updated EvalVisitor, IRMaterializingVisitor to return vectors
- Updated all call sites to extract [0] from the returned vectors

Extracted from https://github.com/google/heir/pull/2655

### DIFF
--- a/lib/Kernel/ArithmeticDagTest.cpp
+++ b/lib/Kernel/ArithmeticDagTest.cpp
@@ -80,45 +80,45 @@ struct FlattenedStringVisitor {
   }
 };
 
-class EvalVisitor : public CachingVisitor<double, double> {
+class EvalVisitor : public CachingVisitor<double, std::vector<double>> {
  public:
   // This is required for this class to see all overloads of the visit function,
   // including virtual ones not implemented by this class.
-  using CachingVisitor<double, double>::operator();
+  using CachingVisitor<double, std::vector<double>>::operator();
 
-  EvalVisitor() : CachingVisitor<double, double>(), callCount(0) {}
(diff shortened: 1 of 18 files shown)


lib/Kernel/ArithmeticDagTest.cpp
```diff
@@ -80,45 +80,45 @@ struct FlattenedStringVisitor {
...
+    return {node.value};
   }
 
-  double operator()(const AddNode<double>& node) override {
+  std::vector<double> operator()(const AddNode<double>& node) override {
     // Recursive calls use the public `process` method from the base class
     // to ensure caching is applied at every step.
     callCount += 1;
-    return this->process(node.left) + this->process(node.right);
+    return {this->process(node.left)[0] + this->process(node.right)[0]};
```
asraa, line 107: wouldn't this need to add each of the LHS and RHS nodes in the vector elementwise?

(or is this NFC just changing the return type but holding the assumption that each node only has one element?)

j2kun: Correct. The only nodes that will have multiple results are loop nodes

asraa: OH gotcha, right - add / etc nodes only have one LHS / RHS, and elementwise vectors for LHS/RHS are handled in visitors that handle vectors as a single element

I've unconfused myself, that makes sense why the arith nodes here don't need to be doing that.

PR #2693, opened 2026-02-25
Add types to some dag nodes

j2kun:
Rebased over #2692

This commit adds types to ArithmeticDag nodes, allowing constant nodes to carry type information needed for proper MLIR materialization. This came from https://github.com/google/heir/pull/2655 where the IRMaterializer now needs to know the appropriate types of certain dag nodes (such as a tensor splat)

Primarily this PR updates ConstantScalarNode and ConstantTensorNode to include a DagType field.

To demonstrate the need for this, this PR also adds a SplatNode dag node.

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/RotationGroupKernel.h
+++ b/lib/Dialect/TensorExt/Transforms/RotationGroupKernel.h
@@ -72,11 +72,14 @@ applyVirtualRotation(ArrayRef<std::shared_ptr<ArithmeticDagNode<T>>> input,
       auto [allZero, allOne] = allZeroAllOne(mask);
       if (allZero) {
         std::vector<double> zeros(ciphertextSize, 0.0);
-        masked.push_back(NodeTy::constantTensor(zeros));
+        masked.push_back(NodeTy::constantTensor(
+            zeros, kernel::DagType::floatTensor(64, {ciphertextSize})));
       } else if (allOne) {
         masked.push_back(ct);
       } else {
-        masked.push_back(NodeTy::mul(ct, NodeTy::constantTensor(mask)));
+        masked.push_back(NodeTy::mul(
+            ct, NodeTy::constantTensor(
+                    mask, kernel::DagType::floatTensor(64, {ciphertextSize}))));
       }
     }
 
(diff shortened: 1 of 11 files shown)


lib/Dialect/TensorExt/Transforms/RotationGroupKernel.h
```diff
@@ -72,11 +72,14 @@ applyVirtualRotation(ArrayRef<std::shared_ptr<ArithmeticDagNode<T>>> input,
       auto [allZero, allOne] = allZeroAllOne(mask);
       if (allZero) {
         std::vector<double> zeros(ciphertextSize, 0.0);
-        masked.push_back(NodeTy::constantTensor(zeros));
+        masked.push_back(NodeTy::constantTensor(
```
asraa, line 75: I was filing https://github.com/google/heir/issues/2700 because Mark told me that the kernel DAGs materialized for mlir-to-bgv contain floating point operations, and that made me conscious of these choices.

Should the DagType passed into the constantTensor pull from the input DAG's type? That way if we need to change the input ciphertexts to match float/int based on the IR, we won't have to update this later

lib/Kernel/RotationCountVisitor.cpp
```diff
@@ -141,6 +141,12 @@ int64_t RotationCountVisitor::operator()(
   return operandCount;
 }
 
+int64_t RotationCountVisitor::operator()(const SplatNode& node) {
+  // Splat constants are always plaintext
```
asraa, line 145: shouldn't a splat node take the secretness of the input operand? (why are splats always constant?)

lib/Dialect/TensorExt/Transforms/RotationGroupKernel.h
j2kun: You're right. That is handled specifically by the changes in https://github.com/google/heir/pull/2696. I just wanted to try to tease apart the different bits from the much larger PR

lib/Kernel/RotationCountVisitor.cpp
j2kun: In this case, the splatted value is static, not dynamic. I suppose we could make it dynamic, but I didn't see a need for it. This is mainly to keep code size small when generating plaintexts of all zeros.

PR #2695, opened 2026-02-25
ArithmeticDag: Add floorDiv node

j2kun:
This is primarily needed for division operations used in index calculations for dynamic rotations (e.g., updating the iter_arg `shift //= 2` in a rotate-and-reduce loop).

### DIFF
--- a/lib/Kernel/ArithmeticDag.h
+++ b/lib/Kernel/ArithmeticDag.h
@@ -118,6 +118,12 @@ struct MultiplyNode {
 };
 
 template <typename T>
+struct FloorDivNode {
+  std::shared_ptr<ArithmeticDagNode<T>> left;
+  int divisor;
+};
+
+template <typename T>
 struct PowerNode {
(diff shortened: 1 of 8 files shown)


lib/Kernel/ArithmeticDag.h
```diff
@@ -117,6 +117,12 @@ struct MultiplyNode {
   std::shared_ptr<ArithmeticDagNode<T>> right;
 };
 
+template <typename T>
+struct DivideNode {
```
asraa: I'd suggest some comment about the semantics here - should this be a floordiv? (can we name this floordiv instead?)

j2kun: Technically (according to IRMaterializingVisitor lowering this to DivSI and DivF) it does both. I only need it for division of int/index values, so I can update it.

j2kun: Fixed

PR #2696, opened 2026-02-25
ArithmeticDag: use dynamic index for LeftRotateNode [NFC]

j2kun:
This commit went a bit down a rabbit hole. The reason is that, by making the index of LeftRotateNode itself a DagNode (required for dynamic rotation indices for rolled kernels), the current deficiencies of the IRMaterializingVisitor started to show through. In particular, when materializing a constant scalar node, the IRMaterializingVisitor decided on the type by using a globally provided "evaluated type," which most users of IRMaterializingVisitor set to a tensor type (e.g., the tensor type of the underlying tensor used in a layout conversion shift network).

To fix this, I had to make the IRMaterializingVisitor use the DagType from a ConstantScalarNode or ConstantTensorNode instead of the evaluatedType.

Fixing this required making a lot more invasive changes across the codebase, including migrating polynomial lowering uses of ArithmeticDag and the shift network.

Along the way, I made some additional opportunistic migrations of code from #2655 to use SplatNodes where more verbose constant tensors were used, and to add SplatNode to IRMaterializingVisitor.

Rebased over #2695 and all its ancestors.

### DIFF
--- a/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
+++ b/lib/Dialect/TensorExt/Transforms/ImplementShiftNetwork.cpp
@@ -216,6 +216,18 @@ LogicalResult convertRemapOp(RemapOp op,
   int64_t ciphertextSize = tensorTy.getDimSize(1);
   auto singleCiphertextType =
       RankedTensorType::get({1, ciphertextSize}, tensorTy.getElementType());
+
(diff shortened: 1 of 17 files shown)


j2kun:
@asraa this one is ready now

asraa (approved):

PR #2699, opened 2026-02-25
Add rolled rotate-and-reduce kernel

j2kun:
This PR adds a new `unroll` API option to the KernelImplementation.h rotate and reduce kernels. Doesn't yet add a pass option to enable it due to the lack of rotation key analysis.

As per discussion with @asraa, I'm leaving out the zeroDiagonals support until a followup PR

### DIFF
--- a/lib/Kernel/KernelImplementation.h
+++ b/lib/Kernel/KernelImplementation.h
@@ -34,6 +34,12 @@ template <typename T>
 using DagExtractor = std::function<std::shared_ptr<ArithmeticDagNode<T>>(
     std::shared_ptr<ArithmeticDagNode<T>>, int64_t)>;
 
+// Dynamic extraction: takes tensor and DAG node representing runtime index
+template <typename T>
+using DagExtractorDynamic = std::function<std::shared_ptr<ArithmeticDagNode<T>>(
+    std::shared_ptr<ArithmeticDagNode<T>>,
+    std::shared_ptr<ArithmeticDagNode<T>>)>;
+
 // Returns an arithmetic DAG that implements a matvec kernel. Ensure this is
 // only generated for T a subclass of AbstractValue.
 template <typename T>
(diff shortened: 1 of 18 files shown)


j2kun:
In 48633c6e7 there is a new problem: supporting the zeroDiagonals argument is not quite so simple with the current way the code is organized. To support this, I think instead of passing a map from index to bool (isZero), we instead want to pass a list of nonzero diagonal indices, and have the loop iterate over that list instead.

j2kun:
Marking as pull_ready so I can patch internal work over this PR

lib/Kernel/TestingUtils.h
```diff
@@ -141,6 +149,37 @@ class MultiplicativeDepthVisitorImpl
...
+
+  double operator()(const VariableNode<LiteralValue>& node) override {
+    // Variables have depth 0 (they are inputs)
+    return 0.0;
+  }
+
+  double operator()(const ForLoopNode<LiteralValue>& node) override {
+    // Return the maximum depth of the init and body
+    double maxDepth = -1.0;
+    for (const auto& init : node.inits) {
```
asraa: why would this include the inits? let's say a for loop initializes something with depth 4, but then returns something with depth 3, shouldn't the result of the for loop be the returned depth 3?

let's say it's something like this, where the iter_arg's initial value isn't used
for i in (0, n) iter_args(depth_4_thing):
   v0 = arith.constant
   v1 = v0 * v0 * v0
   return v1

wouldn't it be depth 3?

lib/Kernel/KernelImplementation.h
```diff
@@ -174,10 +244,102 @@ implementBabyStepGiantStep(
     }
   }
 
-  // Hack to make 0 when there are no nonzero diagonals - only appears in
```
asraa, line 177: omg, thank you - that was my tech debt i think

lib/Kernel/TestingUtils.h
j2kun: Good point, fixed.

PR #2708, opened 2026-02-26
fix fuzztest runner

j2kun:
The fuzz tests weren't actually running!


j2kun:
Ha, I think the internal version was running it, so this is likely a copybara issue.

PR #2733, opened 2026-03-06
[utils] Add getEdges() and contractEdge() to Graph utils

VedantParanjape:
This patch adds two new functions to the Graph class.
1) getEdges() returns all edges as a set of pair of vertices as {source, target}.
2) contractEdge(u, v, fn) merges edge u into edge v and runs a functor fn on u and v to perform custom merge operation on u and v.

### DIFF
--- a/lib/Utils/Graph/Graph.h
+++ b/lib/Utils/Graph/Graph.h
@@ -62,6 +62,15 @@ class Graph {
   bool empty() const { return vertices.empty(); }
 
   const std::set<V>& getVertices() { return vertices; }
+  std::set<std::pair<V, V>> getEdges() const {
+    std::set<std::pair<V, V>> result;
+    for (const auto& [source, targets] : outEdges) {
+      for (const auto& target : targets) {
+        result.insert({source, target});
+      }
+    }
+    return result;
+  }
 
   // Returns the edges that point out of the given vertex.
   std::vector<V> edgesOutOf(V vertex) {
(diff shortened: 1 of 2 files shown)


VedantParanjape:
it seems the linter is failing on some different merged commit.

lib/Utils/Graph/Graph.h
```diff
@@ -85,6 +94,39 @@ class Graph {
     return {};
   }
 
+  bool contractEdge(V source, V target,
```
j2kun, line 110: This could use some documentation, particularly about what `mergeFn` is allowed to do. For instance, I don't think mergeFn is allowed to modify the current graph instance (or is it? if so, in what way?). What are the semantics of the returned `V` from `mergeFn`?

lib/Utils/Graph/Graph.h
```diff
@@ -85,6 +94,39 @@ class Graph {
...
+                    std::function<V(V, V)> mergeFn = nullptr) {
```
j2kun: I suspect this would be better as `std::function<void(const V&, const V&)>` or at least `std::function<void(V&, const V&)>` to signify that the second argument will be deleted and the first argument will be what remains after the merging is complete.

Also, why do you need the `mergeFn` to return a `V`?

j2kun (changes requested):
Generally looks good. Some minor questions

lib/Utils/Graph/Graph.h
VedantParanjape: This seems much better, I just made mergeFn to return V because I wanted to exhibit a semantics that the mergeFn "merged two nodes" and created another node.

lib/Utils/Graph/Graph.h
VedantParanjape: done

VedantParanjape:
> Please squash your commits and we'll get it in. No need to worry about the unrelated formatting error, that shouldn't block merging.

done!

PR #2735, opened 2026-03-06
[utils] Add SCC and condense calls to Graph utils

VedantParanjape:
This patch adds the following API calls to the Graph class:

1) getStronglyConnectedComponents: Implement function to find strongly connected components using Tarjan's algorithm. It returns a vector of sets, each set represents one SCC.

2) condenseGraph: Condenses each strongly connected component into a single vertex, thus making the graph acyclic.

### DIFF
--- a/lib/Utils/Graph/Graph.h
+++ b/lib/Utils/Graph/Graph.h
@@ -14,6 +14,7 @@
 #include <vector>
 
 #include "llvm/include/llvm/ADT/STLExtras.h"          // from @llvm-project
+#include "llvm/include/llvm/ADT/SetVector.h"          // from @llvm-project
 #include "mlir/include/mlir/Support/LogicalResult.h"  // from @llvm-project
 
 namespace mlir {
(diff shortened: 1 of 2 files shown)


lib/Utils/Graph/Graph.h
```diff
@@ -154,6 +155,91 @@ class Graph {
...
+      std::set<V> alive(scc.begin(), scc.end());
+      while (alive.size() > 1) {
+        // needs a snapshot as we modify the alive set while iterating through
+        // it.
+        std::vector<V> aliveSnapShot(alive.begin(), alive.end());
+        for (const V& u : aliveSnapShot) {
+          if (!alive.count(u)) continue;
+          for (V v : edgesOutOf(u)) {
+            if (alive.count(v)) {
+              contractEdge(u, v, mergeFn);
```
j2kun, line 243: While this is correct, I suspect it would be more efficient to construct an entirely new graph (with one node for each SCC component) and then rebuild the edges from scratch in that new graph.

In that setting, it might make more sense to have the callback take as input a list of vertices (one for each vertex in the component), and—contrary to the advice from the last PR!—ask the user to produce a new vertex representing the metadata of the entire component.

I leave it to you to decide if that makes more sense with what you plan to use this for. For small graphs it doesn't matter, and this comment can serve as history for any future optimization ideas.

lib/Utils/Graph/Graph.h
```diff
@@ -154,6 +155,91 @@ class Graph {
...
+  //
+  // V = struct { string name; }
+  // mergeFn = [](V& source, const V& target) { source.name += target.name; }
+  //
+  // and we contract edge (v1, v2) where v1.name = "foo" and v2.name = "bar",
+  // then the mergeFn could update v1.name to "foobar" to preserve some
+  // information from both vertices before v2 is removed from the graph.
+  void condenseGraph(std::function<void(V&, const V&)> mergeFn = nullptr) {
+    auto sccs = getStronglyConnectedComponents();
+    std::map<V, V> oldToNew;
```
j2kun: nit: unused

```suggestion
```

lib/Utils/Graph/Graph.h
```diff
@@ -154,6 +155,91 @@ class Graph {
...
+  // Each SCC is returned as a std::set<V> (consistent with getVertices()).
+  // Reference implementation:
+  // https://cp-algorithms.com/graph/strongly-connected-components.html#implementation_1
+  //
+  // Returns a vector of set of vertices, where each set represents a strong
+  // connected component. The order of the components and the order of vertices
+  // within each component are not preserved.
+  std::vector<std::set<V>> getStronglyConnectedComponents() {
+    std::vector<std::set<V>> result;
+    std::map<V, int> index, lowlink;
```
j2kun, line 173: This looks fine, for the future I want to note that this implementation does not apply Nuutila's modification (which is standard in impls like networkx/LLVM and saves on space).

So if performance is an issue or we hit big graphs, that would be one area for improvement.

lib/Utils/Graph/Graph.h
VedantParanjape: I gave this a thought while working on the implementation, it is okay to construct a new graph. But that means I will have to change what mergeFn does. I can take a look at this again, once I am done with implementing coyote.

lib/Utils/Graph/Graph.h
VedantParanjape: I will add a to-do here.

VedantParanjape:
cleaned up the dead code, squashed the commits. good to go!

PR #2892, opened 2026-04-21
Make LWE ops more robust

crockeea:
While doing other work, I realized that the issue identified [here](https://github.com/google/heir/pull/2671#issuecomment-3975431340) was still present in the LWE dialect. Why wasn't this causing a segfault like it did in Polynomial?

After some investigation, the reason was random chance: there is a real "bug" here, and LWE segfaults were just waiting to happen. I added a test that triggered the segfault, then fixed the issue (by adding explicit return types to ops with attributes and type inference that uses the attributes). There do not appear to be any other occurrences of the "bad" combination of factors in the repo.

I opened issue an with MLIR [here](https://github.com/llvm/llvm-project/issues/193284).

### DIFF
--- a/lib/Dialect/LWE/IR/LWEOps.td
+++ b/lib/Dialect/LWE/IR/LWEOps.td
@@ -317,7 +317,7 @@ def LWE_ExtractSliceOp : LWE_Op<"extract_slice", [Pure, DeclareOpInterfaceMethod
   let results = (outs
     LWERingElt:$output
   );
-  let assemblyFormat = "$input attr-dict `:` type($input)";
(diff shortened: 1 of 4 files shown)


asraa (approved):
> Make the return type explicit in the IR. This seems to defeat the purpose of type inference (since it doesn't compress the IR), but there are still some benefits, namely that you can construct the op in C++ without providing an explicit return type.

True - and I agree with what you did here. Type inference is still nice to have for the C++ code, and it's a bummer that the IR size ends up larger. I think that issue you linked upstream might have been closed with the migration to a props dictionary, but now that props are default, it seems like the issue still exists...

PR #2893, opened 2026-04-21
Implement basis conversion

crockeea:
This PR adds an implementation of basis extension. I made several choices which may not be right in the end, so input on architecture is welcome.

- I implemented basis conversion inside the LWEToPolynomial conversion. An alternate construction would be to lower LWE.convert_basis to polynomial.convert_basis and then replace polynomial.convert_basis with a polynomial.apply_coefficientwise, but that seemed unnecessary to me.
- This means there is no longer a need for Polynomial's ConvertBasisOp, so it has been removed. All of the NTT tests that used ConvertBasis have been converted to use ApplyCoefficientwise instead, since there were no tests for this op previously, and because it is coeff-only like ConvertBasis was.
- I added a helper function in PolynomialOps to properly apply polynomial.apply_coefficientwise, which involves a lot of boilerplate.
- I chose _not_ to add an rns.convert_basis op. My motivation for this and implementing the lowering in LWEToPolynomial is that my goal is to lower the implementation to the polynomial layer _without_ going to ModArith. I'm still not 100% sold on using polynomial.apply_coefficientwise for that reason, since it involves the use of ModArith, but I'm trying it for now. [The alternative would be to add some new ops to the Polynomial dialect, which is also not appealing.] Instead of an rns.convert_basis op, I implemented basis conversion as top-level C++ functions in RNSOps.cpp, which are used by the LWEToPolynomial lowering for LWE.ConvertBasis.
- There are new RNS ops for extracting a single slice (rather than an RNS subset of slices) and for packing individual limbs together into an RNS.

Miscellaneous changes:
1. When programming with GPT in VSCode, I get a .codex file, which should be ignored.
2. I renamed the TypeInterface build targets in RNS and ModArith to match the style of other build targets

The biggest piece missing from this PR is that there are no LLVM tests for basis conversion. This should be added, but I figured I should get an architecture review first, and also get input on where to put the tests. My feeling is that we should have some rns-to-llvm tests for basis conversion.

PR written with AI assistance.

### DIFF
--- a/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
@@ -12,12 +12,14 @@
 #include "lib/Dialect/Polynomial/IR/PolynomialAttributes.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialOps.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialTypes.h"
+#include "lib/Dialect/RNS/IR/RNSOps.h"
 #include "lib/Dialect/Random/IR/RandomEnums.h"
 #include "lib/Dialect/Random/IR/RandomOps.h"
 #include "lib/Dialect/Random/IR/RandomTypes.h"
 #include "lib/Utils/ConversionUtils.h"
 #include "lib/Utils/Polynomial/Polynomial.h"
 #include "llvm/include/llvm/ADT/ArrayRef.h"              // from @llvm-project
+#include "llvm/include/llvm/ADT/DenseMap.h"              // from @llvm-project
 #include "llvm/include/llvm/ADT/TypeSwitch.h"            // from @llvm-project
 #include "llvm/include/llvm/Support/ErrorHandling.h"     // from @llvm-project
 #include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
@@ -610,9 +612,26 @@ struct ConvertConvertBasis : public OpConversionPattern<ConvertBasisOp> {
       ConvertBasisOp op, OpAdaptor adaptor,
       ConversionPatternRewriter& rewriter) const override {
     ImplicitLocOpBuilder b(op->getLoc(), rewriter);
-    rewriter.replaceOp(
-        op, polynomial::ConvertBasisOp::create(b, adaptor.getValue(),
-                                               adaptor.getTargetBasis()));
+    MLIRContext* context = op.getContext();
+
+    polynomial::PolynomialType inputPolyTy =
+        cast<polynomial::PolynomialType>(adaptor.getValue().getType());
+    polynomial::RingAttr ringAttr = inputPolyTy.getRing();
+    rns::RNSType targetBasisTy = dyn_cast<rns::RNSType>(op.getTargetBasis());
+    if (!targetBasisTy) return failure();
+    polynomial::RingAttr outputRingAttr = polynomial::RingAttr::get(
+        context, targetBasisTy, ringAttr.getPolynomialModulus());
+    polynomial::PolynomialType outputPolyTy =
+        polynomial::PolynomialType::get(context, outputRingAttr);
+
+    Value coeffLoop = polynomial::ApplyCoefficientwiseOp::create(
+        b, adaptor.getValue(), outputPolyTy,
+        [&](ImplicitLocOpBuilder& nestedBuilder, Value coeff,
+            Value _) -> Value {
+          return rns::ConvertBasisOp::create(nestedBuilder, coeff,
+                                             targetBasisTy);
+        });
+    rewriter.replaceOp(op, coeffLoop);
     return success();
   }
 };
--- a/lib/Dialect/Polynomial/IR/PolynomialOps.cpp
+++ b/lib/Dialect/Polynomial/IR/PolynomialOps.cpp
@@ -4,6 +4,7 @@
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
+#include <memory>
 #include <optional>
 
 #include "lib/Dialect/ModArith/IR/ModArithAttributes.h"
@@ -68,6 +69,30 @@ LogicalResult coefficientTypeMatchesScalarType(Type polynomialLikeType,
     return failure();
   }
   return success();
+}
+
+void ApplyCoefficientwiseOp::build(
+    OpBuilder& builder, OperationState& state, Value input,
+    PolynomialType outputPolyTy, ApplyCoefficientwiseBuilderFn transformCoeff) {
+  state.addOperands(input);
+  Region* body = state.addRegion();
+  state.addTypes(outputPolyTy);
+
+  PolynomialType inputPolyTy = dyn_cast<PolynomialType>(input.getType());
+  if (!inputPolyTy) {
+    return;
+  }
+
+  ImplicitLocOpBuilder b(state.location, builder);
+  OpBuilder::InsertionGuard guard(b);
+  Block* block = b.createBlock(body);
+  block->addArgument(inputPolyTy.getRing().getCoefficientType(), b.getLoc());
+  block->addArgument(b.getIndexType(), b.getLoc());
+  b.setInsertionPointToStart(block);
+
+  Value coeff = block->getArgument(0);
+  Value degree = block->getArgument(1);
+  YieldOp::create(b, transformCoeff(b, coeff, degree));
 }
 
 void FromTensorOp::build(OpBuilder& builder, OperationState& result,
@@ -499,37 +524,6 @@ LogicalResult INTTOp::inferReturnTypes(
   return inferNTTReturnType(ctx, operands.front().getType(), results);
 }
 
-LogicalResult ConvertBasisOp::inferReturnTypes(
-    MLIRContext* ctx, std::optional<Location> /*loc*/, ValueRange operands,
-    DictionaryAttr attrs, mlir::PropertyRef properties,
-    mlir::RegionRange regions, SmallVectorImpl<Type>& results) {
-  ConvertBasisOpAdaptor op(operands, attrs, properties, regions);
-  if (operands.empty()) return failure();
-  Type inputType = op.getValue().getType();
-  PolynomialType inputPolyType = dyn_cast<PolynomialType>(inputType);
-  RankedTensorType inputTensorType;
-  if (!inputPolyType) {
-    inputTensorType = dyn_cast<RankedTensorType>(inputType);
-    if (!inputTensorType) return failure();
-    inputPolyType = dyn_cast<PolynomialType>(inputTensorType.getElementType());
-    if (!inputPolyType) return failure();
-  }
-  polynomial::RingAttr ringAttr = inputPolyType.getRing();
-  rns::RNSType elementType = dyn_cast<rns::RNSType>(op.getTargetBasis());
-  if (!elementType) return failure();
-
-  polynomial::RingAttr outputRingAttr = polynomial::RingAttr::get(
-      ctx, elementType, ringAttr.getPolynomialModulus());
-  PolynomialType resultType = PolynomialType::get(ctx, outputRingAttr);
-  if (dyn_cast<PolynomialType>(inputType)) {
-    results.push_back(resultType);
-  } else {
-    results.push_back(RankedTensorType::get(
-        inputTensorType.getShape(), resultType, inputTensorType.getEncoding()));
-  }
-  return success();
-}
-
 LogicalResult ExtractSliceOp::inferReturnTypes(
     MLIRContext* context, std::optional<Location> /*loc*/, ValueRange operands,
     DictionaryAttr attrs, mlir::PropertyRef properties,
@@ -599,6 +593,11 @@ LogicalResult ApplyCoefficientwiseOp::verify() {
                          << block.getArgument(1).getType();
   }
 
+  if (!block.mightHaveTerminator()) {
+    // Do not emit an error here. This situation occurs when the op creation
+    // failed; that emits its own (meaningful) error.
+    return failure();
+  }
   auto yieldOp = cast<YieldOp>(block.getTerminator());
   auto outputPolyTy = getOutput().getType();
   auto outputCoeffTy = outputPolyTy.getRing().getCoefficientType();
--- a/lib/Dialect/Polynomial/IR/PolynomialOps.h
+++ b/lib/Dialect/Polynomial/IR/PolynomialOps.h
@@ -4,12 +4,28 @@
 #include "lib/Dialect/Polynomial/IR/PolynomialDialect.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialTraits.h"
 #include "lib/Dialect/Polynomial/IR/PolynomialTypes.h"
-#include "mlir/include/mlir/IR/BuiltinOps.h"    // from @llvm-project
-#include "mlir/include/mlir/IR/BuiltinTypes.h"  // from @llvm-project
-#include "mlir/include/mlir/IR/Dialect.h"       // from @llvm-project
+#include "llvm/include/llvm/ADT/STLFunctionalExtras.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinOps.h"            // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypes.h"          // from @llvm-project
+#include "mlir/include/mlir/IR/Dialect.h"               // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"  // from @llvm-project
 #include "mlir/include/mlir/Interfaces/InferTypeOpInterface.h"  // from @llvm-project
+#include "mlir/include/mlir/Support/LLVM.h"  // from @llvm-project
 
 #define GET_OP_CLASSES
 #include "lib/Dialect/Polynomial/IR/PolynomialOps.h.inc"
+
+namespace mlir {
+namespace heir {
+namespace polynomial {
+
+FailureOr<Value> buildApplyCoefficientwise(
+    ImplicitLocOpBuilder& b, Value input, PolynomialType outputPolyTy,
+    llvm::function_ref<FailureOr<Value>(ImplicitLocOpBuilder&, Value, Value)>
+        transformCoeff);
+
+}  // namespace polynomial
+}  // namespace heir
+}  // namespace mlir
 
 #endif  // LIB_DIALECT_POLYNOMIAL_IR_POLYNOMIALOPS_H_
--- a/lib/Dialect/RNS/IR/RNSOps.cpp
+++ b/lib/Dialect/RNS/IR/RNSOps.cpp
@@ -3,20 +3,152 @@
 #include <cstdint>
 #include <optional>
 
+#include "lib/Dialect/ModArith/IR/ModArithAttributes.h"
+#include "lib/Dialect/ModArith/IR/ModArithOps.h"
+#include "lib/Dialect/ModArith/IR/ModArithTypes.h"
 #include "lib/Dialect/RNS/IR/RNSOps.h"
 #include "lib/Dialect/RNS/IR/RNSTypes.h"
+#include "lib/Utils/APIntUtils.h"
+#include "llvm/include/llvm/ADT/DenseMap.h"              // from @llvm-project
+#include "llvm/include/llvm/ADT/SmallString.h"           // from @llvm-project
 #include "mlir/include/mlir/IR/BuiltinAttributes.h"      // from @llvm-project
 #include "mlir/include/mlir/IR/BuiltinTypeInterfaces.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypes.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/ImplicitLocOpBuilder.h"   // from @llvm-project
+#include "mlir/include/mlir/IR/Location.h"               // from @llvm-project
 #include "mlir/include/mlir/IR/MLIRContext.h"            // from @llvm-project
 #include "mlir/include/mlir/IR/OperationSupport.h"       // from @llvm-project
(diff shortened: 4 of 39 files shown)


j2kun:
(sorry for the delay, I've been at a summit in Zurich; I should have time to review this before Monday)

lib/Dialect/RNS/IR/RNSOps.h
```diff
@@ -53,6 +55,46 @@ RNSType inferExtractSliceReturnTypes(MLIRContext* ctx, Op* op,
...
+
+// Given x \in Z_Q represented as {x_i mod q_i} where Q=\prod q_i,
+// convertBasis returns the target-basis representation obtained by:
+//  1. Compute x's centered integer representative in [-Q/2, Q/2]
+//  2. Reduce the integer representation mod p_i for each p_i in the target
+//  basis.
+//
+// This performs the conversion directly from the RNS representation of x,
+// without first reconstructing x over the integers. Note that both bases must
+// be odd.
```
j2kun:
```suggestion
// have odd entries.
```

Nit: I read this as "have odd length"

lib/Dialect/RNS/IR/RNSOps.td
```diff
@@ -29,5 +29,45 @@ def RNS_ExtractSliceOp : RNS_Op<"extract_slice", [Pure, ElementwiseMappable, Dec
   let hasVerifier = 1;
 }
 
+def RNS_ExtractSingleSliceOp : RNS_Op<"extract_single_slice", [Pure, ElementwiseMappable, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
```
j2kun: Any reason not to name this op `extract_residue` or `extract_limb`? This seems more aligned with the semantics of the op than `single_slice`.

j2kun: I like the name "residue" also because it differentiates from the limb _type_ vs the value in that limb. This could clarify the description of the rns.pack op below: `"Pack multiple limb values into one RNS value"` -> `"Pack multiple residues into one RNS value"`

lib/Dialect/RNS/IR/RNSOps.td
```diff
@@ -29,5 +29,45 @@ def RNS_ExtractSliceOp : RNS_Op<"extract_slice", [Pure, ElementwiseMappable, Dec
...
+  let description = [{
+    Given an RNS-typed value, extract the basis limb at `index` and return it
+    as the corresponding type.
+
+    For shaped inputs, this operation is applied elementwise, preserving the
+    input container shape while replacing each RNS element with the selected
+    limb.
+  }];
+  let arguments = (ins RNSLike:$input, IndexAttr:$index);
+  let results = (outs AnyType:$output);
```
j2kun: I think you can use the `RNSBasisTypeInterface` as a type constraint here instead of `AnyType`, which will provide slightly better error messages if someone tries to use the wrong operand/result types. Same for the variadic operand types to `pack` below.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
```diff
@@ -70,6 +70,35 @@ LogicalResult coefficientTypeMatchesScalarType(Type polynomialLikeType,
   return success();
 }
 
+FailureOr<Value> buildApplyCoefficientwise(
```
j2kun: This boilerplate can be implemented as a custom builder on the `ApplyCoefficientwiseOp` itself, which would allow users to see it alongside other builders when doing `ApplyCoefficientwiseOp::create`. We do this for `secret.generic` to pass a "body builder" function in the same way that it's done here:

The tablegen spec: https://github.com/google/heir/blob/7e114c39eb41fe75f2621a6701adba50d4d73e4e/lib/Dialect/Secret/IR/SecretOps.td#L190-L196 

Generates a header declaration that we implement here:

https://github.com/google/heir/blob/main/lib/Dialect/Secret/IR/SecretOps.cpp#L295-L313

Along with an "extraClassDeclaration" to get the callback type alias in there:
https://github.com/google/heir/blob/7e114c39eb41fe75f2621a6701adba50d4d73e4e/lib/Dialect/Secret/IR/SecretOps.td#L198-L200

In the C++ implementation for secret.generic, it's a little bit more complicated because you have to create the region as well as the block. Also, in that example the callback is required to define the yield op, because what is yielded from a secret.generic is specific to each op, whereas here creating the yield op automatically like you did is better.

The downside of that approach is that it returns void instead of a FailureOr, so if you want to keep your careful error handling, you'd `emitError` and then just return and let the op verifier fail (at least, because it would be missing a yield op).

lib/Dialect/Polynomial/IR/PolynomialOps.h
```diff
@@ -4,12 +4,28 @@
...
+#include "mlir/include/mlir/Support/LLVM.h"  // from @llvm-project
 
 #define GET_OP_CLASSES
 #include "lib/Dialect/Polynomial/IR/PolynomialOps.h.inc"
 
+namespace mlir {
+namespace heir {
+namespace polynomial {
+
+FailureOr<Value> buildApplyCoefficientwise(
```
j2kun, line 22: Nit: add a docstring for this function that specifies what the callback arguments are, in particular which Value is the induction variable and which is the coefficient.

If you go the above route of a custom builder, then you can put the argument description above the type alias for the callback.

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -610,9 +612,47 @@ struct ConvertConvertBasis : public OpConversionPattern<ConvertBasisOp> {
...
+    if (!targetBasisTy) return failure();
+    polynomial::RingAttr outputRingAttr = polynomial::RingAttr::get(
+        context, targetBasisTy, ringAttr.getPolynomialModulus());
+    polynomial::PolynomialType outputPolyTy =
+        polynomial::PolynomialType::get(context, outputRingAttr);
+    auto inputCoeffTy = cast<rns::RNSType>(ringAttr.getCoefficientType());
+
+    // Precompute some values to reduce duplication across each coefficient of
+    // the polynomial Each coefficient uses the same qInvProds
+    ArrayAttr qInvProds = *rns::buildQInvProds(context, inputCoeffTy);
```
j2kun: may as well check for failure here.

lib/Dialect/RNS/IR/RNSOps.cpp
```diff
@@ -3,21 +3,268 @@
...
+    Value qInvConst =
+        mod_arith::ConstantOp::create(b, limbTy, maAttr.getValue());
+    ci = mod_arith::MulOp::create(b, ci, qInvConst);
+    Value liftedCi = mod_arith::ExtractOp::create(b, intType, ci);
+    mrcs.push_back(liftedCi);
+  }
+  return mrcs;
+}
+
+FailureOr<ArrayAttr> buildQInvProds(mlir::MLIRContext* ctx,
```
j2kun: nit: since you end up materializing these qInv values as constants, I suspect you can avoid constructing the `ArrayAttr` and just return a `SmallVector<APint>` of the `qInv` values. I think that should simplify the `computeMixedRadixCoeffs` a little bit. You can even take it a step further and extract the core computation to operate entirely on APInts (and vectors thereof) and unit test the computation in isolation.

You can probably also add a custom `mod_arith::ConstantOp::create` that accepts the type and an APInt as input, and handles the boilerplate of constructing the ModArithAttr and underlying IntegerAttr.

Creating lots of unused attributes also has a performance impact (at heir-opt runtime) because constructing MLIR attributes involves extra hashing and uniquifying steps.

lib/Dialect/RNS/IR/RNSOps.cpp
```diff
@@ -3,21 +3,268 @@
...
+  for (int i = 1; i < numLimbs; i++) {
+    Value xi = ExtractSingleSliceOp::create(b, input, b.getIndexAttr(i));
+    auto limbTy = dyn_cast<mod_arith::ModArithType>(xi.getType());
+    if (!limbTy) {
+      emitError(b.getLoc())
+          << "expected limb type to be mod_arith attribute, got "
+          << xi.getType();
+      return failure();
+    }
+    // Using Horner's method, we compute (c_{i-1}*q_{i-2} + c_{i-2})*q_{i-3} +
```
j2kun, line 73: Offhand, I wonder to what extent folks have looked at improving upon this using something like Estrin's scheme https://en.wikipedia.org/wiki/Estrin%27s_scheme

At least, if basis extension is a computational bottleneck, that should provide a little bit more parallelism opportunities.

Not particularly relevant for this lowering which is meant for functional testing, but I thought I'd mention it, since the use of MacOp reminded me.

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -610,9 +612,47 @@ struct ConvertConvertBasis : public OpConversionPattern<ConvertBasisOp> {
...
+    // We need a map from modulus to index (on the input basis) for
+    // short-circuiting
+    llvm::DenseMap<APInt, size_t> inputBasisIndexByModulus;
+    ArrayRef<Type> inputBasisTypes = inputCoeffTy.getBasisTypes();
+    for (size_t i = 0; i < inputBasisTypes.size(); ++i) {
+      mod_arith::ModArithType inputLimbTy =
+          dyn_cast<mod_arith::ModArithType>(inputBasisTypes[i]);
```
j2kun:
```suggestion
    for (auto [i, basisTy] : llvm::enumerate(inputCoeffTy.getBasisTypes())) {
      auto inputLimbTy = dyn_cast<mod_arith::ModArithType>(basisTy);
```

nit: use `llvm::enumerate`

lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp
```diff
@@ -610,9 +612,47 @@ struct ConvertConvertBasis : public OpConversionPattern<ConvertBasisOp> {
...
+        return rewriter.notifyMatchFailure(
+            op, "expected input basis limb to be ModArithType");
+      }
+      inputBasisIndexByModulus[inputLimbTy.getModulus().getValue()] = i;
+    }
+    FailureOr<Value> coeffLoop = polynomial::buildApplyCoefficientwise(
+        b, adaptor.getValue(), outputPolyTy,
+        [&](ImplicitLocOpBuilder& nestedBuilder, Value coeff,
+            Value degree) -> FailureOr<Value> {
+          (void)degree;
```
j2kun:
```suggestion
            Value _) -> FailureOr<Value> {
```

C++26 treats `_` special for an unused variable, and if that doesn't work for you (we're not on C++26 yet but most tools like the underscore), you can also use `[[maybe_unused]]` in front of the type to make it clearer.

lib/Dialect/RNS/IR/RNSOps.cpp
```diff
@@ -3,21 +3,268 @@
...
+  if (maybeMrcs->empty()) {
+    emitError(b.getLoc()) << "expected non-empty mixed-radix coefficients";
+    return failure();
+  }
+  SmallVector<Value>& mrcs = *maybeMrcs;
+
+  SmallVector<mod_arith::ModArithType> maInputBasisTys;
+  for (int i = 0; i < inputBasisTypes.size(); i++) {
+    mod_arith::ModArithType qi =
+        dyn_cast<mod_arith::ModArithType>(inputBasisTypes[i]);
```
j2kun: nit: same as above (and below), iterate directly over the `inputBasisTypes`, since the index `i` is only used to index into `inputBasisTypes` once.

j2kun (changes requested):
I think the architectural approach is good, though you still seem unsure about relying on ApplyCoefficientwise for the purpose of lowering polynomial to a cornami dialect.

My main question is: do you think we should try to get some intermediate layer of correctness checking for the algorithm before trying to lower all the way to LLVM and running it? We have some facilities for doing this (cf. ArithmeticDag and for a good example see its use in the Horner's method [1]) but it may need some extra work to make it suitable to use with RNS-typed values. In short, doing this would allow us to have the same code path be used to run the computation in a unit test, as well as to generate the MLIR during the lowering.

If you think it's worth it, then I'd say we can submit this PR without an end-to-end test, and follow up to replace the core computational pieces with an ArithmeticDag.

[1]: See https://github.com/google/heir/blob/main/lib/Utils/Polynomial/Horner.h for the implementation, https://github.com/google/heir/blob/main/lib/Utils/Polynomial/HornerTest.cpp for the unit tests, and https://github.com/google/heir/blob/7e114c39eb41fe75f2621a6701adba50d4d73e4e/lib/Transforms/LowerPolynomialEval/Patterns.cpp#L110-L116 for its use in the MLIR lowering.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: I'm torn. Having a nice create constructor is very helpful, but not being able to gracefully handle errors is rough. I had a test that applies invalid arguments (this doesn't allow even moduli in the bases), but that test now causes a segfault due to a missing block terminator, as you predicted. If I make this change, I will have to remove a test, which feels like the wrong solution.

I would love to make HEIR as robust as possible, and for it to produce meaningful errors on invalid input. I'm not sure how best way to resolve this conflict.

j2kun: In theory the verifier should be the guard here, and check robustly for the presence of the terminator. I'm guessing the verifier I wrote does not do that: https://github.com/google/heir/blob/7e114c39eb41fe75f2621a6701adba50d4d73e4e/lib/Dialect/Polynomial/IR/PolynomialOps.cpp#L602

`getTerminator` has an assert inside it. If that's indeed where it's segfaulting (`-c dbg` should show the assertion failure though?) You could check with `block.mightHaveTerminator()` and emit a nicer error message.

lib/Dialect/RNS/IR/RNSOps.cpp
crockeea: There is already an _enormous_ amount of parallelism available in the context of [hybrid] keyswitching: we split the input into multiple "states" (each of which can be basis-extended in parallel), and within each state, we can basis-extend each coefficient in parallel. Concretely, for a ring dimension of 2^16 with 40 moduli in the input basis and 2 key-switch primes, there are (2^16)*20 parallel basis conversions happening. Since Horner "is optimal in the sense that it minimizes the number of multiplications and additions required to evaluate an arbitrary polynomial", I suspect this is more important than adding additional parallelism.

That said, this is the kind of interesting tradeoff that (especially hardware) implementations might want to explore. In my head, I have this idea for a benchmarking suite built into HEIR that explore a high-dimensional space of algorithm combinations to look for the best combination on a particular HW backend.

lib/Dialect/RNS/IR/RNSOps.cpp
crockeea: Agree that the `ArrayAttr` is unnecessary; I've fixed that.

For using `ModArithAttr` vs `APInt`, I erred on the side of type safety: in `computeMixedRadixCoeffs`, I check whether the type of the `ModArithAttr` is what I expect. It's unfortunate that this could have a performance impact, but I would like to keep the type safety if possible (that's one of MLIR's strong points).

> Creating lots of unused attributes

I was thinking about this when I wrote the code and generated the constants once per keyswitch (so outside the `ApplyCoefficientwise` loop) rather than inside. We could go one level higher and make these part of the CKKSAttributes, which would further reduce the number of times they are computed. I think this would make a nice follow-up PR.

On a partially-related note: do we have some kind of benchmarking on individual passes to help identify when a particular pass is slow?

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: This mostly worked. One compromise: when `block.mightHaveTerminator()` fails in` verify`, I just silently fail instead of emitting another error message and assume that the `create` method threw a meaningful error. The failure test is now even cleaner than it was before.

lib/Dialect/RNS/IR/RNSOps.cpp
crockeea: (On that follow-up PR: computing roots of unity for NTTs is another place we should compute the roots _once_ for a parameter set instead of at every NTT)

crockeea:
Addressed comments. Remaining open items:
- `qInvProds` thread
- > though you still seem unsure about relying on ApplyCoefficientwise for the purpose of lowering polynomial to a cornami dialect.

  I am, but I think it's a fine direction to start with. Given that we haven't done any work on moving from HEIR to HW yet, I don't know exactly what the challenges will or won't be.
- > do you think we should try to get some intermediate layer of correctness checking for the algorithm before trying to lower all the way to LLVM and running it?

  I don't see much advantage to doing that; I was happy with the structure of the LLM tests for NTT.
- > In short, doing this would allow us to have the same code path be used to run the computation in a unit test, as well as to generate the MLIR during the lowering.
  
  I think your point here is that in the LLVM lowering, we're using the basis conversion code path _plus_ the ModArith->LLVM code path, and there could be bugs in the latter that affect correctness of the former. That's true, but I think using LLVM is the "easy button" for writing simple tests. If there's a bug in LLVM now, that could be annoying to find. But if someone introduces a bug later, this test would help catch it.
- > If you think it's worth it, then I'd say we can submit this PR without an end-to-end test, and follow up to replace the core computational pieces with an ArithmeticDag.

  My preferred path would be to add an LLVM test to this PR. Given how subtle this algorithm is (having implemented it multiple times and having made mistakes each time), having a test is important :) The open item is where to put the LLVM tests: cram it into the existing Polynomial/Conversions/heir_polynomial_to_llvm tests, or copy that infrastructure to RNS tests?

lib/Dialect/RNS/IR/RNSOps.cpp
j2kun: > do we have some kind of benchmarking on individual passes to help identify when a particular pass is slow

So far just `--mlir-timing` which prints a table of per-pass runtimes after finishing. Useful, but we don't have anything like performance regression testing to ensure changes don't make passes/pipelines slower.

j2kun: I think constructing ArrayAttrs is significantly more expensive than constructing the underlying scalar attributes, in particular since the ArrayAttr hashes as a combination of all its member hashes (unless we're using a dense form of the array attr which isn't implemented for mod_arith member attrs yet). So I'm happy to just omit the ArrayAttr for now.

j2kun:
> My preferred path would be to add an LLVM test to this PR.

Sounds good to me.

> where to put the LLVM tests

I think it's less infrastructure than it seems, so putting it in a new location is fine. Most of the hard work is done in `tests/llvm_runner:llvm_runner.bzl`, and so you can import that macro anywhere and use it as

```
llvm_runner_test(
    name = "target_name",
    heir_opt_flags = [
        "--emit-c-interface",
        // any other passes you need
        "--heir-polynomial-to-llvm",
    ],
    main_c_src = "test_harness.cc",
    mlir_src = "mlir_input.mlir",
)
```

https://github.com/google/heir/blob/main/tests/Dialect/Polynomial/Conversions/heir_polynomial_to_llvm/runner/lower_add_test.cc is a good, simple reference here.

crockeea:
I'll work on adding a test tomorrow. In the meantime, could you help me figure out why CI is failing, but the tests run fine on my local machine? See, e.g,. https://github.com/google/heir/actions/runs/25077913451/job/73475389405?pr=2893

crockeea:
> In the meantime, could you help me figure out why CI is failing, but the tests run fine on my local machine?

Apparently the issue is that CI uses the -c flag, and I introduced a bug that is only triggered with -c. Working on a fix.

j2kun:
> > In the meantime, could you help me figure out why CI is failing, but the tests run fine on my local machine?
> 
> Apparently the issue is that CI uses the -c flag, and I introduced a bug that is only triggered with -c. Working on a fix.

Let me know if there's something I can do to help debug.

crockeea:
Ready for second review:
- e8e4df6: Addresses feedback from the first review
- b7d0126: Following an offline discussion with Jeremy, we decided to add an `RNS::ConvertBasisOp`.
- c94a20a: This is the big change: instead of having the `LWEToPolynomial` conversion completely decompose `LWE::ConvertBasis`, it only lowers to a polynomial coefficient loop of `RNS::ConvertBasisOp`. Meanwhile, a new RNS transform (`--rns-lower-convert-basis`) is introduced to lower `RNS::ConvertBasisOp` to its ModArith implementation. Misc change: removed the `Failure` wrapper from the custom `ApplyCoefficientwiseOp` builder.
- 8984615: While creating a new transform, I realized there were some inconsistencies in other transforms. This commit simplifies the build files for Polynomial and CKKS transforms. I also found some unused imports and declarations in PolyMulToNTT.h, and fixed a deprecated-syntax bug in TargetUtils.cpp.
- d9039bc: More miscellaneous fixes: the polynomialToLLVM tests don't need to explicit use `--emit-c-interface` since that is already included in `--heir-polynomial-to-llvm`.
- 93be6bf: Added LLVM runner test infrastructure for testing RNS::ConvertBasis
- 5a7495e: In order to lower ConvertBasis to LLVM, we need `ModArithToArith` lowerings for `RNS::PackOp` and `RNS::ExtractResidueOp`.
- 7b475d7: `ModArith::EncapsulateOp` doesn't do modular reduction, so add explicit `ModArith::ReduceOp`. This fixes the ConvertBasis LLVM tests.

j2kun (approved):
Great stuff! Very clean

j2kun:
Squash the commits and let's get it in!

crockeea:
Squashed!

j2kun (approved):

j2kun:
One small note for future reference: this PR required an internal patch for a memory leak in the LLVM runner tests: each iteration of the loop was reallocating memory inside the function call without freeing it until after the end of the loop.

When running e2e llvm tests, I recommend adding `--config=asan` to the bazel invocation, and it will then run with instrumentation to catch these sorts of (admittedly minor) issues.

PR #2896, opened 2026-04-23
Add CHEDDAR GPU backend

AlexanderViand:
Part of the work to make Cheddar a first-class backend, extracted to its own PR because the diff was getting out of hand :) 

This adds the backend itself, i.e., the `cheddar` dialect, `--scheme-to-cheddar` passes, emitter, etc and a few basic tests (including some very simple "e2e" ones). It also adds Cheddar itself as an optional (default) off Bazel dependency. 

**Question: Do we want to run the e2e GPU tests on CI? If yes, what runner/etc should we be using?**

This PR does not yet enable HEIR to target more complex programs to Cheddar, because Cheddar has very strict ideas about the acceptable scales of things, and with just this PR, HEIR's mgmt layer still has no idea about the Rational-Rescale-ish "25-30" system Cheddar uses.


MODULE.bazel
```diff
@@ -64,6 +66,18 @@ use_repo(
     "llvm_zstd",
 )
 
+# CHEDDAR GPU FHE library (opt-in, requires CUDA)
+cheddar_extensions = use_extension("//bazel:extensions.bzl", "cheddar_deps")
```
j2kun, line 70: So there's a lot of stuff in this PR around adding the cheddar backend (bazel overlay, opt-in config, etc.) but as far as I can tell nothing tests it because there are no end-to-end tests in this PR.

Can we either include an end-to-end test in this PR, or else have this PR only include the MLIR-only parts? (dialect, transforms, lit tests, emitter).

In particular, when I pull this branch to my local machine and I try running `bazel build @cheddar`, I get

```bash
$ bazel build @cheddar  
ERROR: /usr/local/google/home/jkun/.cache/bazel/_bazel_jkun/886f3804b421b5442165b2f8eb57dad5/external/rules_cuda++toolchain+cuda/BUILD: no such target '@@rules_cuda++toolchain+cuda//:thrust': target 'thrust' not declared in package '' defined by /usr/local/google/home/jkun/.cache/bazel/_bazel_jkun/886f3804b421b5442165b2f8eb57dad5/external/rules_cuda++toolchain+cuda/BUILD
ERROR: /usr/local/google/home/jkun/.cache/bazel/_bazel_jkun/886f3804b421b5442165b2f8eb57dad5/external/+cheddar_deps+cheddar/BUILD.bazel:50:11: no such target '@@rules_cuda++toolchain+cuda//:thrust': target 'thrust' not declared in package '' defined by /usr/local/google/home/jkun/.cache/bazel/_bazel_jkun/886f3804b421b5442165b2f8eb57dad5/external/rules_cuda++toolchain+cuda/BUILD and referenced by '@@+cheddar_deps+cheddar//:cheddar'
ERROR: Analysis of target '@@+cheddar_deps+cheddar//:cheddar' failed; build aborted: Analysis failed
INFO: Elapsed time: 1.340s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
```

j2kun (changes requested):
I will review the IR parts after office hours, but one major request below

lib/Dialect/Cheddar/IR/CheddarAttributes.td
```diff
@@ -0,0 +1,34 @@
...
+
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/DialectBase.td"
+
+class Cheddar_Attribute<string attrName, string attrMnemonic>
+    : AttrDef<Cheddar_Dialect, attrName> {
+    let summary = "Attribute for CHEDDAR";
+    let description = [{
+        This attribute represents configuration values for CHEDDAR.
+    }];
```
j2kun:
```suggestion
```

nit: every `def` will override these fields.

lib/Dialect/Cheddar/IR/CheddarAttributes.td
```diff
@@ -0,0 +1,34 @@
...
+    let summary = "Path to a CHEDDAR parameter JSON file";
+    let description = [{
+        This attribute holds the path to a CHEDDAR parameter JSON file.
+
+        CHEDDAR parameters are loaded from pre-built JSON files that specify
+        ring dimension, prime chains, scale, and other CKKS parameters.
+    }];
+    let parameters = (ins
+      StringRefParameter<"path to JSON parameter file">:$path,
+      DefaultValuedParameter<"bool", "false", "use 64-bit word type">:$use64Bit
```
j2kun, line 26: This field appears to be unused. (In fact, this entire attribute appears to be unused)

lib/Dialect/Cheddar/IR/CheddarDialect.td
```diff
@@ -0,0 +1,25 @@
...
+def Cheddar_Dialect : Dialect {
+  let name = "cheddar";
+  let description = [{
+    The `cheddar` dialect is an exit dialect for generating C++ code against the
+    CHEDDAR GPU FHE library API.
+
+    CHEDDAR is a CKKS-only GPU-accelerated FHE library. It supports both 32-bit
+    and 64-bit word types, with 32-bit being the primary fast path on GPUs.
+
+    See https://github.com/scale-snu/cheddar-fhe
```
j2kun:
```suggestion
    See [the Cheddar GitHub repository](https://github.com/scale-snu/cheddar-fhe)
```

lib/Dialect/Cheddar/IR/CheddarOps.td
```diff
@@ -0,0 +1,497 @@
...
+  let arguments = (ins
+    Cheddar_Context:$ctx,
+    Cheddar_Ciphertext:$input,
+    Cheddar_Ciphertext:$addend,
+    Cheddar_EvalKey:$conjKey
+  );
+  let results = (outs Cheddar_Ciphertext:$output);
+}
+
+def Cheddar_MadUnsafeOp : Cheddar_Op<"mad_unsafe"> {
```
j2kun, line 424: Why "mad" and not "mac"?

lib/Dialect/Cheddar/IR/CheddarOps.h
```diff
@@ -0,0 +1,26 @@
...
+#include "mlir/include/mlir/IR/BuiltinOps.h"  // from @llvm-project
+#include "mlir/include/mlir/Interfaces/InferTypeOpInterface.h"  // from @llvm-project
+// IWYU pragma: end_keep
+
+namespace mlir::heir::cheddar {
+
+// `GetRotKeyOp` stores a static distance attribute, but dynamic
+// `ckks.rotate` lowering still needs a placeholder key op so the emitter can
+// trace back to the `UserInterface`. This sentinel distance marks that case.
+constexpr int64_t kDynamicRotationKeyDistanceSentinel = -1;
```
j2kun, line 19: Is there a reason why you don't want to allow the GetRotKeyOp to take both dynamic and static options as well?

j2kun:
> Question: Do we want to run the e2e GPU tests on CI? If yes, what runner/etc should we be using?

A quick look into this and it appears the Google org does not have any runners configured with a GPU. So this may be an obstacle and might force us to set up a Google-internal CI that runs on GCP and mirrors its results to GitHub.

lib/Pipelines/ArithmeticPipelineRegistration.cpp
```diff
@@ -525,6 +528,39 @@ BackendPipelineBuilder toLattigoPipelineBuilder() {
...
 }
 
+BackendPipelineBuilder toCheddarPipelineBuilder() {
+  return [=](OpPassManager& pm, const BackendOptions& options) {
+    // Convert CKKS to LWE
+    pm.addPass(ckks::createCKKSToLWE());
+
+    if (options.debug) {
+      llvm::errs() << "warning: CHEDDAR backend currently ignores "
+                      "--insert-debug-handler-calls\n";
```
j2kun, line 538: Is there a particular difficulty to allowing this to be supported by default? I'd expect in the worst case it requires additional copying of data to/from the GPU, but in a debug mode that seems OK.

lib/Pipelines/ArithmeticPipelineRegistration.h
```diff
@@ -134,6 +134,11 @@ struct BackendOptions : public PassPipelineOptions<BackendOptions> {
       llvm::cl::desc("Insert function calls to an externally-defined debug "
                      "function (cf. --lwe-add-debug-port)"),
       llvm::cl::init(false)};
+  PassOptions::Option<bool> fuseOps{
+      *this, "fuse-ops",
```
j2kun, line 138: maybe call the flag `cheddar-fuse-ops` to make it clear it's for cheddar only (who reads the description anyway?)

lib/Target/Cheddar/CheddarEmitter.cpp
```diff
@@ -0,0 +1,1781 @@
...
+    logDefaultScale = attr.getInt();
+  if (auto attr = moduleOp->getAttrOfType<IntegerAttr>("cheddar.logN"))
+    logN = attr.getInt();
+  if (auto attr = moduleOp->getAttrOfType<DenseI64ArrayAttr>("cheddar.Q"))
+    qPrimes.assign(attr.asArrayRef().begin(), attr.asArrayRef().end());
+  if (auto attr = moduleOp->getAttrOfType<DenseI64ArrayAttr>("cheddar.P"))
+    pPrimes.assign(attr.asArrayRef().begin(), attr.asArrayRef().end());
+
+  // Collect rotation distances used across the module
+  moduleOp->walk([&](Operation *op) {
```
j2kun, line 320: Isn't the RotationAnalysis supposed to handle this? You should be able to add the `rotation-analysis` pass to the pipeline and then read the rotation distances from the module attribute that pass attaches. It may require implementing the rotation interface on the cheddar ops, but it looks like you had already done that above.

lib/Target/Cheddar/CheddarEmitter.cpp
```diff
@@ -0,0 +1,1781 @@
...
+    }
+  });
+
+  for (Operation &op : moduleOp) {
+    if (failed(translate(op))) return failure();
+  }
+
+  // Emit __configure() if we have the required module-level params
+  if (logN > 0 && !qPrimes.empty()) {
+    os << "std::tuple<CtxPtr, UI> __configure() {\n";
```
j2kun, line 370: Is `__configure()` a cheddar thing? because elsewhere in the PR you have `configNameOs << op.getSymName() << "__configure";` and so here I'd expect a symbol name as well. I also don't see any emitter tests referring to this generated config code, so I'm guessing.

lib/Dialect/Cheddar/Transforms/ConfigureCryptoContext.td
j2kun, line 1: In other backends, this pass does the work of actually inserting the configuration function as IR (create a `func.func`). Is that not feasible here? Probably it is easier to do that here than to do it entirely in the emitter (especially since we hope to migrate to emitC for C++ codegen)

lib/Target/Cheddar/CheddarEmitter.cpp
```diff
@@ -0,0 +1,1781 @@
...
+
+LogicalResult CheddarEmitter::printOperation(AddPlainOp op) {
+  auto name = getName(op.getOutput());
+  auto ctName = getName(op.getCiphertext());
+  auto ptName = getName(op.getPlaintext());
+  os << "Ct " << name << ";\n";
+  // Fix up plaintext scale to exactly match ciphertext scale. This is needed
+  // because CHEDDAR's LinearTransform produces scales via floating-point
+  // arithmetic that may differ slightly from GetScale(level).
+  os << ptName << ".SetScale(" << ctName << ".GetScale());\n";
```
j2kun, line 842: sanity check: to me a plaintext as a concept already has the scaled applied (otherwise it cannot be integer-typed). So does this actively re-encode with a new scale? Or does it just set an internal scale member variable that does not change the underlying values?

Cf. https://github.com/scale-snu/cheddar-fhe/blob/8df8b26ce5411a68b68f0e7c2fb7e9a2f05f3e94/src/core/Container.cpp#L18-L20

lib/Target/Cheddar/CheddarEmitter.cpp
```diff
@@ -0,0 +1,1781 @@
...
+    os << "std::vector<Complex> " << name << ";\n";
+    os << getName(op.getEncoder()) << ".Decode(" << name << ", "
+       << getName(op.getPlaintext()) << ");\n";
+  }
+  return success();
+}
+
+LogicalResult CheddarEmitter::printOperation(EncryptOp op) {
+  auto name = getName(op.getCiphertext());
+  os << "Ct " << name << ";\n";
```
j2kun, line 786: Ideally all the places that emit "Ct" or "Pt" (or any other type) would reuse `convertType`.

lib/Dialect/Cheddar/IR/CheddarOps.td
```diff
@@ -0,0 +1,497 @@
...
+    Calls context->HRot(res, a, rot_key, distance).
+    Single fused GPU kernel for rotation.
+    Supports both static and dynamic shift distances.
+  }];
+  let arguments = (ins
+    Cheddar_Context:$ctx,
+    Cheddar_Ciphertext:$input,
+    Cheddar_EvalKey:$rotKey,
+    Optional<AnySignlessIntegerOrIndex>:$dynamic_shift,
+    OptionalAttr<Builtin_IntegerAttr>:$static_shift
```
j2kun, line 372: This may be a bit redundant. If you have the shift, and you just need to get from this op to the context, the pattern we used in the other backends is to look for the CryptoContext SSA value that is forced to be a function argument of the `func.func` this op is contained in, and then use that in this case to get the UserInterface and look up the rot key. Otherwise having the rot key already, and then using it only has a means to access the user interface to then look up a rotkey seems convoluted.

lib/Target/Cheddar/CheddarEmitter.cpp
```diff
@@ -0,0 +1,1781 @@
...
+  int64_t minNeg = slots;  // smallest negative-rotation index
+  for (auto idx : diagIndices) {
+    if (idx > slots / 2) {
+      hasWrapAround = true;
+      minNeg = std::min(minNeg, static_cast<int64_t>(idx));
+    }
+  }
+  if (hasWrapAround) preRotation = minNeg;
+
+  // Use the same BSGS computation as the rotation-index analysis so that
```
j2kun, line 1057: If it is the same computation and it is not part of RotationAnalysis, then it should probably at least be refactored to a separate function that can be shared by both call sites (and perhaps unit tested independently, the logic here seems nontrivial).

lib/Target/Cheddar/CheddarEmitter.cpp
```diff
@@ -0,0 +1,1781 @@
...
+  }
+  os.unindent();
+  os << "}\n";
+
+  // Construct LinearTransform. CHEDDAR's LT internally rescales (output at
+  // pt_level - 1), so the diagonal plaintext scale should match the OUTPUT
+  // level's scale, not the input level's scale.
+  os << "LinearTransform<word> " << ltName << "(" << ctxName << ", " << matName
+     << ", " << level << ", " << ctxName << "->param_.GetScale(" << level
+     << " - 1), " << bs << ", " << gs << ", " << preRotation << ");\n";
```
j2kun, line 1092: This makes me feel that perhaps there should be a Cheddar-specific scale analysis that knows about all these rules, because it seems a lot of important logic is being stuffed into the emitter.

But maybe I'm mistaken and we can't see the materialized scale values at compile time?

tests/Dialect/LWE/Conversions/lwe_to_cheddar/unsupported_encode_encoding.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: not heir-opt --lwe-to-cheddar %s 2>&1 | FileCheck %s
```
j2kun: Prefer `--verify-diagnostics` instead of piping stderr to stdout.
```suggestion
// RUN: not heir-opt --lwe-to-cheddar --verify-diagnostics %s
```

lib/Target/Cheddar/CheddarEmitter.cpp
j2kun, line 1: Seeing this emitter makes me think we should invest some time into converting these emitters to use the emitc dialect, so that the vast majority of the codegen details can be avoided. 😰

lib/Dialect/LWE/Conversions/LWEToCheddar/LWEToCheddar.cpp
```diff
@@ -0,0 +1,868 @@
...
+      current = cheddar::RescaleOp::create(rewriter, op.getLoc(), cheddarCtType,
+                                           ctx.value(), current);
+      --currentLevel;
+      auto encodedOne = cheddar::EncodeConstantOp::create(
+          rewriter, op.getLoc(), cheddar::ConstantType::get(getContext()),
+          encoder.value(), one, rewriter.getI64IntegerAttr(currentLevel),
+          rewriter.getI64IntegerAttr(*logDefaultScale));
+      current =
+          cheddar::MultConstOp::create(rewriter, op.getLoc(), cheddarCtType,
+                                       ctx.value(), current, encodedOne);
```
j2kun, line 483: A comment to explain why this needs a loop would be helpful.

lib/Dialect/LWE/Conversions/LWEToCheddar/LWEToCheddar.cpp
```diff
@@ -0,0 +1,868 @@
...
+      return op.emitOpError()
+             << "requires inverse-canonical CKKS plaintext encoding for "
+                "CHEDDAR lowering";
+    }
+    // In HEIR's main LWE encoding, the scaling factor stored on
+    // inverse_canonical_encoding is already log2-of-scale (additive for CKKS
+    // multiplies), so it maps directly onto CHEDDAR's logScale field.
+    int64_t logScale = invEncoding.getScalingFactor();
+
+    // lwe.rlwe_encode doesn't carry a plaintext level on main, so encode at
```
j2kun, line 536: It has a RingAttr doesn't it? Isn't that support to encode the level via its RNS coefficient type? We might not instantiate those properly at the moment, but I know for sure the level analysis properly determines the needed levels of plaintexts involved in Pt-Ct ops, cf. `encodeCleartextAsPlaintext`. If there is a gap here please find/file the right GH issue and link it here with a TODO

lib/Dialect/LWE/Conversions/LWEToCheddar/LWEToCheddar.cpp
```diff
@@ -0,0 +1,868 @@
...
+    if (!moduleIsCKKS(module)) {
+      module->emitOpError("CHEDDAR backend only supports CKKS scheme");
+      return signalPassFailure();
+    }
+
+    ConversionTarget target(*context);
+    target.addLegalDialect<cheddar::CheddarDialect>();
+    target.addIllegalDialect<ckks::CKKSDialect>();
+    target
+        .addIllegalOp<lwe::RLWEEncryptOp, lwe::RLWEDecryptOp, lwe::RLWEEncodeOp,
```
j2kun, line 768: Why not make the LWE dialect illegal wholesale? Cheddar has its own types, after all.

lib/Dialect/LWE/Conversions/LWEToCheddar/LWEToCheddar.cpp
```diff
@@ -0,0 +1,868 @@
...
+};
+
+static FailureOr<CheddarLevelReduceBucketKind> getCheddarLevelReduceBucketKind(
+    Operation* op, lwe::LWECiphertextType ctType) {
+  auto logDefaultScale = getCKKSLogDefaultScale(op);
+  if (failed(logDefaultScale)) {
+    return failure();
+  }
+
+  // Scaling factors on CKKS encodings are stored log-additively on main (a
```
j2kun, line 67: nit: comments talking about "on main" will become strange once they are merged into main.

j2kun (changes requested):

lib/Dialect/Cheddar/IR/CheddarOps.td
AlexanderViand: That's (for whatever reason) what Cheddar calls the function: https://github.com/scale-snu/cheddar-fhe/blob/main/include/core/Context.h#L377 and I figure making the exit dialects map closely to the library API is more important than making the naming consistent with other HEIR ops?

lib/Target/Cheddar/CheddarEmitter.cpp
AlexanderViand: Looking at emitC, it's come a long way since I last used it in HECO, and the explicit variable handling looks really nice! I'm kind of tempted to try rewriting the Cheddar emitter into EmitC.... 😅

MODULE.bazel
AlexanderViand: @missing tests: indeed, that's an oopsie on my side while extracting this out of the overall cheddar WIP branch. As we discussed in the office hour yesterday, I'll split this PR into three
1. Dialect
2. Emitter and FileCheck tests (I think I'll re-use this PR for this, to preserve the relevant review comments)
3. Bazel dependency & end-to-end tests

@ local machine: Does your local machine have an Nvidia GPU with CUDA/etc installed? Otherwise, I'm pretty sure that this is an expected failure, due to @rules_cuda being unable to find the required CUDA related stuff?  Maybe there's a way to set up the BUILD file to get a nicer/more helpful error message? Since you're directly asking bezel to build the cheddar target, enable_cheddar isn't "protecting" you here, I guess?

lib/Dialect/Cheddar/IR/CheddarOps.td
j2kun: Persisting what appears to be a typo...is just the right amount of whimsy. Maybe leave a comment for future confused people.

MODULE.bazel
j2kun: From https://bazel-contrib.github.io/rules_cuda/0.3.0/

> For hermetic toolchains, the rules handle toolchain configuration and library downloading automatically. See [cuda.redist_json integration test](https://bazel-contrib.github.io/rules_cuda/0.3.0/tests/integration/toolchain_redist_json) for a comprehensible example.
>
> For locally installed toolchains, [_detect_local_cuda_toolkit](https://github.com/bazel-contrib/rules_cuda/blob/ce98e4ae5c/cuda/private/repositories.bzl#L30-L45) and [detect_clang](https://github.com/bazel-contrib/rules_cuda/blob/ce98e4ae5c/cuda/private/repositories.bzl#L215-L256) determines how they are detected.
>
> Either situation depends on cc toolchain availability, so you must also ensure the cc compiler is properly configured. On Windows, this means that you will also need to set the environment variable BAZEL_VC properly.

I would like to try this again on a machine that has an NVIDIA GPU as well as the hermetic LLVM toolchain. In theory, with the hermetic toolchain we should be able to build the targets even if the machine does not have a GPU or CUDA installed. But yes, we should split that into a standalone PR.

crockeea:
I eagerly await this :)

AlexanderViand:
> I eagerly await this :)

You can get a sneak peek at it (fully working) here: https://github.com/google/heir/compare/main...belfortlabs:heir:ml-cheddar

but I don't have an ETA yet on when it'll all be cleaned up and upstreamed piece-by-piece, sorry!

AlexanderViand:
Closing this old PR in favour of #3334 and its follow-ups

PR #2902, opened 2026-04-25
add defensive guards for eval/coeff forms

toprakmurat:
Fixes #2849 

This PR adds strict `getForm()` checks to all polynomial-to-mod-arith lowerings that assume a specific polynomial form (`COEFF` or `EVAL`).

Previously, operations that only make mathematical sense in `COEFF` form (e.g., `FromTensorOp`, `LeadingTermOp`) would silently miscompile if fed an `EVAL` form polynomial. These conversion patterns now return a `notifyMatchFailure` if the input form is incorrect.

Changes made:

Added `Form::COEFF` checks to: `ConvertFromTensor`, `ConvertConstant`, `ConvertMonomial`, `ConvertMonicMonomialMul`, `ConvertLeadingTerm`, `ConvertApplyCoefficientwise`, and `ConvertNTT`.
Added a `Form::EVAL` check to: `ConvertINTT`.

(Note: I think that the first part of #2849 was previously implemented via the `ConvertMulEvalForm` struct, so no additional changes were needed for that requirement).

### DIFF
--- a/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
+++ b/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
@@ -219,6 +219,11 @@ struct ConvertFromTensor : public OpConversionPattern<FromTensorOp> {
     }
     auto typeInfo = res.value();
 
+    if (typeInfo.polynomialType.getForm() != Form::COEFF) {
+      return rewriter.notifyMatchFailure(
+          op, "FromTensor lowering requires COEFF form");
+    }
+
     auto resultShape = typeInfo.tensorType.getShape()[0];
     auto resultEltTy = typeInfo.tensorType.getElementType();
     auto inputTensorTy = op.getInput().getType();
@@ -275,6 +280,11 @@ struct ConvertConstant : public OpConversionPattern<ConstantOp> {
           op, "failed to construct common conversion info");
 
     auto typeInfo = res.value();
+
+    if (typeInfo.polynomialType.getForm() != Form::COEFF) {
+      return rewriter.notifyMatchFailure(
+          op, "Constant lowering requires COEFF form");
+    }
 
     auto attr = dyn_cast<TypedIntPolynomialAttr>(op.getValue());
     if (!attr)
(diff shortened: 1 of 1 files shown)


lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -276,6 +281,11 @@ struct ConvertConstant : public OpConversionPattern<ConstantOp> {
 
     auto typeInfo = res.value();
 
+    if (typeInfo.polynomialType.getForm() != Form::COEFF) {
```
crockeea, line 284: Constant ops should be allowed to output eval-form polys; see https://github.com/google/heir/blob/main/lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp#L526 and https://github.com/google/heir/pull/2751. The source of truth here is https://github.com/google/heir/blob/main/lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp#L59.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -362,6 +372,11 @@ struct ConvertMonomial : public OpConversionPattern<MonomialOp> {
           op, "failed to construct common conversion info");
     auto typeInfo = res.value();
 
+    if (typeInfo.polynomialType.getForm() != Form::COEFF) {
```
crockeea, line 375: Ditto here; see https://github.com/google/heir/blob/main/lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp#L59. As a CONST op, this output a polynomial in either form.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -600,6 +620,11 @@ struct ConvertLeadingTerm : public OpConversionPattern<LeadingTermOp> {
...
```
crockeea, line 623: LeadingTerm op doesn't output a poly, does it?

crockeea (commented):
I think a COEFF constraint on ConvertBasisOp is missing.

AlexanderViand:
Closing in favor of #3056 due to lack of updates.

crockeea:
All three of my comments were incorrect, in retrospect.

toprakmurat:
Sorry for the late reply. I've been busy with college work and it slipped my mind after a while.

PR #2918, opened 2026-05-04
fix trigger of wheels workflow to use push to tag

j2kun:
Warning: AI-generated.


asraa (approved):
did you want to move the monitoring step to the wheels workflow? or also alert if the release failed?

j2kun:
I think maintainers get an alert when the workflow fails already, and there is little value in having an issue filed in that event.

Issue #2900, opened 2026-04-24
Support 1D convolution in HEIR


j2kun:
This ticket is to support `linalg.conv_1d` in the HEIR compiler. This came up because Marc from Belfort was trying to lower a torch model that used 1d convolution. Together we did a deep dive on the layout system and he will contribute the implementation.

The basic idea is to reduce conv1d to Halevi-Shoup matvec with sparse diagonals, similar to the conv2d work.

The outline of supporting it:

- Add a filter-to-undiagonalized-matrix layout relation in lib/Utils/Layout, and then compose it with the matrix diagonalization relation
- Add a new function in lib/Kernel/KernelImplementation.h that constructs the arithmetic dag for the conv1d kernel, and uses `implementHaleviShoup` as a subroutine. This can be used to run the resulting kernel on test data by instantiating the template type for the ArithmeticDag with a concrete type (std::vector<int>) and using EvalVisitor in a unit test.
- Add a new RewritePattern in ConvertToCiphertextSemantics.cpp (similar to the one for [conv2d](https://github.com/google/heir/blob/d0c64c24bff045d4182103dc103a0335dad0aab8/lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp#L811)) that lowers conv1d using the arithmetic dag, and guards this by the kernel attribute (KernelName::MatvecDiagonal).
- Add support for conv1d in `layout-propagation` (I think this should be easy because the ciphertext input does not change its layout after going through this kernel).
- [optional]: Implement LayoutConversionHoistableOpInterface for linalg.conv_1d, which allows layout conversions to be hoistable through this op in `layout-optimization`.

j2kun:
CC @mdgrs @AlexanderViand

j2kun:
I was looking at the linalg dialect today and noticed that `padding` is not an attribute on most (all?) of the convolution ops. Since we talked about padding in the draft of #2899, I am wondering how @AlexanderViand or @mdgrs saw padding materialize in the IR?

mdgrs:
I also saw 0 instance of padding in the convolution operations for the linalg dialect. My superficial understanding is that one would use a PadOp to add actual padding to the tensor in one of the passes. See https://github.com/llvm/llvm-project/blob/b48aa05f399d16faf90475cfb8131ae8693d8f97/mlir/lib/Dialect/Linalg/TransformOps/LinalgTransformOps.cpp#L2090

[...]

PR #2919, opened 2026-05-05
Add support for conv1D, second step

mdgrs:
Add support for lowering linalg.conv_1d

I wanted to keep things small, but when digging around the lowering from pytorch, I saw other variants as well (one of linalg.conv_1d_ncw_fcw or linalg.conv_1d_nwc_wcf ), so I expect to have a follow up PR adding support for these as well, much like right now there is conv_2d and conv_2d_nchw_fchw .

This is related to #2900 , but I would wait for this follow up work before closing the issue

Regarding the last point of that issue:
```
[optional]: Implement LayoutConversionHoistableOpInterface for linalg.conv_1d, which allows layout conversions to be hoistable through this op in layout-optimization.
```

I started thinking about this, but I am not too sure how to proceed. I understand that a permutation of the packed data vector will induce a permutation of the columns of the filter matrix. This should be possible. However, I think we can have cases where this column permuted filter matrix cannot be lowered to a conv_1d operator (for example if you rotate the data vector).

One solution would be to completely replace the conv1D op with a MatVec op and use the current machinery, but I wanted to check with you, make sure it doesn't break anything else. I wrote a possible implementation in the last commit, but didn't find a good place to test this. If you want to keep it, I would appreciate a suggestion :)

### DIFF
--- a/lib/Kernel/Kernel.cpp
+++ b/lib/Kernel/Kernel.cpp
@@ -25,10 +25,10 @@ static std::unordered_map<KernelName, std::vector<std::string>>
     correspondingOp = {
         {KernelName::MatvecNaive, {"linalg.matvec"}},
         {KernelName::MatvecDiagonal,
-         {"linalg.matvec", "linalg.conv_2d_nchw_fchw"}},
+         {"linalg.matvec", "linalg.conv_2d_nchw_fchw", "linalg.conv_1d"}},
         {KernelName::VecmatDiagonal, {"linalg.vecmat"}},
         {KernelName::MatmulDiagonal, {"linalg.matmul"}},
-        {KernelName::MatmulDiagonal, {"linalg.conv2d"}},
+        {KernelName::MatmulDiagonal, {"linalg.conv_2d"}},
         {KernelName::MatmulBicyclic, {"linalg.matmul"}},
         {KernelName::Dot, {"linalg.dot"}},
 };
--- a/lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
+++ b/lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
@@ -878,6 +878,136 @@ struct ConvertLinalgMatvecLayout
   bool unrollKernels;
 };
 
+struct ConvertLinalgConv1D
+    : public ContextAwareOpConversionPattern<linalg::Conv1DOp> {
+ public:
+  using ContextAwareOpConversionPattern<
+      linalg::Conv1DOp>::ContextAwareOpConversionPattern;
+
+  ConvertLinalgConv1D(
+      const ContextAwareTypeConverter& contextAwareTypeConverter,
+      MLIRContext* context, bool unrollKernels = true)
+      : ContextAwareOpConversionPattern(contextAwareTypeConverter, context,
+                                        /*benefit=*/10),
+        unrollKernels(unrollKernels) {}
+
+  LayoutAttr getLayoutAttr(Value value) const {
+    auto layoutLookup = getTypeConverter()->getContextualAttr(value);
+    if (failed(layoutLookup)) {
+      return nullptr;
+    }
+    return dyn_cast<LayoutAttr>(layoutLookup.value());
+  }
+
+  bool supportsExpandedHaleviShoup(linalg::Conv1DOp op,
+                                   OpAdaptor adaptor) const {
+    Value filter = adaptor.getInputs().back();
+    auto materializedFilterType = cast<RankedTensorType>(filter.getType());
+
+    // If one of these dimensions is not a power of two, then we can't do
+    // the Halevi-Shoup or Squat Packing Matrix Multiplication conversion.
+    auto dimensions = materializedFilterType.getShape();
+    int64_t numRows = dimensions[0];
+    int64_t numCols = dimensions[1];
+    bool isPowerOfTwoDims = isPowerOfTwo(numRows) && isPowerOfTwo(numCols);
+
+    auto kernelAttr = op->getAttrOfType<secret::KernelAttr>(
+        secret::SecretDialect::kKernelAttrName);
+    bool isConv1dAsMatvec =
+        kernelAttr && kernelAttr.getName() == KernelName::MatvecDiagonal;
+
+    LLVM_DEBUG(llvm::dbgs()
+               << "supports expanded conv1d as matvec with halevi-shoup: "
+               << "isPowerOfTwoDims=" << isPowerOfTwoDims
+               << " isConv1dAsMatvec=" << isConv1dAsMatvec << "\n");
+
+    return isPowerOfTwoDims && isConv1dAsMatvec;
+  }
+
+  void haleviShoupKernel(
+      linalg::Conv1DOp op, OpAdaptor adaptor,
+      ContextAwareConversionPatternRewriter& rewriter) const {
+    LLVM_DEBUG(llvm::dbgs()
+               << "Converting linalg.conv1d op with halevi shoup kernel: " << op
+               << "\n");
+
+    TypedValue<RankedTensorType> data =
+        cast<TypedValue<RankedTensorType>>(adaptor.getInputs()[0]);
+    SSAValue vectorLeaf(data);
+    TypedValue<RankedTensorType> filter =
+        cast<TypedValue<RankedTensorType>>(adaptor.getInputs()[1]);
+    SSAValue matrixLeaf(filter);
+
+    // The original matrix shape is the shape of the expanded filter before
+    // diagonalization.
+    RankedTensorType expandedMatrixType = get1dConvFilterExpandedType(
+        cast<RankedTensorType>(op.getInputs()[1].getType()),
+        cast<RankedTensorType>(op.getInputs()[0].getType()), /*stride=*/1,
+        /*padding=*/0);
+
+    // Collect any zero diagonals of the filter matrix.
+    LayoutAttr filterLayout = getLayoutAttr(adaptor.getInputs()[1]);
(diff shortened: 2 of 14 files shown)


j2kun:
> However, I think we can have cases where this column permuted filter matrix cannot be lowered to a conv_1d operator (for example if you rotate the data vector).

I don't quite follow this part. If the data vector is rotated, then the kernel for the conv_1d is unchanged (it still reduces to a matvec), so long as the filter matrix is packed differently (and as you mentioned, the different packing would require rotating the columns of the expanded filter matrix).

mdgrs:
> > However, I think we can have cases where this column permuted filter matrix cannot be lowered to a conv_1d operator (for example if you rotate the data vector).
> 
> I don't quite follow this part. If the data vector is rotated, then the kernel for the conv_1d is unchanged (it still reduces to a matvec), so long as the filter matrix is packed differently (and as you mentioned, the different packing would require rotating the columns of the expanded filter matrix).

Sorry for not being very clear. What I mean is that this new MatVec cannot be written as a Conv1D (I misused lowered). As an example:

Data vector: [1,2,3,4,5]
Filter: [0,1,2]
Resulting vector: [8, 11, 14]

For the new datavector [5,1,2,3,4], I cannot write this as a Conv1D([0,1,2], [5,1,2,3,4])   (or I think any other kernel). I.e. the sliding of the kernel does fit with the rotation of the data vector. I can however write it as a MatVec and forget it ever came from a Conv1D

j2kun:
> Sorry for not being very clear. What I mean is that this new MatVec cannot be written as a Conv1D (I misused lowered). As an example:
> 
> Data vector: [1,2,3,4,5]
> Filter: [0,1,2]
> Resulting vector: [8, 11, 14]
> 
> For the new datavector [5,1,2,3,4], I cannot write this as a Conv1D([0,1,2], [5,1,2,3,4])   (or I think any other kernel). I.e. the sliding of the kernel does fit with the rotation of the data vector. I can however write it as a MatVec and forget it ever came from a Conv1D

It should not be necessary to do this. The operation is still semantically the original conv1d, and all that is changing is the packed layout of the data vector. With respect to that new packing, you only need to specify how the layout of the other arguments (the filter) changes to accommodate, which as you mentioned is applying the same permutation to the columns of the expanded filter matrix.

lib/Kernel/Kernel.cpp
```diff
@@ -25,9 +25,10 @@ static std::unordered_map<KernelName, std::vector<std::string>>
     correspondingOp = {
         {KernelName::MatvecNaive, {"linalg.matvec"}},
         {KernelName::MatvecDiagonal,
-         {"linalg.matvec", "linalg.conv_2d_nchw_fchw"}},
+         {"linalg.matvec", "linalg.conv_2d_nchw_fchw", "linalg.conv_1d"}},
         {KernelName::VecmatDiagonal, {"linalg.vecmat"}},
         {KernelName::MatmulDiagonal, {"linalg.matmul"}},
+        // FIXME is there a _ missing in linalg.conv_2d
         {KernelName::MatmulDiagonal, {"linalg.conv2d"}},
```
j2kun: @asraa isn't it missing a `_` ?

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
```diff
@@ -592,6 +595,81 @@ LogicalResult LayoutPropagation::visitOperation(MatvecOp op) {
...
+  }
+
+  MLIRContext* ctx = &getContext();
+  mlir::IRRewriter builder(ctx);
+
+  // TODO(#1597): a layout optimizer should really be selecting the
+  // layout instead of this pass.
+  LayoutAttr dataLayout = assignedLayouts.at(data);
+  // I should remove this since vector layout is always row major!
+  // FIXME
```
j2kun: I agree but I think it's acceptable to be defensive and leave this branch in

tests/Examples/openfhe/ckks/conv_1d/conv_1d_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+TEST(Conv1DTest, RunTest) {
+  auto cryptoContext = conv_1d__generate_crypto_context();
+  auto keyPair = cryptoContext->KeyGen();
+  auto publicKey = keyPair.publicKey;
+  auto secretKey = keyPair.secretKey;
+  cryptoContext = conv_1d__configure_crypto_context(cryptoContext, secretKey);
+
+  // ct is a length 8 input vector
+  std::vector<float> m(8, 0.1f);
```
j2kun: It would be slightly more robust to have non-uniform values for the input and filter.

tests/Transforms/layout_propagation/conv1d.mlir
```diff
@@ -0,0 +1,26 @@
...
+// CHECK: #kernel = #secret.kernel<name = "MatvecDiagonal", force = false>
+// CHECK: @conv1d
+// CHECK-SAME: %[[arg0:.*]]: !secret.secret<tensor<5xf32>> {tensor_ext.layout = [[rm_layout:.*]]}) ->
+func.func @conv1d(%arg0: !secret.secret<tensor<5xf32>>) -> !secret.secret<tensor<3xf32>> {
+  %cst = arith.constant dense<0.000000e+00> : tensor<3xf32>
+  // CHECK: %[[out:.*]] = arith.constant dense<0.00
+  // CHECK: %[[filter:.*]] = arith.constant
+  // CHECK-SAME: tensor<3xf32>
+
+  // Assign a layout to the matrix and bias
```
j2kun:
```suggestion
  // Assign a layout to the filter and input
```

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -808,6 +808,141 @@ struct ConvertLinalgMatvecLayout
...
+    // the Halevi-Shoup or Squat Packing Matrix Multiplication conversion.
+    auto dimensions = materializedFilterType.getShape();
+    int64_t numRows = dimensions[0];
+    int64_t numCols = dimensions[1];
+    bool isPowerOfTwoDims = isPowerOfTwo(numRows) && isPowerOfTwo(numCols);
+
+    // TODO(#1578): If the matrix has more rows than columns, what kernel
+    // should be used?
+    bool isMatrixCompatible = numRows <= numCols;
+
```
j2kun: https://github.com/google/heir/pull/2920 removes the need for this guard.

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -808,6 +808,141 @@ struct ConvertLinalgMatvecLayout
...
+    SSAValue matrixLeaf(filter);
+
+    // The original matrix shape is the shape of the expanded filter before
+    // diagonalization.
+    RankedTensorType expandedMatrixType = get1dConvFilterExpandedType(
+        cast<RankedTensorType>(op.getInputs()[1].getType()),
+        cast<RankedTensorType>(op.getInputs()[0].getType()), /*stride=*/1,
+        /*padding=*/0);
+
+    // Collect any zero diagonals of the filter matrix.
```
j2kun, line 948: I was thinking about this offhand, when the stride is > 1, the data values will be skewed in the expanded filter matrix in such a way that will mess up the sparsity of this kernel.

e.g. a stride of 2 might produce something like

```
1 1 1 0 0 0 0 0 0 ...
0 0 1 1 1 0 0 0 0 ...
0 0 0 0 1 1 1 0 0...
```

Which has many more diagonals than the corresponding stride 1 kernel. I guess that is compensated for by having fewer total multiplications, but still I'd expect there is a more optimal kernel hidden somewhere here. Nothing immediate for this PR, but something to keep in mind if your application of interest has a larger stride.

j2kun (changes requested):
This looks good, modulo a few nitpicks and the currently failing build.

mdgrs:
> It should not be necessary to do this. The operation is still semantically the original conv1d, and all that is changing is the packed layout of the data vector. With respect to that new packing, you only need to specify how the layout of the other arguments (the filter) changes to accommodate, which as you mentioned is applying the same permutation to the columns of the expanded filter matrix.

Ok yes, that makes sense. Since I am using MatvecDiagonal when converting to ciphertext semantics ( here https://github.com/google/heir/pull/2919/changes#diff-fefcb9de49a7ea688cc6c183d7a0d10c5f51d971df655f0fe30e925530d3a67fR850 ) , I changed the hoisting to also use MatvecDiagonal instead of MatvecNaive (I guess it would break otherwise, because it would be the wrong kernel?)

lib/Kernel/Kernel.cpp
asraa: !!! I cna't tell you at all why the e2e tests for linalg.conv_2d https://github.com/google/heir/blob/e83b861ec71e29c69d2b1a9614320a11024c8cdd/tests/Examples/openfhe/ckks/conv_2d/conv_2d.mlir#L6 are passing... we indeed annotate it with MatmulDiagonal... https://github.com/google/heir/blob/e83b861ec71e29c69d2b1a9614320a11024c8cdd/tests/Transforms/layout_propagation/conv2d.mlir#L18

i'll run through mlir print-ir-after-all very confusing

mdgrs: The only place I see `isSupportedKernel` being used is when trying to apply the hoisting:

https://github.com/google/heir/blob/69a6af0241a8ecd4dd7bbbfcbea3797cc077e333/lib/Transforms/LayoutOptimization/LayoutOptimization.cpp#L148

j2kun (approved):
LGTM!

PR #2920, opened 2026-05-05
Support linalg.matvec matrices with more rows than cols 

### LINKED ISSUE #1578: What kernel is appropriate when a matrix has more rows than columns?
- Halevi-Shoup is for square matvec
- "Squat" ([Juvekar-Vaikuntanathan-Chandrakasan 2018](https://arxiv.org/abs/1801.05507)) packing is when the matrix has more columns than rows.

When there are more rows than columns, what should we do? (making this issue as a TODO for https://github.com/google/heir/pull/1552)

j2kun:
Turns out, the existing matvec kernels works on "tall and skinny" matrices as-is. The wraparound logic we added for "squat" matrices works verbatim, and only superficial changes are required to remove the guards.

Fixes https://github.com/google/heir/issues/1578

Assisted by Jetski

### DIFF
--- a/lib/Kernel/KernelImplementationTest.cpp
+++ b/lib/Kernel/KernelImplementationTest.cpp
@@ -85,6 +85,103 @@ TEST_P(KernelImplementationTest, HaleviShoup3x5) {
   EXPECT_EQ(std::vector<int>(actual.begin(), actual.begin() + 3), expected);
 }
 
+TEST_P(KernelImplementationTest, HaleviShoup4x2) {
+  // Original matrix (4x2):
+  // [0, 1]
+  // [2, 3]
+  // [4, 5]
+  // [6, 7]
+  //
+  // Pre-packed diagonally:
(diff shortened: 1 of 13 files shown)


lib/Kernel/KernelImplementationTest.cpp
```diff
@@ -85,6 +85,69 @@ TEST_P(KernelImplementationTest, HaleviShoup3x5) {
   EXPECT_EQ(std::vector<int>(actual.begin(), actual.begin() + 3), expected);
 }
 
+TEST_P(KernelImplementationTest, HaleviShoup4x2) {
```
asraa, line 88: can you add a non power of two case here? like 5x3?

asraa (approved):

lib/Kernel/KernelImplementationTest.cpp
j2kun: Done

PR #2924, opened 2026-05-06
Support linalg.dot and zero-dimensional tensors in HEIR

j2kun:
This PR adds support for `linalg.dot` and zero-dimensional tensors into the default layout engine.

### DIFF
--- a/lib/Kernel/KernelName.h
+++ b/lib/Kernel/KernelName.h
@@ -30,6 +30,9 @@ enum KernelName : int {
 
   // Ciphertext-ciphertext matmul using the bicyclic packing method.
   MatmulBicyclic,
+
+  // Product and sum of two vectors, using a log2 rotate-and-reduce approach.
+  Dot,
 };
 
 std::ostream& operator<<(std::ostream& os, const KernelName& k);
--- a/lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
+++ b/lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
@@ -62,6 +62,7 @@ namespace heir {
 
 using linalg::Conv2DNchwFchwOp;
 using linalg::Conv2DOp;
+using linalg::DotOp;
 using linalg::MatmulOp;
 using linalg::MatvecOp;
 using linalg::ReduceOp;
@@ -146,6 +147,7 @@ struct LayoutPropagation : impl::LayoutPropagationBase<LayoutPropagation> {
   LogicalResult visitOperation(VecmatOp op);
   LogicalResult visitOperation(MatvecOp op);
   LogicalResult visitOperation(MatmulOp op);
+  LogicalResult visitOperation(DotOp op);
   LogicalResult visitOperation(YieldOp op);
   LogicalResult visitOperation(affine::AffineForOp op);
   LogicalResult visitOperation(func::FuncOp op);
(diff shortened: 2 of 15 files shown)


j2kun:
@aymane-eljerari would you be willing to review this?

lib/Kernel/KernelName.h
```diff
@@ -30,6 +30,9 @@ enum KernelName : int {
 
   // Ciphertext-ciphertext matmul using the bicyclic packing method.
   MatmulBicyclic,
+
+  // Reduction of a tensor to a scalar using rotations.
```
mdgrs: Why does it say single tensor when the dot product involves two tensors? LayoutOptimization also only uses the first operand so I feel like I am missing something. Can you please explain the discrepancy?

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
```diff
@@ -1234,6 +1276,48 @@ void LayoutPropagation::rectifyIncompatibleOperandLayouts(Operation* op) {
...
 
+void LayoutPropagation::rectifyIncompatibleOperandLayouts(DotOp op) {
+  mlir::IRRewriter builder(&getContext());
+  builder.setInsertionPoint(op);
+
+  LayoutAttr lhsLayout = assignedLayouts.at(op.getOperand(0));
+  LayoutAttr rhsLayout = assignedLayouts.at(op.getOperand(1));
+
+  if (lhsLayout != rhsLayout) {
+    // Convert RHS to match LHS layout
```
mdgrs, line 1286: Is there a mechanism to check if it is more advantageous to align the LHS to the RHS or the RHS to the LHS?

mdgrs (approved):
Just a few points to clarify for my understanding, but otherwise LGTM

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
aymane-eljerari: Also curious how this would work. Not sure if this correct, but I assume you need some information about all operands in the operation chain to identify which layout reassignments are more efficient in that context, as well information from the cost model.

aymane-eljerari:
Same comment from the discord thread. The FFN lowering shown in the link below encounters a case where a rank 2 scalar operand is assigned a 0 domain layout. 

https://gist.github.com/aymane-eljerari/1a21d91db9ef4dab0bce24da51cd75aa?permalink_comment_id=6139047#gistcomment-6139047

lib/Transforms/LayoutPropagation/LayoutPropagation.cpp
j2kun: Locally we could do this (e.g., just look at the cost of the layout conversion in both directions and pick the cheapest one), but obviously the more interesting aspect is how that choice propagates further through the program, which is not possible with the default layout optimizer. More to the point, it's definitely out of scope of the `layout-propagation` pass, which is only meant to give a first-pass over the IR to ensure the layouts are all correct and consistent before moving on to the optimizer.

lib/Kernel/KernelName.h
j2kun: It was simply misleading. Fixed!

j2kun: Oh I see what you mean, you're right the cost estimator in layout-optimization should have an extra multiplication in its kernel

j2kun: I was trying to avoid having a new kernel function for the dot product since it's "just" a mul + the existing kernel, but after this it is just better to have one kernel function. Added this.

Issue #2932, opened 2026-05-08
Add codegen for Jaxite CKKS backend


j2kun:
This ticket is to track progress toward upstreaming @Zohaib58's HEIR codegen for CROSS (aka JaxiteWord aka jaxite_ckks).

PR #2933, opened 2026-05-08
Add shell of parameter selection pass for JaxiteWord

j2kun:
Part of #2932, this PR adds a shell parameter selection pass (really a parameter "refinement" pass) that converts the existing CKKS parameters into parameters that are compatible with jaxite_word. In this case, the TPU backend requires <= 30-bit primes, and has a handful of other parameters that should be set.

FIXME: this PR still includes some dummy values for parameters that, I believe, can be hard-coded in the pass for now.


j2kun:
CC @Zohaib58  please review this PR and let me know what values to use for the other currently-dummy parameters.

j2kun:
Note to self:

- r, c: TPU rectangular reshape
- dnum: key switch number of partitions
- composite_degree: "num rescales" (1 for single rescaling, 2 for double-rescaling)
- batch (?)

PR #2938, opened 2026-05-12
contributing: add note about meeting a maintainer

j2kun:
We had been considering and discussing adding a requirement for contributors to meet a maintainer. This PR makes that official.

### DIFF
--- a/docs/content/en/docs/Development/ai_policy.md
+++ b/docs/content/en/docs/Development/ai_policy.md
@@ -66,6 +66,20 @@ squanders the learning opportunity and doesn't add much value to the project.
 **New contributors using AI tools to fix issues labelled as "good first issues"
 is forbidden**.
 
+Finally, new contributors not recognized by the maintainers should first meet a
+project maintainer in a non-textual format before their first contribution will
+be accepted. This includes:
+
+- Meeting a maintainer in person.
+- Attending office hours or a working group meeting (cf. our
+  [community calendar](/community/)).
+- Scheduling a video call with a maintainer (reach out via
+  [Discord](/community/) or email).
+
+New PRs from unrecognized contributors that appear to be AI-generated will be
+closed with a reference to this policy. New issues or issue comments from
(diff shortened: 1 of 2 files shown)


docs/content/en/docs/Development/ai_policy.md
```diff
@@ -66,6 +66,19 @@ squanders the learning opportunity and doesn't add much value to the project.
 **New contributors using AI tools to fix issues labelled as "good first issues"
 is forbidden**.
 
+Finally, new contributors must first meet a project maintainer in a non-textual
+format before their first contribution will be accepted. This includes:
```
asraa: I might push back a little on the "must" - I think the nice part about github and open source is welcoming contributions from anywhere; for e.g. would we require this from someone who regularly chats online with us or is the author of a paper we're porting to HEIR? I think the intent here is to understand the contribution rationale and understand the intent of the contributor (i.e. not just AI contributions to gain internet karma). Maybe we can change must to something more like strongly suggest? Or add the reasoning that we expect some engagement before a first contribution.
Below it says "unrecognized contributors that appear to be AI-generated" - I think that's totally valid.

j2kun: My main goal is to have a means of provenance of identity. So maybe, "if the maintainers don't know who you are, then you must meet a maintainer first. Maintainers may waive this for contributors with a strong prior connection to the community via academic publications or related open source work." 

I.e., I'd be fine with the author of an FHE paper reaching out over email. But such a person would almost certainly quickly come to a video call to discuss what they want to do.

asraa: Yeah, exactly - I understand the intent is to have a lightweight "are you a real human that I'll be interacting with" - I guess I'm just pushing back on *requiring* face to face time

j2kun: Updated.

asraa (approved):

Issue #2923, opened 2026-05-06
Improve torch operator coverage (as exported by torch-mlir)


j2kun:
In a separate repository https://github.com/j2kun/heir_torch_coverage, I created a set of torch-mlir export scripts, one for each torch operator.

For Conv2d for example, the [operators/conv2d](https://github.com/j2kun/heir_torch_coverage/tree/main/operators/conv2d) contains a single-op torch model with garbage weights, exported to MLIR and annotated with secret before being run through `heir-opt` (2026-05-01 release). If it succeeds, the resulting openfhe MLIR is saved, otherwise an erorr log is saved.

I let AI decide on what "representative" set of torch operators should be included in the list (there are over 2k torch operators), so if new ones are needed I can add them.

So this issue exists to track progress on supporting these operators in HEIR. Specifically, if the relevant torch operator compiles to linalg that HEIR can handle _at all_ (even if inefficiently) then it is considered done for this issue.

I also created a new label [torch-support](https://github.com/google/heir/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22torch%20support%22) for tickets related to this.

Comment https://github.com/google/heir/issues/2923#issuecomment-4391907478 has a regularly-updated table containing results of a CI that tests all the torch operators in the test suite.

You can find the failure logs by navigating to the appropriate GH action run of https://github.com/google/heir/actions/workflows/torch_mlir_coverage.yml and downloading the "failure-logs" artifact.

The torch-ci models tested are checked in at https://github.com/google/heir/tree/main/scripts/torch_ci

j2kun:
| Operator | Status | Details |
| --- | --- | --- |
| add | PASS |  |
| avgpool1d | FAIL (heir-opt) | error: 'linalg.conv_2d_nchw_fchw' op Expected equal strides for Conv2DNchwFchwOp     %3 = linalg.poo |
| avgpool2d | PASS |  |
| avgpool3d | FAIL (heir-opt) | error: 'tensor_ext.assign_layout' op requires tensor rank to match the layout's domain size, but fou |
| batchnorm1d | PASS |  |
| batchnorm2d | PASS |  |
| batchnorm3d | PASS |  |
| bmm | FAIL (heir-opt) |  |
| cat | FAIL (heir-opt) | error: unexpected unrealized conversion cast op found     %concat = tensor.concat dim(1) %arg0, %arg |
| conv1d | PASS |  |
[...]

PR #2925, opened 2026-05-06
Simplify convert to ciphertext semantics

j2kun:
Rebased over https://github.com/google/heir/pull/2924

This refactors the implementations of linalg kernels share some common code related to IR materialization.

PR #2941, opened 2026-05-12
Add support for conv1d, Phase 3

mdgrs:
This builds on the previous work #2919 (and that previous work is the first commit) 

It uses the work on Conv2d_nchw_fchw as a blueprint to add support for Con1d_ncw_fcw. In my tests, this is the linalg operation that my pytorch conv1d layer gets lowered to.

It makes conv1d turn green for #2923


lib/Utils/Layout/Convolution.cpp
```diff
@@ -421,13 +544,16 @@ bool isRelationConvFilterDiagonalized(
...
+presburger::IntegerRelation get2dConvRowInterchangeRelation(int64_t c,
+                                                            int64_t h,
+                                                            int64_t w,
+                                                            int64_t g) {
   // 1. Unflatten idx_in (H, W, C*g^2) into (hi, wi, ci).
   // 2. Map to output coordinates by interleaving sub-pixels:
   //    h' = hi * g + (ci % g**2) // g
   //    w' = wi * g + (ci % g)
   // 3. Flatten (gW, gH, C) into idx_out = (c * g * h) * w' + (c) * h' + c'
+  // FIXME why are these interchanged???
```
mdgrs: It seems like h and w were exchanged in the definitions of hOut and wOut

asraa (commented):
Thank you! I'll take a look - could you try rebasing? Something strange happened in the diff where it appears it deletes some recent changes

lib/Kernel/Kernel.cpp
```diff
@@ -28,7 +28,8 @@ static std::unordered_map<KernelName, std::vector<std::string>>
          {"linalg.matvec", "linalg.conv_2d_nchw_fchw", "linalg.conv_1d"}},
         {KernelName::VecmatDiagonal, {"linalg.vecmat"}},
         {KernelName::MatmulDiagonal, {"linalg.matmul"}},
-        {KernelName::MatmulDiagonal, {"linalg.conv_2d"}},
+        // FIXME is there a _ missing in linalg.conv_2d
```
j2kun: Looks like this got messed up in the rebase

lib/Utils/Layout/Convolution.cpp
j2kun: @asraa to comment

lib/Utils/Layout/Convolution.cpp
```diff
@@ -368,6 +460,37 @@ FailureOr<presburger::IntegerRelation> getConvFilterDiagonalizedRelation(
...
+  auto expandedFilterRelation =
+      get1dConvCwFcwFilterRelation(filterType, dataType, stride, padding);
+  // Permutate the rows of the matrix to minimize the number of non-zero
+  // diagonals.
+  if (interchangeRows) {
+    int64_t dataSize = dataType.getDimSize(2);
+    int64_t filterSize = filterType.getDimSize(2);
+    int64_t outputW = (dataSize + 2 * padding - filterSize) / stride + 1;
+
+    auto rowInterchangeRelation = get1dConvRowInterchangeRelation(
```
j2kun, line 475: Sanity check: you want this because the cross-channel filters are stacked a certain way that, without the row interchange, requires more diagonals.

I admit it took us a while to derive the conv2d version, and I am struggling to remember how this interacted with the linalg canonicalization lowering it to a loop (which is also present in the PR for conv1d). @asraa didn't we end up having a problem with converting it to a loop, and didn't we ultimately resort to a dedicated lowering for the Conv2DNchwFchwOp as a whole?

j2kun (commented):
Generally LGTM.

lib/Utils/Layout/Convolution.cpp
mdgrs: I added a test to highlights how the number of diagonals is reduced with the interchange

For the linalg canonicalization, I saw that currently we have for loops that loop over the extra dimensions of Conv2dNchwFchwOp and rewrite it as a series of Conv2dOp. I did the same for Conv1dNcwFcw.

asraa: We don't want to lower to a loop anymore generally because preserving the multi-channel input/output let's us arrange each kernel into blocks of a larger toeplitz matrix - so I don't expect that we'll be wanting to rewrite as a loop of single convolutions often.

The other "problem" with the loop approach is that the conv_nd operations don't support strides so we can't use that to represent pooling / downsampling!

lib/Transforms/LinalgCanonicalizations/LinalgCanonicalizations.cpp
```diff
@@ -729,6 +729,132 @@ struct RewriteAvgPoolAsConv2D
   }
 };
 
+// Lower linalg.conv_1d_ncw_fcw to a loop of linalg.conv_1d operations.
+struct LowerConv1DNcwFcw
```
asraa: see my comment earlier, but I think we might not actualyl want this since you already have the ncw_fcw kernels written. in the end to end pipeline, this pass runs very early in the pipeline and that means later passes will only see the linalg.conv_1d ops

lib/Utils/Layout/Convolution.cpp
asraa: oh no! @mdgrs could you link this into an issue? i don't want to block this but I'm also noticing an error in flattening the final output in the ISL string.

asraa (approved):
thanks! appreciate the test coverage - 
this will conflict on https://github.com/google/heir/pull/2925 depending on which is merged first

mdgrs:
> thanks! appreciate the test coverage - this will conflict on #2925 depending on which is merged first

Great. I'll wait for #2925 to be in, rebase and squash locally and then push again

mdgrs:
It should be good to go now.

j2kun:
There are still merge conflicts but the internal tooling seems to automatically resolve them.

mdgrs:
Hmmm, I hope I did not mess anything up. I still tried to resolve the conflict by hand, which I guess was a mistake. This PR now only deletes a comment, so we can close it.

j2kun:
> Hmmm, I hope I did not mess anything up. I still tried to resolve the conflict by hand, which I guess was a mistake. This PR now only deletes a comment, so we can close it.

This got merged successfully with some additional internal patches

PR #2947, opened 2026-05-13
fix: cleanup scheme parameters in correct place

Tuditi:
I encountered this bug when running a more complex neural network through HEIR with log N = 16 with `--scheme-to-lattigo`. 

IIUC, this is due to the cleanup of the CKKS parameters happening before the bootstrapping parameters are derived. Let me know if you need more information! 

Cheers,
David


j2kun (approved):
Thanks!!

PR #2955, opened 2026-05-16
Add CI for coverage of torch operators

j2kun:
This was largely an AI port of https://github.com/j2kun/heir_torch_coverage to a CI workflow in HEIR directly.

Features:

0. Runs a CI testing heir-opt on all the listed torch operators every two weeks (offset for HEIR monthly release)
1. Does not depend on torch-mlir in HEIR directly, only in CI
2. Updates polynomial-approximation to choose smarter default intervals, hopefully avoiding trivial crashes of non-polynomial functions like sqrt and log2
3. Saves the failure logs to the GH action outputs so they can be inspected
4. Writes a table to the action output and to [a distinguished comment in #2923](https://github.com/google/heir/issues/2923#issuecomment-4391907478)


j2kun:
The PR is AI generated, I want to just give it a quick review before sending it for review.

j2kun:
Trying to reduce the review burden on Asra, adding @AlexanderViand.

lib/Transforms/PolynomialApproximation/PolynomialApproximation.cpp
```diff
@@ -41,6 +41,42 @@ constexpr int64_t kDefaultDegree = 5;
 constexpr double kDefaultDomainLower = -1.0;
 constexpr double kDefaultDomainUpper = 1.0;
 
+template <typename OpTy>
+struct OpDefaultDomain {
```
AlexanderViand: mostly just a nit, but seems a bit verbose - what about `patterns.add<ConvertUnaryOp<math::SqrtOp>>(context, sqrt, 0.0, 2.0);` instead,  defaulting to `kDefaultDomainLower`/`kDefaultDomainUpper` if not provided?

scripts/torch_ci/entrypoint.py
```diff
@@ -0,0 +1,119 @@
...
+
+    cmd = f"{self.heir_opt_path} {input_path}"
+    result = self.executor.run(cmd)
+    return result
+
+  def run_all(self):
+    models = self.find_models()
+    results = {}
+    for model in sorted(models):
+      print(f"Running coverage for {model}...")
```
AlexanderViand: Isn't this going to spam the results table comment?

scripts/torch_ci/test_entrypoint.py
AlexanderViand, line 1: I'd drop this and the other test file, if something will fail, it's probably anyway something GH actions related we can't properly test for locally ^^

scripts/torch_ci/update_issue_comment.py
AlexanderViand, line 1: Using a python script that uses the gh cli to do this seems a bit janky/brittle, why don't we just do this directly from a GH action?

Claude says this can be done via

```
 - name: Find Coverage Comment
    id: fc
    uses: peter-evans/find-comment@<sha>
    with:
      issue-number: 2923
      body-includes: '<!-- TORCH_COVERAGE_TABLE -->'

  - name: Create or Update Coverage Comment
    uses: peter-evans/create-or-update-comment@<sha>
    with:
      comment-id: ${{ steps.fc.outputs.comment-id }}
      issue-number: 2923
      body-path: coverage_table.md
      edit-mode: replace
```

or, if we don't want the 3rd party actions:
<details><summary> with just GH authored actions</summary>
```
 - name: Update Coverage Comment
    uses: actions/github-script@v7   # pin by SHA in practice
    with:
      script: |
        const fs = require('fs');
        const marker = '<!-- TORCH_COVERAGE_TABLE -->';
        const body = `${marker}\n\n` + fs.readFileSync('coverage_table.md', 'utf8');
        const issue_number = 2923;

        const comments = await github.paginate(github.rest.issues.listComments, {
          owner: context.repo.owner,
          repo: context.repo.repo,
          issue_number,
        });
        const existing = comments.find(c => c.body.includes(marker));

        if (existing) {
          await github.rest.issues.updateComment({
            owner: context.repo.owner, repo: context.repo.repo,
            comment_id: existing.id, body,
          });
        } else {
          await github.rest.issues.createComment({
            owner: context.repo.owner, repo: context.repo.repo,
            issue_number, body,
          });
        }
```
</details>

scripts/torch_ci/entrypoint.py
```diff
@@ -0,0 +1,119 @@
...
+      export_result = self.run_export(model)
+      if export_result.returncode != 0:
+        results[model] = ("FAIL (Export)", export_result.stderr)
+        self.save_failure_log(model, "export", export_result.stderr)
+        continue
+
+      heir_opt_result = self.run_heir_opt(model)
+      if heir_opt_result is None:
+        results[model] = ("FAIL (No MLIR)", "model.mlir not found")
```
AlexanderViand: This should probably still be `save_failure_log'ed? Knowing that something breaks in the torch-mlir export seems useful, too.

lib/Transforms/PolynomialApproximation/PolynomialApproximation.cpp
```diff
@@ -41,6 +41,42 @@ constexpr int64_t kDefaultDegree = 5;
...
+  static constexpr double lower = 0.1;
+  static constexpr double upper = 2.0;
+};
+
+template <>
+struct OpDefaultDomain<math::SqrtOp> {
+  static constexpr double lower = 0.0;
+  static constexpr double upper = 2.0;
+};
+
```
AlexanderViand: ❤️  

Though with this fixed, if people forget to fold their BatchNorms, they'll get a bunch of unexpected overhead instead of a cryptic error message about NaN coefficients ;)

AlexanderViand (approved):
Great Idea and mostly looks good to me, but maybe try the comment-finding-and-updating via GH actions (see comment below)? 

Also, I assume "two dozen small python files, each doing just one tiny thing" is a solution a coding agent came up with? (but, hey, it works and is surprisingly clean xD)

j2kun:
> two dozen small python files, each doing just one tiny thing

That was my idea, though implemented by an agent 😅  I just felt it would keep the examples decoupled, and allow us to put larger examples in the same framework as needed.

lib/Transforms/PolynomialApproximation/PolynomialApproximation.cpp
j2kun: I would love to see an example of the input with a bad error message, then we can fix it properly. Probably this pass is not the right place to detect the problem (but maybe)

scripts/torch_ci/test_entrypoint.py
j2kun: I agree, but this is mainly why I moved the logic outside of GH actions: so it can be tested. This ensures, for example, that the entrypoint.py script doesn't get merged with a syntax error, or other basic mistake.

scripts/torch_ci/update_issue_comment.py
j2kun: Having it as a separate script allows us to test and even run the script locally (say, if GH action fails), whereas with this we would be forced to push it to GH in order to test it.

However, I do like these simpler actions better. So I'll update it to do this.

lib/Transforms/PolynomialApproximation/PolynomialApproximation.cpp
j2kun: Done

scripts/torch_ci/entrypoint.py
j2kun: Fixed.

scripts/torch_ci/entrypoint.py
j2kun: Good call

AlexanderViand (approved):

PR #2946, opened 2026-05-13
Feature/heir cross integration

Zohaib58:
Adds end-to-end support for lowering JaxiteWord IR to CROSS (CKKS backend for TPU), including dialect ops, a crypto-context setup pass, and an expanded Python emitter.

This completes the JaxiteWord → CROSS path that was started in earlier JaxiteWord work. The LWE → JaxiteWord conversion pass is intentionally out of scope here and tracked separately in [#2964](https://github.com/google/heir/pull/2964).

### **Changes**
**JaxiteWord** dialect
- Extend JaxiteWordOps.td with CROSS-oriented ops (e.g. gen_params, program_initialization, mul/relin/rot helpers aligned with the CROSS API).
- Simplify dialect verification by removing redundant MulOp / RotOp verifiers where TableGen already enforces types/attributes.

**jaxiteword-configure-crypto-context pass**

- New transform pass that analyzes an entry function and synthesizes helper functions to generate and configure the CROSS crypto context.
- Infers CKKS parameters such as mul depth, rotation indices, and relinearization needs from the IR.
- Exposes pass options for manual overrides (mul depth, TPU tiling r/c, batch size, composite rescaling, scaling factor, etc.).
- JaxiteWord emitter (heir-translate --emit-jaxiteword)
- Substantially expand JaxiteWordEmitter and templates to emit CROSS-compatible Python.
- Cover homomorphic ops (add/sub/mul/relin/rot/mod_reduce, etc.) and crypto setup (gen_params, keygen, eval keys).

PR #2964, opened 2026-05-19
Add LWE to JaxiteWord conversion pass

Zohaib58:
Adds the lwe-to-jaxiteword conversion pass, which lowers the lwe dialect (and related ckks / bgv ops handled by the pass) to the jaxiteword dialect for CROSS / JaxiteWord codegen.

This is a follow-up to **#2946** (JaxiteWord dialect, emitter, and configure-crypto-context). That PR landed the target dialect and Python emission; this PR wires the LWE → JaxiteWord lowering so HEIR pipelines can reach CROSS backend ops from scheme-level IR.

### Changes

- New pass: lib/Dialect/LWE/Conversions/LWEToJaxiteWord/
- Pass flag: --lwe-to-jaxiteword
- Inserts jaxiteword.crypto_context and jaxiteword.eval_key on func.func when the function contains crypto ops
- Lowers homomorphic ops (e.g. add/sub/mul/rotate/relin/mod-switch, encode/encrypt/decrypt/decode, plaintext variants) to jaxiteword ops
- Register pass in heir-opt (tools/heir-opt.cpp, tools/BUILD)
- Lit tests under tests/Dialect/LWE/Conversions/lwe_to_jaxiteword/:
- smoke_test.mlir - encode + encrypt path runs without error
- radd.mlir - lwe.radd lowers to jaxiteword.add with context/eval-key args

### Notes
Intended to mirror patterns from existing LWE lowering passes (lwe-to-openfhe, lwe-to-polynomial).

### DIFF
--- a/lib/Dialect/LWE/Conversions/LWEToJaxiteWord/BUILD
+++ b/lib/Dialect/LWE/Conversions/LWEToJaxiteWord/BUILD
@@ -1,0 +1,35 @@
+load("@heir//lib/Transforms:transforms.bzl", "add_heir_transforms")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+package(
+    default_applicable_licenses = ["@heir//:license"],
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "LWEToJaxiteWord",
+    srcs = ["LWEToJaxiteWord.cpp"],
+    hdrs = ["LWEToJaxiteWord.h"],
+    deps = [
+        ":pass_inc_gen",
+        "@heir//lib/Dialect/BGV/IR:Dialect",
+        "@heir//lib/Dialect/CKKS/IR:Dialect",
+        "@heir//lib/Dialect/JaxiteWord/IR:Dialect",
+        "@heir//lib/Dialect/LWE/IR:Dialect",
+        "@heir//lib/Dialect/ModArith/IR:Dialect",
+        "@heir//lib/Dialect/RNS/IR:Dialect",
+        "@heir//lib/Utils:ConversionUtils",
+        "@llvm-project//mlir:FuncDialect",
+        "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:Support",
+        "@llvm-project//mlir:TensorDialect",
+        "@llvm-project//mlir:Transforms",
+    ],
+)
+
+add_heir_transforms(
+    header_filename = "LWEToJaxiteWord.h.inc",
+    pass_name = "LWEToJaxiteWord",
+    td_file = "LWEToJaxiteWord.td",
+)
--- a/lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
+++ b/lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
@@ -1,0 +1,381 @@
+#include "lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.h"
+
+#include <cstdint>
+#include <utility>
+
+#include "lib/Dialect/BGV/IR/BGVDialect.h"
+#include "lib/Dialect/BGV/IR/BGVOps.h"
+#include "lib/Dialect/CKKS/IR/CKKSDialect.h"
+#include "lib/Dialect/CKKS/IR/CKKSOps.h"
+#include "lib/Dialect/JaxiteWord/IR/JaxiteWordDialect.h"
+#include "lib/Dialect/JaxiteWord/IR/JaxiteWordOps.h"
+#include "lib/Dialect/JaxiteWord/IR/JaxiteWordTypes.h"
+#include "lib/Dialect/LWE/IR/LWEOps.h"
+#include "lib/Dialect/LWE/IR/LWETypes.h"
+#include "lib/Utils/ConversionUtils.h"
+#include "lib/Utils/Utils.h"
+#include "mlir/include/mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Func/Transforms/FuncConversions.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinAttributes.h"      // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinOps.h"             // from @llvm-project
+#include "mlir/include/mlir/IR/BuiltinTypes.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/PatternMatch.h"           // from @llvm-project
+#include "mlir/include/mlir/IR/TypeUtilities.h"          // from @llvm-project
+#include "mlir/include/mlir/Pass/Pass.h"                 // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace lwe {
+
+class JaxiteWordTypeConverter : public TypeConverter {
+ public:
+  JaxiteWordTypeConverter(MLIRContext *ctx) {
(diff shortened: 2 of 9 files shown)


lib/Dialect/LWE/Conversions/LWEToJaxiteWord/BUILD
```diff
@@ -0,0 +1,36 @@
...
+        "@heir//lib/Dialect/RNS/IR:Dialect",
+        "@heir//lib/Utils:ConversionUtils",
+        "@llvm-project//mlir:FuncDialect",
+        "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Pass",
+        "@llvm-project//mlir:Support",
+        "@llvm-project//mlir:TensorDialect",
+        "@llvm-project//mlir:Transforms",
+    ],
+    alwayslink = 1,
```
j2kun:
```suggestion
```

nit: should not be needed

tests/Dialect/LWE/Conversions/lwe_to_jaxiteword/radd.mlir
```diff
@@ -0,0 +1,24 @@
...
+!rns_L1_ = !rns.rns<!Z1095233372161_i64_, !Z1032955396097_i64_>
+#ring_Z65537_i64_1_x1024_ = #polynomial.ring<coefficientType = !Z65537_i64_, polynomialModulus = <1 + x**1024>>
+#ring_rns_L1_1_x1024_ = #polynomial.ring<coefficientType = !rns_L1_, polynomialModulus = <1 + x**1024>>
+#ciphertext_space_L1_ = #lwe.ciphertext_space<ring = #ring_rns_L1_1_x1024_, encryption_type = lsb>
+!ct_L1_ = !lwe.lwe_ciphertext<plaintext_space = <ring = #ring_Z65537_i64_1_x1024_, encoding = #full_crt_packing_encoding>, ciphertext_space = #ciphertext_space_L1_, key = #key, modulus_chain = #modulus_chain_L5_C1_>
+
+module {
+  // CHECK: func.func @test_radd
+  // CHECK-SAME: !jaxiteword.crypto_context
+  // CHECK-SAME: !jaxiteword.eval_key
```
j2kun:
```suggestion
  // CHECK-SAME: (%{{[^:]*}}: !jaxiteword.crypto_context, %{{[^:]*}}: !jaxiteword.eval_key
```

This will assert that they are the first two operands, which is a relatively important property of the PR that should be tested.

tests/Dialect/LWE/Conversions/lwe_to_jaxiteword/smoke_test.mlir
j2kun, line 1: Perhaps for the smoke test you could include a few more operations and add checks just for the expected name of the lowered op? (`// CHECK: jaxiteword.add`, etc.) I think it would be appropriate to have an LLM generate a suitably expressive test input here.

lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
```diff
@@ -0,0 +1,399 @@
...
+#include "mlir/include/mlir/Pass/Pass.h"                 // from @llvm-project
+#include "mlir/include/mlir/Transforms/DialectConversion.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace lwe {
+
+namespace jaxiteword = ::mlir::heir::jaxiteword;
+namespace ckks = ::mlir::heir::ckks;
+namespace bgv = ::mlir::heir::bgv;
```
j2kun: I believe these are no-ops, because it is already within the `mlir::heir` namespace. Also feel free to add `using` declarations for individual types if you want to cut down on the namespace verbosity at the usage sites.

```suggestion
```

lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
```diff
@@ -0,0 +1,399 @@
...
+    auto cryptoContextType = jaxiteword::CryptoContextType::get(getContext());
+    auto evalKeyType = jaxiteword::EvalKeyType::get(getContext());
+
+    rewriter.modifyOpInPlace(op, [&] {
+      bool hasContext = op.getFunctionType().getNumInputs() > 0 &&
+                        mlir::isa<jaxiteword::CryptoContextType>(
+                            op.getFunctionType().getInput(0));
+      if (!hasContext) {
+        (void)op.insertArgument(0, evalKeyType, nullptr, op.getLoc());
+        (void)op.insertArgument(0, cryptoContextType, nullptr, op.getLoc());
```
j2kun:
```suggestion
        if (failed(op.insertArgument(0, evalKeyType, nullptr, op.getLoc()))) return failure();
        if (failed(op.insertArgument(0, cryptoContextType, nullptr, op.getLoc()))) return failure();
```

lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+    Value dynamicShift = adaptor.getDynamicShift();
+    IntegerAttr staticShift = op.getStaticShiftAttr();
+    if (!staticShift && !dynamicShift) {
+      return rewriter.notifyMatchFailure(
+          op, "rotate op must have either static or dynamic shift");
+    }
+    if (dynamicShift) {
+      return rewriter.notifyMatchFailure(
+          op, "jaxiteword rotation requires static shift");
```
j2kun, line 182: This will end up being a bit of an obstacle for supporting looped linalg kernels. @JianmingTONG is this a strict limitation of CROSS?

No need to make any changes to this PR, I just want to know for tracking future work.

lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+    rewriter.replaceOpWithNewOp<jaxiteword::RotOp>(
+        op, this->getTypeConverter()->convertType(op.getOutput().getType()),
+        ctx.value(), adaptor.getInput(), evalKey.value(), staticShift);
+    return success();
+  }
+};
+
+template <typename SourceOp, typename TargetOp>
+struct ConvertPlainBinOp : public OpConversionPattern<SourceOp> {
```
j2kun: Is this identical to ConvertBinOp? Could we just use ConvertBinOp for both ct-ct and ct-pt ops?

j2kun (changes requested):
Looks great! Only some minor nitpicks

j2kun (changes requested):
LGTM! Please squash the commits (removing my co-authorship or replacing my email with jkun@google.com, I'm not sure why it doesn't recognize my account in the CLA...)

j2kun (approved):
LGTM!

lib/Dialect/LWE/Conversions/LWEToJaxiteWord/LWEToJaxiteWord.cpp
JianmingTONG: Hey Jeremy

Thank you for the comment and wish you a great memorial weekend!

IIUC, the computation in HE is preknown ahead of execution, which could be used to generate all rotation keys. And this dynamicShift seems telling that we need some rotation indice to be generated in the runtime. Might I get some insights on what cases would we need this runtime dynamity? And correct me if I miss understand it

j2kun: If we express a program using loops, then the SSA value representing the desired shift is statically inferrable, but dynamically dependent on the loop iteration in question. So even if we know all the rotation keys at compile time, the operation syntax may still need to express a non-attribute for the shift operand.

JianmingTONG: Aha got it!

It's the actual shift used at runtime that depends on which loop iteration is executing

I could confirm that is not the limitation of CROSS to support dynamic rotation, as it'll be the same API as OpenFHE

PR #2977, opened 2026-05-21
Resolve TODOs from #2157

### LINKED ISSUE #2157: Lower ckks.relinearize to polynomial dialect

crockeea:
It's unclear if this actually resolves #2157. The TODOs from that issue are no longer relevant, and have been removed. But that issue also mentions 
- getting the key switching key added to the IR at a higher level
- getting parameter-selected params forwarded to this lowering

neither of which are done.


j2kun:
It appears there is a related test failure. Also happy to chat live or email about the next steps

crockeea:
The test failure is related to a test which was previously disabled, and which, with this PR, became enabled. The test looked *very* old (it used `polynomial.key_switch_inner`, an op which I'm not sure ever existed). The appropriate place for this test is in CKKS/Transforms, and that test already exists. As a result, I deleted this test. Should be ready to merge after CI.

j2kun (approved):

PR #2978, opened 2026-05-21
CKKS Dialect Improvements

crockeea:
Cleanup in the CKKS dialect:
- Folded the `--ckks-decompose-keyswitch` transform into --ckks-to-lwe. There was no need for a separate transform since `--ckks-decompose-keyswitch` just decomposed a CKKS op into LWE ops.
- Moved `--ckks-to-lwe` from the Transform directory to the Conversion directory. It is now a full-fledged dialect lowering.
- Removed duplicate CKKSOp attributes that are inherited.

Assisted by GPT.


j2kun:
While this is technically acceptable, the main reason this pass was a "transform" and not a "dialect conversion" is because there were no type changes. The only reason to use the dialect conversion infrastructure is because it handles the complexity of changing types, and the invalid IR produced in intermediate steps of the process of lowering types.

So basically doing dialect conversion without a type converter is the same as doing the walk rewrite driver with the corresponding OpRewritePatterns.

Maybe you felt that because it is changing dialects from CKKS to lwe, it should be called a "conversion" and live in the Conversions directory? I have no issue with that, but the pass can remain unchanged and it will be less complex overall to debug.

crockeea:
No, I just forgot about the criteria for what makes a transform and what makes a conversion. The goal was to make CKKS/LWE/Polynomial more symmetric, but I see that's not really the correct thing to do. Closing this PR

PR #2979, opened 2026-05-21
Orbit's Level-Scale-Aware ILP Bootstrap and Rescale Placement

edwjchen:
The ILP formulation is inspired by [Orbit](https://eprint.iacr.org/2026/213.pdf)

This PR extends the ILP bootstrap placement model to solve for both bootstrap and rescale placement. 

Changes:
- level and scale constraints for CKKS placement decisions
- weighted objective function, rather than solely minimizing the bootstrap count 
- a toy JSON cost model from Orbit 
- updates to the mgmt dialect to annotate the input level and scale

### DIFF
--- a/lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
+++ b/lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
@@ -1,17 +1,17 @@
 #include "lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.h"
 
-#include <cassert>
 #include <cmath>
-#include <cstddef>
 #include <sstream>
 #include <string>
 #include <utility>
 
 #include "absl/status/statusor.h"  // from @com_google_absl
 #include "lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h"
+#include "lib/Dialect/Mgmt/IR/MgmtAttributes.h"
 #include "lib/Dialect/Secret/IR/SecretOps.h"
 #include "llvm/include/llvm/ADT/DenseMap.h"                // from @llvm-project
 #include "llvm/include/llvm/ADT/STLExtras.h"               // from @llvm-project
+#include "llvm/include/llvm/ADT/SmallVector.h"             // from @llvm-project
 #include "llvm/include/llvm/Support/Debug.h"               // from @llvm-project
 #include "llvm/include/llvm/Support/raw_ostream.h"         // from @llvm-project
 #include "mlir/include/mlir/Analysis/DataFlowFramework.h"  // from @llvm-project
@@ -23,25 +23,41 @@
 #include "ortools/math_opt/cpp/math_opt.h"  // from @com_google_ortools
 
 // The level describes the remaining multiplicative depth of a ciphertext.
-// Bootstrap operations reset the level to a maximum value (or a target level
-// if specified), allowing further operations to be performed.
+// Bootstrap operations reset the level to a maximum value, after which decoded
+// management ops may reduce it to the target level chosen by the ILP. In CKKS
+// mode this analysis also tracks scale bits:
+//   Sw: waterline/base scale bits (scaleWaterline)
+//   Sf: scale bits dropped by one rescale/modreduce (scaleFactorBits)
 //
-// This implementation tracks the level state through the computation
-// and uses ILP to determine optimal bootstrap placement to minimize costs
-// while ensuring level constraints are satisfied.
+// This implementation tracks level/scale state through one secret.generic body
+// and uses ILP to determine cost-minimal bootstrap and rescale placement while
+// keeping producer values, operand edges, and op results mutually feasible.
+// A later transform decodes the solution into mgmt operations.
 //
-// ILP Formulation:
-// - Variables: level[value] for each SSA value, input_level[op] for each op,
-//   bootstrap[op] for each operation
+// ILP formulation:
+// - Variables:
+//   * level[value], scale[value] (in bits) for each secret SSA value
+//   * input_level[op], input_scale[op] for each tracked op
+//   * bootstrap[op], node_rescale[op] for management after an op
+//   * edge_rescale[use] and, for CKKS multiplication operands, edge_scale[use]
+//     for management before a consuming op
+// - Initialization:
+//   * secret.generic body args are initialized from associated mgmt.mgmt attrs
+//     when present, otherwise from (bootstrapWaterline, Sw)
 // - Constraints:
-//   * Level bounds: 0 <= level[value] <= bootstrapWaterline (levels
-//   0..bootstrapWaterline)
-//   * Operand matching: for each op, operand_i_level == input_level[op]
-//   * Multiplication: output_level = input_level - 1
-//   * Non-mult: output_level = input_level
-//   * Bootstrap (big-M): if bootstrap[op] = 1, level_after = bootstrapWaterline
-//                        else level_after = level_before
-// - Objective: minimize sum of bootstrap decisions
+//   * bounds: levels are 0..bootstrapWaterline; CKKS scales are Sw..scaleMax
+//   * level-only mode fixes all live scales to Sw
+//   * operand edges allow level/scale reduction before the consuming op
+//   * non-multiplication operands share the consumer's input_scale
+//   * CKKS multiplication input_scale is the sum of operand edge scales;
+//     plaintext constants contribute Sw
+//   * node transitions relate each op's input state to each result state by
+//     either direct rescale/modswitch management or a bootstrap transition
+//   * yielded result scales are constrained to explicit nonzero mgmt.mgmt
+//   scales
+//     on corresponding secret.generic results
+// - Objective: minimize bootstrap and rescale costs, with a small tie-breaker
+//   favoring higher remaining levels.
 
 namespace math_opt = ::operations_research::math_opt;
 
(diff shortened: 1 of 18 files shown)


j2kun:
I will review this PR this week, sorry for the delay!

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -31,14 +33,18 @@
 // while ensuring level constraints are satisfied.
 //
 // ILP Formulation:
-// - Variables: level[value] for each SSA value, input_level[op] for each op,
-//   bootstrap[op] for each operation
+// - Variables: level[value] and scale[value] for each SSA value,
```
```suggestion
// - Variables: level[value] and scale[value] (in bits) for each SSA value,
```

nit: may be worth specifying the units on scale.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -124,54 +146,118 @@ LogicalResult ILPBootstrapPlacementAnalysis::solve() {
...
     std::string opName = uniqueName(op);
 
-    // Get secret operands
-    SmallVector<Value> secretOperands;
-    getSecretOperands(op, solver, secretOperands);
-
-    // Operand matching: all operands must be at the same level when consumed.
-    // inputLevel = level at which operands are consumed.
+    // Operand matching: all operands must be consumed at the same level.
+    // Unlike the original model, a producer may stay at a higher level and a
```
j2kun: nit: mentioning "the original model" is not particularly informative, given there is no explanation of why the difference to the original model is useful. (Is "the original model" the earlier implementation? an external impl? best to just not mention it and leave discussion of the changes to the PR/commit description)

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -26,20 +33,90 @@ namespace heir {
...
+  if (!parsed) {
+    llvm::consumeError(parsed.takeError());
+    return failure();
+  }
+
+  const llvm::json::Object* root = parsed->getAsObject();
+  if (!root) return failure();
+
+  std::optional<int> parsedBootstrapCost =
+      averagePositiveLatency(*root, "earth.bootstrap_single");
```
j2kun: Is there a hope to make this more fine-grained than using a blended cost? If not, perhaps the json schema itself should just provide a blended latency cost for each op. I think supporting a per-level cost would require pretty dramatic changes to the model.

tests/Transforms/ilp_bootstrap_placement/orbit_cost_model.json
j2kun, line 1: It would be helpful to have documented somewhere (perhaps in a README.md in the lib/Transforms/ILPBootstrapPlacement or analysis directory) the specification of the cost model json file, and to include some breadcrumbs (maybe in the pass tablegen) linking back to that document, so the meaning of the fields of this json can be understood in the future.

Also some nits on the json spec itself, which would help to either fix or explain in the readme:

- Why use `earth.*_single`, which I think came from DaCapo (?) instead of our own op names?
- Why are there multiple values for each op in the latency table? Are these latencies per-level, and if so (a) what is the ordering of the levels and (b) why do they have a different number of values (bootstrap has 16, rescale 15). 16 also seems quite low, so I'm wondering what you'd expect a pass to do when they want the cost of an op at level 30.
- What is the unit of latency?
- How do the other fields influence the rest of the cost model? (They seem unused, so maybe remove them and re-introduce them in the future when they are used)

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -26,20 +33,90 @@ namespace heir {
...
   LogicalResult processSecretGenericOp(
       secret::GenericOp genericOp, DataFlowSolver* solver,
-      SmallVector<Value, 32>* valuesToBootstrap) {
+      SmallVector<Value, 32>* valuesToBootstrap,
+      SmallVector<ILPBootstrapPlacementAnalysis::OutputLevelReduction, 32>*
+          outputLevelReductions,
+      SmallVector<ILPBootstrapPlacementAnalysis::OperandLevelReduction, 32>*
+          operandLevelReductions) {
     genericOp->walk([&](mgmt::BootstrapOp op) {
       op.getResult().replaceAllUsesWith(op.getOperand());
```
j2kun: sanity check: should you also remove any existing mgmt.modreduce and mgmt.level_reduce ops?

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -26,20 +33,90 @@ namespace heir {
...
+    int effectiveBootstrapCost = bootstrapCost;
+    int effectiveRescaleCost = rescaleCost;
+    if (!orbitCostModel.empty()) {
+      FailureOr<OrbitCostModel> loadedCostModel =
+          loadOrbitCostModel(orbitCostModel);
+      if (failed(loadedCostModel)) {
+        llvm::errs() << "failed to load Orbit cost model from `"
+                     << orbitCostModel << "`\n";
+        genericOp->emitError() << "failed to load Orbit cost model from `"
+                               << orbitCostModel << "`";
```
j2kun, line 155:
```suggestion
```

Nit, unnecessary.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -48,26 +125,69 @@ struct ILPBootstrapPlacement
...
+      ArrayRef<ILPBootstrapPlacementAnalysis::OutputLevelReduction>
+          outputLevelReductions) {
+    OpBuilder b(&getContext());
+    for (auto reduction : outputLevelReductions) {
+      auto [toReduce, insertAfter] =
+          followRelinearizeModReduceChain(reduction.value);
+      if (!insertAfter) continue;
+
+      b.setInsertionPointAfter(insertAfter);
+      auto levelReduceOp = mgmt::LevelReduceOp::create(
```
j2kun, line 228: `mgmt.level_reduce` drops the last limb, while `mgmt.modreduce` is what lowers to a CKKS rescale op. Is `level_reduce` what you want?

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
j2kun: Since this function is now 500 lines long, I think it would help greatly to split out the component pieces into separate functions. Maybe one function to construct the model and one to solve it, and then in the model construction function you can split up the relevant sub-tasks into more readable pieces (e.g., generating and indexing the variables, generating each type of constraint, applying compression).

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -124,54 +146,118 @@ LogicalResult ILPBootstrapPlacementAnalysis::solve() {
...
 
     // Create level variables for results
     for (OpResult result : op.getResults()) {
       if (!isSecret(result, solver)) continue;
 
       std::stringstream ss;
       ss << "level" << opName << result.getResultNumber();
-      auto levelVar =
-          model.AddContinuousVariable(0, bootstrapWaterline, ss.str());
+      auto levelVar = model.AddIntegerVariable(0, bootstrapWaterline, ss.str());
```
j2kun: If the level input constraints are integral, and the level drop variables are integral, then the level variables can be left continuous and they will force the state variables to be integral (as they were before this change). I don't see any newly added constraints in this PR that would contradict this.

j2kun (changes requested):
Generally it looks good. I have some comments on organization, one about the integrality of the state variables, and then I'm still a little bit confused on where the rescale is happening. It seems like this model is not yet handling rescale decisions, but instead is using level_reduce to align mis-aligned operand levels (which is good but not quite the same as a rescale decision). It may be that refactoring the code will make it easier for me to see exactly which constraint is controlling the scale updates.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
edwjchen: updated!

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
edwjchen: The next PR is to make this a level-dependent model (as this PR is getting quite long).

tests/Transforms/ilp_bootstrap_placement/orbit_cost_model.json
edwjchen: Added a README for the cost modeling. It might be nice to also include the the cost modeling script we used for Lattigo in the next PR . 

- Renamed the cost model key names
- Simplified the model for now, but the ordering would be the cost from level 1 to level n, in ascending order.
- Unit of latency is in microseconds, but this can be changed!

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
edwjchen: refactored!

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.td
```diff
@@ -6,18 +6,25 @@ include "mlir/Pass/PassBase.td"
...
+           mgmt.modreduce becomes ckks.rescale.
+        4) Inserts mgmt.bootstrap at the positions chosen by the ILP.
+        5) Inserts mgmt.relinearize after each mul. Resulting order is mul ->
+           relinearize, with rescale/bootstrap chains inserted after
+           relinearize when the ILP chooses them on a mul result.
+
+        The ILP formulation is inspired by [Orbit](https://eprint.iacr.org/2026/213.pdf).
+        CKKS scale modeling is selected from CKKS module attributes. Modules
+        without CKKS attributes use the generic level-only model, where the
+        effective waterline scale and rescaling factor are equal.
```
j2kun, line 30:
```suggestion
        This pass uses an integer linear program to determine feasible levels
        and Orbit-style scales for each term in the MLIR, and thus the
        placement of mod-down, rescale, upscale, and bootstrap operations.

The pass runs on [ciphertext-semantic](https://heir.dev/docs/design/layout/#data-semantic-and-ciphertext-semantic-tensors)
IR (secret.generic with arith ops operating on pre-packed tensors). It

        1. Inserts `mgmt.level_reduce` for decoded level-only mod-down.
        2. Inserts `mgmt.adjust_scale` for decoded upscale.
        3. Inserts `mgmt.modreduce` for decoded rescale. In CKKS lowering,
           `mgmt.modreduce` becomes `ckks.rescale`.
        4. Inserts `mgmt.bootstrap` at the positions chosen by the ILP.
        5. Inserts `mgmt.relinearize` after each mul. Resulting order is mul ->
           relinearize, with rescale/bootstrap chains inserted after
           relinearize when the ILP chooses them on a mul result.

        The ILP formulation is inspired by [Orbit](https://eprint.iacr.org/2026/213.pdf).
        CKKS scale modeling is selected from CKKS module attributes. Modules
        without CKKS attributes use the generic level-only model, where the
        effective waterline scale and rescaling factor are equal.
```

nit: the descriptions here render as markdown on the website.

tests/Transforms/ilp_bootstrap_placement/ilp_relin_modreduce_placement.mlir
```diff
@@ -1,52 +1,48 @@
...
-    %l7 = arith.mulf %l3, %l4 : tensor<8xf32>
-    %l8 = arith.mulf %l5, %l6 : tensor<8xf32>
-    %l9 = arith.mulf %l6, %l7 : tensor<8xf32>
-    %l10 = arith.mulf %l8, %l9 : tensor<8xf32>
-    secret.yield %l10 : tensor<8xf32>
-  } -> !secret.secret<tensor<8xf32>>
-  return %0 : !secret.secret<tensor<8xf32>>
+      %arg4: !secret.secret<tensor<8xf32>>) -> !secret.secret<tensor<8xf32>> {
+    %0 = secret.generic(
+        %arg0: !secret.secret<tensor<8xf32>>,
```
j2kun: Just FYI for future reference, you can make type aliases in the IR to make this easier to read/write

```
!tens = tensor<8xf32>
!ty = !secret.secret<!tens>


func.func  @bootstrap_placement_test(%arg0: !ty, %arg1: !ty, ...
```

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -50,217 +66,441 @@ namespace math_opt = ::operations_research::math_opt;
...
+    objective += bootstrapCost * bootstrapVar;
+  }
+  for (auto& [op, rescaleVar] : state.nodeRescaleVars) {
+    objective += rescaleCost * rescaleVar;
+  }
+  for (auto& [operand, rescaleVar] : state.edgeRescaleVars) {
+    objective += rescaleCost * rescaleVar;
+  }
+  for (auto& [value, levelVar] : state.valueLevelVars) {
+    objective += -0.001 * levelVar;
```
j2kun, line 441: Can you add a comment explaining the reason for this coefficient? When you're minimizing an expression that includes a negative coefficient, my spidey senses go off worrying about unbounded solutions :)

j2kun (approved):
Looks great! Just a few more nits.

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
edwjchen: This was used as a tie-breaker to favor solutions with the same bootstrap and rescale cost to have higher output levels. I've added a comment that states we'll remove this condition on the next PR as our objective function should be also based on the per-level operation costs (e.g., for multiplication, addition, and rotations).

j2kun (approved):

PR #2982, opened 2026-05-21
Remove Jupyter infra

j2kun:
Now that we have both a PyPI distribution and rules_heir, there is no longer a significant reason to have a Jupyter integration.

Also, since it is not tested, it is almost surely out of date and broken. I'd prefer we instead direct our efforts toward improving the Python frontend to encapsulate whatever usability might still be desired from the jupyter cell magics.


AlexanderViand (approved):
100% agree - in place of the old heir_play package, I've been using this snippet in notebooks:

```python
from heir.heir_cli import heir_cli, heir_cli_config

# expose heir-opt to Python (bypassing the normal HEIR Python frontend)
heir_opt = heir_cli.HeirOptBackend(heir_cli_config.from_pip_installation().heir_opt_path)
```

which can be used like this:
```python

# put MLIR in here
program = """
func.func @example(%x : f32) -> f32 {
  %0 = math.exp %x {degree = 5 : i32, domain_lower = -4.0 : f64, domain_upper = 4.0 : f64} : f32
  return %0 : f32
}
"""

print(heir_opt.run_binary(
            input=program,
            options=["--secretize", "--torch-linalg-to-ckks"]
        ))
````

It might be a good idea to provide the `heir_cli.HeirOptBackend(heir_cli_config.from_pip_installation().heir_opt_path)` under a nicer name for exactly this use case?

PR #2984, opened 2026-05-22
Update nightly builds

j2kun:
This PR updates the nightly release to look more like the full releases. It:

- Builds the wheels in manylinux containers for better compatibility and to support more platforms
- Uploads the binaries from the built wheels into the GH release
- Keeps running 90-day nightlys with different tags and release names
- Makes `nightly` a floating release tag that always points to the latest nightly release

The scripts are a little bit messy, but the new parts are starting at "Create Dated Release"

Context: I am doing this so I can add a feature to rules_heir to have it opt-in to using the nightly instead of a pinned release.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,75 +1,179 @@
 name: Nightly release
 
-permissions: read-all
-
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


asraa (approved):
we might be able to use a reusable workflow to package up the common steps (build wheels, create the source archive)

PR #2986, opened 2026-05-26
Add test for non square filter

mdgrs:
This fixes a mistake in the row interchange permutation computation.


asraa (approved):
thank you! is there any way you can try out another test point - maybe something on the domain coordinate (1, 0, 1, 2) (h != w != 0)?

but thank you so much for adding this test!

mdgrs:
I added a point check as a sanity check

asraa (approved):
lgtm! thank you!

PR #2944, opened 2026-05-13
Add CHEDDAR dialect

AlexanderViand:
Extracted (w/ some cleanup) from #2896

PR #2994, opened 2026-05-27
Add cheddar emitter

AlexanderViand:
Split out from #2896, follow-up to #2944

### DIFF
--- a/lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
+++ b/lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
@@ -537,7 +537,8 @@ std::string floatArrayLit(ArrayAttr a) {
   std::string s = "{";
   for (size_t i = 0; i < a.size(); ++i) {
     if (i > 0) s += ", ";
-    s += floatLit(cast<FloatAttr>(a[i]));  // full %.17g precision, not truncated
+    s +=
+        floatLit(cast<FloatAttr>(a[i]));  // full %.17g precision, not truncated
   }
   return s + "}";
 }
@@ -888,9 +889,13 @@ struct ConvertMadUnsafe : public OpConversionPattern<cheddar::MadUnsafeOp> {
   }
 };
 
-// `ctx->Boot(res, input, evk_map);`. The CHEDDAR runtime resolves whether
-// `ctx` is a regular `Context` or a `BootContext`; the dialect carries only
-// a single Context type today.
+// `static_cast<BootContext<word>*>(ctx)->Boot(res, input, evk_map);`.
+// The dialect models a CHEDDAR Context and a BootContext as one
+// `!cheddar.context` type (see CheddarTypes.td) -- a module that bootstraps
+// must be run with a BootContext, which the harness creates and passes in as
+// the Context. `Boot` is a (non-virtual) BootContext method, so we recover the
+// derived type with a static_cast here; the cast is safe by construction
+// because the op only exists in modules that require a BootContext at runtime.
 struct ConvertBoot : public OpConversionPattern<cheddar::BootOp> {
   using OpConversionPattern::OpConversionPattern;
   LogicalResult matchAndRewrite(
@@ -898,13 +903,25 @@ struct ConvertBoot : public OpConversionPattern<cheddar::BootOp> {
       ConversionPatternRewriter& rewriter) const override {
     Type t = this->typeConverter->convertType(op.getResult().getType());
     Value out = declareLocal(rewriter, op.getLoc(), t);
-    emitOutParamCall(rewriter, op.getLoc(), adaptor.getCtx(), "Boot", out,
-                     ValueRange{adaptor.getInput(), adaptor.getEvkMap()});
+    VerbatimOp::create(rewriter, op.getLoc(),
+                       "static_cast<BootContext<word>*>({})->Boot({}, {}, {});",
+                       ValueRange{adaptor.getCtx(), out, adaptor.getInput(),
+                                  adaptor.getEvkMap()});
     rewriter.replaceOp(op, loadAfter(rewriter, op.getLoc(), t, out));
     return success();
   }
 };
 
+// cheddar.linear_transform -> CHEDDAR's LinearTransform extension. Unlike every
+// other op there is no Context *method* to call: LinearTransform is a *class*
+// (constructed from a StripedMatrix of diagonals, then Evaluate()'d). The
+// HEIR-side emitter prelude provides a thin `RunLinearTransform<W, word>` shim
+// that packs the cleartext diagonals operand into a StripedMatrix, constructs
+// the LinearTransform at the op's level + canonical plaintext scale
+// (ctx->param_.GetScale(level)), and evaluates it -- so we emit one uniform
+// structured call_opaque (no verbatim, no Context/CHEDDAR change):
+//   RunLinearTransform<W, word>(out, ctx, in, evk_map, diagonals,
+//                               {indices}, level, bs, gs);
 struct ConvertLinearTransform
     : public OpConversionPattern<cheddar::LinearTransformOp> {
   using OpConversionPattern::OpConversionPattern;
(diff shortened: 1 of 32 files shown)


tools/heir-translate.cpp
```diff
@@ -12,9 +12,14 @@
 #include "lib/Target/TfheRustHL/TfheRustHLEmitter.h"
 #include "lib/Target/Verilog/VerilogEmitter.h"
 #include "llvm/include/llvm/Support/LogicalResult.h"  // from @llvm-project
+#include "mlir/include/mlir/InitAllTranslations.h"    // from @llvm-project
 #include "mlir/include/mlir/Tools/mlir-translate/MlirTranslateMain.h"  // from @llvm-project
 
 int main(int argc, char** argv) {
+  // MLIR-to-C++ via the EmitC dialect: used as the final stage of the CHEDDAR
+  // backend (`scheme-to-cheddar | --cheddar-to-emitc | heir-translate
+  // --mlir-to-cpp`), and available standalone for any EmitC IR.
```
AlexanderViand, line 21: Looks like there's still a bit of 🤖-y bla-bla in here - I'll come back tomorrow and clean those up

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,996 @@
...
+    // `emitc.size_t` and leave a `builtin.unrealized_conversion_cast` at the
+    // boundary. Hooking `index -> emitc.size_t` here lets our memref-op
+    // patterns consume the converted index directly via the adaptor and the
+    // dialect-conversion framework reconciles the cast away.
+    addConversion(
+        [ctx](IndexType) -> Type { return emitc::SizeTType::get(ctx); });
+  }
+};
+
+// emitc::OpaqueType doesn't implement MemRefElementTypeInterface upstream,
```
j2kun, line 240: I'm mildly surprised you aren't doing an array of pointers instead of an array of values.

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,996 @@
...
+Value declareLocal(OpBuilder& b, Location loc, Type t) {
+  return VariableOp::create(b, loc, LValueType::get(t),
+                            OpaqueAttr::get(b.getContext(), ""));
+}
+
+Value loadAfter(OpBuilder& b, Location loc, Type t, Value lvalue) {
+  return mlir::emitc::LoadOp::create(b, loc, t, lvalue);
+}
+
+// Emit `receiver->method(out, args..., extra);`.
```
j2kun: Cf. https://discourse.llvm.org/t/method-calls-in-emitc/90898 I may end up adding support for method calls so as to support OpenFHE more naturally as well

tools/heir-opt.cpp
```diff
@@ -278,6 +284,21 @@ int main(int argc, char** argv) {
   registerPass(
       []() -> std::unique_ptr<Pass> { return createConvertToLLVMPass(); });
 
+  // SCF/MemRef/Arith -> EmitC, used by the cheddar pipeline to lower
+  // bufferized loop kernels through EmitC to C++.
+  registerPass([]() -> std::unique_ptr<Pass> { return createSCFToEmitC(); });
```
j2kun: This doesn't have its own registration hook already because the `--convert-to-emitc` pass uses dialect interface registrations to handle all dialects at once. Cf. https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Conversion/ConvertToEmitC/ConvertToEmitCPatternInterface.td

I'm guessing you found this and, due to the complexity of the emitc patterns for Cheddar, you weren't able to use it?

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
j2kun: ah this is the move-only thing?

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,1408 @@
...
+Value findUi(Operation* op, const TypeConverter& tc) {
+  Type uiType =
+      tc.convertType(cheddar::UserInterfaceType::get(op->getContext()));
+  auto r = getContextualArgFromFunc(op, uiType);
+  if (failed(r)) return Value{};
+  return r.value();
+}
+
+struct ConvertHRot : public OpConversionPattern<cheddar::HRotOp> {
+  using OpConversionPattern::OpConversionPattern;
```
j2kun, line 826: Most of these uses of Verbatim seem like they should also be MemberCallOpaque. Any reason why not?

lib/Conversions/CheddarToEmitC/CheddarToEmitC.td
```diff
@@ -0,0 +1,32 @@
...
+    `UserInterface` method call expressed in the EmitC dialect.
+    `mlir-translate --mlir-to-cpp` renders the resulting IR as host-side
+    C++ against the CHEDDAR library API.
+
+    Bufferized loop kernels are handled in the same conversion: the SCF and
+    Arith EmitC patterns are run alongside the cheddar patterns so that the
+    shared type converter sees cheddar types (e.g. an `scf.for` carrying an
+    `!cheddar.*` iter_arg lowers to a move-assigned `emitc.variable`, and
+    loop-index values feed `memref` subscripts without a stranded
+    `index -> emitc.size_t` cast).
```
j2kun, line 20: Please add a doctest

patches/llvm.patch
j2kun, line 1: Send me an upstream patch for this please!

tests/Conversions/CheddarToEmitC/loop.mlir
j2kun, line 1: I'm still a bit confused why you're not using the emitc conversion interfaces and https://mlir.llvm.org/docs/Passes/#-convert-to-emitc. Is there a concrete reason for that?

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,1408 @@
...
+
+    // A move-only memref arg that is *stored into* is an output/in-place
+    // buffer: at the C++ boundary it must become a mutable `std::array<T, N>&`,
+    // not the `const std::array<T, N>&` that a read-only input array gets.
+    // memref.store ops are gone after conversion (rewritten to subscript + a
+    // std::move verbatim), so capture which entry-block-arg memrefs are written
+    // *now*, keyed by func op (whose identity persists through the in-place
+    // signature conversion). Consumed by Pass 1 below.
+    llvm::DenseMap<Operation*, llvm::SmallDenseSet<unsigned>> writtenArrayArgs;
+    getOperation()->walk([&](func::FuncOp fn) {
```
j2kun: This loop can be simplified by walking just the StoreOps and using op->getParentOfType<FuncOp>() to get the map key and for the owner check, instead of a nested walk of funcs and bodies.

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,1408 @@
...
+    // stranding an `index -> emitc.size_t` cast at a pass boundary.
+    //
+    // These op-level illegality markings must come *after*
+    // addStructuralConversionPatterns: that helper calls
+    // populateSCFStructuralTypeConversionsAndLegality, which marks the SCF ops
+    // dynamically legal (type-conversion only). Op-level legality overrides the
+    // dialect-level setting regardless of call order, so a dialect-level
+    // addIllegalDialect<SCFDialect> would be silently shadowed and the EmitC
+    // for/if patterns would never fire. Marking the ops illegal here forces the
+    // SCFToEmitC patterns to win.
```
j2kun, line 1121: Perhaps it would be simpler to replace `addStructuralConversionPatterns` with the subset of patterns from the implementation of `addStructuralConversionPatterns` that are relevant to this pass, then you don't need implicit tie-breaking with a long comment.

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,1408 @@
...
+    // of the receiver methods.
+    getOperation()->walk([](emitc::LoadOp load) {
+      StringRef name;
+      if (!isMoveOnlyOpaque(load.getType(), name)) return;
+      load.getResult().replaceAllUsesWith(load.getOperand());
+      load.erase();
+    });
+
+    auto& ctxRef = *ctx;
+    getOperation()->walk([&ctxRef, &writtenArrayArgs](func::FuncOp op) {
```
j2kun, line 1216: I feel this section would greatly benefit from having the steps inside the walk split into sub-functions.

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,1408 @@
...
+          else if (argIdx != static_cast<int>(ba.getArgNumber())) {
+            ok = false;
+            break;
+          }
+        }
+        if (ok && argIdx >= 0) {
+          resultIsInout[i] = true;
+          argIsInout[argIdx] = true;
+        }
+      }
```
j2kun:
```suggestion
      for (unsigned i = 0; i < numResults; ++i) {
        StringRef name;
        if (!isMoveOnlyOpaque(funcType.getResult(i), name)) 
          continue;
      
        if (returns.empty()) 
          continue;
      
        auto firstBa = dyn_cast<BlockArgument>(returns.front().getOperand(i));
        if (!firstBa || firstBa.getOwner() != &entry) 
          continue;
      
        unsigned expectedArgIdx = firstBa.getArgNumber();
        
        bool allMatch = llvm::all_of(returns, [&](auto ret) {
          auto ba = dyn_cast<BlockArgument>(ret.getOperand(i));
          return ba && ba.getOwner() == &entry && ba.getArgNumber() == expectedArgIdx;
        });
      
        if (allMatch) {
          resultIsInout[i] = true;
          argIsInout[expectedArgIdx] = true;
        }
      }
```

nit: The nesting makes this quite hard to read, plus the use of llvm::all_of simplifies the structure a bit.

lib/Conversions/CheddarToEmitC/CheddarToEmitC.cpp
```diff
@@ -0,0 +1,1408 @@
...
+        if (load.use_empty()) load.erase();
+      }
+    });
+
+    // Strip leftover `tensor_ext.*` metadata (original-type / layout
+    // annotations carried on the client function boundaries). They are
+    // HEIR-internal and reference the unregistered `tensor_ext` dialect, which
+    // mlir-translate (mlir-to-cpp) cannot parse.
+    getOperation()->walk([](func::FuncOp fn) {
+      auto stripTensorExt =
```
j2kun, line 1412: Why not use AttributeUtils::clearAttrs ?

j2kun (changes requested):
This will have to wait for the LLVM commit bump to hit `main`, but otherwise some minor organizational comments. I admit I don't fully understand all the work that went into accommodating the move-only style.

AlexanderViand:
Closing this old PR in favour of #3334  and its follow-ups

PR #2995, opened 2026-05-27
Add support for pooling_ncw_sum

mdgrs:
I play the same trick as in the 2D convolution case and rewrite the pooling operation as a convolution.

While trying to lower a small convolutional network, I also came across some dimensionality mismatch (expand/collapse dims). I have included fixes in this PR, as well as tests.


j2kun (approved):
Looks great! Please squash the commits and we'll get it in

j2kun (approved):

PR #2999, opened 2026-05-28
support DenseResourceElementsAttr in memref'ed lattigo emitter

AlexanderViand:
Missing support for model weights in dense resources popped up when we rebased our internal ML pipeline onto the new bufferizied lattigo emitter. 🤖-generated fix.


asraa (commented):
could you add some test coverage for this please?

j2kun (changes requested):
Agreed, looks great but just add a small lit test so it doesn't get refactored into oblivion.

j2kun (approved):
I think I missed that this got updated until the morning of the WG meeting. LGTM!

PR #3000, opened 2026-05-28
extend lattigo.chebyshev with domain start/end

AlexanderViand:
While it's usually a good idea to massage your model/application into a shape where it only requires approximations over the default [-1,1], sometimes you do need a different interval and HEIR wasn't exposing the options from the Lattigo API in the IR.


j2kun:
This looks good to me. Add a test and we can merge it.

AlexanderViand:
> This looks good to me. Add a test and we can merge it.

Done + some small cleanup (now an optional size-2 dense array instead of two required float attrs)

j2kun (approved):

j2kun:
Due to a change that updated a copybara behavior (3a41d2f9ff48e744d18010eb1a9a29d63e2d2860), I will need you to rebase over main, sorry!

AlexanderViand:
@j2kun Done!

PR #3001, opened 2026-05-28
add bufferization-preparation passes to lattigo pipeline

AlexanderViand:
Bufferization is pretty picky about what ops it will support, for example it does not accept `arithf.add %t1, %t2 : tensor<...>` but instead wants the `linalg.generic` version. We already do this in other bufferization-using pipelines, this PR simply gives that set of passes a nice name/helper and adds it to the lattigo pipeline, too.

PS: Sorry for the piece-meal PRs, I didn't realize the rabbit hole I'd be going down as part of rebasing our stuff onto the new bufferized lattigo backend


j2kun (approved):
LGTM! I think I discovered this while working on something similar today where arith.truncf was not supported

PR #3019, opened 2026-06-02
Added cache to AttachNTTRoots pass to speed it up

crockeea:
For a test application with 83 NTT/iNTT ops, the --attach-ntt-roots pass took 4.77s.

After this PR, the pass took 0.32s on the same application, a 14.8x improvement.

It's still not blazing fast, which I'm guessing is due to `findPrimitive2nthRoot`.


j2kun (approved):
LGTM!

PR #3026, opened 2026-06-03
python: move front/backend to extras

### LINKED ISSUE #3024: Prune frontend-only dependencies from `heir-py` package
The heir-py package currently serves as both a nice clean distribution of built binaries of heir, and also as the package for the python frontend. The latter has a bunch of complex requirements (python version, exact numba version, etc) that can very easily lead to unresolvable dependency conflicts when pulling heir-py into a project, even if the project only uses the binaries, not the actual frontend.

This isn't just a hypothetical, @superlopuh and his student ran into this issue when they wanted to use `heir-opt` for some polynomial approximation stuff.

I suggest we put the frontend behind an optional/extra, so that you need to install `pip install heir-py[frontend]` if you DO want it. Alternatively, we could also create heir-py-core and heir-py-frontend packages and have heir-py be little more than a wrapper around them, but that seems a bit overkill.

@j2kun what do you think?

AlexanderViand:
Closes #3024 

The actual extras stuff is, imho, fairly nice and clean, but "hiding" the imports is a bit of a hack to minimise disruption/code changes while still making heir-py "work" (or, at least, import cleanly) when the extras aren't there.


AlexanderViand:
pinging @superlopuh

j2kun (changes requested):
I think it would be better to use importlib to find the module that is supposed to be there. Otherwise this lookup will (a) be slower and (b) potentially mask other import errors in the code it's importing.

What if we have a python helper module like

```python
def verify_module_exists(module_relative_path, extra_name):
    """Checks if the internal module file exists."""
    # not sure here, maybe it would be best to always use absolute paths as input
    absolute_module_name = f"{__package__}{module_relative_path}"
    if importlib.util.find_spec(absolute_module_name) is None:
        raise ImportError(
            f"Failed to import {absolute_module_name}. This suggests "
            f"heir was installed with pip and missing the {extra_name} extra; "
            f"install it via `pip install heir_py[{extra_name}]`."
        )
```

And then the usage could be

```python
from heir_extras_util   # not wedded to this name

__all__ = ["compile"]

# Imported lazily so that `import heir` works without the [python] extra.
def __getattr__(name):
  return match name:
    case "compile":
      heir_extras_util.verify_module_exists(".pipeline", extra_name="python")
      from .pipeline import compile  # pylint: disable=g-import-not-at-top
      return compile
    case _:
      raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

And

```python
  if not backend:
    heir_extras_util.verify_module_exists("heir.backends.openfhe", extra_name="openfhe")
    from heir.backends.openfhe import OpenFHEBackend, config as openfhe_config  # pylint: disable=g-import-not-at-top
```

AlexanderViand:
Oh, I realised I also forgot to fix the other thing @superlopuh was running into, which was python version requirements - we should have basically no (or extremely, extremely broad) python version requirements on the main thing (since it'd be just the binaries and a bit of basic python, like your suggested helper), and then require-version the extras down to the range they need (especially the frontend is suuuuper picky, iirc).

And while writing this, I realised this PR should probably also update the wheel building action, otherwise anyone asking for extras would probably end up building from source?

j2kun:
> Oh, I realised I also forgot to fix the other thing @superlopuh was running into, which was python version requirements - we should have basically no (or extremely, extremely broad) python version requirements on the main thing (since it'd be just the binaries and a bit of basic python, like your suggested helper), and then require-version the extras down to the range they need (especially the frontend is suuuuper picky, iirc).
> 
> And while writing this, I realised this PR should probably also update the wheel building action, otherwise anyone asking for extras would probably end up building from source?

any need to keep the two changes in the same PR?

superlopuh:
Why not catch an import error if the package is not found?

frontend/heir/__init__.py
```diff
@@ -27,6 +27,18 @@ def find_spec(self, fullname, path, target=None):
...
+  if name == "compile":
+    try:
+      from .pipeline import compile  # pylint: disable=g-import-not-at-top
+    except ImportError as e:
+      raise ImportError(
+          "Failed to import the HEIR python frontend. Its dependencies are"
+          " likely missing; install them via `pip install heir_py[python]`."
+      ) from e
+    return compile
+  raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
superlopuh: If you move the optional dependency import into `compile` then this bit of code is no longer necessary, right?

superlopuh:
Not sure how strict you are with your testing but we have two CIs: one with optional dependencies and one without to check that we handle missing optional dependencies gracefully, and still succeed with the core tests

AlexanderViand:
> Not sure how strict you are with your testing but we have two CIs: one with optional dependencies and one without to check that we handle missing optional dependencies gracefully, and still succeed with the core tests

I think for us there are no meaningful (python) tests that work only on "core" (everything interesting needs at least a backend, most also the frontend), so I just added a minimal test that'll trigger if someone adds code that leaks an optional/extra dependency into the core.

j2kun (approved):
LGTM! Let's just make sure we can build a dev wheel properly before merging.

j2kun:
> Not sure how strict you are with your testing but we have two CIs: one with optional dependencies and one without to check that we handle missing optional dependencies gracefully, and still succeed with the core tests

Is there a way to specify a test matrix in cibuildwheel? If so, we could have that be the mechanism to split into groups of extras.

superlopuh:
Not entirely sure, we do this just via two CIs with different installs, and check that the tests pass (or are skipped as appropriate) without the extra installs, but it seems quite likely that it should be matrixable.

j2kun:
I guess we could do it locally, outside of the cibuildwheel workflow, with `pip install .`...

PR #3039, opened 2026-06-05
OpenFHE codegen for torch sigmoid + cf.assert on plaintexts

j2kun:
This is mirroring an example from a user involving a torch sigmoid operator. They are lowering to OpenFHE.

### DIFF
--- a/lib/Target/OpenFhePke/BUILD
+++ b/lib/Target/OpenFhePke/BUILD
@@ -33,6 +33,7 @@ cc_library(
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:AffineDialect",
         "@llvm-project//mlir:ArithDialect",
+        "@llvm-project//mlir:ControlFlowDialect",
         "@llvm-project//mlir:FuncDialect",
         "@llvm-project//mlir:IR",
         "@llvm-project//mlir:SCFDialect",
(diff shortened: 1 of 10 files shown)


lib/Target/OpenFhePke/BUILD
```diff
@@ -33,6 +33,7 @@ cc_library(
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:AffineDialect",
         "@llvm-project//mlir:ArithDialect",
+        "@llvm-project//mlir:ControlFlowDialect",
```
AlexanderViand, line 36: Doesn't the emitter target also need this new dependency? I mean, it compiled without it I guess...

lib/Target/OpenFhePke/OpenFhePkePybindEmitter.cpp
```diff
@@ -92,6 +92,16 @@ LogicalResult OpenFhePkePybindEmitter::printOperation(func::FuncOp funcOp) {
       }
     }
   } else {
+    // This branch isneeded to support split preprocessing, which has multiple
```
AlexanderViand:
```suggestion
    // This branch is needed to support split preprocessing, which has multiple
```

AlexanderViand (approved):
LGTM! All the plaintext op handling really makes the point for why we should switch to emitC, though.

lib/Target/OpenFhePke/BUILD
j2kun: yeah, the import to google3 will enforce that more strictly, so this PR will probably land with some extra build changes.

asraa (commented):
late to the party but nice work!

Issue #3003, opened 2026-05-28
Failing integration test related to convolutions


mdgrs:
I am investigating a bug where I am getting the wrong output values with an offset of -1024 (other examples give other (negative) powers of 2). 

You can find a minimal reproducible example here:

https://github.com/belfortlabs/heir/blob/16b70b94611cb878500e04dab8c28c7b1d036306/tests/Examples/lattigo/ckks/bug_minrepro/bug_minrepro_test.go#L14

Interestingly, if I change the value from i*i+1 to i or i+1, the test passes.

I'll keep looking into it, but I figured someone might have an intuition as to where the problem lies. Any ideas @asraa @AlexanderViand @j2kun ?

asraa:
I'm going to patch your test - I fixed a bug I found in the Codegen in testing the convolution yesterday, and I can see if that fixes it.

Thank you for the min repro code! That's very helpful

j2kun:
@mdgrs sorry we had a team summit all week. I think Asra is still working on this, I will sync with her next week about it.

asraa:
I've now realized that the floor divisions and modulus operations are sometimes resulting in incorrect loop generation from ISL (if I apply the layout relation by manually computing points, it's correct, but if I use ISL i'll hit index out of range issues). I don't think that's the issue in this particular case (since the floor division isn't happening unless interchangeRows = true)

asraa:
I'll push my PR, but this fixes it - one of the expected outputs 3081.0. The default scale is 2^45 and the default first mod (lowest modulus) is 2^55, so when we decrypt and remove the scale we only got 2^10 bits, and we were wrapping around by -1024 because of that.
      "--mlir-to-ckks=ciphertext-degree=1024 first-mod-bits=60"

mdgrs:
That makes sense. Thank you for looking into this.

PR #3041, opened 2026-06-05
Rewrite dilated convolutions by expanding the kernel

mdgrs:
The dilation argument adds 0 between the kernel values:

[1,2,3], dilation=3 -> [1,0,0,2,0,0,3], dilation=1

This pass rewrites a dilated convolution operator to a convolution operator with dilation=1 by explicitly inserting the 0s in the kernel.

If you agree with approach, I can also port it to the 2d convolution operator.

I am opening this as a draft PR because it is blocked by #3003 . The bug is unrelated to these changes, but it is making the integration test fail. Let me know if I can help you somehow with the resolution of #3003 @asraa . I am happy to sit down and try to figure this out together if you think that would be helpful.

### DIFF
--- a/lib/Transforms/LinalgCanonicalizations/LinalgCanonicalizations.cpp
+++ b/lib/Transforms/LinalgCanonicalizations/LinalgCanonicalizations.cpp
@@ -1002,6 +1002,173 @@ struct DropCfAssertInLinalg : public OpRewritePattern<linalg::GenericOp> {
   }
 };
 
+// Rewrites a linalg.conv_1d_ncw_fcw operation with dilation > 1 into an
+// equivalent dilation=1 convolution This materializes the filter to insert
+// `dilation-1` zeros between the filter entries See
+// https://github.com/vdumoulin/conv_arithmetic/blob/af6f818b0bb396c26da79899554682a8a499101d/gif/dilation.gif
+// for a visualization of dilation
+struct UndilateConv1DNcwFcw
+    : public OpRewritePattern<mlir::linalg::Conv1DNcwFcwOp> {
+ public:
+  UndilateConv1DNcwFcw(MLIRContext* context)
+      : OpRewritePattern<mlir::linalg::Conv1DNcwFcwOp>(context) {}
+
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(mlir::linalg::Conv1DNcwFcwOp convOp,
+                                PatternRewriter& rewriter) const override {
+    auto dilations = convOp.getDilations();
+    if (!dilations)
+      return rewriter.notifyMatchFailure(convOp, "no dilations attribute");
+    int64_t d = *dilations.getValues<int64_t>().begin();
+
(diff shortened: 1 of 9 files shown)


lib/Transforms/LinalgCanonicalizations/LinalgCanonicalizations.cpp
```diff
@@ -1002,6 +1002,84 @@ struct DropCfAssertInLinalg : public OpRewritePattern<linalg::GenericOp> {
   }
 };
 
+// Rewrites a linalg.conv_1d_ncw_fcw operation with dilation > 1 into an
+// equivalent dilation=1 convolution This materializes the filter to insert
+// `dilation-1` zeros between the filter entries See
+// https://github.com/vdumoulin/conv_arithmetic/blob/master/README.md for a
```
j2kun: This is great. Let's use the permalink:

https://github.com/vdumoulin/conv_arithmetic/blob/af6f818b0bb396c26da79899554682a8a499101d/gif/dilation.gif

lib/Transforms/LinalgCanonicalizations/LinalgCanonicalizations.cpp
```diff
@@ -1002,6 +1002,84 @@ struct DropCfAssertInLinalg : public OpRewritePattern<linalg::GenericOp> {
...
+    auto filterTy = cast<RankedTensorType>(filter.getType());
+    if (!filterTy.hasStaticShape())
+      return rewriter.notifyMatchFailure(convOp, "dynamic filter shape");
+
+    // filter is (F, C, KW); dilation acts on KW. New KW' = d*(KW-1) + 1.
+    ArrayRef<int64_t> shape = filterTy.getShape();
+    int64_t f = shape[0], c = shape[1], kw = shape[2];
+    int64_t newKw = d * (kw - 1) + 1;
+    Type eltTy = filterTy.getElementType();
+    auto newFilterTy = RankedTensorType::get({f, c, newKw}, eltTy);
```
j2kun:
```suggestion
    auto newFilterTy = filterTy::clone({f, c, newKw});
```

Cf. https://github.com/llvm/llvm-project/blob/1ef2f3b98b130bb99ecd412094de44dda89bd392/mlir/include/mlir/IR/BuiltinTypes.h#L80-L82

tests/Transforms/linalg_canonicalizations/conv1d_dilated.mlir
j2kun, line 1: The loop looks correct to me, but for a little bit extra coverage, could you add a lit test of dilation=3 or higher?

j2kun (approved):
LGTM! I think with Asra's other fix you should be unblocked here, either by making the input values smaller or by increasing the mod bits.

mdgrs:
> LGTM! I think with Asra's other fix you should be unblocked here, either by making the input values smaller or by increasing the mod bits.

I changed the test parameters and added the 2d version, which was pretty straightforward. If you are happy with it, I'll squash down to 1 commit

j2kun (approved):
LGTM! Let's squash and get it in

PR #3043, opened 2026-06-05
OpenFHE configuration improvements

j2kun:
This change ensures that the openfhe backend properly receives config and overrides from CLI flags

### DIFF
--- a/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
+++ b/lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
@@ -235,8 +235,21 @@ struct ConfigureCryptoContext
   LogicalResult getConfig(func::FuncOp op) {
     auto module = op->getParentOfType<ModuleOp>();
 
-    // remove bgv.schemeParam attribute if present
-    // fill encryptionTechniqueExtended and plaintextModulus
+    config.ringDim = ringDim;
+    config.batchSize = batchSize;
+    config.firstModSize = firstModSize;
+    config.scalingModSize = scalingModSize;
+    config.digitSize = digitSize;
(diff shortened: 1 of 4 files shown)


lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
```diff
@@ -267,53 +276,66 @@ struct ConfigureCryptoContext
...
+          config.mulDepth = mulDepth;
+        }
+      });
     }
 
-    // pass option could override mulDepth
-    if (mulDepth != 0) {
-      config.mulDepth = mulDepth;
+    config.hasBootstrapOp = hasBootstrapOp(op);
+    if (needMulDepthAnalysis && config.hasBootstrapOp) {
```
asraa, line 337: why is this gated on needMulDepthAnalysis? that's true when mulDepth == 0 - but what happens if a mulDepth is set AND there's a boostrap op? then i feel like config.mulDepth should be checked to ensure that it was set to handle the bootstrap depth? or maybe the suggestion is that that shouldn't occur, or we should warn that bootstrap requires additional multiplicative depth.

asraa (approved):

lib/Dialect/Openfhe/Transforms/ConfigureCryptoContext.cpp
j2kun: Mainly I want to avoid adding the bootstrapDepth if they set the depth manually. We add it because we don't consider it as part of the mul depth analysis, but if a user sets it they should know and incorporate the bootstrap depth.

j2kun: Added code comments to clarify.

PR #3048, opened 2026-06-09
Fix hermetic toolchain

### LINKED ISSUE #3047: Properly enable the hermetic C++ toolchain
Contributors have noted that I failed to fully enable the hermetic LLVM build.

https://github.com/google/heir/pull/3045/  is a prototype fix, and I am going to try to whittle it down to a minimal set of improvements to allow us to properly enable the hermetic build.

One big obstacle is the python frontend shells out to clang on the CLI, and the hermetic toolchain doesn't make that easy.

j2kun:
Fixes https://github.com/google/heir/issues/3047

A simplification and borrowing parts of https://github.com/google/heir/pull/3045/

Change summary:

1. Bump hermetic llvm to a version that has openmp
2. Patch openfhe locally to pull from the hermetic llvm's openmp
3. Bump fmt to a working version
4. Move zlib to MODULE.bazel to avoid the need to patch
5. Patch bison same as https://github.com/google/heir/pull/3045/
6. Use `@hermetic_llvm//openmp:libomp` in the places in HEIR that need openmp
7. Enable BAZEL_TRACK_SOURCE_DIRECTORIES to silence some warnings introduced by hermetic llvm (this will go away by default in bazel 9, but we're still on bazel 8)
8. Add `-no-pie` for golang tests
9. Enable remote build execution and caching with buildbuddy, but don't yet update main CI workflow to use it.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,7 +3,6 @@ common --cxxopt='-std=c++20'
 common --copt=-fdiagnostics-color=always
 common --test_output=errors
 common --test_summary=terse
-
 common --enable_platform_specific_config
 
 # compilation_mode options below are `dbg` (debug symbols) `fastbuild`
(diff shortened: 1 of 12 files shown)


.bazelrc
```diff
@@ -14,19 +13,42 @@ common --enable_bzlmod
...
+# extra flags for hermetic LLVM toolchain
+common --repo_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+common --repo_env=BAZEL_DO_NOT_DETECT_SWIFT_TOOLCHAIN=1
+common --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
+common --@rules_python//python/config_settings:bootstrap_impl=script
+# Needed to silence hermetic LLVM build warnings, until this is enabled by
+# default in bazel 9
+startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
+
+# buildbuddy
```
asraa, line 25: might consider adding the API key here directly, so that users can just set their env var with their read token and then use --config remote seamlessly

tools/BUILD
```diff
@@ -185,6 +185,7 @@ cc_binary(
         "@heir//lib/Transforms/TensorToScalars",
         "@heir//lib/Transforms/UnusedMemRef",
         "@heir//lib/Transforms/ValidateNoise",
+        "@hermetic_llvm//openmp:libomp",
```
asraa, line 188: shouldn't this be gated on `config_enable_openmp` too?

asraa (commented):
quick questions

tools/BUILD
j2kun: we added enable_openmp because some people's systems didn't support it. But now that the hermetic LLVM build handles everything, I think it is not needed and we can actually remove all the openmp gating code.

I will try running the macos CI to at least confirm this doesn't break MacOS.

.bazelrc
j2kun: I will defer this to a standalone buildbuddy PR

tools/BUILD
j2kun: I will also add an issue to cleanup the OpenMP gating code

asraa (approved):

j2kun:
The macos build ran out of memory, so this suggests the openmp was not blocking the build. I think buildbuddy will reduce RAM needs for the macos CI when it's working

PR #3056, opened 2026-06-11
Add guards to prevent lowering eval-form polynomial constants

crockeea:
PolyMulToNTT can create EVAL-form constants, but the PolynomialToModArith lowering didn't look at the form when lowering the constants, resulting in bogus lowerings (e.g., the constant polynomial f(x)=5 with the EVAL tag would be lowered to the tensor [5, 0, 0, ..., 0]). This PR adds guards to prevent such lowerings. It also adds support for lowering eval-form constants in one special case, namely when the constant is a degree-zero polynomial, because the NTT of a degree-zero polynomial with constant-term `c` is [c, c, c, c, ...., c].

Note that there are discussions about changing the behavior of PolyMulToNTT so that constants are _forced_ into COEFF form, which forces PolyMulToNTT to insert explicit NTTs, which would subsequently be removed by an sccf pass. Until we get that working, I'm keeping the behavior of PolyMulToNTT on constants unchanged.

I realized that this is more or less what #2902 was trying to do, so I incorporated those changes here. Here's the rundown:
- PolyMulToNTT allows constants to be emitted in either form, since it is efficient materialize EVAL-form constants rather than performing a runtime NTT. However, we don't have the code to actually *do* the compile-time NTT, so this resulted in invalid lowerings. That's what the initial PR guarded against.
- For non-constant Polynomial ops, PolyMulToNTT already enforces the expected input/output forms. For example, the pass requires a COEFF-form polynomial as input to `LeadingTermOp`. #2902 added guards for those ops as well. This is the safe thing to do, although the current code never emitted EVAL-form inputs to these ops.
- ToTensorOp and FromTensorOp are a bit special: if you know what you're doing, there's nothing wrong with extracting the NTT coefficients of a polynomial. In fact, we [need to] do so in the NTT tests. The PolyMulToNTT pass requires COEFF-form inputs/outputs for these ops, but I do *not* enforce that in the PolynomialToModArith lowering since we need EVAL-form inputs/outputs for the tests.

The implementation of the degree-zero NTT was done by AI.

### DIFF
--- a/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
+++ b/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
@@ -298,6 +298,11 @@ struct ConvertConstant : public OpConversionPattern<ConstantOp> {
           op, "failed to construct common conversion info");
 
     auto typeInfo = res.value();
+    // TODO(#97): support compile-time NTT
+    if (typeInfo.polynomialType.getForm() == Form::EVAL) {
+      return rewriter.notifyMatchFailure(
+          op, "unsupported eval-form polynomial constant");
+    }
 
     auto attr = dyn_cast<TypedIntPolynomialAttr>(op.getValue());
     if (!attr)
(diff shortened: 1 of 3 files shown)


crockeea:
Actually, I see that this is more or less precisely what #2902 was trying to solve. That PR appears abandoned, so I will add the extra guards from that PR to this one.

crockeea:
Ready for review.

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -393,6 +398,45 @@ struct ConvertMonomial : public OpConversionPattern<MonomialOp> {
...
+    // We don't have proper support for EVAL-form constants, but we can
+    // at least support degree-zero polynomial constants in EVAL form. The
+    // NTT of a degree-zero polynomial is a vector where each coefficient is the
+    // constant term.
+    if (typeInfo.polynomialType.getForm() == Form::EVAL) {
+      auto degree = adaptor.getDegree().getDefiningOp<arith::ConstantIndexOp>();
+      if (!degree || degree.value() != 0) {
+        return rewriter.notifyMatchFailure(
+            op, "unsupported eval-form non-constant monomial");
+      }
```
j2kun: Slightly more general is to use Matchers.h, which eagerly applies constant folding on the defining ops.

```
IntegerAttr degreeAttr;
matchPattern(op.adaptor.getDegree(), m_Constant(&degreeAttr))
```


```suggestion
      IntegerAttr degreeAttr;
      if (!matchPattern(op.adaptor.getDegree(), m_Constant(&degreeAttr)) || degreeAttr.getInt() != 0) {
        return rewriter.notifyMatchFailure(
            op, "unsupported eval-form non-constant monomial");
      }
```

lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp
```diff
@@ -393,6 +398,45 @@ struct ConvertMonomial : public OpConversionPattern<MonomialOp> {
...
+      Value result;
+      if (auto modQTy = dyn_cast<ModQTypeInterface>(typeInfo.coefficientType)) {
+        Type extractedType = modQTy.getLoweringType();
+        Value extracted = mod_arith::ExtractOp::create(
+            b, extractedType, adaptor.getCoefficient());
+        Value replicatedStorage;
+        if (isa<ShapedType>(extractedType)) {
+          auto init = tensor::EmptyOp::create(b, storageTensorType.getShape(),
+                                              typeInfo.coefficientStorageType);
+          replicatedStorage = linalg::BroadcastOp::create(b, extracted, init,
```
j2kun, line 424:
```suggestion
          // Broadcast an RNS constant coefficient along the degree axis
          replicatedStorage = linalg::BroadcastOp::create(b, extracted, init,
```

j2kun (approved):

crockeea:
Applied suggested changes and squashed.

j2kun (approved):

PR #3058, opened 2026-06-11
Remove bundled openfhe from PyPI wheel

j2kun:
This change modifies the PyPI distribution of HEIR to no longer build and ship OpenFHE. The user must now provide a self-provided OpenFHE installation (shared libs and heaers) via environment variables.

As a consequence, we updated the CI to:

1. Not run frontend tests in bazel
2. After the main bazel build/test, add new steps that install openfhe, `pip install -e .`, and runs `pytest` to evaluate the frontend tests outside of bazel.

I also discovered that more recent manylinux containers have a sanctioned clang, so I updated the cibuildwheel container to use 2_34. We may be able to go back to supporting 2_28 after the hermetic LLVM lands, because I believe they have hermetic glibc that you can pick explicitly.

part of #3047


j2kun:
Miraculously, it works. I will give this code a bit of a cleanup tomorrow and then we can merge it in, which unblocks hermetic LLVM (and buildbuddy)

j2kun:
@AlexanderViand this should be ready now.

frontend/heir/backends/openfhe/config.py
```diff
@@ -86,7 +115,52 @@ def development_openfhe_config() -> OpenFHEConfig:
...
+  for p in lib_path.iterdir():
+    if not p.is_file():
+      continue
+    name = p.name.lower()
+    if not (".so" in name or ".dylib" in name):
+      continue
+    if "openfhe" in name and not any(
+        x in name for x in ["core", "pke", "binfhe"]
+    ):
+      return [str(p.resolve())]
```
asraa: do we want to return a list of all the possible resolved p's? (core, pke, and binfhe)

asraa (approved):

frontend/heir/backends/openfhe/config.py
j2kun: Done

pyproject.toml
```diff
@@ -12,13 +12,14 @@ authors = [
 ]
 description = "The HEIR compiler"
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
AlexanderViand, line 15: Do we really need this version bump?

pyproject.toml
```diff
@@ -58,7 +62,8 @@ dev = [
   # For ML integration tests
   "torch>=2.8.0",
   "setuptools>=80.9.0",
-  "cibuildwheel>=3.0.0",
+  "cibuildwheel>=4.0.0",
```
AlexanderViand, line 65:
```suggestion
  "cibuildwheel>=4.0.0" ; python_version >= '3.11'",
```
cibuildhweel 3+ does need python 3.11+ but since it's not a core dependency we can just gate it away.
(and maybe specify cibuildwheel>=2.0.0 ; python_version < 3.11 but I don't think that's even necessary

frontend/heir/mlir/types.py
```diff
@@ -1,14 +1,8 @@
 """Defines Python type annotations for MLIR types."""
 
 from abc import ABC, abstractmethod
-import sys
 from typing import Generic, Optional, TypeVar, get_args, get_origin
-
-if sys.version_info >= (3, 11):
-  from typing import TypeVarTuple, Unpack
-else:
-  # TypeVarTuple/Unpack landed in typing in 3.11; use the backport on 3.10.
```
AlexanderViand, line 10: Why drop the 3.10 compatibility shim here?

AlexanderViand (changes requested):
Sorry to be annoying about this, but can we avoid bumping minimum python from 3.10 to 3.11 again? This is a strong preference of the xdsl folks and wasn't hard to achieve in the past.

pyproject.toml
j2kun: Ah, I did that because we didn't actually test 3.10 or build any 3.10 wheels. I can put it back on Monday with a proper build.

pyproject.toml
j2kun: Can't you use a newer cibuildwheel but build for an older python? I don't think there's a conflict here.

AlexanderViand: You can build for really old target versions, but the package itself still requires a relatively new version to run on

AlexanderViand: So this suggestion is purely about making sure uv sync doesn't error on 3.10

j2kun:
This got merged just as I was getting into work. I'll push another fix restoring 3.10

j2kun:
It may also be worth noting that Python 3.10 is EOL this October (about 4 months away), so the xdsl folks should consider upgrading.

PR #3080, opened 2026-06-15
Use per-level logqi for CKKS mod-reduce scale analysis.

Zohaib58:
Partial fix for #1640: evalModReduceScale and evalModReduceScaleBackward now subtract/add logqi[level] instead of a fixed logDefaultScale. This matters when Q moduli have non-uniform bit widths, e.g. CROSS scheme params with a small last modulus. Falls back to logDefaultScale when the level is out of range.


j2kun:
@AlexanderViand you mentioned this may be of interest to your group. want to take a crack at the review?

j2kun (approved):

PR #3100, opened 2026-06-19
Emitter Optimization

Zohaib58:
lower CKKS rotations when a dynamic shift is defined by an arith
constant.

Rework the JaxiteWord Python emitter around tiled Polynomial values:
- emit func.call operations
- add _ensure_poly and _assign_poly helpers
- normalize ciphertext operands before add/sub/mul/square/rotate/rescale/decrypt
- use he_mul[level].mul for relined multiplication
- emit he_rescale for modulus reduction across levels
- implement plaintext add/sub against the c0 limb
- generate rotation keys via key_gen.gen_rotation_key
- store key material in generated CKKS params instead of mutating program init


j2kun:
I will take a closer look today. But to start, you should not check in these large generated python files (matvec_8x8_cross.py and matvec_8x8_jaxiteword.py). The lit tests should suffice for now, and then we can add a bazel-based macro (like https://github.com/google/heir/blob/138853859486b7c2a55a8088466f32ce253a0084/tools/heir-jaxite.bzl#L7) to combine heir-opt, heir-translate, and wrapping up the resulting py_library/py_test rule.

lib/Dialect/JaxiteWord/IR/JaxiteWordOps.td
```diff
@@ -80,6 +80,9 @@ def GenParamsOp : JaxiteWord_Op<"gen_params"> {
     - numEvalMult: Number of evaluation multiplications
   }];
   let arguments = (ins
+    JaxiteWord_PublicKey:$publicKey,
+    JaxiteWord_PrivateKey:$secretKey,
```
j2kun: Doesn't the server need to generate parameters independently of any particular user? Taking a public/private/eval key as input seems like it would make it impossible to have a user-independent FHE program.

And for that matter, wouldn't you need to decide on all these parameters before generating key material? Surely the evaluation key depends on the data pre-computed by the crypto context at initialization...

j2kun (commented):
I am a bit confused why the codegen logic is so complicated all of a sudden. The JaxiteWord ops are supposed to be a direct mirror of the CROSS API, so each emitter should be at most 1-2 lines of code to call the corresponding API. But these emitters are full of reshaping, member accesses, etc. What happened?

Zohaib58:
> I am a bit confused why the codegen logic is so complicated all of a sudden. The JaxiteWord ops are supposed to be a direct mirror of the CROSS API, so each emitter should be at most 1-2 lines of code to call the corresponding API. But these emitters are full of reshaping, member accesses, etc. What happened?

Hi Jeremy,

Lowering the matvec example helped identify a few places where the HEIR emitter currently needs extra compatibility logic around CROSS APIs, mainly for reshaping, type normalization, and layout handling. Since the goal is to keep the emitter as simple and close to the generated operations as possible, I wanted to summarize the places where small CROSS-side API adjustments could make the integration cleaner.

@JianmingTONG I have mentioned the issues, followed by the possible CROSS-side changes that would help keep the HEIR emitter clean: 

### 1. Missing ciphertext add/sub operators

**Issue**: CROSS has level-specific operators such as he_mul, he_rot, and he_rescale, but there is no equivalent he_add or he_sub.

Because of this, the emitter has to manually manipulate ciphertext limbs for Add, Sub and also handle modular reduction itself.

**Proposed solution**: Add he_add[level] and he_sub[level] operators that:

accept tiled Polynomial ciphertexts,
return tiled Polynomial ciphertexts,
handle modular reduction internally.

**Alternatively**, CROSS could provide add/sub APIs that infer the level from ct.moduli.

### 2. Mixed return types: Polynomial vs raw jnp.ndarray

**Issue**: Most CROSS operators return Polynomial, but hemul_no_relin() returns a raw 3-element jnp.ndarray, while relinearize() returns a Polynomial. This makes the multiplication path type-inconsistent:

Because of this, the emitter has to handle both cases and check whether the input has a .polynomial field or is already raw array data.

**Proposed solution**: Make hemul_no_relin() return a Polynomial, or introduce a typed 3-element ciphertext wrapper for unrelinearized multiplication results. This would make the MulNoRelinOp -> RelinOp chain consistent with the rest of the CROSS operator API.

### 3. Inconsistent data layout: flat vs tiled

**Issue**: After program_initialization, most CROSS operators expect tiled ciphertext layout:
(batch, 2, r, c, num_moduli)
But some APIs still produce or expect flat layout:
(batch, 2, degree, num_moduli)

This forces the emitter to repeatedly reshape between flat and tiled layouts.

**Proposed solution**: Standardize the layout used after program_initialization. Ideally, CROSS operators should consistently accept and return tiled Polynomial objects.

### 4. HERot.rotate() returns a mutable internal reference

**Issue**: HERot.rotate() currently returns self.ct_in. This means the returned object is not an independent result. Two consecutive rotations can overwrite the first result because the same internal object is reused.


**Proposed solution**: HERot.rotate() should return a fresh copy, similar to how relinearize() already does:

copy.copy(self.ct_obj)

This looks like a small fix around herot.py line 328 in CROSS_DEV.

### 5. encrypt() returns non-tiled Polynomial

**Issue**: encrypt() currently returns a Polynomial with shape:
(batch, 2, degree, num_moduli)
However, after program_initialization, the rest of the operators expect tiled layout:
(batch, 2, r, c, num_moduli)
So every op after encryption needs _ensure_poly or manual reshaping.


**Proposed solution**: When program_initialization has been called, encrypt() should return ciphertexts in tiled layout directly.

### 6. decrypt() expects flat layout only

**Issue**: decrypt() currently assumes flat ciphertext layout:
(batch, num_elements, degree, num_moduli)
But after program_initialization, ciphertexts are usually tiled:
(batch, 2, r, c, num_moduli)
So the emitter has to manually reconstruct a flat Polynomial before calling decrypt.


**Proposed solution**: decrypt() should accept tiled Polynomial ciphertexts directly and reshape internally when needed.

### 7. Plaintext layout mismatch with tiled ciphertext

**Issue**: Plaintexts produced by encode() have shape:
(1, 1, degree, num_moduli)
But ciphertexts after program_initialization use tiled layout:
(batch, 2, r, c, num_moduli)

So for plaintext-ciphertext ops, the emitter has to manually slice and reshape the plaintext, for example:

pt.polynomial[0, 0, :, :num_moduli].reshape(r, c, num_moduli)


**Proposed solution**: Plaintext-ciphertext operators should accept plaintexts in the layout produced by encode() and internally reshape them to match the tiled ciphertext layout.

### 8. Decode returns complex slots for real-valued outputs

**Issue**: CROSS `decode()` returns complex CKKS slot values, while the lowered HEIR examples currently expect real-valued tensors. Because of this, the emitter has to append `.real` and reshape the decoded vector back to the MLIR result shape.

JianmingTONG:
Thank you for clearly organizing these change requests!
Since this is the first I’m hearing of them, I’d like to better understand the rationale before making any changes. I suggest we review them in our next meeting with @j2kun  so we can align on why these specific updates are necessary.

PR #3101, opened 2026-06-22
restore python3.10 compatibility

j2kun:
This change restores python3.10 compatibility and adds python3.10 and python3.14 to the cibuildwheel target builds.


AlexanderViand (approved):
Thanks!

EDIT: not sure why the build is failing, but I assume it's not because of the 3.11->3.10 changes here

j2kun:
Yeah that fix is incoming

PR #3104, opened 2026-06-22
Enable buildbuddy in CI workflows build_and_test.yml and docs.yml

j2kun:
This change adds support for buildbuddy in the CI workflows.

However, due to GitHub's security isolation, we can't directly load the buildbuddy API key used for remote builds from a GH pull request. So we have a (public) read-only key that at least allows one to pull in cached build artifacts.

The configuration here also (hopefully) sets up the build_and_test workflow that runs on main to use RBE, which I can't test until this is merged. But once that workflow runs, then the Content addressable storage cache should reflect in faster pull_request builds with the read-only key.

https://heir.buildbuddy.io/invocation/4dd62bd8-bb88-4049-8f0a-2988d008d655

Includes instructions on using buildbuddy as a developer in the README.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -22,17 +22,26 @@ common --@rules_python//python/config_settings:bootstrap_impl=script
 # default in bazel 9
 startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
 
-# buildbuddy
-# for remote execution of the build. Add `common --config=remote` to
-# .bazelrc.user
-common:remote --bes_backend=grpcs://heir.buildbuddy.io
-common:remote --bes_results_url=https://heir.buildbuddy.io/invocation/
-common:remote --remote_cache=grpcs://heir.buildbuddy.io
-common:remote --remote_cache_compression
-common:remote --remote_timeout=10m
+# BuildBuddy
+build --bes_results_url=https://heir.buildbuddy.io/invocation/
+build --bes_backend=grpcs://heir.buildbuddy.io
+common --noremote_upload_local_results # Uploads logs & artifacts without writing to cache
+common --remote_cache=grpcs://heir.buildbuddy.io
+common --remote_cache_compression
+common --remote_header=x-buildbuddy-api-key=IqGGzfrvMgDJh927qtrw
+common --remote_timeout=10m
+
+# This enables remote build execution and writing to the remote cache, which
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on:
       labels: ubuntu-24.04-64core
     env:
-      BUILDBUDDY_API_KEY: ${{ github.event_name != 'pull_request' && secrets.BUILDBUDDY_API_KEY || vars.BUILDBUDDY_READ_ONLY_API_KEY }}
+      BUILDBUDDY_API_KEY: ${{ github.event_name != 'pull_request' && secrets.BUILDBUDDY_API_KEY || '7eYz4UY70YSrT55wmjWV' }}
```
asraa, line 18: wdyt about putting the read only key string directly into the .bazelrc for `build:common` or ci? that way we sync across all the workflows

.bazelrc
```diff
@@ -22,17 +22,26 @@ common --@rules_python//python/config_settings:bootstrap_impl=script
...
-common:remote --remote_cache=grpcs://heir.buildbuddy.io
-common:remote --remote_cache_compression
-common:remote --remote_timeout=10m
+# BuildBuddy
+build --bes_results_url=https://heir.buildbuddy.io/invocation/
+build --bes_backend=grpcs://heir.buildbuddy.io
+common --noremote_upload_local_results # Uploads logs & artifacts without writing to cache
+common --remote_cache=grpcs://heir.buildbuddy.io
+common --remote_cache_compression
+common --remote_header=x-buildbuddy-api-key=IqGGzfrvMgDJh927qtrw
```
asraa, line 31: this one is different than the one in CI "7eYz4UY70YSrT55wmjWV"? could you add a comment for this (is this read-only as well? hopefully this isn't your personal one)

asraa (approved):

.github/workflows/build_and_test.yml
j2kun: I tried that, but then this env setting overrides it via the flag with an empty string.

.bazelrc
j2kun: Indeed, two separate read-only keys. I wanted to make them different for metrics tracking purposes (if one key starts being abused, we can know where it's coming from).

asraa (approved):

PR #3109, opened 2026-06-23
Made ConversionUtils more robust and cover more of the tensor dialect

crockeea:
- Added explicit uniform checks for unshaped tensors or tensors with a dynamic shape to robustly limit the coverage of these conversions. Note that `ConvertAny` still supports dynamic shapes, but it is the only one that does.
- Added support for `tensor.from_elements` where the output shape is multi-dimensional; previously it only handled the case of a 1D output. Added a test for this case.
- Added support for `tensor.extract_slice` via `ConvertOffsetSizeStrideOp`
- Added support for `tensor.insert_slice` via `ConvertOffsetSizeStrideOp`
- Added support for `tensor.reshape`

### DIFF
--- a/lib/Utils/ConversionUtils.cpp
+++ b/lib/Utils/ConversionUtils.cpp
@@ -3,6 +3,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <type_traits>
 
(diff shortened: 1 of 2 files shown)


crockeea:
Is CI using BuildBuddy? 42 minutes and counting to validate the PR.

j2kun (approved):

PR #3126, opened 2026-06-25
Parse ModArith constants as modulus's bit-width

crockeea:
This simple test fails on both lines, for different reasons.
```
func.func @test_large_decimal_constant_syntax() {
  %scalar = mod_arith.constant 12535824225335233 : !Zp_i64
  %dense = mod_arith.constant dense<[12535824225335233, 0]> : !Zp_i64_vec
  return
}
```

For the first line, 12535824225335233 is a 53-bit constant, but it triggers an error in the ConstantOp verifier that the constant is too big to fit in the modulus. For some reason, MLIR's int parser makes it a 68-bit value.

For the second line, existing code in the ConstantOp parser sets 0 to 4 bits, but this is incompatible with a non-zero constant (of any size other than 4 bits).

This PR fixes these issues by forcing all ModArith constants to have the same bit-width as the modulus (after checking that it in fact fits).

Assisted by AI.

### DIFF
--- a/lib/Dialect/ModArith/IR/ModArithOps.cpp
+++ b/lib/Dialect/ModArith/IR/ModArithOps.cpp
@@ -223,7 +223,23 @@ LogicalResult BarrettReduceOp::verify() {
 }
 
 ParseResult ConstantOp::parse(OpAsmParser& parser, OperationState& result) {
-  unsigned minBitwidth = 4;  // bitwidth assigned by parser to integer `1`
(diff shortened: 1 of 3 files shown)


j2kun:
Wow, this led me down a bit of rabbit hole.

https://github.com/llvm/llvm-project/pull/205947/changes should tighten the bit width upstream, but I still think this code should be defensive and so I will review it now.

j2kun (approved):

PR #3129, opened 2026-06-26
Fixed ModDown step of lwe::keySwitchInner

### LINKED ISSUE #2987: CKKS: basis conversion and modulus switching
While having a look at the [CKKS key switch lowering](https://github.com/google/heir/blob/main/lib/Dialect/CKKS/Transforms/DecomposeKeySwitch.cpp#L34), I noticed that the last step (Step 4: Remove the key-switch primes) incorrectly performs an [LWE level basis conversion op](https://github.com/google/heir/blob/main/lib/Dialect/LWE/IR/LWEOps.td#L324) from `QP` to `Q` instead of a modulus switch from `QP` to `Q`. These operations would correspond to `Conv` (basis conversion) and `ModDown` (modulus switching) in Appendix A of [Faster Bootstrapping for CKKS with Less Modulus Consumption](https://eprint.iacr.org/2025/1403.pdf).

While investigating how `ModDown` could be implemented I also found the following:

- The [LWE level basis conversion op](https://github.com/google/heir/blob/main/lib/Dialect/LWE/IR/LWEOps.td#L324) gets [lowered](https://github.com/google/heir/blob/main/lib/Dialect/LWE/Conversions/LWEToPolynomial/LWEToPolynomial.cpp#L631) to the [RNS level basis conversion](https://github.com/google/heir/blob/main/lib/Dialect/RNS/IR/RNSOps.td#L72).
- The RNS level basis conversion op then gets lowered to further RNS and ModArith level operations [here](https://github.com/google/heir/blob/main/lib/Dialect/RNS/Transforms/LowerConvertBasis.cpp#L55) with Garner's algorithm.

I propose the following fixes:

- Remove the LWE level basis conversion op. The direct translation from LWE basis conversion to RNS basis conversion is an indication that basis conversion should live in the RNS dialect and not the LWE one.
- Define a `ModUp` (basis extension from `Q` to `QP`) and a `ModDown` (modulus switching from `QP` to `Q`) operation on the LWE dialect following the standard naming convention from Section 2.2 of [HK19](https://eprint.iacr.org/2019/688.pdf). These ops will use the RNS level base conversion when lowered.
- Replace the usages ([basis extension](https://github.com/google/heir/blob/main/lib/Dialect/CKKS/Transforms/DecomposeKeySwitch.cpp#L120) and [modulus switching](https://github.com/google/heir/blob/main/lib/Dialect/CKKS/Transforms/DecomposeKeySwitch.cpp#L146)) of the current LWE level basis conversion with the newly added `ModUp` and `ModDown` ops.
- Reimplement the RNS level basis conversion lowering to use the fast base conversion algorithm ([Section 2.2](https://eprint.iacr.org/2019/688.pdf)), which is the standard algorithm used in CKKS (hybrid) key switching implementations. The Garner's algorithm implementation could optionally be kept as an exact basis conversion algorithm.

If you think this is a good direction I can start putting together PRs for these. I am not confident about the compiler/MLIR side of things, so any suggestions are welcome.

tagging @AlexanderViand, @crockeea, @j2kun

crockeea:
This should fix #2987.

In short, the final step of keyswitch should be a "mod down", which means
- pull off the special-prime RNS components of the dot product result
- basis-convert them to the input RNS type
- subtract the previous result from the input-basis-components of the dot-product result

Architecture questions:
- It's annoying to have to repack the ring elements into a ciphertext just to do two ops (subtract and mul_scalar) and then unpack them again. However, LWE doesn't have ops that work on individual ring elements. Should it? If so, we could either add new ops that work specifically on ring elements, or make the existing ops work on ring elements. Note that I need this functionality for some other stuff I'm working on, so this isn't a one-off need.
- I would really have *liked* to make an `lwe::ModDown` op. But it would have been annoying to use: DecomposeKeySwitch would product an `lwe::ModDown` op, which would then need to be decomposed *again* (via a separate transform) into smaller LWE ops. It seems weird to make yet-another transform to decompose a single op in the LWE pipeline; I feel like I must be missing some convenient technique for this. One option is to make the LWeToPolynomial conversion pass iterative instead of single-pass. Then we could dispense with the existing DecomposeKeySwitch op and have KeySwitchInner lower to other LWE ops (including ModDown), ModDown would lower to yet more LWE ops, and finally, those could be lowered directly to Polynomial. Is that better? Unclear.

Also note that there still aren't end-to-end tests for keyswitch, which are still needed.

### DIFF
--- a/lib/Dialect/LWE/IR/LWEOps.td
+++ b/lib/Dialect/LWE/IR/LWEOps.td
@@ -176,7 +176,7 @@ def LWE_RNegateOp : LWE_Op<"rnegate", [SameOperandsAndResultType, ElementwiseMap
 def LWE_MulScalarOp : LWE_Op<"mul_scalar", [ElementwiseMappable,
     AllTypesMatch<["ciphertext", "output"]>]> {
   let summary = "Multiply an LWE ciphertext by a scalar";
-  let arguments = (ins LWECiphertextLike:$ciphertext, AnyInteger:$scalar);
+  let arguments = (ins LWECiphertextLike:$ciphertext, AnyType:$scalar);
   let results = (outs LWECiphertextLike:$output);
   let hasFolder = 1;
 }
@@ -337,6 +337,23 @@ def LWE_ConvertBasisOp : LWE_Op<"convert_basis", [Pure, DeclareOpInterfaceMethod
     LWERingElt:$output
   );
   let assemblyFormat = "$value attr-dict `:` type($value) `->` type($output)";
+}
+
+def LWE_ModDownOp : LWE_Op<"mod_down", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Scale a plaintext by the ratio of the input and output (target) moduli.";
+  let description = [{
+    Given an encryption of `x` with ciphertext modulus `Q` and target modulus `Q'`, produces
+    an encryption of round(Q' / Q * x).
+
+    The target RNS basis must be a prefix of the input's RNS basis.
+  }];
+  let arguments = (ins
+      LWECiphertext:$input,
+      TypeAttr:$targetBasis
+  );
+  let results = (outs LWECiphertext:$output);
+  let assemblyFormat = "$input attr-dict `:` type($input) `->` type($output)";
+  let hasVerifier = 1;
 }
 
 def LWE_KeySwitchInnerOp : LWE_Op<"key_switch_inner", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
--- a/lib/Dialect/LWE/Transforms/DecomposeLWEOps.cpp
+++ b/lib/Dialect/LWE/Transforms/DecomposeLWEOps.cpp
@@ -1,5 +1,6 @@
-#include "lib/Dialect/LWE/Transforms/DecomposeKeySwitch.h"
+#include "lib/Dialect/LWE/Transforms/DecomposeLWEOps.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <utility>
 
@@ -7,8 +8,11 @@
 #include "lib/Dialect/LWE/IR/LWEDialect.h"
 #include "lib/Dialect/LWE/IR/LWEOps.h"
 #include "lib/Dialect/LWE/IR/LWETypes.h"
+#include "lib/Dialect/ModArith/IR/ModArithAttributes.h"
 #include "lib/Dialect/ModArith/IR/ModArithTypes.h"
+#include "lib/Dialect/RNS/IR/RNSOps.h"
 #include "lib/Dialect/RNS/IR/RNSTypes.h"
+#include "lib/Utils/APIntUtils.h"
 #include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"    // from @llvm-project
 #include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"  // from @llvm-project
 #include "mlir/include/mlir/IR/Builders.h"               // from @llvm-project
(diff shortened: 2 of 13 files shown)


j2kun:
> LWEToPolynomial conversion pass iterative instead of single-pass

Generally speaking, this is how most passes that use `RewritePattern` already operate. The rewrite driver determines the logic. With the exception of `walkAndApplyPatterns` (which `DecomposeKeySwitch` currently uses), all the standard rewrite drivers will iteratively apply rewrite patterns until some condition is met. For dialect conversion, the iteration is until the operation produced is identified as "legal" by the `ConversionTarget`. For the greedy driver, patterns are applied until either all matches fail (it reaches a fixed point of the IR under the operation of applying all patterns).

The benefit of a separate pass like `DecomposeKeySwitch` is that you can test the individual pattern directly, rather than implicitly by testing the composition. But you can also generally include the same pattern from `DecomposeKeySwitch` in `LWEToPolynomial` alongside the other patterns, and make the standalone pass just for testing purposes.

Or, you could have `DecomposeKeySwitch` include the patterns to lower `lwe::ModDown`, and switch from the walk driver to the greedy rewrite driver (`applyPatternsGreedily`).

Generally I think it's a good architecture to separate the patterns that require type conversion from the patterns that operate within a dialect but do not convert types. So the DialectConversion pass `LWEToPolynomial` will only lower the "mos t primitive" ops within a dialect in some sense, and require as a precondition that all other ops are reduced to the primitives. The other transformations to achieve that can be organized as needed, to ensure enough granularity to add optimization passes at the right abstraction layers.

j2kun:
I don't know if the above comment will change how you approach this PR. Let me know if you still want me to review this PR as-is (and note that `//tests/Dialect/LWE/Conversions/lwe_to_polynomial:lower_relinearize_to_polynomial.mlir.test` is failing in CI)

crockeea:
I fixed the test failure (there wasn't a lowering for lwe.mul_scalar). I will ponder the architecture decision and get back to you when I'm ready for review.

crockeea:
I ended up adding `lwe::ModDownOp` and changed the `lwe-decompose-keyswitch` transform to `lwe-decompose` since it now decomposes both `KeySwitchInnerOp` and `ModDownOp`.

Ready for review.

lib/Dialect/LWE/IR/LWEOps.td
```diff
@@ -339,6 +339,19 @@ def LWE_ConvertBasisOp : LWE_Op<"convert_basis", [Pure, DeclareOpInterfaceMethod
   let assemblyFormat = "$value attr-dict `:` type($value) `->` type($output)";
 }
 
+def LWE_ModDownOp : LWE_Op<"mod_down", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>]> {
+  let summary = "Scale a plaintext by the ratio of the input and output moduli.";
+  let description = [{
```
j2kun, line 344: This could use a slightly more detailed description. Maybe make the terms use ("output moduli" vs "target basis") consistent, and include in the description the prefix constraint applied in the verifier.

tests/Dialect/LWE/Transforms/DecomposeLWEOps/keyswitch.mlir
j2kun, line 1: Nit: move these into a subdirectory and rename

```
tests/Dialect/LWE/Transforms/DecomposeLWEOps/
   keyswitch.mlir
   keyswitch_multiple.mlir
   moddown.mlir
```

lib/Dialect/LWE/Transforms/DecomposeLWEOps.cpp
```diff
@@ -20,15 +24,130 @@
...
+#define GEN_PASS_DEF_DECOMPOSELWEOPS
 #include "lib/Dialect/LWE/Transforms/Passes.h.inc"
 
+LogicalResult DecomposeModDownPattern::matchAndRewrite(
+    ModDownOp op, PatternRewriter& rewriter) const {
+  ImplicitLocOpBuilder b(op.getLoc(), rewriter);
+
+  auto ctType = dyn_cast<lwe::LWECiphertextType>(op.getInput().getType());
+  if (!ctType)
+    return op->emitOpError("Expected value to be an LWECiphertextType");
```
j2kun:
```suggestion
  auto ctType = cast<lwe::LWECiphertextType>(op.getInput().getType());
```

The tablegen forces this, so you can cast if needed. That said, since the tablegen has a single type, it should generate an accessor like

```
::mlir::TypedValue<::mlir::heir::lwe::LWECiphertextType> getInput()
```

Which would preclude the need to cast at all.

j2kun (changes requested):
Just a few nits. LGTM!

lib/Dialect/LWE/IR/LWEOps.td
crockeea: Forgot to go back and fill that in!

crockeea:
Addressed your comments and squashed.

j2kun (approved):

PR #3139, opened 2026-06-29
ci: remove unused workflows

j2kun:
This change removes now-unused runners (clang-matrix and docker-dev),
which are obviated by the hermetic llvm toolchain. Also reduces the
runner size for build_options_matrix.yml since it also runs on main
always, and hence gets the benefit of the RBE.


asraa (approved):

AlexanderViand:
If we're no longer testing the Dockerfile, should we just remove it?

PR #3148, opened 2026-06-30
Prevent optimize relinearization from mixing key basis degree in ct-pt ops

### LINKED ISSUE #3140: BFV gives a wrong result when a multiplication result is used by a later operation
I found a possible issue in HEIR's Python frontend when using BFV. The program below computes `x * y + 1`. With `x = 2` and `y = 3`, the plaintext result is `7`, but the FHE result is `-5611`.

version: heir-py 2026.6.1

```python
from heir import compile
from heir.mlir import I16, Secret

@compile(scheme="bfv")
def multiply_then_add(x: Secret[I16], y: Secret[I16]):
    return x * y + 1

if __name__ == "__main__":
    x, y = 2, 3
    print(f"plaintext::{multiply_then_add.original(x, y)}")

    multiply_then_add.setup()
    encrypted_x = multiply_then_add.encrypt_x(x)
    encrypted_y = multiply_then_add.encrypt_y(y)
    encrypted_result = multiply_then_add.eval(encrypted_x, encrypted_y)
    print(f"fhe::{multiply_then_add.decrypt_result(encrypted_result)}")
```

The output I get is:

```text
plaintext::7
fhe::-5611
```

I also checked a few nearby cases:

```python
@compile(scheme="bfv")
def multiply_only(x: Secret[I16], y: Secret[I16]):
    return x * y
```

For `x = 2` and `y = 3`, this returns:

```text
plaintext::6
fhe::6
```

```python
@compile(scheme="bgv")
def multiply_then_add(x: Secret[I16], y: Secret[I16]):
    return x * y + 1
```

For `x = 2` and `y = 3`, this returns:

```text
plaintext::7
fhe::7
```

This does not look like a generic BFV arithmetic failure, because simple BFV addition, subtraction, and direct multiplication work. It seems to happen when a ciphertext-ciphertext multiplication result is used by subsequent operations.

j2kun:
Fixes #3140

### DIFF
--- a/lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
+++ b/lib/Analysis/OptimizeRelinearizationAnalysis/OptimizeRelinearizationAnalysis.cpp
@@ -221,30 +221,48 @@ LogicalResult OptimizeRelinearizationAnalysis::solve() {
   // Some ops require a linear key basis. Yield is a special case
   // where we require returned values from funcs to be linearized.
   // TODO(#1398): determine whether we need linear key basis for modreduce.
+  // TODO(#3149): Investigate which backends actually support higher-degree
+  // ciphertext inputs to ciphertext-plaintext operations. For now, require
+  // degree-1 ciphertext inputs for all ct-pt operations.
   opToRunOn->walk([&](Operation* op) {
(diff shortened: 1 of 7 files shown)


edwjchen (approved):

edwjchen:
The conservative path forcing secret operands to be relinearized looks good!

PR #3151, opened 2026-06-30
change the default build mode to fastbuild

j2kun:
I think this should improve linker run time by stripping unused symbols.


asraa (approved):
+1

PR #3152, opened 2026-07-01
layout: Support arbitrary layout permutation for rank-2 tensors

### LINKED ISSUE #2666: Support adding arbitrary layout permutation for rank-2 tensor

VedantParanjape:
Permutation layout attributes can be used to assign random permutations of scalar values to ciphertext slots. This patch adds support for adding permutations to rank-2 tensors.

Fixes #2666

### DIFF
--- a/lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
+++ b/lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
@@ -134,7 +134,7 @@ static FailureOr<Value> implementUnpackOpStep(
 
 static FailureOr<Value> implementAssignLayoutPermutation(
     Value input, DenseIntElementsAttr permutation, Type targetTypeTy,
-    int64_t ciphertextSize, ImplicitLocOpBuilder& builder,
+    ImplicitLocOpBuilder& builder,
     const std::function<void(Operation*)>& createdOpCallback) {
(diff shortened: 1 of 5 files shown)


lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -133,63 +133,67 @@ static FailureOr<Value> implementUnpackOpStep(
 }
 
 static FailureOr<Value> implementAssignLayoutPermutation(
-    Value input, DenseIntElementsAttr permutation, int64_t ciphertextSize,
-    ImplicitLocOpBuilder& builder,
+    Value input, DenseIntElementsAttr permutation, Type targetTypeTy,
+    int64_t ciphertextSize, ImplicitLocOpBuilder& builder,
```
j2kun: Is the ciphertext size argument needed now that you have targetTypeTy (targetType?) whose last axis should be of size ciphertextSize right?

j2kun (changes requested):
LGTM, one little nit

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
VedantParanjape: yeah, I can remove it.

VedantParanjape:
Done!

j2kun (approved):

PR #3157, opened 2026-07-01
layout: Add support for permutation layout in unpackOp

VedantParanjape:
Permutation layout is supported by assign_layout op. The inverse of this op is unpack op. It is required by client decryption handles to decrypt the data and store it according to the layout attribute. This patch adds support for permutation layout to work with unpackOp.

### DIFF
--- a/lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
+++ b/lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
@@ -132,7 +132,7 @@ static FailureOr<Value> implementUnpackOpStep(
   return loop.value().getResults()[0];
 }
 
-static FailureOr<Value> implementAssignLayoutPermutation(
+static FailureOr<Value> implementLayoutPermutation(
     Value input, DenseIntElementsAttr permutation, Type targetTypeTy,
     ImplicitLocOpBuilder& builder,
     const std::function<void(Operation*)>& createdOpCallback) {
@@ -150,8 +150,11 @@ static FailureOr<Value> implementAssignLayoutPermutation(
   createdOpCallback(zeroCtxt);
   Value result = zeroCtxt.getResult();
 
-  int64_t ctBound = (tensorType.getRank() == 2) ? tensorType.getDimSize(0) : 1;
+  int64_t srcCtBound =
+      (tensorType.getRank() == 2) ? tensorType.getDimSize(0) : 1;
   int64_t srcSlotBound = tensorType.getDimSize(tensorType.getRank() - 1);
+  int64_t dstCtBound =
+      (targetType.getRank() == 2) ? targetType.getDimSize(0) : 1;
   int64_t dstSlotBound = targetType.getDimSize(targetType.getRank() - 1);
 
   for (auto it = permutation.value_begin<APInt>();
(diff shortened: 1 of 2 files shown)


lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -150,8 +150,11 @@ static FailureOr<Value> implementAssignLayoutPermutation(
   createdOpCallback(zeroCtxt);
   Value result = zeroCtxt.getResult();
 
-  int64_t ctBound = (tensorType.getRank() == 2) ? tensorType.getDimSize(0) : 1;
+  int64_t srcCtBound =
```
j2kun, line 153: The naming is a little strange here because only one of the source and destination are semantically "ct, slot," depending on the caller's context. Moreover, the non-ct-slot side is restricted (somewhat arbitrarily) to be dimension 2.

Maybe don't name the variables "ct, slot", and then document the function with a code comment for the dim-2 restriction and for the semantic ambiguity of which one is the data and which is ciphertext.

tests/Transforms/add_client_interface/dense_unpack.mlir
```diff
@@ -0,0 +1,64 @@
+// RUN: heir-opt --lower-unpack --canonicalize %s | FileCheck %s
+
+// Tests `tensor_ext.unpack` with a dense-permutation layout attribute
+
+// -----------------------------------------------------------------------------
+// Case 1: single dense permutation with a rank-1 data-semantic target.
```
j2kun, line 6: I think the design of this explicit layout attribute is insufficient for what you're trying to use it for, and this PR is showing the strain.

The original permutation attribute seems incorrect in retrospect:

```
  // A list of tuples [a, b, c, d] representing an explicit map (ct, slot) ->
  // (ct, slot) defined by f(a, b) = (c, d).
```

In fact, it's not a mapping (ct, slot) -> (ct, slot), because one side is data semantic (and restricted to rank 2) and the other is ciphertext semantic.

And then here, one must further hack around it by having the permutation attribute use rank 2 indices for the data semantic side, even though it has semantic rank 1.

Probably what it should be is

```
  // A list of tuples [a, b, ..., c, d] representing an explicit map 
  // (d0, d1, ...) -> (ct, slot)
  // defined by f(a, b, ...) = (c, d)
```

With a verifier to assert that when this is attached to an op, the ranks agree (the data semantic tensor involved has rank `denseEltsAttr.getType().getShape()[1] - 2` and numElements = `denseEltsAttr.getType().getShape()[0]`).

Then the layout attribute would naturally support any rank of data-semantic tensor.

j2kun (commented):
I think we can merge this as-is, but I think the attribute as a whole needs to be changed to avoid the special cases and confusing aspects of this PR.

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
VedantParanjape: I just realized this, I can rework this when I address the overall issue with the design of layout attribute.

tests/Transforms/add_client_interface/dense_unpack.mlir
VedantParanjape: Yes that seems like a more generic implementation. But the subtle issue with `f(a, b, ...) = (c, d)` is that it cannot be used on result attributes. How about I implement a n-dim to n-dim mapping and then let downstream passes in the ciphertext semantic verify its correctness.

j2kun: > But the subtle issue with f(a, b, ...) = (c, d) is that it cannot be used on result attributes.

I don't see why not...

VedantParanjape: > > But the subtle issue with f(a, b, ...) = (c, d) is that it cannot be used on result attributes.
> 
> I don't see why not...

because for a result attribute, we are going from ciphertext to a n-dim array, hence we would need `f(a, b) = (c, d, e, ...)`

j2kun: The layout attribute is defined to always map from the data-semantic domain to the ciphertext-semantic domain. See for example https://github.com/google/heir/blob/main/tests/Transforms/lower_unpack/lower_unpack.mlir

Then `unpack` is supposed to be smart enough to know that it needs to invert the mapping.

PR #3161, opened 2026-07-03
Change complement definition

mdgrs:
The ISL complement operation was very slow when the dimension was not a power of 2.

In my test, this reduces the timing for the convert-to-ciphertext-semantics pass of //tests/Examples/lattigo/ckks/pooling from 16s to 0.9 seconds.

### DIFF
--- a/lib/Utils/Layout/Utils.cpp
+++ b/lib/Utils/Layout/Utils.cpp
@@ -764,24 +764,29 @@ void getCtComplementPoints(const presburger::IntegerRelation& relation,
   assert(relation.getNumRangeVars() == 2 && "Expected 2 range vars.");
   assert(outputType.getRank() == 2 && "Expected 2D output type.");
 
-  // Get the range set for the ct var.
-  auto* bmap = convertRelationToBasicMap(relation, collector.ctx);
-  isl_set* set = isl_set_from_basic_set(isl_basic_map_range(bmap));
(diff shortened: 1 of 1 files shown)


j2kun:
I am surprised this is faster. Do we have any idea why?

mdgrs:
The LLM explains it like this, which makes sense to me: 

```
Two of those steps are genuinely costly on this kind of set:

1. project_out(slot) — the range set carries the layout's modulo constraint ((-196·i1 - 14·i2 - i3 + slot) mod 1024 = 0) plus the existentially-quantified domain vars. ISL models a mod with an extra existential "div" variable. Eliminating slot from a set full of existential/div variables is a quantifier-elimination step (Fourier–Motzkin over locals), not a cheap operation.
2. isl_set_complement — negating a conjunction of constraints produces a union of many disjuncts (DNF blow-up), and to make the complement exact ISL must again eliminate the existential/div variables. This is super-polynomial in the worst case, and the modulo constraints are exactly the pathological input for it.
```

When the spatial dimension is a power of two, things align nicely and there are not too many sets to keep track of, while non powers of two blow things up

j2kun (approved):

j2kun:
Will get merged Monday

asraa (approved):
interesting - nice! was this causing the regression Alex mentioned?

asraa:
merging with a small fix in `lib/Utils/Layout:UtilsTest` to add bounding dimensions on the slot variable in the test relation

PR #3162, opened 2026-07-03
use copyOnWrite for OneShotBufferize

mdgrs:
This is a bit hackish (as a way to only use copyOnWrite for preprocessing). There might very well be a better way. However, it significantly reduces compile time for the lattigo pooling test (and for all convolutional networks we are working on)


j2kun:
My concern is that this makes us need to do codegen for mixed tensor and memref semantics, when our goal is to eliminate tensor semantics at the backend. Can you give an example preprocessing function that is too slow to bufferize? Maybe we can figure out a root cause for what makes that slow and improve it.

mdgrs:
I ran this against some of the larger tests, and got some speed ups (4.45s -> 0.44s) for conv2d_dilated. However, on the Lola test, I gain nothing. I think the rolled kernels make the benefits go away. When I run conv2d_dilated test with the `experimental-disable-loop-unroll` option, all benefits from this PR evaporate. I am closing this as no longer relevant.

Since I had already made the changes, I made a small PR to make conv2d_dilated use rolled kernels:https://github.com/google/heir/pull/3259

PR #3163, opened 2026-07-05
Added support for the linalg.broadcast operation and its kernel

### LINKED ISSUE #2876: Torch-MLIR Reduction + Broadcast Pattern
Torch-MLIR generates a pattern of reduce, element-wise operation, broadcast. In vector FHE, the rotate-and-reduce sum populates all slots in the vector with the sum, not just one element. It could be helpful to remove the broadcast operation, but linalg.reduce discards dimensions, forcing the broadcast. What would be a good way to symbolize a reduction that keeps the dimensions?

Sample from torch.layernorm:
```
module {
  func.func @main(%arg0: tensor<2x64x768xf32> {secret.secret}) -> tensor<2x64x768xf32> {
    %cst = arith.constant dense<0.000000e+00> : tensor<2x64xf32>
    %cst_0 = arith.constant dense<7.680000e+02> : tensor<2x64xf32>
    %reduced = linalg.reduce ins(%arg0 : tensor<2x64x768xf32>) outs(%cst : tensor<2x64xf32>) dimensions = [2] 
      (%in: f32, %init: f32) {
        %2 = arith.addf %in, %init : f32
        linalg.yield %2 : f32
      }
    %0 = arith.divf %reduced, %cst_0 : tensor<2x64xf32>
    %1 = tensor.empty() : tensor<2x64x768xf32>
    // Additional broadcast
    %broadcasted = linalg.broadcast ins(%0 : tensor<2x64xf32>) outs(%1 : tensor<2x64x768xf32>) dimensions = [2] 
    return %broadcasted : tensor<2x64x768xf32>
  }
}

```

Could become something like:

```module {
  func.func @main(%arg0: tensor<2x64x768xf32> {secret.secret}) -> tensor<2x64x768xf32> {
    %cst = arith.constant dense<0.000000e+00> : tensor<2x64x768xf32>
    %cst_0 = arith.constant dense<7.680000e+02> : tensor<2x64x768xf32>
    // Something similar that keeps dimensions
    %reduced = linalg.reduce ins(%arg0 : tensor<2x64x768xf32>) outs(%cst : tensor<2x64x768xf32>) dimensions = [2] 
      (%in: f32, %init: f32) {
        %2 = arith.addf %in, %init : f32
        linalg.yield %2 : f32
      }
    %0 = arith.divf %reduced, %cst_0 : tensor<2x64x768xf32>
    return %0 : tensor<2x64x768xf32>
  }
}
```

Someone117:
This patch resolves https://github.com/google/heir/issues/2876 by implementing a linalg broadcast kernel, rewerting some linalg.generic operations, and removing unnecessary linalg.broadcast operations.


lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -707,7 +703,60 @@ class ConvertLinalgReduce : public ConversionBase<linalg::ReduceOp> {
...
+          broadcastsToErase.push_back(broadcastOp);
+        }
+      }
+      for (auto broadcastOp : broadcastsToErase) {
+        broadcastOp.getResults()[0].replaceAllUsesWith(op.getResults()[0]);
+      }
+    }
+
+    // if the reduce is immediately broadcasted, the broadcast is a NoOp
+    if (checkReductionThenBroadcast(op)) {
```
j2kun, line 726: This block looks like a duplicate of the previous block

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -707,7 +703,60 @@ class ConvertLinalgReduce : public ConversionBase<linalg::ReduceOp> {
...
     // Add the initial  value.
     Value result = adaptor.getInits()[0];
-    addBiasAndReplace(rewriter, op, finalOutput, result, layoutAttr);
+    Operation* addBias = makeAppropriatelyTypedAddOp(rewriter, op->getLoc(),
+                                                     finalOutput, result);
+    addBias->setAttr(kLayoutAttrName, layoutAttr);
+    setMaterializedAttr(addBias);
+    rewriter.replaceOp(op, addBias);
+
+    // if the reduce is immediately broadcasted, the broadcast is a NoOp
```
j2kun, line 712: Why exactly is it a no-op? I suspect this depends on the chosen layout. For example, we have some cases where we want to do row-summations, but the matrix is laid out in the "bicyclic" layout, which involves a diagonal traversal of the matrix. This led me to define a new, dedicated op for the combined reduce+broadcast op in https://github.com/google/heir/pull/3269

- Could we add a check on which kinds of layouts this no-op holds for?
- Would it make sense to instead use the "broadcast swap" in combination with looking for reduce + broadcast to produce the `broadcasted_reduce` op as a single op, and then match on it with a dedicated pattern?

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
```diff
@@ -754,6 +803,91 @@ class ConvertLinalgReduce : public ConversionBase<linalg::ReduceOp> {
...
+      linalg::BroadcastOp op, OpAdaptor adaptor,
+      ContextAwareConversionPatternRewriter& rewriter) const final {
+    Value input = adaptor.getInput();
+
+    LayoutAttr inputLayout = getLayoutAttr(input);
+    if (!inputLayout)
+      return rewriter.notifyMatchFailure(
+          op, "missing new layout attribute for input");
+
+    // TODO: support multi-dimension broadcasts
```
j2kun, line 869: Create a GH issue and tag the issue as `// TODO(#dddd): message`, which will allow GH to link the line of code to the issue.

j2kun (changes requested):
So overall I think it looks good. I was hoping for a few small changes:

1. Can you add the patterns from the two new passes (broadcast-canonicalizations and broadcast-swap) to `linalg-canonicalizations` instead? (And ideally in a separate PR)
2. Can you add a simple end-to-end test in tests/Examples/lattigo/ckks that runs against the Lattigo backend?
3. As far as I understand, the `implementRotateAndBroadcast` has a limitation on the layouts of its inputs: if the data matrix is laid out among multiple ciphertexts and the broadcast spills over from one ciphertext to another, it will not work with this kernel. Can we guard against this? Imagine for example having a tensor of size 256x256 packed into two ciphertexts of size 32768 in a row-major fashion. reducing along the leading axis and broadcasting would require mapping from one ciphertext to another (I believe I didn't get that backwards...). I think the "right" way to handle this generally is to create a tensor_ext.remap op and write down the broadcast as an explicit permutation (or an ISL mapping string) and then use the existing lowerings to generate a shift network. But that can also be beyond the scope of this PR.

lib/Transforms/ConvertToCiphertextSemantics/ConvertToCiphertextSemantics.cpp
j2kun: If we do want to merge the reduce + broadcast ops, then the KernelImplementation code in this PR can probably be merged with https://github.com/google/heir/blob/cfae7f2d702990b202f2a321e6628e2c60e3f16a/lib/Kernel/KernelImplementation.h#L517

To my eye, your implementation is more general than the linked one (because it supports non-power-of-two and is smarter about its axis analysis).

PR #3168, opened 2026-07-06
Improved batchnorm fusing

j2kun:
This PR improves the support for fusing batch norm into matmul or other prior linalg ops. I found a few cases where it wasn't properly fusing (involving broadcasts between the dense constant and the mul op). I also noticed that the `outs` accumulator was being overridden, and even though we don't typically have nonempty/nonzero `outs`, this PR adds support for that case.

### DIFF
--- a/lib/Transforms/LinalgFuseLinearOps/LinalgFuseLinearOps.cpp
+++ b/lib/Transforms/LinalgFuseLinearOps/LinalgFuseLinearOps.cpp
@@ -12,6 +12,7 @@
 #include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"      // from @llvm-project
 #include "mlir/include/mlir/Dialect/Linalg/IR/Linalg.h"    // from @llvm-project
 #include "mlir/include/mlir/Dialect/Tensor/IR/Tensor.h"    // from @llvm-project
+#include "mlir/include/mlir/Dialect/Traits.h"              // from @llvm-project
 #include "mlir/include/mlir/Dialect/Utils/StructuredOpsUtils.h"  // from @llvm-project
 #include "mlir/include/mlir/IR/AffineExpr.h"    // from @llvm-project
 #include "mlir/include/mlir/IR/AffineMap.h"     // from @llvm-project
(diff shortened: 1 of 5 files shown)


tests/Transforms/linalg_fuse_linear_ops/preexisting_bias.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: heir-opt --linalg-fuse-linear-ops %s | FileCheck %s
+
+// Tests that fusing a multiplication (scale) into a linalg.matmul updates the
+// pre-existing outs accumulator appropriately.
+
+// CHECK: func.func @matmul_preexisting_bias
+// CHECK: %[[EMPTY_W:.*]] = tensor.empty() : tensor<3x4xf32>
+// CHECK: %[[BROADCAST_W:.*]] = linalg.broadcast ins(%arg3 : tensor<4xf32>) outs(%[[EMPTY_W]] : tensor<3x4xf32>) dimensions = [0]
+// CHECK: %[[SCALED_W:.*]] = arith.mulf %arg1, %[[BROADCAST_W]] : tensor<3x4xf32>
```
asraa:
```suggestion
// CHECK: func.func @matmul_preexisting_bias
// CHEK-SAME: %[[arg1:.*]]: tensor<3x4xf32>
// CHECK-SAME: %[[arg3:.* tensor<4xf32>
// CHECK: %[[EMPTY_W:.*]] = tensor.empty() : tensor<3x4xf32>
// CHECK: %[[BROADCAST_W:.*]] = linalg.broadcast ins(%[[arg3]] : tensor<4xf32>) outs(%[[EMPTY_W]] : tensor<3x4xf32>) dimensions = [0]
// CHECK: %[[SCALED_W:.*]] = arith.mulf %[[arg1]], %[[BROADCAST_W]] : tensor<3x4xf32>
```

lib/Transforms/LinalgFuseLinearOps/LinalgFuseLinearOps.cpp
```diff
@@ -36,30 +37,104 @@ namespace heir {
...
   Value rhs = op.getRhs();
 
-  auto defOp = lhs.getDefiningOp();
-  Value other = rhs;
-  if (!defOp || !isa<linalg::LinalgOp>(defOp)) {
-    defOp = rhs.getDefiningOp();
-    other = lhs;
+  auto isLinearOp = [](Operation* defOp) {
+    return defOp && isa<linalg::LinalgOp>(defOp) &&
+           !isa<linalg::BroadcastOp, linalg::FillOp, linalg::TransposeOp>(
```
asraa, line 107: i feel like this is too permissive - e.g. a generic op would survive this check and allow anything

lib/Transforms/LinalgFuseLinearOps/LinalgFuseLinearOps.cpp
```diff
@@ -36,30 +37,104 @@ namespace heir {
...
+    moveSliceBefore(operand.getDefiningOp(), target);
+  }
+  op->moveBefore(target);
+}
+
+// Given a value, walk the use-def chain backwards to try to find a scalar SSA
+// value, passing through ops that only splat/reshape a scalar as a dense
+// tensor. Alternatively, if the end of the chain is a tensor type, check
+// to see if it can be broadcast to a given weightsType.
+FailureOr<OpFoldResult> findOriginalScalar(Value scaleVal, Type weightsType) {
```
asraa, line 141: just a curiousity: could you try OpBuilder::tryFold? I've always been curious about using that to simplify / fold the value

asraa (approved):

lib/Transforms/LinalgFuseLinearOps/LinalgFuseLinearOps.cpp
j2kun: So I don't think that is a problem for `fuseScaleOrDivIntoLinearOp` the caller follows up this function with an explicit type switch over the set of allowed ops. For `fuseAddOrSubIntoLinearOp` it supports any linalg op that is a `DestinationStyleOpInterface`, and so what I need to do is opt-in to ops that actually accumulate to their `outs` additively. TIL that not all ops do this. So I will figure out which ones do, and indeed I could even support a `linalg.generic` if I see an addition into the block argument corresponding to the out param.

lib/Transforms/LinalgFuseLinearOps/LinalgFuseLinearOps.cpp
j2kun: Done. I went one step further and used `rewriter.createOrFold` which does the folding at op creation time.

PR #3169, opened 2026-07-07
Rotom: Layout Representation updates for tiled layouts

edwjchen:
Three attribute-layer changes preparing the layout search for split and tiled packings:

Part 1 of https://github.com/google/heir/pull/2980.

 - preprocessLayoutData produces a LayoutPiece list and a deduped axes table that maps one ISL domain variable per logical tensor dim. A traversal piece is a mixed-radix digit (i / divBy) mod modBy of its axis's index, so an axis split into several pieces (tiled or ct/slot-straddling layouts) shares one ISL domain variable. 
 - The axes table canonicalizes to ascending dim id. Previously row- and column-major materialized identical relations, making conversions between them silently free when lowered to the ISL domain.
 - The dim assembly separates ciphertext dims from slot dims with `|` (e.g., `[ct pieces | slot pieces]`). Unused slots is written as an explicit gap piece. 
- Replication and gap dims print as R and G ([R:4:1], [G:4:1]); the numeric ids -1/-2 are still accepted on input and round-trip to the letter forms.

### DIFF
--- a/lib/Dialect/Rotom/IR/RotomAttributes.cpp
+++ b/lib/Dialect/Rotom/IR/RotomAttributes.cpp
@@ -2,9 +2,11 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <map>
 #include <utility>
 #include <vector>
 
+#include "llvm/include/llvm/ADT/DenseMap.h"           // from @llvm-project
 #include "llvm/include/llvm/ADT/DenseSet.h"           // from @llvm-project
 #include "llvm/include/llvm/ADT/STLExtras.h"          // from @llvm-project
 #include "llvm/include/llvm/ADT/SmallVector.h"        // from @llvm-project
@@ -20,9 +22,7 @@ namespace mlir {
 namespace heir {
 namespace rotom {
 
-namespace {
-
-static size_t inferCtPrefixLen(ArrayRef<DimAttr> dims, int64_t n) {
+size_t inferCtPrefixLen(ArrayRef<DimAttr> dims, int64_t n) {
   int64_t nRem = n;
   size_t i = dims.size();
   while (i > 0) {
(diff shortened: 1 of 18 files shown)


lib/Dialect/Rotom/IR/RotomAttributes.cpp
```diff
@@ -54,12 +53,29 @@ static int64_t computeImplicitFrontGap(ArrayRef<DimAttr> dims, int64_t n) {
...
+// descriptor used to emit ciphertext addresses; also the validity check
+// behind `LayoutAttr::verify`.
+//
+// On success `LayoutData` describes, in packing order: `pieces`, each tagged
+// Traversal, Replication, or Gap with `pieceIndex` into its per-kind dim list;
+// for each traversal piece the mixed-radix digit (i / pieceDivBy) mod
+// pieceModBy read from tensor index i, so an axis's pieces share one domain
+// variable; and `ctPrefixLen`, the ciphertext/slot split (with an implicit
+// front gap when packed content leaves slots unfilled). See `LayoutData` in
+// the header for exact field semantics.
```
j2kun: I find the indirection of `pieceIndex` a little bit confusing here. What is stopping you from having a `LayoutData` have a vector of `LayoutPiece` that contains all the information for one piece, and perhaps a `map` from axis index to the set of pieces for that axis?

Even if you have to add a second pass over the pieces to construct the map, it seems worth it, and the data sizes here are quite small.

lib/Dialect/Rotom/IR/RotomAttributes.cpp
```diff
@@ -69,25 +85,118 @@ static FailureOr<LayoutData> preprocessLayoutData(ArrayAttr dims, int64_t n,
...
+      if (inserted) data.traversalDims.push_back(d);
+      data.pieceIndex.push_back(it->second);
       data.pieces.push_back(LayoutPieceKind::Traversal);
+      data.pieceDivBy.push_back(d.getStride());
       continue;
     }
     return failure();
   }
 
+  // Canonicalize the deduped traversal dims to ascending dim id. This is a
```
j2kun: Is this a good time to use `std::map` for `traversalIndexForDim`, which already keeps the keys in sorted order? (unlike DenseMap)

lib/Dialect/Rotom/IR/RotomAttributes.cpp
```diff
@@ -69,25 +85,118 @@ static FailureOr<LayoutData> preprocessLayoutData(ArrayAttr dims, int64_t n,
...
+    const int64_t dim = data.originalDims[p].getDim();
+    ++pieceCount[dim];
+    auto [it, inserted] = dimFullExtent.try_emplace(dim, 1);
+    it->second *= data.originalDims[p].getSize();
+  }
+
+  for (size_t p = 0; p < data.pieces.size(); ++p) {
+    if (data.pieces[p] != LayoutPieceKind::Traversal) continue;
+    if (pieceCount[data.originalDims[p].getDim()] == 1) {
+      // A whole dim packed as one piece keeps the legacy behavior: the stride
```
j2kun: nit: don't talk about "legacy" behavior; someone reading this in the future will have no idea what the legacy behavior was. The comments should only describe the current behavior/reasons.

j2kun (changes requested):

j2kun:
I guess one thing I don't quite understand about the design here is explained with this example:

```
#rotom.layout<n = 4, dims = [[0:8:1]]> 
```

Here you'd split the single dim into two pieces because it has size 8 and there are only 4 slots, right?

What is the benefit of doing that instead of, say, explicitly representing what piece is slots and what piece is ciphertext?

```
#rotom.layout<n = 4, ct_dims = [[0:2:4]], slot_dims = [[0:4:1]]> 
```

I guess what I'm saying is: if you explicitly represented this, would you be able to avoid most of the work in this PR?

lib/Dialect/Rotom/IR/RotomAttributes.cpp
edwjchen: Refactored to have a list of `LayoutPiece`s! Previously, we had separate vectors corresponding to different traversal dimensions kinds and used this index to track a piece's position in the list.

edwjchen:
The layout representation is now updated to have a clear separation between ct and slot dimensions:
```
#rotom.layout<n = 4, dims = [[0:2:4]  | [0:4:1]]>
```
where LayoutPieces before the `|` correspond to ct dims and after the `|` correspond to slot dims.  

The remaining changes in this PR are for adding robustness checks to constructing a Rotom layout (in `preprocessLayoutData`) and facilitating lowering a Rotom layout into an ISL relation (by canonicalizing the relationship between LayoutPieces and the original tensor axes.

j2kun (approved):

PR #3170, opened 2026-07-07
Rotom: Rolled Layout Representation

edwjchen:
Part 2 of https://github.com/google/heir/pull/2980  and stacked on https://github.com/google/heir/pull/3169 

Two extensions to roll(from, by) semantics that make diagonal packings expressible beyond the square, equal-extent case.

1. Roll partners may be replication or gap dims. Previously both roll operands had to be traversal dims. Now the by (second) dim may also be a replication or gap dim.

2. Roll extents need not match (mod-from). roll(i, j) now rewrites dims[i]'s index to (idx_i  - idx_j) mod extent(dims[i]), well-defined for any partner extent - a smaller partner covers a prefix of the rotations, a larger one wraps. This unlocks rolled placements for rectangular (non-square-k) contractions.

3. Rolled-by-gap bound. A rolled-by gap partner is bounded to the rolled dim's extent.

PR #3174, opened 2026-07-08
 Rotom: Roll arguments can name a whole tensor axis

edwjchen:
Part 3 of https://github.com/google/heir/pull/2980 and stacked on https://github.com/google/heir/pull/3170

A Rotom roll decorates a layout: `roll(from, by)` rewrites the `from` index by subtracting the `by` argument's index. Until now a roll argument could only be a **piece**, a position in the `dims` list. This PR lets it also be a whole tensor **axis**, spelled `axis N` and stored as `-(axis+1)` in the flat rolls array.

```mlir
// Piece arguments: Halevi-Shoup diagonal of a 4x4 matrix, slot j of
// ciphertext i holds A[i, (j - i) mod 4].
#diag  = #rotom.layout<n = 4, rolls = [(1, 0)], dims = [[0:4:1] | [1:4:1]]>

// Axis 1 is split across two pieces, so the whole-axis roll names it.
// How to read this: `axis 1` refers to `[1:4:4], [1:4:1]` and `2` refers to piece 2 or `[0:16:1]`
#split = #rotom.layout<n = 16, rolls = [(axis 1, 2)],
                       dims = [[1:4:4], [1:4:1] | [0:16:1]]>
```
- Piece FROM rewrites only that piece's mixed-radix digit, leaving the axis's other digits untouched. This is the original per-piece reading, now materialized correctly on split axes (no borrow leaks across digits).
- Axis FROM rewrites the whole axis index modulo its full extent; each piece then takes its digit of the rolled index, so the shift borrows across digits.

### DIFF
--- a/lib/Dialect/Rotom/IR/RotomAttributes.cpp
+++ b/lib/Dialect/Rotom/IR/RotomAttributes.cpp
@@ -227,25 +227,40 @@ static ParseResult parseLayoutDims(AsmParser& parser,
   }
 }
 
+// One argument of a roll pair: a bare non-negative integer is a dims-list
+// position (one piece); `axis N` names the whole tensor axis N.
+static ParseResult parseRollArg(AsmParser& parser, int64_t& encoded) {
+  const bool isAxis = succeeded(parser.parseOptionalKeyword("axis"));
+  int64_t value;
+  if (parser.parseInteger(value)) return failure();
+  if (value < 0) {
+    return parser.emitError(parser.getNameLoc())
+           << (isAxis ? "an axis roll argument must name a non-negative "
+                        "tensor axis"
+                      : "a piece roll argument must be a non-negative dims "
+                        "position (spell a whole-axis argument as `axis N`)");
+  }
+  encoded = encodeRollArg({isAxis, value});
+  return success();
+}
+
 static ParseResult parseLayoutRolls(AsmParser& parser,
                                     SmallVector<int64_t>& rolls) {
   if (parser.parseLSquare()) return failure();
   if (succeeded(parser.parseOptionalRSquare())) return success();
 
+  // Each entry is one `(from, by)` pair, parenthesized -- the form the
+  // printer emits, so written and round-tripped layouts read alike.
   while (true) {
-    if (succeeded(parser.parseOptionalLParen())) {
-      int64_t from;
-      int64_t to;
-      if (parser.parseInteger(from) || parser.parseComma() ||
-          parser.parseInteger(to) || parser.parseRParen())
-        return failure();
-      rolls.push_back(from);
-      rolls.push_back(to);
-    } else {
-      int64_t value;
-      if (parser.parseInteger(value)) return failure();
-      rolls.push_back(value);
+    int64_t from;
+    int64_t by;
+    if (parser.parseLParen() || failed(parseRollArg(parser, from)) ||
+        parser.parseComma() || failed(parseRollArg(parser, by)) ||
+        parser.parseRParen()) {
+      return failure();
     }
+    rolls.push_back(from);
+    rolls.push_back(by);
 
     if (succeeded(parser.parseOptionalComma())) continue;
     return parser.parseRSquare();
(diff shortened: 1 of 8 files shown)


j2kun:
@edwjchen Could you rebase this one?

lib/Dialect/Rotom/IR/RotomAttributes.td
```diff
@@ -55,11 +55,23 @@ def Rotom_LayoutAttr : Rotom_Attr<"Layout", "layout"> {
...
 
-    Optional **rolls** encode a `roll(i,j)` metadata object: each pair `(i, j)`
-    indexes into the `dims` array (the flattened `ct_dims + slot_dims` list) and
-    rewrites `dims[i]`'s index to `(idx_i - idx_j) mod size(dims[i])`. The two
-    extents need not match: the shift reduces modulo the rolled dim's extent, so
-    a smaller partner covers a prefix of the rotations and a larger one wraps.
+    Optional **rolls** encode `roll(from, by)` metadata objects, applied left
+    to right. Each endpoint is either a *piece* -- a position in the `dims`
+    list, spelled as a bare integer -- or a whole tensor *axis*, spelled
+    `axis N` (legal only when axis `N` is packed as more than one piece; the
```
j2kun, line 61: Put an example or two in a code block.

lib/Dialect/Rotom/IR/RotomAttributes.td
```diff
@@ -55,11 +55,23 @@ def Rotom_LayoutAttr : Rotom_Attr<"Layout", "layout"> {
...
+    piece spelling is canonical for an unsplit axis, where the two coincide).
+
+    A piece FROM rewrites that piece's own mixed-radix digit in place,
+    `digit(from) <- (digit(from) - shift(by)) mod extent(from)`, leaving
+    the axis's other digits untouched. An `axis` FROM rewrites the whole axis
+    index modulo its full extent, and each piece then takes its digit of the
+    rolled index -- the shift borrows across digits, which no combination of
```
j2kun: I also feel this could use an example, maybe with just two digits, to help clarify what it means to 'take its digit of the rolled index' and 'borrow across digits'

lib/Dialect/Rotom/IR/RotomAttributes.td
```diff
@@ -55,11 +55,23 @@ def Rotom_LayoutAttr : Rotom_Attr<"Layout", "layout"> {
     with an explicit gap piece (e.g. `[G:4:1]`).
     See [Section 4.2 of the Rotom paper](https://eprint.iacr.org/2025/1319.pdf).
 
-    Optional **rolls** encode a `roll(i,j)` metadata object: each pair `(i, j)`
-    indexes into the `dims` array (the flattened `ct_dims + slot_dims` list) and
-    rewrites `dims[i]`'s index to `(idx_i - idx_j) mod size(dims[i])`. The two
-    extents need not match: the shift reduces modulo the rolled dim's extent, so
-    a smaller partner covers a prefix of the rotations and a larger one wraps.
+    Optional **rolls** encode `roll(from, by)` metadata objects, applied left
+    to right. Each endpoint is either a *piece* -- a position in the `dims`
```
j2kun: "endpoint" is a bit of a weird term to use here, because the arguments to `roll` are not entities with a geometry (interval, segment) that has endpoints.

I see `RollSpec` above for the pair of `from` and `by`, so maybe call `RollEndpoint` instead `RollArg` and you can refer to it in prose as "the `by` argument" or "the `from` roll arg"?

lib/Dialect/Rotom/IR/RotomAttributes.cpp
```diff
@@ -227,25 +227,44 @@ static ParseResult parseLayoutDims(AsmParser& parser,
...
+      if (failed(parseRollEndpoint(parser, from)) || parser.parseComma() ||
+          failed(parseRollEndpoint(parser, to)) || parser.parseRParen())
         return failure();
-      rolls.push_back(from);
-      rolls.push_back(to);
     } else {
-      int64_t value;
-      if (parser.parseInteger(value)) return failure();
-      rolls.push_back(value);
+      if (failed(parseRollEndpoint(parser, from)) || parser.parseComma() ||
```
j2kun: nit: I sort of missed this last time, but I think it would be better to just have one option for parsing and printing, and leaving the parens explicit seems better than leaving them out.

lib/Dialect/Rotom/Utils/RotomTensorExtLayoutLowering.cpp
```diff
@@ -18,20 +18,40 @@
...
   }
   return failure();
 }
 
+/// Product of the extents of `dim`'s pieces in the layout's dims list (the
+/// logical axis extent; a single whole-dim piece gives its own size).
+static int64_t fullExtentOfDim(ArrayAttr rotomDims, int64_t dim) {
+  int64_t extent = 1;
+  for (Attribute a : rotomDims) {
+    auto d = cast<DimAttr>(a);
```
j2kun:
```suggestion
  for (auto d : rotomDims.getAsRange<DimAttr>()) {
```

j2kun: Same below.

lib/Dialect/Rotom/Utils/RotomTensorExtLayoutLowering.h
```diff
@@ -10,10 +10,26 @@ namespace mlir {
...
 namespace rotom {
 
+/// A kernel-plan encoding composed into the materialized relation AFTER the
+/// layout's own rolls: `fromPiece`'s axis index is rewritten to
+/// (idx + multiplier * digit(byPiece)) mod extent. This is the
+/// baby-step/giant-step giant pre-rotation -- a kernel-schedule shift by a
+/// MULTIPLE of a digit, deliberately not layout vocabulary (layout rolls
+/// shift by exactly the partner index). The plan that wants the packing
+/// carries it; layouts stay unit-step and alignable endpoint-for-endpoint.
+struct PreRotation {
```
j2kun: I'm a little bit confused by this.

Is this essentially a workaround for a limitation of the rotom layout descriptor?

Is the rotom-produced plan accounting for this, i.e., is the operation that comes before this BSGS producing output that pre-computes the non-unit-step roll, so that the layout is already eagerly in this form? And the issue is just that we can't describe that with the rotom layout descriptor as-is?

What concerns me slightly: this deficiency would presumably propagate, so you need to track pre-rotations throughout the IR, and then the layout descriptor attribute needs to include one or more of these pre-rotations as well. But I don't really see where this is going: this PR doesn't update the attribute to have this, and it's not clear who would be calling this new method overload `lowerToTensorExtIsl` with a `preRotation`.

edwjchen: The original problem was regarding how to express the BSGS shift in the layout representation. The goal was to generalize this optimization in the layout representation for easier cost analysis. My first attempt was to put it in the layout representation as a per-roll step. However, that approach conflated the `stride` attribute in the layout representation, making it harder to understand. The follow-up fix was to introduce this `PreRotation`, keeping the layout representation clean and inject the BSGS shift into the ISL relation when lowering.

Upon further thought, BSGS really shouldn't be a part of the layout representation. The layout representation should strictly describe how tensor elements are laid out. Instead, BSGS should be incorporated as optimization pass(es) when constructing the kernel. resulting in lowering the BSGS shifts as rotations on the plaintext weights. Further optimization passes could then hoist these rotations into the plaintext packings. 

tldr: The updated changes drop this `PreRotation`, and BSGS will be incorporated in later Rotom PRs.

j2kun (approved):

PR #3177, opened 2026-07-09
Add Poulpy exit dialect types and ops

martonmoro:
Implements step 1 and 2 (Add Poulpy as a Bazel Dependency and Define the poulpy Exit Dialect) from #3096.

- Adds types: `module`, `ciphertext`, `unnormalized_ciphertext`,
  `plaintext`, `scratch`, `secret_key`, `tensor_key`,
  `automorphism_key_map`, `bootstrapping_keys`, `bootstrapping_context`
- Adds ops: `module_create`, `scratch_alloc`, `encode`, `decode`,
  `encrypt`, `decrypt`, `add`, `sub`, `mul`, `rotate`, `rescale` and
  their in-place assign variants, `add_unnormalized`, `sub_unnormalized`,
  `normalize`, `compact_limbs`
- All ciphertext/plaintext operands use memrefs as described in the issue ("The poulpy exit dialect should be designed to work with memrefs")
- Adds syntax tests in `tests/Dialect/Poulpy/IR/ops.mlir`

Not in scope for this PR:
PoulpyEmitter (step 3), configure-crypto-context pass (step 4),
parameter selection (step 5), SecretToPoulpy lowering (step 6),
management passes (step 7), end-to-end tests (step 8).

Assisted-by: Claude

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -217,5 +217,13 @@ crate.spec(
     package = "tfhe",
     version = "1.5.0",
 )
+crate.spec(
+    package = "poulpy-ckks",
+    version = "0.7.0",
+)
+crate.spec(
+    package = "poulpy-cpu-ref",
+    version = "0.7.0",
+)
 crate.from_specs()
 use_repo(crate, "crates")
--- a/lib/Dialect/Poulpy/IR/PoulpyOps.td
+++ b/lib/Dialect/Poulpy/IR/PoulpyOps.td
@@ -1,0 +1,307 @@
+#ifndef LIB_DIALECT_POULPY_IR_POULPYOPS_TD_
+#define LIB_DIALECT_POULPY_IR_POULPYOPS_TD_
+
+include "PoulpyDialect.td"
+include "PoulpyTypes.td"
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+class Poulpy_Op<string mnemonic, list<Trait> traits = []> :
+        Op<Poulpy_Dialect, mnemonic, traits> {
+  let assemblyFormat = [{
+    operands attr-dict `:` functional-type(operands, results)
+  }];
+}
+
+def Poulpy_CiphertextMemRef : MemRefOf<[Poulpy_CiphertextType]>;
+def Poulpy_UnnormalizedCiphertextMemRef : MemRefOf<[Poulpy_UnnormalizedCiphertextType]>;
+def Poulpy_PlaintextMemRef : MemRefOf<[Poulpy_PlaintextType]>;
+
+// maps to ModuleNew::<BE>::new(n)
+def Poulpy_ModuleCreateOp : Poulpy_Op<"module_create", [Pure]> {
+  let summary = "Create a new poulpy module";
+  let description = [{
+    See https://docs.rs/poulpy-hal/latest/poulpy_hal/layouts/struct.Module.html
+  }];
+  let arguments = (ins I64Attr:$N);
+  let results = (outs Poulpy_ModuleType:$module);
+}
+
+// maps to ScratchOwned::<BE>::alloc(size)
+def Poulpy_ScratchAllocOp : Poulpy_Op<"scratch_alloc", [Pure]> {
+  let summary = "Allocate a scratch arena of the given byte size";
+  let description = [{
+    See https://docs.rs/poulpy-hal/latest/poulpy_hal/layouts/struct.ScratchOwned.html
+  }];
+  let arguments = (ins I64Attr:$size);
+  let results = (outs Poulpy_ScratchType:$scratch);
+}
+
+// maps to host_module.ckks_pt_vec_alloc + pt.set_meta + encoder.encode_reim
+def Poulpy_EncodeOp : Poulpy_Op<"encode"> {
+  let summary = "Encode real and imaginary slot vectors into a plaintext";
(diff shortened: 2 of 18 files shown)


martonmoro:
I have two questions I'd appreciate guidance on:
1. ScratchOwned vs ScratchArena
The dialect models scratch as !poulpy.scratch mapping to ScratchOwned<BE> (the owned allocation). All Poulpy ops take ScratchArena (a borrow of it). I assumed the emitter would generate .borrow() at each call site. Is this the right split, or should the IR represent the arena directly?

2. encode/decode in the exit dialect
I've included encode and decode ops (mapping to encoder.encode_reim / encoder.decode_reim), but I'm not sure if these belong here. The alternative is for SecretToPoulpy to handle encoding entirely in the emitter without emitting explicit encode ops. What's the preferred approach?

MODULE.bazel
```diff
@@ -217,5 +217,13 @@ crate.spec(
     package = "tfhe",
     version = "1.5.0",
 )
+crate.spec(
+    package = "poulpy-ckks",
+    version = "0.6.0",
```
j2kun: 0.7.0 just released today, is it possible to bump the version now as opposed to later?

lib/Dialect/Poulpy/IR/PoulpyOps.td
```diff
@@ -0,0 +1,365 @@
...
+    Poulpy_ScratchType:$scratch
+  );
+}
+
+// maps to ckks_decrypt
+def Poulpy_DecryptOp : Poulpy_Op<"decrypt"> {
+  let summary = "Secret-key decryption of a CKKS ciphertext.";
+  let description = [{
+    Recovers the plaintext polynomial from a CKKS ciphertext under the given
+    secret key.  The result is a [`CKKSPlaintext`](crate::layouts::CKKSPlaintext)
```
j2kun:
```suggestion
    secret key.  The result is a `CKKSPlaintext`
```

Looks like some of the descriptions here were copied from the rust docs, which is fine, but remove the (unresolvable) links. Even better would be to just have the description of each op link directly to the appropriate page on https://docs.rs/ rather than repeat the contents here.

lib/Dialect/Poulpy/IR/PoulpyOps.td
```diff
@@ -0,0 +1,365 @@
...
+def Poulpy_UnnormalizedCiphertextMemRef : MemRefOf<[Poulpy_UnnormalizedCiphertextType]>;
+def Poulpy_PlaintextMemRef : MemRefOf<[Poulpy_PlaintextType]>;
+
+// maps to ModuleNew::<BE>::new(n)
+def Poulpy_ModuleCreateOp : Poulpy_Op<"module_create", [Pure]> {
+  let summary = "Create a new poulpy module";
+  let description = [{
+    This operation creates a new poulpy module, which is an evaluator context
+    for the poulpy backend. The module is parameterized by a ring degree `N`
+    and a backend-specific handle that holds any required precomputed state.
```
j2kun: Should this correspond to an argument (attribute operand) that specifies the desired "handle"? I'm not sure what the options are, but maybe it can be an optional attribute with CPU as the default.

lib/Dialect/Poulpy/IR/PoulpyOps.td
```diff
@@ -0,0 +1,365 @@
...
+    Poulpy_CiphertextMemRef:$dst,
+    Poulpy_CiphertextMemRef:$src,
+    Poulpy_ScratchType:$scratch
+  );
+}
+
+
+
+
+
```
j2kun: nit
 
```suggestion

```

j2kun (changes requested):
Generally looks good!

> ScratchOwned vs ScratchArena

I think it's fair to have the emitter generate `.borrow()` at each call site.

> encode/decode in the exit dialect

This should be needed IMO, because we will be generating helper functions that compute a memref of preprocessed plaintexts without any local context of how they are used.

lib/Dialect/Poulpy/IR/PoulpyOps.td
martonmoro: I added the backend parameter to the module type so the emitter can always tell from a value's type. I hope this is what you meant.

j2kun: That makes sense if this parameter needs to be printed as part of the type when emitting it, otherwise it may not be necessary and can remain an operand without putting it in the type itself.

j2kun: Given the backends [I see in cpu-ref](https://github.com/poulpy-fhe/poulpy/tree/main/poulpy-cpu-ref0:

```
// f64 FFT backend
let module: Module<FFT64Ref> = Module::<FFT64Ref>::new(1 << log_n);

// Q120 NTT backend (CRT over four ~30-bit primes)
let module: Module<NTT4x30Ref> = Module::<NTT4x30Ref>::new(1 << log_n);
```

It might be worth it to make an `EnumAttr` backed by an enum that matches these two?

j2kun:
@martonmoro

When you're ready please ping me for review again, and note that you will need to manually squash the commits as per [the review flow docs](https://heir.dev/docs/development/#pull-request-review-flow)

j2kun (approved):
LGTM! Just needs to be squashed and we can merge it

j2kun (approved):

PR #3179, opened 2026-07-09
Remove OpenMP gating

### LINKED ISSUE #3103: Clean up OpenMP gates
I believe after the hermetic LLVM toolchain is merged, we can remove all of the gates around `openmp`, since the hermetic toolchain has native OpenMP support.

- Remove enable_openmp bazel flags
- Simplify CI to not need to install openmp (except for workflows that build and test OpenFHE for the frontend)

j2kun:
Fixes #3103


j2kun:
The build passes, but tests fail for lattigo, which cannot possibly be related to openmp changes...

PR #3186, opened 2026-07-10
make two small changes to trigger CI

j2kun:
DONOTSUBMIT=just testing CI


j2kun:
For posterity: it appears the default public runner is less performant even with RBE (9 minute build with also proportionally slower pip install + openfhe tests), but this doesn't seem too bad to live with. We could also have the default runner be a 8-core, which still incurs costs but maybe not too much.

PR #3187, opened 2026-07-11
Orbit: Level-dependent Cost Model

edwjchen:
### Level-dependent cost model

`orbit-cost-model=PATH` loads a JSON cost model whose `latencyTable` maps
each key to a per-level latency array (index `i` = latency at level `i + 1`).
All keys are required:
- `bootstrap`, `rescale`: constant per-decision costs in the objective
  (positive-sample average / per-level max), overriding the `bootstrap-cost`
  and `rescale-cost` options.
- `addCtCt`, `addCtPt`, `mulCtCt`, `mulCtPt`, `rotate`, `negate`: each array
  is least-squares fitted to `cost(level) = slope * level + intercept`, and
  every tracked op is charged its fitted cost at its ILP-chosen input level.
  Ciphertext-ciphertext vs ciphertext-plaintext is decided per op from
  secretness analysis.
  
### Solver configuration

The ILP is solved to a fixed 1% relative optimality gap with no time limit,
matching Orbit's solver setup (Gurobi `MIPGap` / CBC `gapRel` = 0.01).


lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -72,10 +74,36 @@ static bool isMultiplication(Operation* op) {
...
+    if (isSecret(operand, solver)) ++count;
+  return count;
+}
+
+// The level-dependent latency term for one op, or nullopt if the op class has
+// no level-dependent cost.
+static std::optional<LinearCost> levelCostForOp(Operation* op,
+                                                DataFlowSolver* solver,
+                                                const OpCostModel& costModel) {
+  bool ctCt = countSecretOperands(op, solver) >= 2;
```
j2kun: nit: use `llvm::count_if` and remove `countSecretOperands` since it is only used once here. 

```suggestion
  bool ctCt = llvm::count_if(
     op->getOperands(), [&](auto opd) { return isSecret(opd, solver); }
     ) >= 2;
```

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -33,32 +34,73 @@ namespace heir {
...
-
-static std::optional<int> averagePositiveLatency(const llvm::json::Object& root,
-                                                 llvm::StringRef opName) {
-  const llvm::json::Object* latencyTable = root.getObject("latencyTable");
-  if (!latencyTable) return std::nullopt;
-
-  const llvm::json::Array* latencies = latencyTable->getArray(opName);
+// Read one op's per-level latency array (index i holds the latency at level
+// i + 1).
+static std::optional<SmallVector<double>> readLatencies(
```
j2kun, line 39: just FYI, my intention is to eventually replace the json config/parsing code with a tablegen-based backend configuration.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
```diff
@@ -33,32 +34,73 @@ namespace heir {
...
+  return maxValue;
+}
+
+// Least-squares fit of cost(level) = slope * level + intercept over a
+// per-level latency array, where array index i holds the latency at level
+// i + 1.
+static std::optional<LinearCost> fitLinearCost(
+    const llvm::json::Object& latencyTable, llvm::StringRef opName) {
+  auto values = readLatencies(latencyTable, opName);
+  if (!values) return std::nullopt;
```
j2kun:
```suggestion
  if (!values || values->empty()) return std::nullopt;
```

tests/Transforms/ilp_bootstrap_placement/orbit_cost_model.json
j2kun, line 1: do we have a real cost model file to use? I think it would be worth checking in at some point (doesn't have to be this PR)

j2kun (changes requested):
Great work! Only minor nits and notes.

lib/Transforms/ILPBootstrapPlacement/ILPBootstrapPlacement.cpp
edwjchen: Sounds good!

tests/Transforms/ilp_bootstrap_placement/orbit_cost_model.json
edwjchen: Perhaps on a separate PR, I could pull in the cost modeling script used in Orbit for Lattigo? The gist of how it works is it finds the average latency of each HE operation at various levels. 

On another note, it seems worthwhile to build a general cost model script for both the layout assignment,  bootstrap placement, and any foreseeable optimization passes.

j2kun: Yes, please see the neighboring PR on backend configuration: https://github.com/google/heir/pull/2607

IMO the better long-term design does not compute any values on the fly, as in this PR, and instead the construction of the cost model and any postprocessing is a separate script from the loading and use in the compiler.

j2kun (approved):

j2kun:
For some reason it doesn't like my coauthorship on these commits. Please amend that commit and remove my coauthorship and it should get merged.

PR #3200, opened 2026-07-16
Convolution speedup

mdgrs:
Three small improvements/special cases for improving compile time for convolution networks:

- project out dimension in getCtComplementPoints

- avoid folding large extract_slice

- pack constant tensors using memcpy

### DIFF
--- a/lib/Utils/Layout/Utils.cpp
+++ b/lib/Utils/Layout/Utils.cpp
@@ -709,9 +709,6 @@ void getRangePoints(const presburger::IntegerRelation& relation,
 isl_stat pointPairCallback(__isl_take isl_point* pnt, void* user) {
   PointPairCollector* collector = static_cast<PointPairCollector*>(user);
 
-  isl_space* space = isl_point_get_space(pnt);
-  isl_space_free(space);
-
   std::vector<int64_t> domainPoint(collector->domainDims);
   std::vector<int64_t> rangePoint(collector->rangeDims);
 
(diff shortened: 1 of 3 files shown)


lib/Utils/Layout/Utils.cpp
```diff
@@ -785,18 +782,20 @@ void getCtComplementPoints(const presburger::IntegerRelation& relation,
...
-  PointPairCollector present(relation.getNumDomainVars(),
-                             relation.getNumRangeVars());
-  enumeratePoints(relation, present);
+  // We only need the distinct ct indices (range var 0) in the relation's image.
+  // Projecting out the slot dimension first lets us enumerate at most numCts
+  // points instead of the full (ct, slot) grid.
+  PointCollector ctPoints;
+  isl_basic_map* bmap = convertRelationToBasicMap(relation, ctPoints.ctx);
+  isl_set* ctSet = isl_set_from_basic_set(isl_basic_map_range(bmap));
+  ctSet = isl_set_project_out(ctSet, isl_dim_set, /*first=*/1, /*n=*/1);
```
asraa: I think before we also had project out for ct dimension here :https://github.com/google/heir/commit/f3dbc9c868a097329bd744873b4421206a0f22a1

but what you're saying is that it was slow there before because taking the set complement would introduce floordivs and here you're now counting the seen points

asraa (approved):

j2kun (changes requested):
Could you patch in and run the benchmarks from #3204 before and after this change? Only -c opt or -c fast build needed.

j2kun:
I patched it and ran it myself, I see 

  - Relation 1: ~8ms -> ~8ms (No change)
  - Relation 2: ~384ms -> ~111ms (~3.5x speedup)
  - Relation 3: ~2.88s -> ~2.42s (~1.2x speedup)

So this looks good. Let me know if you see something significantly different if you run the benchmark.

j2kun (approved):

j2kun:
Please squash and rebase and we'll get it in

mdgrs:
> I patched it and ran it myself, I see
> 
> * Relation 1: ~8ms -> ~8ms (No change)
> * Relation 2: ~384ms -> ~111ms (~3.5x speedup)
> * Relation 3: ~2.88s -> ~2.42s (~1.2x speedup)
> 
> So this looks good. Let me know if you see something significantly different if you run the benchmark.

I get the same results

j2kun (approved):

j2kun:
LGTM! FYI, I am working on more performance improvements in #3206, with some encouraging partial results

PR #3219, opened 2026-07-20
Add folding for TypedIntPolynomialAttr

crockeea:
This is a follow-up on #3135 that adds support for folding in the following situtaions:
- MonomialOp gets folded to a ConstantOp when the arguments are attributes
- Added support for folding ModArith polynomials (previously, only RNS polynomials had folding support)
- As an essentially-free consequence of adding support for ModArith polynomials, we get folding support for int polynomials.

As a result of adding support for constant folding, this PR is able to remove two hacks:
- PolyMulToNTT previously created bogus EVAL-form constants that it optimistically hoped would be computed at compile time, but in reality, never were. Now it just inserts (free) NTT/iNTT ops around constants, which later get folded away.
- PolynomialToModArith handled a special case of MonomialOp in EVAL form, which is no longer needed due to the changes in PolyMulToNTT.

After merging this, we should revisit #97 to see if it can be closed.

Heavily assisted by AI.


crockeea:
I added additional folders for `MulScalarOp` and `MonicMonomialOp`.

I also discovered a subtlety in `PolyMulToNTT`: `MonomialOp` _sometimes_ folds to a `ConstantOp`, and sometimes doesn't (e.g., if its inputs are not themselves static constants). This means that `PolyMulToNTT` can't just treat `MonomialOp`s as constants (which implies free NTT conversions). The solution is to run the folder *before* `PolyMulToNTT`. Now any ops that are static constants get folded into `ConstantOp`s. `PolyMulToNTT` treats NTT conversions on `ConstantOp` (and only `ConstantOp`) as free, and in fact, they are the only places where NTTs will get folded away.

Note this means that the guards added in #3056 are still correct: it's not correct to have an, e.g., `MonomialOp` that is in EVAL form; there can only be a `MonomialOp` producing a COEFF form polynomial followed by an NTT, or a `MonomialOp` that gets folded into an EVAL-form `ConstantOp`.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -513,24 +522,62 @@ void PolyMulToNTT::runOnOperation() {
...
-      if (needCoeff) {
+      Type coeffTy = typeToForm(v.getType(), Form::COEFF);
+      Type evalTy = typeToForm(v.getType(), Form::EVAL);
+      if (!coeffTy || !evalTy) {
+        signalPassFailure();
+        return;
+      }
+
+      auto repairConstantValueAttr = [&](ConstantOp constantOp, Type newTy) {
+        Attribute value = constantOp.getValue();
```
j2kun, line 543: This may be slightly cleaner using llvm::TypeSwitch

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
```diff
@@ -639,88 +736,507 @@ static RNSPolynomialAttr getRNSPolynomialAttr(MLIRContext* context,
...
+    monomial.setExponent(APInt(apintBitWidth, i));
+    monomials.push_back(monomial);
+  }
+
+  auto result = IntPolynomial::fromMonomials(monomials);
+  assert(succeeded(result) && "Construction guarantees unique exponents");
+  return TypedIntPolynomialAttr::get(resultType, result.value());
+}
+
+static SmallVector<APInt> getCoefficientVector(const IntPolynomial& poly) {
```
j2kun: Let's make this a method `IntPolynomial::getDenseCoefficientVector`

Generally speaking, any functions that involve only IntPolynomial or RNSPolynomial should belong in their respective helper files in lib/Utils/Polynomial/ and have unit tests.

Anything connecting the ops/attributes to use those folding calculations belongs here. I think that will also make it easier to understand to what extent the folding calculations are intended to be general or specialized.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
```diff
@@ -639,88 +736,507 @@ static RNSPolynomialAttr getRNSPolynomialAttr(MLIRContext* context,
...
+}
+
+static APInt normalizeMod(APInt value, uint64_t modulus) {
+  APInt modulusValue(apintBitWidth, modulus);
+  APInt result = value.sextOrTrunc(apintBitWidth).srem(modulusValue);
+  if (result.isNegative()) result += modulusValue;
+  return result;
+}
+
+static std::optional<RNSPolynomial> monomialMulCoeffModulo(
```
j2kun, line 846: A bunch of these helpers seem like they're trying to be generic, while RNSPolynomial is currently hard-coded to support only x^N+1 modulus (due to the NTT/INTT impl and mul implementation via NTT). It might not be worth trying to make this generic, and just specializing for the moment.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
```diff
@@ -639,88 +736,507 @@ static RNSPolynomialAttr getRNSPolynomialAttr(MLIRContext* context,
...
+  if (poly.getRepresentation() != Form::COEFF) return std::nullopt;
+
+  uint64_t numCoeffs = poly.getNumCoeffs();
+  SmallVector<uint64_t> result;
+  result.reserve(poly.getData().size());
+  for (size_t limbIndex = 0; limbIndex < poly.getNumLimbs(); ++limbIndex) {
+    SmallVector<APInt> shifted(numCoeffs + exponent, APInt(apintBitWidth, 0));
+    for (size_t coeffIndex = 0; coeffIndex < numCoeffs; ++coeffIndex) {
+      shifted[coeffIndex + exponent] =
+          APInt(apintBitWidth, poly.getElement(limbIndex, coeffIndex));
```
j2kun, line 858: For RNSPolynomial I think you can skip using APInt, since the underlying data is int64_t and it doesn't really make sense to have arbitrary-precision RNS limbs.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
```diff
@@ -639,88 +736,507 @@ static RNSPolynomialAttr getRNSPolynomialAttr(MLIRContext* context,
...
+    }
+  }
+
+  return RNSPolynomial(
+      std::move(result),
+      SmallVector<uint64_t>(poly.getModuli().begin(), poly.getModuli().end()),
+      Form::COEFF);
+}
+
+static std::optional<RNSPolynomial> scalarMul(const RNSPolynomial& poly,
```
j2kun: ditto above: a method on RNSPolynomial

j2kun (changes requested):
I think lib/Dialect/Polynomial/IR/PolynomialOps.cpp could use some cleanup and separation of concerns, but otherwise LGTM.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: I'm not sure what you mean here; I think `monomialMulCoeffModulo` supports arbitrary polynomial moduli.

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: The next thing we do after the highlighted block is call `reduceModuloPolynomial`, which takes an `ArrayRef<APInt>`, so it's easier to just use APInt here (otherwise we have to construct the arrayref anyway).

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: Fixed

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
j2kun: RNSPolynomial does not support arbitrary moduli. So the generic-ness of this implementation will not be exercised.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -513,24 +523,66 @@ void PolyMulToNTT::runOnOperation() {
...
+              constantOp->setAttr("value", TypedIntPolynomialAttr::get(
+                                               newTy, intAttr.getValue()));
+            })
+            .Case<TypedFloatPolynomialAttr>([&](auto floatAttr) {
+              constantOp->setAttr("value", TypedFloatPolynomialAttr::get(
+                                               newTy, floatAttr.getValue()));
+            })
+            .Case<RNSPolynomialAttr>([&](auto rnsAttr) {
+              auto newPolyTy = dyn_cast<PolynomialType>(newTy);
+              assert(newPolyTy && "Expected polynomial type for constant op");
```
j2kun:
```suggestion
              auto newPolyTy = cast<PolynomialType>(newTy);
```

j2kun (approved):

lib/Dialect/Polynomial/IR/PolynomialOps.cpp
crockeea: Got it. I simplified `reduceModuloPolynomial` so that it only supports moduli of the form `X^N+1`; it's now quite readable. This function didn't change.

crockeea:
I think we should be okay now. Commits are squashed.

j2kun (approved):

PR #3226, opened 2026-07-22
lattigo: lattigo.bootstrap needs ResetsLevelOpInterface

AlexanderViand:
Driven by #3225, where I ran into 

```
--- FAIL: TestLoLA (20.83s)
    lola_test.go:65: Lola__preprocessing took 522.491416ms
panic: cannot Rescale: input Ciphertext level is too low [recovered, repanicked]
```

The bug had a really simple fix (`lattigo.bootstrap` seems to have lost its level-analysis annotation when we switched from ResetsMulDepthOpInterface to ResetsLevelOpInterface) but really annoying to track down. As a result of the missing level reset annotation, `AllocToInplace` thought that a mul op after a bootstrap was still happening at basically level 0, so tried to re-use level-0 buffers for them. Lattigo, in its wisdom, would not throw an error upon using a too-small/too-low-level output buffer, but instead just force things down to level 0 and hence in the next step, we'd get an attempt to rescale a level 0 ciphertext 🤦

### DIFF
--- a/lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
+++ b/lib/Dialect/Lattigo/IR/LattigoCKKSOps.td
@@ -331,7 +331,9 @@ def Lattigo_CKKSRotateOp : Lattigo_CKKSUnaryInPlaceOp<"rotate", [
   let hasVerifier = 1;
 }
 
-def Lattigo_CKKSBootstrapOp : Lattigo_CKKSUnaryOp<"bootstrap"> {
+def Lattigo_CKKSBootstrapOp : Lattigo_CKKSUnaryOp<"bootstrap", [
+  DeclareOpInterfaceMethods<ResetsLevelOpInterface, ["getOperandToReset"]>
+]> {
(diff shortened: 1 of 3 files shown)


asraa (approved):
while i know this is right (hence lgtm).. it broke my compilation of lenet when i patched it :|

j2kun (approved):

j2kun:
It looks like this PR still has `//third_party/heir/tests/Dialect/Lattigo/Transforms:alloc_to_in_place_levels.mlir.test` failing

AlexanderViand:
> It looks like this PR still has `//third_party/heir/tests/Dialect/Lattigo/Transforms:alloc_to_in_place_levels.mlir.test` failing

Oh, good catch, I had a simple error in the FileCheck, but now I'm very confused why the CI passed?
image

EDIT: is it possible that with pull_request_target, github.sha points to main, rather than the PR?

asraa:
> EDIT: is it possible that with pull_request_target, github.sha points to main, rather than the PR?

I noticed the same thing with CI :|

j2kun (approved):

PR #3232, opened 2026-07-23
ci: give explicit ref for pull request target

asraa:
The base version of the workflow is used so I can't really "test" it within this PR https://github.com/orgs/community/discussions/27084


asraa:
cc @AlexanderViand

j2kun (approved):

AlexanderViand:
Lovely, thanks! I think the docs workflow needs the same treatment, too :)

PR #3241, opened 2026-07-24
Fold vector permutation into matrix layout

mdgrs:
Also unifies it with the hoisting operation through a small helper. Unfortunately, they are not quite the same operation (hoisting composes on the slots, while folding composes on the matrix columns), so I was not able to completely unify these two operations

I can squash the two commits if you agree with the changes, but felt it was easier to review with two distinct commits to see the refactoring more clearly

### DIFF
--- a/lib/Transforms/LayoutOptimization/BUILD
+++ b/lib/Transforms/LayoutOptimization/BUILD
@@ -85,12 +85,16 @@ cc_library(
     srcs = ["Patterns.cpp"],
     hdrs = ["Patterns.h"],
     deps = [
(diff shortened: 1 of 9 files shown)


mdgrs:
There is no more space on the GitHub runner. I am not too sure what I can do about this...

j2kun:
Fair enough, thanks for giving the refactoring a shot!

j2kun (approved):
LGTM, please squash and I'll get it in

j2kun:
> There is no more space on the GitHub runner. I am not too sure what I can do about this...

when I pull it into Google it will re-run CI internally, and I'll look into the GHA issues separately.

PR #3250, opened 2026-07-26
Poulpy Emitter

martonmoro:
Step 3 (Implement Code Generation) from #3096.

Adds lib/Target/Poulpy/{PoulpyEmitter.h,.cpp,BUILD,PoulpyTemplates.h}
and registers the --emit-poulpy translation in heir-translate.

- ModuleOp printer: Rust prelude, backend detection from
  !poulpy.module argument types, type BE = ...; and derived type
  aliases (Ct/Tsk/Akm)
- func::FuncOp / func::ReturnOp printers: typed, mutability-aware
  argument list, Result<T> return, zero/one-operand return handling
- convertType: !poulpy.module, !poulpy.scratch, !poulpy.tensor_key,
  !poulpy.automorphism_key_map, and rank-0
  memref<!poulpy.ciphertext> (&Ct/&mut Ct/Ct depending on position
  and mutation)
- Argument mutability analysis: decides &Ct vs &mut Ct from which
  values are written by a compute op in the function body
- Lazy memref.alloc materialization: a fresh ciphertext buffer's
  Rust allocation is synthesized at the first op that writes into
  it, deriving its layout at runtime from a sibling operand
- Rejects an _assign op whose dst is still unmaterialized, since
  that would silently read uninitialized memory
- Op printers: add, add_assign, sub, sub_assign, mul, mul_assign,
  rotate, rotate_assign

Adds tests/Emitter/Poulpy/{BUILD,emit_poulpy.mlir,
emit_poulpy_invalid_assign.mlir}. Output has also been verified with
cargo check against the real poulpy 0.7.0 crates, not just
pattern-matched.


lib/Target/Poulpy/PoulpyEmitter.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+LogicalResult PoulpyEmitter::printOperation(func::FuncOp funcOp) {
+  computeMutatedValues(funcOp);
+  encoderEmitted = false;
+  os << "pub fn " << funcOp.getName() << "(\n";
+  os.indent();
+  for (Value arg : funcOp.getArguments()) {
+    auto argName = variableNames->getNameForValue(arg);
+    os << argName << ": ";
+    bool isMutated = mutatedValues.contains(arg);
```
j2kun, line 207: It looks like the purpose of this is to use mutable outparams instead of returning multiple values, yes? That works, but it seems unnecessary since both MLIR and rust support multi-op results and tuple-typed return values (respectively).

IMO it would be simpler to have the return type be packed up as a tuple and use `let (a, b, c) = ...` for any calls to multi-return-type funcs. Then you can remove the search for mutated values. At least, it appears that there are no other places in the emitter that query `mutatedValues`.

In particular, the compiler inputs to this step would return values instead of mutating them as outparams, unless we have a poulpy-specific pass that modifies a function to hoist its return values to outparams.

On the other hand, maybe something like the ScratchOwned type should always be emitted as mutable? I just say this because I see both the scratch arg being marked mutable in the tests, as well as some instances of `&mut Ct` in the tests.

lib/Target/Poulpy/PoulpyEmitter.cpp
```diff
@@ -0,0 +1,589 @@
...
+  auto dst = mulOp.getDst();
+  auto a = mulOp.getA();
+  auto b = mulOp.getB();
+  auto tsk = mulOp.getTsk();
+  auto scratch = mulOp.getScratch();
+
+  materializeIfPending(dst, module, a, /*useSemanticWidth=*/false,
+                       /*wrapUnnormalized=*/false);
+
+  os << variableNames->getNameForValue(module) << ".ckks_mul_into("
```
j2kun: It might be worth adding one or two templated `emitBinaryOp`/`emitBinaryAssignOp` helper functions to reduce some of the duplication among these printers.

j2kun (changes requested):
Looks pretty good! I think it would be worth sharing a sample of the generated code with @pro7ech to get a validation check on the API usage and the prelude.

martonmoro:
> Looks pretty good! I think it would be worth sharing a sample of the generated code with @Pro7ech to get a validation check on the API usage and the prelude.

will do!

lib/Target/Poulpy/PoulpyEmitter.cpp
martonmoro: I might be misunderstanding something, but the aim of this was for `mutatedValues` only to decide whether a function argument needs `&mut` or `&` in the generated signature (e.g. a caller-allocated `Ct` buffer gets written in place by `ckks_add_into`). Every Poulpy op's `dst` is a memref operand it writes into, not something the op computes and returns, so I think this is the way if we want to model the target library's API 1:1. 
Do you have a recommended better way for this? I agree that it is fragile to having to always update `computeMutatedValues` by hand. After some research, I found `MemoryEffectOpInterface`, do you think using it would be a good direction?

Regarding scrathc buffers: it is always emitted as `&mut ScratchOwned<BE>` uncondigitonally in `convertType`.

j2kun: Ah, I think I understand now. I suppose the alternative to making the function arguments `mut` is to eagerly clone them before doing any computation, and just replace references to func args with references to the clones. I think either way would be fine, so your current approach can stay as-is if you prefer that.

j2kun:
Please rebase and squash your changes and we can optimistically get it merged pending feedback from Jean-Philippe

Pro7ech:
I went over the example, the emitted code looks good, **assuming that hard-coded randomness was intentional**. A future optimization would be to use the `<op>_tmp_bytes(...)` and `<struct>.bytes_of(...)` API to assess the scratch space required instead of hard-coding it's size, and borrow the tmp values from it instead of doing heap allocations.

Also, just to avoid ambiguity, in Poulpy Galoiskeys are keyed by their signed Galois element.

martonmoro:
> Please rebase and squash your changes and we can optimistically get it merged pending feedback from Jean-Philippe

I've squashed the commits. Also i left the randomness as is for now and added a comment so the emitter prints code that can be compiled at this stage.

j2kun (approved):

PR #3254, opened 2026-07-29
rename ciphertext-degree flags (and similar) to min-slot-count

j2kun:
Renamed flags: 

- ciphertext-degree -> min-slot-count
- slot-number -> min-slot-count
- poly-mod-degree -> min-slot-count
- message-size -> min-slot-count 

Lots of downstream updates for C++ camelcase variants, docs, and test files.


lib/Pipelines/ArithmeticPipelineRegistration.h
```diff
@@ -34,11 +34,11 @@ struct MlirToRLWEPipelineOptions : public LoopOptions {
...
       llvm::cl::init(true)};
-  PassOptions::Option<int> ciphertextDegree{
-      *this, "ciphertext-degree",
-      llvm::cl::desc("The degree of the polynomials to use for ciphertexts; "
-                     "equivalently, the number of messages that can be packed "
-                     "into a single ciphertext."),
+  PassOptions::Option<int> minSlotCount{
+      *this, "min-slot-count",
+      llvm::cl::desc("The minimum number of slots to use for packing messages; "
+                     "equivalently, the degree of the polynomials to use for "
```
asraa: this isn't always true that it's the degree of the polynomials though - for CKKS it'll be half of it, and for security heir might configure a larger N

asraa (approved):

lib/Pipelines/ArithmeticPipelineRegistration.h
j2kun: You're right, that description was LLM generated :( sorry!!

j2kun: Fixed

PR #3255, opened 2026-07-29
Clean up mgmt flags in prep for orbit integration

j2kun:
This change prepares heir-opt for a world in which there are multiple parallel ciphertext management pipelines. It adds a new `ciphertext-management-style` enum option (with `greedy` and `orbit-ilp` variants for now) and prefixes the flags specific to each style with the name (e.g., `greedy-bootstrap-waterline` vs `orbit-bootstrap-waterline`).

This change also adds validation that the subset of chosen flags are mutually exclusive across mgmt styles, so you won't accidentally set an orbit flag when trying to use greedy or vice versa.

### DIFF
--- a/lib/Pipelines/ArithmeticPipelineRegistration.h
+++ b/lib/Pipelines/ArithmeticPipelineRegistration.h
@@ -16,6 +16,9 @@ namespace mlir::heir {
 // RLWE scheme selector
 enum RLWEScheme { ckksScheme, bgvScheme, bfvScheme };
 
+// Ciphertext management style selector
+enum CiphertextManagementStyle { greedy, orbitIlp };
+
 struct LoopOptions : public PassPipelineOptions<LoopOptions> {
   PassOptions::Option<bool> experimentalDisableLoopUnroll{
       *this, "experimental-disable-loop-unroll",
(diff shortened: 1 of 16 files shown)


lib/Pipelines/ArithmeticPipelineRegistration.h
```diff
@@ -85,16 +78,64 @@ struct MlirToRLWEPipelineOptions : public LoopOptions {
...
+  PassOptions::Option<bool> greedyModulusSwitchBeforeFirstMul{
+      *this, "greedy-modulus-switch-before-first-mul",
+      llvm::cl::desc("Modulus switching right before the first multiplication "
+                     "(greedy style)"),
+      llvm::cl::init(false)};
+  PassOptions::Option<int> greedyLevelBudget{
+      *this, "greedy-level-budget",
       llvm::cl::desc(
-          "The level budget excluding levels required for bootstrap"),
+          "Level budget for greedy management (excluding bootstrap levels)"),
```
asraa: While we're at it, can we re-explain this as the "Maximum bound on the level for greedy management (excluding bootstrap levels)"? Level budget has always confused me with what we call the bootstrap waterline, because I assume the level "budget" means how many levels we're willing to have in the FHE scheme

asraa (approved):
I don't *love* the prefixed options and would love to have a nicer configuration, but that seems to be out of scope for a development CLI

lib/Pipelines/ArithmeticPipelineRegistration.h
j2kun: Done

PR #3263, opened 2026-07-31
Suggestions for externalizing constants

mdgrs:
I found a few things to change. Since I was writing the tests anyway to make sure everything was fine, I figured you might want to include them as well.


j2kun:
Thanks!! I incorporate this into #3260 and added you as a coauthor to that PR.

PR #3295, opened 2026-08-04
Implementation of loop support in PolyMulToNTT

### LINKED ISSUE #2685: Add region support in PolyMulToNTT
The PolyMulToNTT pass inserts NTTs and INTTs in a function where the AST is composed of Polynomial and some subset of tensor ops. There are likely many ops that should reasonably be handled (e.g., loops), but which currently would produce an error.

crockeea:
Fixes #2685 
This PR adds support for `RegionBranchOpInterface` to PolyMulToNTT, enabling NTT optimization inside things like `for` loops, `if` statements, and `while` loops.

Assisted by AI.


tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt_region_branch.mlir
```diff
@@ -0,0 +1,138 @@
...
+
+!Zq0 = !mod_arith.int<1095233372161 : i64>
+#ring_1 = #polynomial.ring<coefficientType = !rns.rns<!Zq0>, polynomialModulus = <1 + x**1024>>
+!poly_ty_1 = !polynomial.polynomial<ring=#ring_1, form=coeff>
+!ntt_poly_ty_1 = !polynomial.polynomial<ring=#ring_1, form=eval>
+
+module {
+  // Covers: an scf.for iter_arg that needs both forms -- coeff to match its
+  // entry operand/loop result (consumed by to_tensor, coeff-only) and eval to
+  // feed the eval-only MulOp in the loop body. The NTT/INTT pair is hoisted
```
j2kun, line 11: Wouldn't it be more optimal to apply NTT once before the loop even begins, and then INTT once after the loop ends, rather than in each loop iteration?

tests/Dialect/Polynomial/Transforms/poly_mul_to_ntt_region_branch.mlir
```diff
@@ -0,0 +1,138 @@
...
+  // CHECK: [[x2e:%.+]] = polynomial.ntt [[x2]] : [[poly_ty_1]]
+  // CHECK: [[r2:%.+]] = scf.if [[cond]] -> ([[ntt_poly_ty_1]]) {
+  // CHECK: scf.yield [[x2e]] : [[ntt_poly_ty_1]]
+  // CHECK: } else {
+  // CHECK: scf.yield [[x2e]] : [[ntt_poly_ty_1]]
+  // CHECK: }
+  // CHECK: [[t2:%.+]] = polynomial.to_tensor [[x2]] : [[poly_ty_1]] -> tensor<1024x[[RNS]]>
+  // CHECK: [[m2:%.+]] = polynomial.mul [[r2]], [[r2]] : [[ntt_poly_ty_1]]
+  // CHECK: return [[t2]], [[m2]] : tensor<1024x[[RNS]]>, [[ntt_poly_ty_1]]
+  func.func @if_yields_both_branches(%cond: i1, %x: !poly_ty_1) -> (tensor<1024x!rns.rns<!Zq0>>, !poly_ty_1) {
```
j2kun, line 91: I think it would be useful to have a test case where an if statement requires two different forms in the two different branches.

lib/Dialect/Polynomial/Transforms/PolyMulToNTT.cpp
```diff
@@ -593,6 +812,49 @@ void PolyMulToNTT::runOnOperation() {
...
+  // arbitrarily fix one native form into the IR -- preferring coeff form
+  // when both are needed -- and, if the other form is also required by some
+  // use, insert a single conversion right where the value comes into
+  // existence: at the start of the owning block for a block argument, or
+  // right after the op for a value returned to the parent. We remember which
+  // form each successor input was fixed to so Step 4b can rewrite the
+  // forwarding operands that feed it to match.
+  llvm::DenseMap<Value, Form> successorNativeForm;
+  for (Value target : polySuccessorInputs) {
+    Form f = soln.needsForm(target, Form::COEFF) ? Form::COEFF : Form::EVAL;
```
j2kun, line 829: I feel like this is what's making the `for_iter_arg_needs_both_forms` test feel incorrect to me.

In my view, it's not that the successor input must have COEFF form if it has uses of both in different branches, but that all region successor operands and all corresponding successor inputs must have the same form (you can't change form during a control flow branch), and then allow the solver to choose which form it is based on the implied cost of changing form before/after the branch.

In the `for_iter_arg_needs_both_forms` test, it would require `x`, `acc`, `sq` and `r` to all have the same form. If they're all coeff, the ntt/intt would be in the loop and have cost given by the lb/ub (which I guess is the default 1k iterations in that example?), and if they're all eval, then it would have to insert conversions for `x` and for `r`.

j2kun (changes requested):
[edit]: looks mostly good! I see one issue that I think would have a big impact on the solver's output optimality.

crockeea:
Your review exposed some major gaps (thank you!) which will take some time to fix. I'll let you know when I'm ready again.

j2kun:
Given that the gaps are optimizations, we could merge what you have and follow up with the improvements in a second PR. Up to you.

PR #3302, opened 2026-08-05
Improve single element tensor handling

mdgrs:


lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
```diff
@@ -367,6 +387,25 @@ static FailureOr<Value> implementAssignLayoutStep(
     return constantOp.getResult();
   }
 
+  // A single-element tensor packed into a dense layout is a broadcast of its
+  // lone element. The loop-generator path below handles it correctly, but emits
+  // one iteration per slot whose body is loop-invariant; extract the element
+  // once and splat it instead.
+  if (dataSemanticType && dataSemanticType.hasStaticShape() &&
+      dataSemanticType.getNumElements() == 1 &&
+      isDenseLayout(rel, targetType)) {
```
asraa: i think isDenseLayout takes some time and there's a small chance that it can both run up on line 374 and here - I think in the case that it's a splat tensor of some Value (not constant) with size one.

asraa (approved):

lib/Transforms/ConvertToCiphertextSemantics/AssignLayout.cpp
mdgrs: Good point. I folded this if into the preceding one, to avoid this

j2kun (approved):

PR #3303, opened 2026-08-05
improve handling of padding for conv1d

mdgrs:
Fixes issues with the padding where it was not consistently handled

### DIFF
--- a/lib/Kernel/KernelImplementationTest.cpp
+++ b/lib/Kernel/KernelImplementationTest.cpp
@@ -1,6 +1,7 @@
 #include <cstdint>
 #include <functional>
 #include <map>
+#include <string>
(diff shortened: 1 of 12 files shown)


mdgrs:
Once we agree on the basic structure, I can port this to the 2d version

j2kun (commented):
I was considering this today, because indeed I found a bug with the hotword compilation in that it was dropping the padding op completely (!!), and then I refreshed an old change that supports padding.

I agree the best thing to do is to combine the conv op with its prior padding and treat is as a joint kernel. I feel a slight preference to do this fusion as a separate op + pass, but the change here is pretty solid and I think we can just go with this for a while.

j2kun (approved):

PR #3305, opened 2026-08-05
lwe: add (optional) level and scale attributes to `lwe.rlwe_encode`

AlexanderViand:
Preliminary for Cheddar work (where encoding a plaintext effectively requires knowing what level to encode it to, since the canonical scale is different for each level).

### DIFF
--- a/lib/Dialect/LWE/IR/LWEOps.td
+++ b/lib/Dialect/LWE/IR/LWEOps.td
@@ -211,7 +211,9 @@ def LWE_RLWEEncodeOp : LWE_Op<"rlwe_encode", [
   let arguments = (ins
     SignlessIntegerOrFloatLike:$input,
     AnyPlaintextEncodingAttr:$encoding,
-    Polynomial_RingAttr:$ring
+    Polynomial_RingAttr:$ring,
+    OptionalAttr<I64Attr>:$level,
+    OptionalAttr<I64Attr>:$scale
   );
 
   let results = (outs LWEPlaintextLike:$output);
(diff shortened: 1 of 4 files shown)


j2kun:
The CI failure looks legit to me.

AlexanderViand:
> The CI failure looks legit to me.

Yep, looks like something broke when I cherry-picked it :(

mdgrs (approved):

j2kun (commented):
Can't the level be inferred from the result type of the encode op?

AlexanderViand:
> Can't the level be inferred from the result type of the encode op?

@j2kun Nope, unfortunately the info on the ptxt type isn't sufficient. In #3305 @ZenithalHourlyRate suggested also adding a ctxt space to the op as a solution, I instead picked this approach with an optional Attr when I needed it "just to make cheddar work" since its a lot lower-impact.

In the meantime, @mdgrs is looking into actually doing smart encoding-target-level selection in general, so maybe it is now actually worth doing a "real" solution rather than the optional attr workaround?

j2kun:
I'm fine with this approach, though I suspect we will also want a scale attribute so that plaintext encoding can target a specific scale. Can you add that as another optional attribute, then we can get this in.

AlexanderViand:
> I'm fine with this approach, though I suspect we will also want a scale attribute so that plaintext encoding can target a specific scale. Can you add that as another optional attribute, then we can get this in.

@j2kun  done :)

j2kun (approved):

PR #3306, opened 2026-08-05
cheddar: switch to DPS

AlexanderViand:
This makes the cheddar dialect one-shot-bufferize compatible by making it Destination-Passing-Style, in preparation for a bufferize + emitC based emitter.

This PR also adds a few ops that were missing, especially around setup/parameter generation.

Since I still find DPS hard to get my head around, I left a lot of the verbose comments in, even if they're sometimes a bit LLM-y, just to make things a bit easier to follow along.

### DIFF
--- a/lib/Dialect/Cheddar/IR/BUILD
+++ b/lib/Dialect/Cheddar/IR/BUILD
@@ -27,8 +27,10 @@ cc_library(
         ":types_inc_gen",
         "@heir//lib/Dialect:HEIRInterfaces",
         "@llvm-project//llvm:Support",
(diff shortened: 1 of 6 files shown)


j2kun:
I think if you rebase over main the CI failure will go away. Specifically 7ea540ba9296efcffbee4b7b92a113a0e1dc9967

j2kun (approved):

j2kun:
IMO this PR makes me want to go study enough to write an MLIR tutorial on the memory effects interfaces and the bufferization pipeline.

PR #3316, opened 2026-08-07
Add composite relu

mdgrs:
Ports the 3 stage RELU approximation from the Orion paper. It is hidden behind a flag.

### DIFF
--- a/tests/Transforms/activation_canonicalizations/relu_generic_domain.mlir
+++ b/tests/Transforms/activation_canonicalizations/relu_generic_domain.mlir
@@ -1,0 +1,111 @@
+// RUN: heir-opt --split-input-file --activation-canonicalizations %s | FileCheck %s
+
+#map = affine_map<(d0, d1) -> (d0, d1)>
+
+// A torch ReLU imports as a linalg.generic carrying the polynomial-approximation
+// domain on the generic op, with cmpf+select in its body. The float select
+// must canonicalize to arith.maximumf AND inherit the generic's
+// domain_lower/domain_upper so PolynomialApproximation reads the right domain.
+
+// The domain must move onto the maximumf and NOT remain on the generic
+// (leaving it on both collides during later activation lifting).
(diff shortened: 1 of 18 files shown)


tests/Transforms/activation_canonicalizations/relu_generic_domain.mlir
```diff
@@ -0,0 +1,111 @@
...
+    linalg.yield %3 : f32
+  } -> tensor<1x3xf32>
+  return %1 : tensor<1x3xf32>
+}
+
+// -----
+
+#map = affine_map<(d0, d1) -> (d0, d1)>
+
+// A fused elementwise generic can carry more than one ReLU. Every maximumf must
```
j2kun, line 36: In your opinion, would it make sense in the future to aim for a general handler of linalg.generic for canoncalizing activations?

j2kun (approved):
Fantastic work! I will try to carve out some time in the future to do a more detailed analysis of the tradeoff between the composite method and the vanilla approximation.

j2kun:
I did notice that enabling this flag for the HE-LRM model actually causes a segfault, but I think I will not have the time to investigate until after the blog deadline + my 2 weeks of vacation. However, next week you will be able to see the exact invocation and I can tell you where I flipped this flag and saw an error

tests/Transforms/activation_canonicalizations/relu_generic_domain.mlir
mdgrs: Just to make sure we are on the same page, you mean grouping operations that belong to the same activation function together, so they can be treated as a unit. I think that would indeed make sense since we typically want to treat each activation function as a unit (since we think in terms of layers in the neural network)

j2kun: Right, I was mainly surprised that the linalg produced by torch-mlir can produce generics with multiple ops operating on different operands within.

PR #3332, opened 2026-08-12
Encode plaintext at highest used level instead of max level

mdgrs:
For the Lola test that uses Lattigo, the encode time goes from 357 ms to 288 ms, and the memory use goes from 125 MB to 84 MB.

### DIFF
--- a/lib/Dialect/LWE/Transforms/AnnotatePlaintextLevel.cpp
+++ b/lib/Dialect/LWE/Transforms/AnnotatePlaintextLevel.cpp
@@ -1,0 +1,154 @@
+#include "lib/Dialect/LWE/Transforms/AnnotatePlaintextLevel.h"
+
+#include <algorithm>
+#include <cstdint>
+#include <optional>
+
+#include "lib/Dialect/LWE/IR/LWEOps.h"
+#include "lib/Dialect/LWE/IR/LWETraits.h"
+#include "lib/Dialect/LWE/IR/LWETypes.h"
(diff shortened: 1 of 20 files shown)


lib/Dialect/LWE/Transforms/Passes.td
```diff
@@ -69,6 +69,42 @@ def ImplementTrivialEncryptionAsAddition : Pass<"implement-trivial-encryption-as
...
+    than a lower bound. Annotating an encryption therefore changes what the
+    client emits, not just how many limbs it encodes: where the entry level sits
+    below the top of the chain, the ciphertext now enters at the level its type
+    declares instead of at `MaxLevel`. A plaintext that both feeds an encryption
+    and is combined with a higher-level ciphertext has no correct level, and is
+    left to the fallback.
+
+    Run this after CSE. Identical encode ops that CSE has merged are cheaper as
+    one plaintext at the highest level than as several at their own levels,
+    since each encode pays a fixed cost on top of the per-limb work.
```
j2kun, line 99: TIL!

lib/Dialect/LWE/Transforms/AnnotatePlaintextLevel.cpp
j2kun, line 1: This works fine, though this kind of problem is the poster-child for a backwards dataflow analysis (lattice join is to take the max). I'd say only worry about it if this ends up having performance issues.

j2kun (approved):

PR #3334, opened 2026-08-13
cheddar: add optional scale-snu CHEDDAR dependency

AlexanderViand:
PR 3/? on Cheddar :) 

This just adds an optional cheddar bazel dependency (EDIT: only expected to work on linux with CUDA)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -45,6 +45,7 @@ bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.8.2", dev_dependenc
 # implicitly depends upon the target '//:license'. How bizarre.
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "platforms", version = "1.1.0")
+bazel_dep(name = "rules_cuda", version = "0.3.0")
 bazel_dep(name = "rules_go", version = "0.61.1")
 bazel_dep(name = "rules_python", version = "1.7.0")
 bazel_dep(name = "googletest", version = "1.17.0")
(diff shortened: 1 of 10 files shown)


tools/heir-opt.bzl
```diff
@@ -5,10 +5,10 @@ def executable_attr(label):
...
         default = Label(label),
         executable = True,
-        # commenting this out breaks cross-compilation, but this should not be a problem
-        # for developer builds
-        # cfg = "exec",
-        cfg = "target",
+        # The compiler runs during the build, so it must use the execution
+        # platform rather than inheriting the generated program's target
+        # platform and C++ runtime.
+        cfg = "exec",
```
j2kun: IIRC this will double HEIR's build time because it has to build heir-opt twice: once for the toolchain and once for the target.

So I'm a bit nervous to enable this, and I think preserving hermeticity would avoid the need to do that.

MODULE.bazel
```diff
@@ -45,6 +53,16 @@ bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.8.2", dev_dependenc
 # implicitly depends upon the target '//:license'. How bizarre.
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "platforms", version = "1.1.0")
+bazel_dep(name = "rules_cuda", version = "0.3.0")
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
```
j2kun: I am skeptical that using the native toolchain/foreign_cc is necessary (you added rules_cuda, after all!).

Clang can natively compile cuda, so we don't need `nvcc`, and rules_cuda talks to rules_cc just fine with clang as far as I have read. I'm guessing the main obstacle is writing the bazel overlay that uses cc_library and cuda_library directly. Is that right? If so, I'd say let's give that a shot instead of breaking hermeticity.

j2kun (changes requested):

MODULE.bazel
j2kun: I took a quick look and here's a starting point: https://github.com/scale-snu/cheddar-fhe/compare/main...j2kun:cheddar-fhe:bazel-overlay

At this commit, the project builds with the hermetic toolchain (I can't run it because I don't have a GPU on this machine). I think I could just fork cheddar and point HEIR to that fork with the bazel overlay already there, or else use a patch file to add the overlay via that commit's diff.

Please give that a try and see if you can run on your machine's GPU with that overlay. If so we can find the right way to integrate it.

AlexanderViand: > I'm guessing the main obstacle is writing the bazel overlay that uses cc_library and cuda_library directly. Is that right? If so, I'd say let's give that a shot instead of breaking hermeticity.

I figured it'd be easier to use the existing CMake setup (and with pre-hermetic heir, it was quite easy), but you're right, a bazel overlay is much nicer - thanks for drafting one!

AlexanderViand: >  fork cheddar and point HEIR to that fork with the bazel overlay already there, or else use a patch file to add the overlay via that commit's diff.

I'd love to avoid having to keep a fork of scale-snu/cheddar around - can we just keep the required build files in the HEIR repo for now (and maybe eventually move to a BCR entry for Cheddar?)

patches/mlir_forward_call_result_out_params.patch
j2kun, line 1: Is this something that can be feasibly upstreamed? I'm certain it will break the next time an LLVM version bump touches this file.

patches/cheddar.patch
j2kun, line 1: Optional: can we also try to upstream the C++ changes here?

j2kun (approved):
FYI, my plan is to set up buildkite in September so that we can run the cheddar tests on GPU.

patches/mlir_forward_call_result_out_params.patch
AlexanderViand: Wait, what the heck is that even doing in this PR - I must have messed something up during rebasing the stack :/

AlexanderViand: (If anything, this should be in the emitC/bufferization branch, and even there I'm 90% sure I removed the need for it)

patches/cheddar.patch
AlexanderViand: Sure, I'll open a PR against scale-snu/cheddar

patches/mlir_forward_call_result_out_params.patch
AlexanderViand: Removed!

j2kun (approved):

AlexanderViand:
Added some slight fixes to make macOS builds happy

(Well, at least my local macOS build - the CI macOS build hasn't been happy for a few weeks, apparently :(

PR #3335, opened 2026-08-14
Absorb ciphertext layout changes into plaintext matrix for convolutions

mdgrs:
We should be able to avoid rotating the ciphertext after convolutions and push the permutation into the plaintext. For our convolutional networks, this is particularly beneficial because they often look like: 

Activation -> Convolution -> Activation' -> Convolution'->...

This feels a bit too complex, but I have not been able to pull it apart into simpler pieces.


j2kun:
I will be on vacation for two weeks. If Alex can review the code I can give it a brief scan and then we can merge it

mdgrs:
@AlexanderViand tells me that this is quite large to review and I agree. I want to try to simplify this further, so I'll mark it as draft in the meantime. In all cases, I expect to be busy with other unrelated things for the next week or so

PR #3340, opened 2026-08-17
Fix --//:enable_yosys=0

AlexanderViand:
Trying to build without yosys failed because of missing includes (probably overly eagerly removed by some automated tool because they're only used in preprocessor-gated code?)

This has been around for a bit, and the weekly CI testing no-yosys has been failing because of it https://github.com/google/heir/actions/runs/31916922440/job/95090212968

EDIT: Did that CI ever pass? A bunch of tests do not seem to work for the non-yosys fallback branch, so I've tagged them as "requiring yosys" for now, but it looks like the non-yosys path is quite out of date?

### DIFF
--- a/bazel/lit.bzl
+++ b/bazel/lit.bzl
@@ -5,7 +5,7 @@ load("@rules_python//python:py_test.bzl", "py_test")
 
 _DEFAULT_FILE_EXTS = ["mlir"]
 
-def lit_test(name = None, src = None, size = "small", tags = None, data = None):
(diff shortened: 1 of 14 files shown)


j2kun:
In my view we should just remove the yosys flag entirely. The build is hermetic now and there should be no issues with the build including yosys always.

AlexanderViand:
Closing in favour of https://github.com/google/heir/pull/3342

PR #3346, opened 2026-08-19
preprocessing: make resource loading DPS

AlexanderViand:
`preprocessing.load_resource` simply returned the file contents into a tensor SSA value, which would then bufferize into a kind-of "floating" memref without a real allocation backing it (the file was kind of acting as the conceptual "allocation" if I understand the previous bufferization interface implementation?). Eventually, EmitC (e.g., as in the Cheddar backend) would take this slightly-weird-bufferized state and somehow turn it into "auto x = loadResource(...)" or similar. This is fine for smaller resources, but for larger resources, the local variable's automatic stack-based storage would immediately cause stack overflows. 

This PR changes it so that `preprocessing.load_resource` becomes DPS and gets an explicit target allocation (`memref.alloc`, so on the heap) during bufferization. ~~For small resources, the old stack allocation behaviour can be  automatically recovered by `--promote-buffers-to-stack` during the bufferization pipeline.~~
EDIT: I removed this - we only externalise large things to begin with, so there's no point to do this, and it also removes the need to support `memref.alloca` in the emitters, and means this basically an NFC for the existing emitters, just making live easier for Cheddar (and other potential EmitC backends, I'd assume?)

~~Unfortunately, this does require making the existing emitters aware of this, so sorry for the noisy diff~~

### DIFF
--- a/lib/Dialect/Preprocessing/IR/BUILD
+++ b/lib/Dialect/Preprocessing/IR/BUILD
@@ -24,6 +24,7 @@ cc_library(
         ":types_inc_gen",
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:BytecodeOpInterface",
+        "@llvm-project//mlir:DestinationStyleOpInterface",
         "@llvm-project//mlir:IR",
         "@llvm-project//mlir:InferTypeOpInterface",
         "@llvm-project//mlir:SideEffectInterfaces",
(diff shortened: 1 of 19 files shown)


AlexanderViand:
@j2kun just tagging you already so I don't forget, but please interpret this (and all the others) as request for a review _after_ you're back from vacation ;)

j2kun (approved):
LGTM. I appreciate keeping the change lightweight and I agree with your reasoning.

PR #3347, opened 2026-08-19
tfhe-rust-hl: handle affine loop iter_args

### LINKED ISSUE #3344: heir-translate --emit-tfhe-rust-hl aborts on affine.for without iter_args
# How to reproduce

```mlir
// repro.mlir
func.func @f() {
  affine.for %i = 0 to 16 {
  }
  return
}
```

$ bazel build //tools:heir-translate
$ ./bazel-bin/tools/heir-translate --emit-tfhe-rust-hl repro.mlir
heir-translate: llvm/include/llvm/ADT/ArrayRef.h:439:
  Assertion `Index < this->size() && "Invalid index!"' failed.
#15 mlir::heir::tfhe_rust::TfheRustHLEmitter::printOperation(mlir::affine::AffineYieldOp)
#19 mlir::heir::tfhe_rust::TfheRustHLEmitter::printOperation(mlir::affine::AffineForOp)

lib/Target/TfheRustHL/TfheRustHLEmitter.cpp:434

```cpp
LogicalResult TfheRustHLEmitter::printOperation(affine::AffineYieldOp op) {
  if (op->getNumResults() != 0) {
    return op.emitOpError() << "AffineYieldOp has non-zero number of results";
  }

  os << variableNames->getNameForValue(op->getOperand(0)) << "\n";

  return success();
}
```

affine.yield is a terminator and never has results, so the guard is always false and the operand access below it is effectively unguarded. The loop-carried values are the yield's operands. Reading operand 0 is right for the iter_args form, where printOperation(AffineForOp) emits (lo..hi).fold(init, |mut acc, i| { ... }) and the yielded value is the closure's tail expression — but a loop without iter_args yields nothing, and the same printer runs on its terminator.

# Minimal correction

generate by Claude Code 

```diff
 LogicalResult TfheRustHLEmitter::printOperation(affine::AffineYieldOp op) {
-  if (op->getNumResults() != 0) {
-    return op.emitOpError() << "AffineYieldOp has non-zero number of results";
+  // A loop with no iter_args yields nothing.
+  if (op->getNumOperands() == 0) {
+    return success();
+  }
+
+  if (op->getNumOperands() == 1) {
+    os << variableNames->getNameForValue(op->getOperand(0)) << "\n";
+    return success();
   }

-  os << variableNames->getNameForValue(op->getOperand(0)) << "\n";
+  // Multiple iter_args are folded as a tuple.
+  os << "("
+     << commaSeparatedValues(
+            op->getOperands(),
+            [&](Value value) { return variableNames->getNameForValue(value); })
+     << ")\n";

   return success();
 }
```

AlexanderViand:
Fixes #3344

New logic:
*  no operands / iter_args -> return success / do nothing
* 1 operand / iter_arg-> old behaviour
* multiple operands/ iter_args -> target the (fixed up  by 🤖 ) tuple emitter

note: the previous check for non-zero results was dead (`affine.yield` has zero results by definition/verifier) and got removed.

### DIFF
--- a/lib/Target/TfheRustHL/TfheRustHLEmitter.cpp
+++ b/lib/Target/TfheRustHL/TfheRustHLEmitter.cpp
@@ -387,7 +387,13 @@ LogicalResult TfheRustHLEmitter::printOperation(affine::AffineForOp op) {
   if (!op.getInits().empty()) {
     suffix = " });\n";
 
-    emitAssignPrefix(op->getResult(0));
+    if (op.getNumResults() == 1) {
+      emitAssignPrefix(op.getResult(0));
+    } else {
+      os << "let (" << commaSeparatedValues(op.getResults(), [&](Value value) {
+        return variableNames->getNameForValue(value);
+      }) << ") = ";
+    }
 
     os << "(" << op.getConstantLowerBound() << ".."
        << op.getConstantUpperBound() << ")" << ".fold(";
@@ -402,10 +408,11 @@ LogicalResult TfheRustHLEmitter::printOperation(affine::AffineForOp op) {
                                  [&](Value value) {
                                    return variableNames->getNameForValue(value);
                                  })
-         << "), |mut ("
+         << "), |("
          << commaSeparatedValues(op.getRegionIterArgs(),
                                  [&](Value value) {
-                                   return variableNames->getNameForValue(value);
+                                   return "mut " +
+                                          variableNames->getNameForValue(value);
                                  })
          << "), ";
     }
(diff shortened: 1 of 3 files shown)


WoutLegiest (approved):
lgtm
Maybe we should sync. I have these and many other extension done for the tfhe-rs in my private repo

AlexanderViand:
> Maybe we should sync. I have these and many other extension done for the tfhe-rs in my private repo

@WoutLegiest That would be great! Any chance the office hour slot on Monday works for you?

j2kun (approved):

PR #3352, opened 2026-08-23
fix: lower chained secret Chebyshev evaluations natively

AlexanderViand:
In a situation like poly.eval(poly.eval(x)) which occurs, e.g., in composite ReLU lowerings, only the inner one would be lowered to kernel.chebyshev, the second one would incorrectly be detected to be non-secret:

Replacing the inner eval creates a new operand with no lattice entry for the outer eval, so we need to query on the result rather than the operand.


j2kun (approved):

j2kun:
After this change, the relu_composite test takes 3 minutes to run and exceeds RAM on the standard google-internal test runner. Looking into it, it results in logN=16 and 23 levels in the modulus chain.

I think there's an underlying bug here that HEIR thinks we need to bootstrap. So in the internal change I will add a hard-coded level budget of 14 to this test. That makes the test run without bootstrapping.

PR #3354, opened 2026-08-23
RotationAnalysis: model nested affine.for in arithmetic DAGs

AlexanderViand:
RotationAnalysis rebuilds an enclosing scf.for as an ArithmeticDag and enumerates the shifts its rotations use, so configure-crypto-context can generate exactly the Galois keys a program needs. DagBuilder covered scf.for, scf.if and the tensor ops, but not an affine.for nested in that reconstructed loop. The analysis still reports success, but the rotations inside the affine loop disappear. 

This teaches DagBuilder to model constant-bound affine.for and affine.yield just like their scf counterparts. The lit test exercises both the induction variable and induction-variable-plus-constant forms inside an affine.for nested in scf.for; the unit test covers DAG evaluation directly.

(If you can't already tell from the language - this one is straight from the 🤖. It checks out to me and works for us, but I'm not sure if there are subtleties about either the RotationAnalysis or the DagBuilder I'm missing)

### DIFF
--- a/lib/Analysis/RotationAnalysis/BUILD
+++ b/lib/Analysis/RotationAnalysis/BUILD
@@ -36,6 +36,7 @@ cc_library(
         "@heir//lib/Kernel:ArithmeticDag",
         "@heir//lib/Kernel:Utils",
         "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:AffineDialect",
         "@llvm-project//mlir:ArithDialect",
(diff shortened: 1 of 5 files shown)


j2kun (approved):
For some reason I was expecting all affine loops were converted to scf loops by the time this is run (during ConfigureCryptoContext). But maybe that is not the case for elementwise-to-affine?

PR #3357, opened 2026-08-24
Record where a ciphertext argument starts on the modulus chain

AlexanderViand:
A ciphertext argument does not have to start at the top of the modulus chain. HEIR asks the client to encrypt at the level the computation actually begins at, which for a model whose first op is a linear transform can be well below the top. (e.g., `func.func @foo(%0: !lwe.lwe_ciphertext<..., modulus_chain = <elements = <11 moduli>, current = 2>>, ...)`

The lattigo lowering then gives every ciphertext the same opaque `!lattigo.rlwe.ciphertext`, and `LevelAnalysis::setToEntryState` joins every entry lattice to `LevelState(0)` (top of the chain).
As a result, the level analysis used by `--lattigo-alloc-to-inplace` is off, which causes incorrect (or, at least inefficient) buffer reuse.

This PR simply adds a `lwe.entry_level_depth` attribute to function arguments while still on the LWE types and uses that to determine the level at the `! lattigo.rlwe.ciphertext ` level.

### DIFF
--- a/lib/Analysis/LevelAnalysis/LevelAnalysis.h
+++ b/lib/Analysis/LevelAnalysis/LevelAnalysis.h
@@ -8,6 +8,7 @@
 
 #include "lib/Analysis/SecretnessAnalysis/SecretnessAnalysis.h"
 #include "lib/Dialect/Mgmt/IR/MgmtOps.h"
+#include "lib/Utils/AttributeUtils.h"
 #include "lib/Utils/Utils.h"
 #include "llvm/include/llvm/Support/Debug.h"        // from @llvm-project
 #include "llvm/include/llvm/Support/raw_ostream.h"  // from @llvm-project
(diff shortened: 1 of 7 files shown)


lib/Analysis/LevelAnalysis/LevelAnalysis.h
```diff
@@ -202,7 +208,32 @@ class LevelAnalysis
...
+    // than it is, which later reads as "this buffer still has levels left".
+    propagateIfChanged(
+        lattice,
+        lattice->join(LevelState(getEntryLevelDepth(lattice->getAnchor()))));
+  }
+
+  /// Levels already consumed by `value` on entry, from the attribute the
+  /// backend lowering leaves behind. Zero (fresh) when unannotated.
+  static int getEntryLevelDepth(Value value) {
+    auto arg = dyn_cast_or_null<BlockArgument>(value);
```
j2kun: use `findAttributeAssociatedWith`, returning 0 on failure.

j2kun (changes requested):
LGTM, but is there a particular reason that you're using depth here instead of a zero-indexed level starting from the base of the chain? Then you can just pass along the `current` index instead of inverting it relative to the max level.

lib/Analysis/LevelAnalysis/LevelAnalysis.h
AlexanderViand: Done! 

For the inversion, I think there *was* a reason... Oh, because the attribute is used as starting state for the level analysis and that uses the 0-is-top convention.

j2kun (approved):

PR #3194, opened 2026-07-15
Orbit: Compression & SISO Partitioning

edwjchen:
Builds on top of https://github.com/google/heir/pull/3187 

### Features:
- Compression to merge equivalent ops into groups that share one set of ILP variables.
- Single-input single-output partitioning to break large circuits into smaller groups for faster ILP solves. 
- Dynamic programming to stitch the local solutions of each partition together into a global solution.

PR #3383, opened 2026-08-31
 Orbit: Bypass handling for residual connections

edwjchen:
Stacks on https://github.com/google/heir/pull/3194.
 
A residual connection is a fork that feeds a deep "main" path and a shallow path which rejoin at an addition. No single value is live across that span, so the region has no single-input single-output cut and SISO partitioning would leave it as one monolithic ILP. When the two rejoined operands differ in multiplicative depth by at least bypass-depth-threshold (Orbit's bpsdepth, default 15), the region is instead split and solved separately.

## How it works

- Detection (detectBypass): finds the fork/join in a region, e.g., an output addition with two secret operands from distinct producers whose mul-depth gap clears the threshold.
- Split solve (solveBypass): the deep main path is solved for every enumerated output level to build a QBP table; the shallow path plus the join is then solved once per realized main-output state under the region's own output boundary; the cheapest main + bypass combination is kept.
- Runs on every region (solveRegion dispatcher)
- Residual-join carve-out: the join addition is kept out of any squashed addition tree (via per-op multiplicative depth in OpGrouping) so its two-predecessor structure stays visible to detection.
- Contested fork: the fork is pinned to the region's DP-enumerated input state and each path rescales it independently per edge, so a rescale common to both paths is charged once (absorbed into the previous partition's output) rather than duplicated.

To make bypass regions first-class, the partition driver was generalized from a contiguous [begin, end) range to an explicit op-set (groupIds), so main and bypass can be arbitrary subsets of a region.

## Default
On by default at threshold 15, matching Orbit; set bypass-depth-threshold=0 to disable. Existing circuits without a large-gap residual are unaffected.


lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -252,8 +256,9 @@ struct PartitionSolution {
 };
 
 struct Partition {
-  int groupBegin;
-  int groupEnd;
+  // Group indices in this partition, ascending. A contiguous range for a SISO
+  // partition; bypass regions (M4) supply arbitrary subsets.
```
j2kun: what is "M4"?

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -796,9 +800,14 @@ static SmallVector<Partition> computePartitions(const OpGrouping& grouping,
                                                 DataFlowSolver* solver,
                                                 int partitionMinSize) {
   int numGroups = grouping.groups.size();
+  auto rangeIds = [](int lo, int hi) {
+    SmallVector<int> ids;
+    for (int i = lo; i < hi; ++i) ids.push_back(i);
+    return ids;
+  };
```
```
auto rangeIds = [](int lo, int hi) {
    return llvm::to_vector(llvm::seq(lo, hi));
  };
```

Needs `#include "llvm/ADT/STLExtras.h"` as well

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -796,9 +800,14 @@ static SmallVector<Partition> computePartitions(const OpGrouping& grouping,
...
   SmallVector<Partition> partitions;
   if (numGroups == 0) {
-    partitions.push_back({0, 0, Value(), 0});
+    partitions.push_back({rangeIds(0, 0), Value(), 0});
```
j2kun: Maybe instead of calling just explicitly pass an empty list `{}`?

lib/Analysis/ILPBootstrapPlacementAnalysis/ILPBootstrapPlacementAnalysis.cpp
```diff
@@ -885,21 +894,316 @@ static SmallVector<Partition> computePartitions(const OpGrouping& grouping,
...
+  DenseSet<int> mainAnc =
+      ancestorsWithin(grouping, solver, regionSet, deep.second);
+  DenseSet<int> bypassAnc;
+  if (shallowInRegion)
+    bypassAnc = ancestorsWithin(grouping, solver, regionSet, shallow.second);
+
+  // Clean single fork: the two paths share no in-region ancestor, and together
+  // with the join they cover the whole region.
+  for (int g : mainAnc)
+    if (bypassAnc.contains(g)) return result;
```
j2kun: Intersection?

PR #3395, opened 2026-09-03
update llvm dep for macos build

j2kun:
Fixes the MacOS build in CI by:

- updating a variety of bazel dependencies
- Fixing some test failures triggered on arm but not x86:
  - Memory corruption and use-after-free in AllocToInPlaceUtils
  - Properly walk blocks in AllocToInPlaceUtils
  - Relax CHECK-NEXT in ilp_relin_modreduce_placement.mlir for test with multiple optimal solutions.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -25,16 +25,7 @@ llvm.configure(
 )
 use_repo(llvm, hermetic_llvm = "llvm-project")
 
-# Apple compilation support.
-# Note this requires full xcode, not just commandline tools
-#
-# Tested with:
-#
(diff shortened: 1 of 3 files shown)


j2kun:
Low priority: @AlexanderViand can you try building with this checked out on your machine? I figure if it works on our two macos machines and the CI, it should be good enough.

AlexanderViand (approved):
Works for me, too! 

https://heir.buildbuddy.io/invocation/456a9af4-db0f-4c3c-a6c0-fe6e396dc2de

PR #3404, opened 2026-09-07
cheddar: add destination-style bufferization support

AlexanderViand:
PR 4/? on Cheddar.

This is an attempt to do bufferization for Cheddar with as much of the existing upstream stuff as possible, while still trying to account for the std::move based API of Cheddar and minimising the amount of needless deep copying. 

I'd love a thorough review of this, because I've had to lean heavily on AI here and still don't feel like I _really_ get how a bufferization like this is supposed to happen 😅

### DIFF
--- a/lib/Dialect/Cheddar/Transforms/CheddarBufferize.cpp
+++ b/lib/Dialect/Cheddar/Transforms/CheddarBufferize.cpp
@@ -1,0 +1,36 @@
+#include "lib/Dialect/Cheddar/Transforms/CheddarBufferize.h"
+
+#include "mlir/include/mlir/Dialect/Bufferization/Transforms/Passes.h"  // from @llvm-project
+#include "mlir/include/mlir/Dialect/MemRef/Transforms/Passes.h"  // from @llvm-project
+#include "mlir/include/mlir/Pass/PassManager.h"   // from @llvm-project
+#include "mlir/include/mlir/Transforms/Passes.h"  // from @llvm-project
+
+namespace mlir {
+namespace heir {
+namespace cheddar {
+
+void buildCheddarBufferizationPipeline(OpPassManager& pm) {
+  pm.addPass(bufferization::createEmptyTensorEliminationPass());
+  bufferization::OneShotBufferizePassOptions oneShot;
+  oneShot.bufferizeFunctionBoundaries = true;
+  oneShot.functionBoundaryTypeConversion =
+      bufferization::LayoutMapOption::IdentityLayoutMap;
+  pm.addPass(bufferization::createOneShotBufferizePass(oneShot));
+  pm.addPass(memref::createFoldMemRefAliasOpsPass());
+  pm.addPass(createCSEPass());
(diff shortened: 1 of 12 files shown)


lib/Dialect/Cheddar/Transforms/CheddarBufferize.cpp
```diff
@@ -0,0 +1,90 @@
...
+  bufferization::DropEquivalentBufferResultsPassOptions dropEquivalent;
+  dropEquivalent.modifyPublicFunctions = true;
+  pm.addPass(
+      bufferization::createDropEquivalentBufferResultsPass(dropEquivalent));
+  bufferization::BufferResultsToOutParamsPassOptions outParams;
+  outParams.hoistStaticAllocs = true;
+  outParams.addResultAttribute = true;
+  outParams.modifyPublicFunctions = true;
+  pm.addPass(bufferization::createBufferResultsToOutParamsPass(outParams));
+  pm.addNestedPass<func::FuncOp>(createCheddarElideOutParamCopies());
```
j2kun: As I understand, this is only intended to clean up out-param copies. But nothing appears to prevent the insertion of additional copies in the function body due to control flow (i.e., where an actual cheddar.copy is unavoidable, from the perspective of one-shot-bufferize). I think a loop like a BSGS might trigger this.

I think it might be worthwhile to try using the custom copy function on BufferizationOptions in the original one-shot bufferization pass. Is there anything that stops us from having that emit cheddar.copy directly (when the inputs are cheddar types)? Then we can have a final pass here that fails early if it detects cheddar.copy ops we didn't handle elsewhere? Otherwise, it seems like the cheddar copy op is not even used in this PR, so I'm not sure where it will come into the picture.

https://github.com/llvm/llvm-project/blob/de2a0dd540f714f5e5ae2d3017fe22e2aa6de002/mlir/include/mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h#L343

I guess I just find it hard to believe that we will be able to categorically eliminate ALL copies, so I'm trying to figure out a good way to proceed with guardrails.

j2kun (changes requested):
I admit I am not an expert either, but it seems reasonable how much you were able to get with just upstream machinery. I have one comment to explore, but otherwise I don't think it would be outrageous to merge this.

lib/Dialect/Cheddar/Transforms/CheddarBufferize.cpp
AlexanderViand: It's OK for copies to survive, they can be converted to (annoying/somewhat expensive) deep copies during the emitC lowering pass, but your comments made me take another look at this, and it turns out this can be way simplified, with no need for a `cheddar.copy` op or an elide-copy pass. Instead, we now just have (in the next PR in the stack) a custom emitC lowering for `mermef.copy` that defaults to a deep copy, but has a few simple checks that turns the vast majority of emitted copies (which are of values that'll be dead code afterwards) into `std::move`.

j2kun (approved):

PR #3407, opened 2026-09-10
Preserve logical client/server interface roles

AlexanderViand:
Work towards making it possible for HEIR to actually generate separate client-side and server-side binaries for a given model/program.


lib/Transforms/SplitPreprocessing/SplitPreprocessing.cpp
```diff
@@ -534,6 +534,10 @@ struct SplitPreprocessingPass
             builder.getNamedAttr(kClientHelperFuncName,
                                  builder.getStringAttr(op.getName())),
         }));
+    funcOp->setAttr(
+        kServerEvaluateFuncAttrName,
+        preprocessingFuncOp->getAttr(kServerPreprocessingFuncAttrName));
+    op->removeAttr(kServerEvaluateFuncAttrName);
```
mdgrs: Is it safe to already remove this Attr here?

mdgrs (approved):
Looks good to me

j2kun:
My two questions are:

1. Why have multiple dictionary attrs on the func instead of just one?
2. Could you provide a little bit more information about what the end-result of this work will look like? Will we be splitting HEIR into two `module`s? Producing two `cc_library` targets? Generating actual binaries with RPC interfaces or similar?

AlexanderViand:
> 1. Why have multiple dictionary attrs on the func instead of just one?

No particular reason, changed to one

> 2. Could you provide a little bit more information about what the end-result of this work will look like? Will we be splitting HEIR into two `module`s? Producing two `cc_library` targets? Generating actual binaries with RPC interfaces or similar?

The idea is that with this info, the backends can actually generate two sets of, e.g., *.h/*.cpp files, one for the client and one for the server. As I said in yesterday's meeting, I'll try and push a Lattigo example of what that might look like.

j2kun (approved):

PR #3410, opened 2026-09-11
disable loop unrolling by default

j2kun:
This PR enables loop support by default in HEIR.

`experimental-disable-loop-unroll` is changed to be default `true`, and legacy tests that require unrolling loops (such as older tests of the HECO transforms) opt-in to using it.

Remaining changes are updating test harness to pass through encryptions of zero as needed.


AlexanderViand (commented):
Maybe it's a good time to rename the option, to something (a) without "experimental" and (b) flipped, so that it's "--unroll-loops"?

j2kun:
@AlexanderViand fixed! I'm also not sure why the rotom_mnist test was failing, so I opted it back into loop unrolling, but I will look into that later. CC @edwjchen

tests/Examples/lattigo/ckks/lola/lola_test.go
```diff
@@ -61,14 +61,17 @@ func TestLoLA(t *testing.T) {
...
+	ctZero2 := Lola__encrypt__zero__2(evaluator, params, ecd, enc)
 
 	// The embedded constant weights are encoded into plaintexts once, up front.
 	startPre := time.Now()
 	weightPlains := lola_utils.Lola__preprocessing(params, ecd)
 	t.Logf("Lola__preprocessing took %s", time.Since(startPre))
 
 	start := time.Now()
-	resultCt := Lola__preprocessed(evaluator, params, ecd, ct0, weightPlains)
+	resultCt := Lola__preprocessed(evaluator, params, ecd, ct0, ctZero0, ctZero1, ctZero2, weightPlains)
```
AlexanderViand, line 74: What's the state of #3398 ? In an ideal world, I think we'd merge that first, before enabling looped kernels by default?

AlexanderViand (approved):
Nice!

tests/Examples/lattigo/ckks/lola/lola_test.go
j2kun: It's just waiting on me (I plan to get to it this week), but I don't see updating this API as a major obstacle.

AlexanderViand: I was more thinking it'd be nice to not change the user-code-facing API twice in a row, but it's not a strong enough feeling to not merge this one for now.

j2kun: I will try to get both in today

PR #3415, opened 2026-09-15
bazel: patch rules_cuda breakage for CUDA 12.9

AlexanderViand:
The CUDA 12.9 redistributable lists cufile and nvidia_fs, for which rules_cuda 0.3.0 registers no targets, yet its generated toolkit aggregates their headers, so a hermetic build fails to analyze @cuda//:cuda_headers. 

This has been fixed already upstream in rules_cuda (this just ports their fix into a local patch), so I assume it's actually easier for @j2kun to just poke them to put out a patch release and for us to bump to rules_cuda 0.3.1 when it comes out?


j2kun (approved):
This is probably the fastest way, looks like they have a very manual release process, LGTM.

PR #3416, opened 2026-09-15
Fix infinite looping frontend tests

j2kun:
BGV/BFV schemes don't support loops; in particular their noise analyses never converge during parameter selection. This causes the loop test in the frontend to hang, breaking CI.

For now, the frontend will force loop unrolling
for all schemes that are not CKKS.


AlexanderViand (approved):
frontend fix LGTM 👍  but I'm surprised a frontend test was the only one that failed - did we not have a single loop-using filecheck test for BGV/BFV?

j2kun:
> frontend fix LGTM 👍 but I'm surprised a frontend test was the only one that failed - did we not have a single loop-using filecheck test for BGV/BFV?

we did, but I opted them specifically in to unroll loops, and I am not sure how I missed the test failure for the frontend tests (we don't run them internally because we are not allowed to run clang as a subprocess during a test, but it should have failed on github and prevented me from submitting).

PR #3417, opened 2026-09-15
nfc: improve print statements of openfhe build script for local repro

j2kun:
Helped during debugging of https://github.com/google/heir/pull/3416


AlexanderViand (approved):
Ideally, this wouldn't rely on the echo line and the `>> "GITHUB_ENV"` line being kept in sync manually... but then again, who's going to touch this file ;)

